Issue #11, opened 2016-01-28
Search responder queries are not deterministic


benjamingeer:
Two problems:

1. In a full-text search (e.g. search for 'Orationes' in Incunabula), if a resource has a matching label and a matching value, Fuseki and GraphDB return different results: Fuseki returns the match for the resource label, while GraphDB returns the one for the text value. Both appear to be correct: we are using SAMPLE, so each triplestore is returning a different random result. Try to find another approach so that they return the same result.
2. GraphDB returns ontology entity labels in the wrong language (again, this seems to be because we have erroneously requested a random one, rather than the one in the user's preferred language as intended). A good solution would probably be to remove all ontology-related things from the search queries, and instead to have the search responder ask the ontology responder for that information.

benjamingeer:
`wip/ontology-in-search`

tobiasschweizer:
I have just thought about the handling of valueHasComment in the search templates #22

tobiasschweizer:
Summarizing the changes you made, I mention the following steps:
- in the SPARQL templates for fulltext and extended search, the language dependent labels for resource classes were removed, instead we will now deal with the Iris of the resource class and the property type returned by the query
- the SearchResultsRowV1s are wrapped in another case class called SearchResultRowWithEntityIris / SearchResultRowWithResourceClassIri exposing the resource class IRI (and the property type Iri in case of a fulltext search)
- these Iris are passed to the ontology responder
- the original SearchResultRowV1s are copied and completed with the information returned by the ontology responder

I have the following comments:

- in the search template, a `resourceLabel` is returned containing the `rdfs:label` of the resource found. This is absolutely fine in the RDF-terminology, but it is easily confused with a language dependent label like a "Ein Buch". Fot this reason, I found firstProp clearer (but this is some kind of SALSAH legacy).

-  in the fulltext template, there is a variable called `?valueObjectResource`. This is the resource relating to the found value object via a property. The name is not very clear to me.

- now, we have two steps: first the query is done and the results are put in a bunch of `SearchResultsRowV1`, which are wrapped into `SearchResultRowWithEntityIris` / `SearchResultRowWithResourceClassIri `. When the ontology responder has been called, the missing information are added to `SearchResultsRowV1` (the case class is copied). Could this be done without creating `SearchResultsRowV1` before calling the ontology responder?

benjamingeer:
[...]

PR #21, opened 2016-02-02
Get ontology info from ontology responder, not from search queries

benjamingeer:
This branch changes search SPARQL queries and the search responder so the search queries don't get ontology information. Instead, the search responder first does the search, then asks the ontology responder for the ontology information, which it merges into the search results.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/v1respondermessages/searchmessages/SearchMessagesV1.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/v1respondermessages/searchmessages/SearchMessagesV1.scala
@@ -140,7 +140,7 @@ case class SearchPreviewDimensionsV1(nx: Int, ny: Int)
   * @param value (text) value of the first property.
   */
 case class SearchResultRowV1(obj_id: IRI,
-                             preview_path: String,
+                             preview_path: Option[String],
                              iconsrc: Option[String],
                              icontitle: Option[String],
                              iconlabel: Option[String],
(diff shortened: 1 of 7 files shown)


benjamingeer:
This is for issue #11 .

benjamingeer:
I think there needs to be an additional test to make sure that the labels are being returned in the correct language, as per issue #11. I seem to remember that when we did a full-text search for `Orationes`, we got a label in French. Also, this pull request doesn't do anything about the non-deterministic effect of SAMPLE.

tobiasschweizer:
Summarizing the changes you made, I mention the following steps:
- in the SPARQL templates for fulltext and extended search, the language dependent labels for resource classes were removed, instead we will now deal with the Iris of the resource class and the property type returned by the query
- the SearchResultsRowV1s are wrapped in another case class called SearchResultRowWithEntityIris / SearchResultRowWithResourceClassIri exposing the resource class IRI (and the property type Iri in case of a fulltext search)
- these Iris are passed to the ontology responder
- the original SearchResultRowV1s are copied and completed with the information returned by the ontology responder

I have the following comments:

- in the search template, a `resourceLabel` is returned containing the `rdfs:label` of the resource found. This is absolutely fine in the RDF-terminology, but it is easily confused with a language dependent label like a "Ein Buch". Fot this reason, I found firstProp clearer (but this is some kind of SALSAH legacy).

-  in the fulltext template, there is a variable called `?valueObjectResource`. This is the resource relating to the found value object via a property. The name is not very clear to me.

- now, we have two steps: first the query is done and the results are put in a bunch of `SearchResultsRowV1`, which are wrapped into `SearchResultRowWithEntityIris` / `SearchResultRowWithResourceClassIri `. When the ontology responder has been called, the missing information are added to `SearchResultsRowV1` (the case class is copied). Could this be done without creating `SearchResultsRowV1` before calling the ontology responder?

benjamingeer:
* I added comments to the SPARQL templates to clarify that `resourceLabel` is called `firstprop` in the API. I'm making a distinction between `resourceLabel` and `resourceClassLabel`.
* `?valueObjectResource` is now called `?containingResource`.
* In the case of full-text search, I think it makes sense to construct `SearchResultsRowV1` objects first and then add ontology information to them, because we have to parse the SPARQL query results anyway to get the property IRIs. However, in the case of extended search, this is not necessary (because we already have the property IRIs), so I changed extended search to just create the `SearchResultsRowV1` objects once, rather than copying them.

tobiasschweizer:
Great, your changes look fine an flatter my ego! 

The code compiles and tests run successfully.

Let's merge it!

Issue #19, opened 2016-02-01
Value version history should not refer to hidden values


benjamingeer:
The values responder hides values that the user doesn't have permission to see, but the IRI of a hidden values can still be returned as `previousValue`.

PR #25, opened 2016-02-03
fix (values responder): Hide value IRIs that the user doesn't have permission to see.

benjamingeer:
Fixes issue #19.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
@@ -737,12 +737,18 @@ class ValuesResponderV1 extends ResponderV1 {
                     valuePermissions.nonEmpty
             }
 
+            // Make a set of the IRIs of the versions that the user has permission to see.
+            visibleVersionIris = filteredVersionRowsVector.map(_("value")).toSet
+
             versionV1Vector = filteredVersionRowsVector.map {
                 rowMap =>
                     ValueVersionV1(
                         valueObjectIri = rowMap("value"),
                         valueCreationDate = rowMap.get("valueCreationDate"),
-                        previousValue = rowMap.get("previousValue")
+                        previousValue = rowMap.get("previousValue") match { // Don't refer to a previous value that the user doesn't have permission to see.
+                            case Some(previousValueIri) if visibleVersionIris.contains(previousValueIri) => Some(previousValueIri)
+                            case _ => None
+                        }
                     )
             }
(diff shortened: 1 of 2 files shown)


tobiasschweizer:
Ok, If I understand correctly, the scenario in the test (`versionHistoryWithHiddenVersion`) is this:

- we look at the title of a book (http://data.knora.org/21abac2162/values/f76660458201)
- this title has a previous value version which is not visible to the user (with Iri 22222222) with the value "previous 1"
- this value (222222) has a previous value version which is visible to the user (with Iri 11111111) with the value "previous 2"

The user will see the value with the current title and the previous version (11111111) "previous 2", but not "previous 1" with the Iri 22222222. 

To make this consistent, you set the "previous value" pointer of the current title to None rather than mentioning an Iri the user is not allowed to dereference. However, the user knows that he cannot see a version because otherwise there would never be a None in previousValue but in the first value.

So you actually check if (1) there is a previous value (nonEmtpy) and (2) it is visible.

I agree.

tobiasschweizer:
Ok, If I understand correctly, the scenario in the test (`versionHistoryWithHiddenVersion`) is this:

- we look at the title of a book (http://data.knora.org/21abac2162/values/f76660458201)
- this title has a previous value version which is not visible to the user (with Iri 22222222) with the value "previous 1"
- this value (222222) has a previous value version which is visible to the user (with Iri 11111111) with the value "previous 2"

The user will see the value with the current title and the previous version (11111111) "previous 2", but not "previous 1" with the Iri 22222222. 

To make this consistent, you set the "previous value" pointer of the current title to None rather than mentioning an Iri the user is not allowed to dereference. However, the user knows that he cannot see a version because otherwise there would never be a None in previousValue but in the first value.

So you actually check if (1) there is a previous value (nonEmtpy) and (2) it is visible.

I agree.

tobiasschweizer:
Ok, If I understand correctly, the scenario in the test (versionHistoryWithHiddenVersion) is this:
- we look at the title of a book (http://data.knora.org/21abac2162/values/f76660458201)
- this title has a previous value version which is not visible to the user (with Iri 22222222) with the value "previous 1"
- this value (222222) has a previous value version which is visible to the user (with Iri 11111111) with the value "previous 2"

The user will see the value with the current title and the previous version (11111111) "previous 2", but not "previous 1" with the Iri 22222222.

To make this consistent, you set the "previous value" pointer of the current title to None rather than mentioning an Iri the user is not allowed to dereference. However, the user knows that he cannot see a version because otherwise there would never be a None in previousValue but in the first value.

So you actually check if (1) there is a previous value (nonEmtpy) and (2) it is visible.

I agree.

Issue #23, opened 2016-02-03
Ontology responder caches entity info filtered by language


benjamingeer:
So if someone requests information about a resource class in German, the ontology responder will cache that, and then if someone requests information about the same resource class in French, they'll get the cached German version.

The ontology responder should either:

* Cache the raw SPARQL query results, and filter them by language on each request, or
* Cache them separately by language.

benjamingeer:
I chose the second option because it's easier, and memory is probably not an issue with ontology data.

benjamingeer:
Fixed by pull request #26, but see issue #28.

PR #26, opened 2016-02-03
fix(ontology responder): Include preferred language in ontology cache keys

benjamingeer:
Fixes issue #23.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/v1/OntologyResponderV1.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v1/OntologyResponderV1.scala
@@ -59,53 +59,58 @@ class OntologyResponderV1 extends ResponderV1 {
       * Keys for the ontology cache.
       */
     private object OntologyCacheKeys {
+        import scala.math.Ordered.orderingToOrdered
 
         /**
           * The type of ontology cache keys for instances of [[ResourceEntityInfoV1]].
           *
           * @param resourceClassIri the IRI of the resource class described.
+          * @param preferredLanguage the user's preferred language, which the key's value will use if it was available.
           */
-        class ResourceEntityInfoKey(val resourceClassIri: IRI) extends Ordered[ResourceEntityInfoKey] {
+        class ResourceEntityInfoKey(val resourceClassIri: IRI, val preferredLanguage: String) extends Ordered[ResourceEntityInfoKey] {
             def compare(that: ResourceEntityInfoKey): Int = {
-                this.resourceClassIri.compare(that.resourceClassIri)
+                (this.resourceClassIri, this.preferredLanguage) compare (that.resourceClassIri, that.preferredLanguage)
             }
 
             override def equals(that: Any): Boolean = {
                 that match {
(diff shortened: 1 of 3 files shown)


tobiasschweizer:
You extended the Ontology Cache so it does not take one argument (the Iri of the entity), but two: also the language. The key for the cache is now a combined one consisting of entity Iri and the preferred language of the user.

My comment:

In the templates `getEntityInfoForResource` and `getEntityInfoForProperty`, all the possible languages are returned, and each result row has a variable `lang` indicating the language. 

However, all the languages but the user's preferred one are filtered out (a fallback default language is provided):

```scala
// Filter the query results to get text in the user's preferred language (or the application's default language),
                // if possible.
                resourcesFilteredByLanguage = SparqlUtil.filterByLanguage(response = resourcesResponse,
                    langSpecificColumnName = "o",
                    preferredLanguage = userProfile.userData.lang,
                    settings = settings
                )
```
(queryResourceEntityInfoResponse)

If I understand correctly, then only the preferred language is cached and the same query is run again if a user requests the same entity Iri providing another preferred language.

It would be nice to automatically create several combined cache keys for each language if this is feasible. Of course, we would have redundancy then for the non language dependent information.

But probably that would make it too complicated (regarding the possibly non expensive queries we save).

benjamingeer:
> only the preferred language is cached

If the preferred language is French, the ontology responder will try to return information in French. But if the information was available only in German, you'll get German instead. However, the caching mechanism doesn't look at the contents of the data, so those results will still be cached under French. The cache key contains the language code of the preferred language, not the actual language that was returned.

If someone else then requests the same information in German, the triplestore will be queried again, and the results will also be cached, this time under German.

This way, if we get repeated requests for French data, but there is no French data, we don't keep doing the same query again and again. At most, the number of queries is equal to the number of entity IRIs in the ontologies times the number of preferred languages.

The other, perhaps better option would be to cache the raw query results for each entity IRI, and filter them by language each time they are read from the cache.

> It would be nice to automatically create several combined cache keys for each language if this is feasible.

I don't understand what you mean by 'combined cache keys for each language'.

> Of course, we would have redundancy then for the non language dependent information.

We have that already. If different users request the same information in different languages, all the results will be cached, and the non-language-specific information will be cached redundantly.

tobiasschweizer:
> The other, perhaps better option would be to cache the raw query results for each entity IRI, and filter them by language each time they are read from the cache.

Yes, this would be an improvement since we would have to query each entity Iri only once in the triplestore.

What I meant before was that you could turn one entity query automatically in several entries in the cache (following the existing architecture). For example, if an entity is queried for the first time, the cache would be build up for every possible combination of this Iri and language keys. So the behaviour would be the same like for the solution you proposed above.

benjamingeer:
> What I meant before was that you could turn one entity query automatically in several entries in the cache (following the existing architecture). For example, if an entity is queried for the first time, the cache would be build up for every possible combination of this Iri and language keys. So the behaviour would be the same like for the solution you proposed above.

We could do this but we would need a list of all possible languages.

tobiasschweizer:
That's a good point! 

I suggest that you note the proposal you made (caching the raw results) and put it in an issue. I do not think this change is urgent, but we should not forget about it either.

I think you can merge your changes back to develop. :-)

PR #27, opened 2016-02-03
perf(search): Move MINUS clause to improve extended search performance.

benjamingeer:
This search responder test was taking 8 seconds on GraphDB SE 6.6.2:

```
should return 79 pages when we search for all pages that have a sequence number greater than 450 in the Incunabula test data
```

I moved the MINUS clause that filters out deleted resources, and it now takes 350 ms. It's still fast with Fuseki 2.3.0, too.

### DIFF
--- a/webapi/src/main/twirl/queries/sparql/v1/searchExtended.scala.txt
+++ b/webapi/src/main/twirl/queries/sparql/v1/searchExtended.scala.txt
@@ -104,10 +104,6 @@ WHERE {
 
         @if(searchCriterion.valueType != "http://www.knora.org/ontology/knora-base#Resource") {
             ?resource <@{searchCriterion.propertyIri}> ?valueObjectIri@index .
-
-            MINUS {
-                ?resource knora-base:isDeleted true .
-            }
         }
 
         # Search criterion @index: operator @searchCriterion.comparisonOperator
@@ -193,10 +189,6 @@ WHERE {
 
                         ?resource <@{searchCriterion.propertyIri}> ?targetResource@index .
 
-                        MINUS {
-                            ?resource knora-base:isDeleted true .
-                        }
-
                         ?targetResource@index rdfs:label ?anyLiteral@index .
 
                         MINUS {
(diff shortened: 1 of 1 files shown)


subotic:
Just out of curiosity, why is the commit called ```perf(search)```? If applying the scheme that we loosely agreed upon, then shouldn't it be more something like ```feature (webapi)```

benjamingeer:
Ivan, I think we need a label for performance, because improving performance doesn't add a feature, it just makes an existing feature faster. Also, in the scheme we have, I thought that the label in parentheses was supposed to be the specific component affected. If we use `webapi`, nearly all our commit messages will say `webapi`. Using `search` here is more specific and therefore more informative.

subotic:
For clarification, the structure of the commit message as discussed was ```<type>(<scope>): <subject>```.

I agree that performance improvement is not a new feature. This is why there is the ```fix``` type. We could certainly rename ```fix``` to something more positive sounding like ```improvement```.

The scope, as far as I understand it, should provide the reader with a quick overview of where in the codebase the change was introduced. Beside webapi, we have also salsah, docs and knora-ontologies. As soon as we start to work more on salsah, we will be glad to be able to easily discern commits.

The subject line sums up the change nicely, where you describe it as an performance improvement of search. I don't think that by writing it again as a type and scope gives the reader more information.

Just my thoughts.

benjamingeer:
If I remember correctly, when you first suggested this commit message scheme, you gave `query resource` as an example of a scope, so that's the sort of thing I've been writing ever since. I don't mind if we change it now, but let's have a complete, documented list of scopes to choose from.

I think it's important to have `fix`. You don't 'improve' a bug, you fix it. But improved performance isn't a bug fix. If there's a bug, it means that something was working incorrectly. But performance is neither correct nor incorrect, only better or worse.

benjamingeer:
Also, 'improvement' doesn't seem to add any information, since every commit should contain an improvement.

tobiasschweizer:
Are you seriously arguing about this?

subotic:
It denotes the type of the commit. Broadly speaking, it gives a clue if this change is affecting the user (feature, fix), or not. The commit messages with feature and fix could then be used do generate a change log for the release.

Tobias, how else are we going to decide how a commit message should look like if we don't speak (or in this case write) about it?

benjamingeer:
I was talking about 'scope', not 'type'. Let's talk about this on Skype later.

benjamingeer:
Better yet, let's talk about it in person next week.

subotic:
ok :-)

PR #31, opened 2016-02-08
fix (webapi): Update a resource's lastModificationDate when one of its values is changed

benjamingeer:
Fixes issue #9.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
@@ -972,7 +972,7 @@ class ValuesResponderV1 extends ResponderV1 {
                 ).toString()
             }
 
-            _ = println(sparqlQuery)
+            // _ = println(sparqlQuery)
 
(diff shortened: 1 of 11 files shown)


benjamingeer:
While doing this, I realised that if a resource has more than one `lastModificationDate` (which should never happen), it can cause SPARQL updates to insert duplicate data. I opened another issue for this, #32.

tobiasschweizer:
In the SPARQL templates that handle adding values or new versions of values to a resource, you delete (Delete clause) and recreate (Insert clause) the resource's `lastModificationDate` so it reflects the time when the last modification has been made (as the name suggests) :-)

Why does this have to be included in the Where-clause of the templates? It is optional, so it won't affect the matching to a specific resource.

```
 @* Get the resource's last modification date, if it has one, so we can update it. *@

    OPTIONAL {
        ?resource knora-base:lastModificationDate ?resourceLastModificationDate .
    }
```

benjamingeer:
The WHERE clause binds variables that are used in the DELETE and INSERT clauses. If a variable is unbound, any statement in the DELETE clause that uses that variable will have no effect. We have to delete any existing `lastModificationDate` before we can insert a new one. Therefore we need to know the current `lastModificationDate` so it can be deleted.

PR #35, opened 2016-02-10
Wip/subotic cleanup

subotic:
I did some overdue cleanup and documentation


benjamingeer:
Looks good, thanks! Maybe we can also remove `Obfuscate.scala`?

Can you check these:

* `webapi-production.rst`: It talks about running Fuseki in Tomcat, but not about running the built-in Fuseki. Should we still support Fuseki in Tomcat? And do you still want to use supervisord?
* `test-tags.rst`: I put a TODO there because I'm not sure it's up to date. Also, now that we use SBT configurations to choose triplestores for tests, does the DbTest tag still make sense?

Could you also add at least a little bit to `development-tools.rst`, listing the different SBT configurations that are available for different triplestores (`fuseki:`, `graphdb:`, etc.), saying that these are defined in `KnoraBuild.sbt`, and giving examples of `xyz:test` and `xyz:test-only`?

subotic:
Yes, of course. Very good suggestions. Thanks. Maybe you can wait with the merge until I do the other things.

subotic:
> Should we still support Fuseki in Tomcat?

Fuseki under Tomcat is probably the preferred way of running it in production, like running GraphDB under Tomcat. We can have a short documentation describing both cases. But at the end, the system administrator responsible for deployment will need to read/know more than just this short description. 

> And do you still want to use supervisord?

Good question. Maybe it will be the easiest/fastest way of running it manually at the beginning, at least until we don't have a RPM package and dedicated systemd scripts.

> ... does the DbTest tag still make sense?

At the moment we probably don't need it. We never really used it. Shall I remove it, or should we leave it? Maybe we will find it useful some day?

benjamingeer:
> Fuseki under Tomcat is probably the preferred way of running it in production, like running GraphDB under Tomcat. We can have a short documentation describing both cases.

OK, can you write it?

I think it's a good idea to remove unneeded code (less code means less code that can have bugs, and less potential confusion for developers), so let's remove the DbTestTag. I think we could actually use a SipiTestTag instead, because some tests will require Sipi.

subotic:
>> Fuseki under Tomcat is probably the preferred way of running it in production, like running GraphDB under Tomcat. We can have a short documentation describing both cases.

> OK, can you write it?

Doing it just now.

I have also updated the documentation regarding the section header levels, and going through the whole documentation and adjusting them. Also we talked about the ToDo list a while ago. I added it also. 

> I think we could actually use a SipiTestTag instead, because some tests will require Sipi.

Then I can rename DbTestTag to SipiTestTag.

subotic:
What I still didn't cover is the documentation surrounding ```KnoraBuild.sbt``` and the different tests that can be called. This will change completely when I make the switch to unforked testing. I will write the documentation when I introduce this change.

I think we should merge this now as a lot of files where changed. I'm on vacation from tomorrow on and probably wont work on this for more than a week.

benjamingeer:
OK, but the ` wip/subotic_cleanup` branch doesn't compile:

```
/Users/ben/git/Knora/webapi/src/main/scala/org/knora/webapi/store/triplestore/TriplestoreManagerActor.scala:31: object JenaGraphDBActor is not a member of package org.knora.webapi.store.triplestore.embedded
[error] import org.knora.webapi.store.triplestore.embedded.{JenaGraphDBActor, JenaTDBActor}
```

```
[error] /Users/ben/git/Knora/webapi/src/main/scala/org/knora/webapi/store/triplestore/TriplestoreManagerActor.scala:31: object JenaGraphDBActor is not a member of package org.knora.webapi.store.triplestore.embedded
[error] import org.knora.webapi.store.triplestore.embedded.{JenaGraphDBActor, JenaTDBActor}
[error]        ^
```

subotic:
hmm, not good. I will check.

subotic:
Sorry, forgot to run ```clean```. Now everything should compile.

Issue #37, opened 2016-02-11
Default Permissions for knora-base:hasStillImageFileValue is missing in the ontology


tobiasschweizer:
For `knora-base:hasStillImageFileValue`, no default permissions are set.

This is because in `incunabula-onto.ttl`, `incunabula:page` is made a subclass of `knora-base:StillImageRepresentation` and inherits its property `knora-base:hasStillImageFileValue` (defined in `knora-base.ttl`), but without any default permission.

I suggest to define those default permissions in `incunabula-onto.ttl` because this should be project specific.

CC @benjamingeer

tobiasschweizer:
Is there actually a way to define project specific default permissions on a knora-base property? Would that not have a global effect?

Would we have to make subproperty and attach the default permissions to that?

benjamingeer:
I'm confused. Why does `StillImageRepresentation` need default permissions? And what do permissions have to do with the property `hasStillImageFileValue`? Permissions relate to resources and values, not to properties. The class `incunabula:page` has default permissions. Why isn't that enough?

tobiasschweizer:
As far as I understand, default permissions are set for resources and for values.

Example from `incunabula-onto.ttl`. These are the statements that relate to the resource (page):

```turtle
### ###########################################
### incunabula:page

:page rdf:type owl:Class ;

      rdfs:subClassOf knora-base:StillImageRepresentation ,
                      [
                         rdf:type owl:Restriction ;
                         owl:onProperty :pagenum ;
                         owl:cardinality "1"^^xsd:nonNegativeInteger                      ] ,
                      [
                         rdf:type owl:Restriction ;
                         owl:onProperty :partOfValue ;
                         owl:cardinality "1"^^xsd:nonNegativeInteger                      ] ,
                      [
                         rdf:type owl:Restriction ;
                         owl:onProperty :partOf ;
                         owl:cardinality "1"^^xsd:nonNegativeInteger                      ] ,
                      [
                         rdf:type owl:Restriction ;
                         owl:onProperty :seqnum ;
                         owl:cardinality "1"^^xsd:nonNegativeInteger                      ] ,
                      [
                         rdf:type owl:Restriction ;
                         owl:onProperty :description ;
                         owl:maxCardinality "1"^^xsd:nonNegativeInteger                      ] ,
                      [
                         rdf:type owl:Restriction ;
                         owl:onProperty :citation ;
                         owl:minCardinality "0"^^xsd:nonNegativeInteger                      ] ,
[...]

PR #43, opened 2016-02-12
Wip/sipi integration

tobiasschweizer:
I have still a few open questions:
- SipiMessages -> object SIPI -> FileType: can we make that a enum?
- SipiResponder -> callSipiConvertRoute: can we use a try catch block here to catch and handle HTTP errors? For example, if Sipi host is not reachable, it throws `spray.can.Http$ConnectionAttemptFailedException: Connection attempt to localhost:1024 failed` or `spray.httpx.UnsuccessfulResponseException: Status: 400 Bad Request` if Sipi does not return a HTTP 200 (conversion unsuccessful)
- do we actually need a Sipi status code? 
- is there a better solution for SipiMessages `SipiConversionResponse`. This is meant to be generic and thus everything is optional which makes it inconvenient to use.


tobiasschweizer:
at the moment, file values are created without any permissions #37

benjamingeer:
Initial comments (I'll look at the code in more detail another day):

> SipiMessages -> object SIPI -> FileType: can we make that a enum?

Yes. Your comment says `I had problems to use it in match case statement`. What problems did you have exactly? The StackOverflow post you mention in the comment isn't relevant, because it's about Java enumerations, not Scala enumerations.

For an example of using a Scala enumeration in a match-case statement, see `DateUtilV1.julianDay2DateString`, which does a match-case over the enumeration `KnoraCalendarV1.Value`.

> SipiResponder -> callSipiConvertRoute: can we use a try catch block here to catch and handle HTTP errors?

In a method that returns a `Future`, any error should result in a failed future. There's no need for a try-catch block unless you want to catch one type of exception and re-throw it as another type of exception (which would still result in a failed `Future`). Do you mean that `callSipiConvertRoute` is throwing an exception rather than returning a failed `Future`? That could happen if:

* `conversionRequest.toFormData() throws an exception`
* the constructor of `FormData` throws an exception
* the constructor of `Post` throws an exception
* your `pipeline` function throws an exception

The solution is to ensure that a `Future` is constructed before you do anything that could throw an exception. Please see `Handling Errors with Futures` in `futures-with-akka.rst`.

The error handling could also be improved here:

```scala
            _ = if (conversionResult.status != 0) throw BadRequestException("Sipi error when trying to convert the file")
```

Here, the `BadRequestException` tells the client nothing about what sort of error occurred or why. If the error code returned by Sipi gives some information about the error, that information should be passed on to the client.

> do we actually need a Sipi status code?

If Sipi returns a status code indicating an error, I suggest that `SipiResponderV1` should return a failed `Future` containing an exception with information about the error. There's no need for Sipi's status codes to be visible to anything except `SipiResponderV1`.

> is there a better solution for SipiMessages SipiConversionResponse. This is meant to be generic and thus everything is optional which makes it inconvenient to use.

You could just parse Sipi's response to a `JsValue` and check the status code. If the status code is OK, you could then convert the `JsValue` into a case class instance.

The formatting of `SipiResponderV1.receive` looks strange to me:

```scala
    def receive = {
        case SipiFileInfoGetRequestV1(fileValueIri, userProfile) => future2Message(sender(), getFileInfoForSipiV1(fileValueIri, userProfile), log)
        case convertPathRequest: SipiResponderConversionPathRequestV1
        => future2Message(sender(), convertPathV1(convertPathRequest), log)
        case convertFileRequest: SipiResponderConversionFileRequestV1
        => future2Message(sender(), convertFileV1(convertFileRequest), log)
        case other => sender ! Status.Failure(UnexpectedMessageException(s"Unexpected message $other of type ${other.getClass.getCanonicalName}"))
    }
```

Please put the `=>` on the same line as the `case`.

benjamingeer:
About `webapi/_test_data/test_route/create_page_with_binaries.py`: I think it's good that there's a test for this, but it should be part of the existing test suite instead of a separate script. There's an example of a test that connects to the API over HTTP in `org.knora.webapi.e2e.v1.ResourcesV1E2ESpec`. There's also a tag called `SipiTest` that you could tag your test with, to enable people to suppress it if they want to run only the tests that don't require Sipi.

In `application.conf`, you've changed the default triplestore to GraphDB, but it should be Fuseki.

In `SipiResponderConversionRequestV1`, the comment says: `For each type of conversion request, an implementation of `toFormData` must be provided.` This is not clear to me. Why must it be provided? What does it do? Please write a Scaladoc comment for the method `def toFormData(): Map[String, String]` explaining what implementations of the method are expected to do, why it is an abstract method, and what is supposed to be in the `Map` that it returns.

The comments for the constructors of `SipiResponderConversionPathRequestV1` and `SipiResponderConversionFileRequestV1` mention the `non GUI-case` and the `GUI-case`, but this isn't going to be clear to anyone who hasn't been following in detail the discussions about the design of Knora-Sipi interaction. Please write longer comments explaining what these two cases are, assuming that the reader knows nothing about the SALSAH GUI.

In `SipiResponderConversionPathRequestV1`, since `source` is an absolute path, I suggest using a `java.io.File` object instead of a string. The Java and Scala IO libraries support doing things with `File` objects directly, and this can prevent bugs (e.g. if someone forgets that a particular string represents an absolute path rather than a relative path).

In `SipiMessagesV1.scala`, there's an object called `Sipi`, but its name is confusing, because it doesn't represent Sipi; it just contains some Sipi-related constants. Perhaps `SipiConstants` would be a better name (like `OntologyConstants`).

To fix issue #37, we may not be able to depend anymore on having just one Knora property per file type. This would mean that something like `Sipi.fileType2FileValueProperty` couldn't work, at least not the way you're using it in `ResourcesResponderV1.createResourceAndCheck`. But I think we should talk to Lukas about how to deal with #37.

The method `ResourcesResponderV1.createResourceAndCheck` refers several times to `GUI-case` and `non-GUI case`, not only in comments, but also in error messages. This isn't going to mean anything to developers who are new to the code, and still less to clients who receive these error messages. Please find a clearer way to say what you mean, without assuming so much knowledge on the part of the reader.

In `ValueUtilV1`, why is the method `makeFileValueV1FromCreateFileV1` commented out?

In `SipiResponderV1`, when you use spray pipelines, please don't just provide a link to the Spray documentation. Explain in a comment what each pipeline is supposed to do.

In `ResourcesRouteV1`, again, please clarify the comments that refer to `GUI case` and `non-GUI case`. This file is really complex and hard for me to understand. It needs a lot more comments in the code.

tobiasschweizer:
I could do some of the things but I am not done yet.

The code in `SipiResponderV1.scala` on lines 112 to 135 is executed (handling of tmp files), but still the error message received by the client is directly the one from spray.

So re-throwing an error does not seem to reach the client. What is the difference to your example (https://github.com/dhlab-basel/Knora/blob/wip/transaction-management/webapi/src/main/scala/org/knora/webapi/responders/v1/TransactionUtil.scala#L52)?

I would like to replace the message originally thrown by spray with our own exceptions. For example, I made a case class to nicely represent Sipi error messages: `SipiErrorConversionResponse`.

Is the spray error sent before our callback is ever called? In their example (https://github.com/spray/spray/blob/release/1.2/examples/spray-client/simple-spray-client/src/main/scala/spray/examples/Main.scala), they use `onComplete`.

benjamingeer:
I bet it's because of what I said above:

> That could happen if:
>
> * `conversionRequest.toFormData() throws an exception`
> * the constructor of `FormData` throws an exception
> * the constructor of `Post` throws an exception
> * your `pipeline` function throws an exception
>
> The solution is to ensure that a `Future` is constructed before you do anything that could throw an exception. Please see `Handling Errors with Futures` in `futures-with-akka.rst`.

tobiasschweizer:
solved in https://github.com/dhlab-basel/Knora/commit/28138790e0330c0a67a3355c9a789396012c5e5b

benjamingeer:
The test `ResourcesV1E2ESpec` fails for me, with two different errors:

```
2016-02-23 10:37:24,415 - ERROR - org.knora.webapi.e2e.v1.ResourcesV1E2ESpec - pool-1-thread-1-ScalaTest-running-ResourcesV1E2ESpec - ResourcesV1E2ESpec(akka://org-knora-webapi-e2e-v1-ResourcesV1E2ESpec) - Unable to run route
org.knora.webapi.FileWriteException: Data directory /tmp/webapi_tmp/ does not exist on server

2016-02-23 10:37:25,978 - ERROR - akka.dispatch.Dispatcher - org-knora-webapi-e2e-v1-ResourcesV1E2ESpec-akka.actor.default-dispatcher-9 - akka.dispatch.Dispatcher - Sipi not reachable
```

This seems to be because it now depends on Sipi. That's not good, because Sipi hasn't been released yet. The tests have to pass without Sipi. Please talk to Ivan to figure out the most convenient way to enable you to run the tests with Sipi, while enabling everyone else to run the tests without Sipi.

The Scaladoc comment on `SipiConstants.lookup` says:

```
Given the name of a file type in this enumeration, returns the file type. If the file type is not found, throws an BadRequestException.
```

But the method actually throws `SipiException`, which seems correct, so the comment needs to be changed.

In `sipi-and-knora.rst`, your RST syntax is incorrect. To make an inline literal in RST, you need double backticks, not single backticks as in Markdown. See the [RST quick reference](http://docutils.sourceforge.net/docs/user/rst/quickref.html#inline-markup).

The sections **GUI-Case** and **Non GUI-Case** both end with a message being sent to the resources responder, without explaining what happens next (the resources responder talks to the Sipi responder, which talks to Sipi). I found this confusing, because it wasn't immediately obvious that there's a separate section further down, called **Further Handling of the GUI and the non GUI-case in the Resources Responder**. Please add a label just before that section, like this:

```
.. _resources-responder-and-sipi:
```

Then at the end of each of the sections **GUI-Case** and **Non GUI-Case**, please add a sentence like this:

```
See :ref:`resources-responder-and-sipi` for details of how the resources responder then handles the request.
```

tobiasschweizer:
Thx, I will look at how to do the fake Sipi responder together with Ivan now.

benjamingeer:
OK. Also, when you're done, please have IntelliJ IDEA format your code (menu item Code -> Reformat Code).

tobiasschweizer:
Ok, done. Ivan tries to setup a fake Sipi actor.

tobiasschweizer:
Ivan figured out a way to setup fake actors. I can now implement a sipi responder that simulates sipi's behaviour for our test!

benjamingeer:
Wonderful, looking forward to seeing it.

tobiasschweizer:
Ok, now you should be able to run the test `SipiV1E2ESpec` without a running Sipi server.

I will have a look at the ResourcesResponderSpec because there we have a test dealing with the creation of an image that involves the Sipi Responder.

Issue #34, opened 2016-02-09
Don't use rdfs:range to check data consistency


benjamingeer:
Currently we use `rdfs:range` for consistency checks when creating values: if a property P has an `rdfs:range` of class C, we check that the value submitted is a C. In other words, we use `rdfs:range` to mean that if a value is not a C, it's invalid to make it an object of P.

Ontotext pointed out that this isn't actually what `rdfs:range` means. As [its definition](https://www.w3.org/TR/rdf-schema/#ch_range) says, it really means that if something is an object of P, it is a C. In other words, `rdfs:range` is for inferring the type of the value, not for imposing constraints. They suggest we use another predicate to avoid confusion.

In fact, as far as I can tell, there's no standard predicate for this, so perhaps we should add one to `knora-base`.

lrosenth:
Yes, I think I understand. If "something" has a property of P, the triplestore knows that this "something" must be a C. Right?

benjamingeer:
> Yes, I think I understand. If "something" has a property of P, the triplestore knows that this "something" must be a C. Right?

That would be `rdfs:domain`. With `rdfs:range`, if something has a property P with value V, the triplestore knows that V must be a C.

Instead of `rdfs:domain` (which Knora doesn't currently check) and `rdfs:range`, we could add properties with names like `knora-base:subjectConstraint` and `knora-base:objectConstraint`. Then you could have:

```
:hasStillImageFileValue rdf:type owl:ObjectProperty ;
              rdfs:subPropertyOf :hasFileValue ;
              :subjectConstraint :StillImageRepresentation ;
              :objectConstraint :StillImageFileValue .
```

lrosenth:
Yes, of course – I always confuse the two ways :-)
I like Your idea with :subjectConstraint and :objectConstraint !

By the way, do we understand the cardinalities correctly??? I'm always a little unsure about this. I will have to learn it for my lecture next semester – do You want to help me ???

subotic:
I find ```subjectConstraint``` and ```objectConstraint``` a good idea. Now that you explained the meaning of ```rdfs:domain```and ```rdfs:range```, I can vaguely remember that we already talked about it a long while ago.

benjamingeer:
I think we do understand cardinalities correctly. GraphDB can use them for constraint checks; they've just been explaining to me how to do it (see issue #33), and that was the context in which they pointed out that we `rdfs:range` means something else. I'd be glad to help you with your class.

[...]

Issue #33, opened 2016-02-09
Check the triplestore for inconsistencies


benjamingeer:
This could be either:

* A program that we can run periodically to find inconsistencies.
* A consistency checker that runs each time an update is performed. But this would pose a problem. To create a resource, we first create an empty resource, and then we add each value, one by one. The constraints won't be met until all the values are added.

Ideally, it would use the existing OWL constraints that are defined in the ontologies in the triplestore.

GraphDB [can do consistency checks](http://graphdb.ontotext.com/documentation/standard/reasoning.html#consistency-checks), but the documentation is not clear to me. I've emailed them to ask for clarification.

benjamingeer:
To make it possible to run a consistency checker on each update, we could implement our own application-level transaction management in the store package. SPARQL update syntax allows you to [concatenate multiple operations together using a semicolon](https://www.w3.org/TR/sparql11-update/#updateLanguage). We could implement something in the store package that would:

* Accept a `StartTransaction` message with an API request ID (which we already have).
* Accept one or more `SparqlUpdateRequest` messages with the same request ID, and store them in memory.
* On receipt of a `CommitTransaction` message, concatenate all the requests together into a valid SPARQL update, and run it in a single triplestore transaction.
* On receipt of a `RollbackTransaction` message, discard the saved update requests.

If someday we use the Sesame API instead of HTTP, this could be implemented to use Sesame's transaction management rather than concatenation.

benjamingeer:
cc @subotic

benjamingeer:
Ontotext have clarified the way their consistency checker works. It looks as if we should be able to check for consistency with OWL cardinalities after every update. I'll try to get it working.

subotic:
> If someday we use the Sesame API instead of HTTP, this could be implemented to use Sesame's transaction management rather than concatenation.

This also holds true with the embedded Jena triplestore, when we get it to work some day.

benjamingeer:
The transaction management idea I suggested above would require a change in the values responder. Currently when it gets a `CreateMultipleValuesRequestV1`, it creates each value and then checks whether the value was created. With transaction management, this check would have to be done after all the whole transaction was committed, which means that it would have to be done by the resources responder.

benjamingeer:
[...]

PR #49, opened 2016-02-23
fix (ontologies): Use Knora-specific properties instead of rdfs properties to check consistency during updates

benjamingeer:
This pull request replaces `rdfs:domain` and `rdfs:range`, which we were misusing for consistency checking, with Knora-specific properties, following the discussion in #34 and #33.

* `rdfs:domain` is replaced with `knora-base:subjectClassConstraint`.
* `rdfs:range` is replaced with `knora-base:objectClassConstraint` and `knora-base:objectDatatypeConstraint`.

A subsequent pull request will enable us to use GraphDB's built-in consistency checking for both `knora-base:subjectClassConstraint` and `knora-base:objectClassConstraint`.

I've also cleaned up the ontologies a bit in general.

The old SALSAH export script will also need to be changed to output these Knora-specific properties instead of `rdfs:domain` and `rdfs:range`.


tobiasschweizer:
In `knora-base.ttl`, shouldn't there be a `knora-base:objectDataTypeConstraint` on `resourceIcon` replacing the original `rdfs:range` (`xsd:string`)? As far as I understood, `knora-base:objectClassConstraint` is used to replace `rdfs:range` in case of an OWL-class, whereas `knora-base::objectDataTypeConstraint` should be used for a datatype like `xsd:string`. I understood that the latter case is not involved in consistency checking, but should still be there for documentation purposes.

tobiasschweizer:
Could you update the knora-base documentation (`docs/latex/knora-base/knora-base.tex`)?

benjamingeer:
> In `knora-base.ttl`, shouldn't there be a `knora-base:objectDataTypeConstraint` on `resourceIcon` replacing the original `rdfs:range` (`xsd:string`)?

Yes. Fixed.

> Could you update the knora-base documentation (`docs/latex/knora-base/knora-base.tex`)?

Done.

benjamingeer:
I missed a few `rdfs:range` in the LaTeX doc, fixed now.

tobiasschweizer:
Did you replace `rdfs:domain` and `rdfs_range` manually using a text editor?

I am concerned that - given the multitude of changes made to the ontology files - we cannot really assure that we actually did not change the ontology by accident.

benjamingeer:
> Did you replace `rdfs:domain` and `rdfs_range` manually using a text editor?

Yes.

> I am concerned that - given the multitude of changes made to the ontology files - we cannot really assure that we actually did not change the ontology by accident.

If I made a mistake, it's your job as reviewer to find it. :) But all the tests still pass.

benjamingeer:
Other things I changed, besides `rdfs:domain` and `rdfs:range`:

* Shorten a lot of `rdfs:comment` strings (and correct the English in them). We don't need such long comments in the ontologies now that we have separate documentation.
* Add copyrights to the ontology files.
* Use the `knora-base` prefix in `knora-dc.ttl` instead of full IRIs.

benjamingeer:
And I took out the statements that imported ontologies into other ontologies. These seemed to be there only because of some limitation of Protégé, and that didn't seem like a good enough reason to keep them.

tobiasschweizer:
We should be sure that when we make a lot changes that they were actually done in a consistent way. 

What about counting all the original occurrences of `rdfs:range` and `rdfs:domain` and comparing them to the occurrences of `subjectClassConstraint` as well as `objectClassConstraint` and `objectDataTypeConstraint`summed up?

Since it is a search and replace operation, the numbers should be the same.

In the actual branch:

```bash
grep "subjectClassConstraint" knora-base.ttl | wc -l
```

On development branch:

```bash
grep "rdfs:domain" knora-base.ttl | wc -l
```

As a reviewer, it is my job to mistakes and inconsistencies. But I cannot go through every line you changed.

benjamingeer:
> What about counting all the original occurrences... Since it is a search and replace operation, the numbers should be the same.

It wasn't just a search and replace operation. I had to look at every case manually, and I corrected a lot of things that were either incorrect or impractical, so just counting the changes won't work. For example, I removed `owl:unionOf`, which was used like this:

```
                   rdfs:domain [ rdf:type owl:Class ;
                                 owl:unionOf ( :Resource
                                               :User
                                               :Value
                                             )
                               ] .
```

I removed these because I think they probably won't work with constraint checking. Once constraint checking works, I can try to put them back.

I've just checked through all the changes in `knora-base` manually again.

These properties had `owl:unionOf` in `rdfs:domain`, which I removed:

* `attachedToProject`
* `attachedToUser`
* `hasPermission`
* `hasRestrictedViewPermission`
* `hasViewPermission`
* `hasModifyPermission`
* `hasDeletePermission`
* `hasChangeRightsPermission`
* `seqnum`
* `isDeleted`
* `preferredLanguage`

These properties had `owl:unionOf` in `rdfs:range`, which I removed:

* `hasValue`
* `hasExtResValue`

The property `deleteDate` had no `rdfs:domain`, but the comment says it only applies to resources, so I gave it a `subjectClassConstraint`.

The property `phone` had no `rdfs:range`, but is clearly a string, so I gave it an `objectDatatypeConstraint`.

These properties had an `rdfs:domain` of `owl:Thing`, which is meaningless, so I removed it:

* `description`
* `name`

The property `protocol` had a list of literals as an rdfs:range. I don't think this can work with constraint checking, and I don't see the point of having a list of all possible Internet protocols in the ontology, so I replaced it with an `objectDatatypeConstraint` of `xsd:string`.

Other changes:

I changed `owl:qualifiedCardinality`, `owl:minQualifiedCardinality`, and `owl:maxQualifiedCardinality` to `owl:cardinality`, `owl:minCardinality`, and `owl:maxCardinalities`, because Knora doesn't support qualified cardinalities.

I changed some cardinalities that were clearly wrong (e.g. in `LinkValue`, where `subject` and `object` were optional but should be required).

I removed `knora-base:mimetypes`, which was taking a huge amount of space in the ontology, but isn't used in the application.

I removed the `root` user, which doesn't belong in `knora-base`.

> As a reviewer, it is my job to mistakes and inconsistencies. But I cannot go through every line you changed.

What do you think I do when I review your changes? :)

tobiasschweizer:
Ok, let me have a final look at it :-)

tobiasschweizer:
In `knora-base.tex`, the examples still use `rdfs:domain` in chapter 3.

tobiasschweizer:
Otherwise, the changes seem fine to me. In the Scala code, you simply had to replace one constant by another and adapt method, parameter and variable names.

benjamingeer:
> In `knora-base.tex`, the examples still use `rdfs:domain` in chapter 3.

Thanks, fixed now.

Issue #40, opened 2016-02-11
Password checking does not seem to work (authentication)


tobiasschweizer:
When I submit credentials (username and password), the password does not seem to be checked correctly. Only the existence of the username seems to be checked.

This happens both with HTTP header auth. and when the params are set in the ULR (Get params).

subotic:
I have started working on a spec for testing authentication in the wip/authentication branch. Both tests in 89cf2cd6f633a7493a30ba2101ddd4ad935deee6 pass. I will add more tests until I catch your problem.

subotic:
Fixed in 0e900a2eddf6974beb0c9d41e861fc67431ebbb4

Issue #13, opened 2016-01-28
Userdata includes the user's password hash


benjamingeer:
When `authenticator.getUserProfileV1` returns a `UserProfileV1`, that object includes a `UserDataV1` containing a hash of the user's password. This hash can easily end up being returned to the client in an API response. Can we remove it from `UserDataV1`?

subotic:
In 0e900a2eddf6974beb0c9d41e861fc67431ebbb4 I have changed the Authenticator trait so that the UserProfileV1 returned to the responders does not contain any sensitive information (token, password).

PR #51, opened 2016-02-25
Wip/authentication

subotic:
Ben,

As far as I can tell the authentication issue is fixed. I also added tests that catch the error.

These tests should pass, but there seems to be another problem with the returned HTTP status code. The code that I expect is 401 (as with all BadCredentialExceptions) but 400 is returned. I cannot find where the problem is. Can you maybe take a look?

The test in question is ```AuthenticationV1E2ESpec.scala```.

### DIFF
--- a/KnoraBuild.sbt
+++ b/KnoraBuild.sbt
@@ -150,7 +150,7 @@ lazy val webApiLibs = Seq(
     "org.joda" % "joda-convert" % "1.8",
     // testing
     "com.typesafe.akka" %% "akka-testkit" % "2.4.0" % "test, fuseki, fuseki-tomcat, graphdb, tdb",
-    "org.scalatest" %% "scalatest" % "2.2.5" % "test, fuseki, fuseki-tomcat, graphdb, tdb",
+    "org.scalatest" %% "scalatest" % "2.2.6" % "test, fuseki, fuseki-tomcat, graphdb, tdb",
     "io.spray" %% "spray-testkit" % "1.3.3" % "test, fuseki, fuseki-tomcat, graphdb, tdb"
 )
 
--- a/webapi/src/main/resources/logback.xml
+++ b/webapi/src/main/resources/logback.xml
@@ -35,7 +35,7 @@
 
     <logger name="org.knora" level="DEBUG"/>
     <logger name="org.knora.webapi" level="DEBUG"/>
-    <logger name="org.knora.webapi.util.authentication" level="INFO"/>
+    <logger name="org.knora.webapi.util.authentication" level="DEBUG"/>
     <logger name="org.knora.webapi.util.cache" level="INFO"/>
     <logger name="org.knora.webapi.store.StoreManager" level="INFO"/>
     <logger name="org.knora.webapi.store.triplestore" level="INFO"/>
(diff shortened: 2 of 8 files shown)


subotic:
Contains fixes for #13 and #40

benjamingeer:
I'm looking into the 400 vs. 401 error now, but in the meantime, I have a suggestion: [let's not use SHA-1](https://en.wikipedia.org/wiki/SHA-1).

> SHA-1 is no longer considered secure against well-funded opponents. In 2005, cryptanalysts found
> attacks on SHA-1 suggesting that the algorithm might not be secure enough for ongoing use, and
> since 2010 many organizations have recommended its replacement by SHA-2 or SHA-3.
> Microsoft, Google and Mozilla have all announced that their respective browsers will
> stop accepting SHA-1 SSL certificates by 2017.

subotic:
Yes, I agree. The only reason why we use SHA-1 is that the demo data comes like this. Beside SHA-1 we have already implemented BCrypt. According to this [blogpost](https://paragonie.com/blog/2016/02/how-safely-store-password-in-2016) BCrypt seems to be acceptable. Apparently Argon2 is now the top dog. The libsodium library that they recommend is written in C, so we would need to find a Java/Scala implementation of Argon2 or write it ourselves.

benjamingeer:
There's a Java binding for the C implementation of Argon2, which runs on Linux and Mac OS X:

https://github.com/phxql/argon2-jvm

subotic:
Cool. Shall we then use Argon2? On the other hand, this will make using Knora a bit more difficult, i.e. this library would need to be installed first. Maybe for now BCrypt is more than sufficient?

benjamingeer:
I guess if bcrypt is good enough for BSD it's good enough for us. But:

http://blog.ircmaxell.com/2012/12/seven-ways-to-screw-up-bcrypt.html

subotic:
Very interesting blog post, thanks.
At the moment we don't actually have a way of creating new users. Maybe I can start a route and extend the users responder so that we can actually create new users?

benjamingeer:
I fixed the incorrect HTTP status code, it was a bug in `ApiStatusCodesV1.fromException`. `BadCredentialsException` is a subclass of `RequestRejectedException`, but `RequestRejectedException` was first in the match-case expression, so `BadCredentialsException` was never matched. Now you should get the 401 status code.

subotic:
cool, thanks :-)

benjamingeer:
> Maybe I can start a route and extend the users responder so that we can actually create new users?

Good idea!

This pull request looks good to me, you can go ahead and merge it.

subotic:
Thanks. I will clean up first (log levels, etc.) and then merge.

subotic:
I have added a few more test cases surrounding cookie/session handling and dialed down the log levels. If it's ok, I will merge it.

benjamingeer:
Looks good, could you just add Scaladoc comments to `SessionResponse` and `MyJsonSessionResponseProtocol` to explain what they are, and maybe think of a clearer name for the latter? (Why is it called `My` something?) Then you can go ahead and merge.

subotic:
Better?

benjamingeer:
Great, thanks! Looks good to merge now.

subotic:
fixes #13 
fixes #40

PR #52, opened 2016-02-25
test: Ensure that empty strings aren't used as IRIs in tests

benjamingeer:
The resources responder and values responder test specs include a number of tests that generate IRIs that are used in subsequent tests. These IRIs have just been stored as strings in mutable variables, initialised to the empty string. A problem I've encountered is that if one test fails, a subsequent test may try to use an empty string as an IRI, resulting in an unbound SPARQL variable, which leads to a very busy triplestore, a massive unwanted query result, and a confusing delay before the test fails.

This pull request replaces those mutable variables with a little class called `MutableTestIri`, which has `get` and `set` methods that throw exceptions if the IRI is an empty string, is invalid, or hasn't been set when it's needed. So if a test fails and doesn't set an IRI, a subsequent test that needs the IRI will fail gracefully.


tobiasschweizer:
Why did we have to check for an empty resource Iri in `getResourceInfoV1` and what did you replace this behaviour with? Was this just there for the test cases (in case the previous operation failed which was supposed to return a valid resource Iri) and for the live cases we would have checked for a valid resource Iri already in the route?

If it was just there for the test cases, we do not need it anymore, and so it is correct as it is now. 

In general, we should avoid putting code in the responders that handle cases that only occur in tests.

tobiasschweizer:
In `ResourcesResponderV1Spec.scala`, couldn't you make `private var newResourceIri = new MutableTestIri` (line 186) a `val` as it points always to the same instance (immutable).

tobiasschweizer:
Otherwise it seems fine to me. This is a much better than we had before.

benjamingeer:
> Why did we have to check for an empty resource Iri in getResourceInfoV1 and what did you replace this behaviour with? Was this just there for the test cases

Yes, I had just added it yesterday. :)

> In general, we should avoid putting code in the responders that handle cases that only occur in tests.

Yes, it was a hack and it bothered me, that's why I made this pull request. :)

> In `ResourcesResponderV1Spec.scala`, couldn't you make `private var newResourceIri = new MutableTestIri` (line 186) a `val` as it points always to the same instance (immutable).

Yes, thanks, fixed.

tobiasschweizer:
great, this was fast

PR #56, opened 2016-02-29
Wip/setup fake responders

tobiasschweizer:
Covenient way to provide mock responders so tests are self contained.


benjamingeer:
Looks good. Please just comment the Python test script (what it's for, why it's in Python and not in Scala) and then you can merge.

PR #58, opened 2016-03-01
Salsah Integration

subotic:
I added a script for loading the ```images-demo-data``` as this kind of data is simpler and should be easier to get working first.

### DIFF
--- a/webapi/_test_data/demo_data/images-demo-data.ttl
+++ b/webapi/_test_data/demo_data/images-demo-data.ttl
@@ -52597,15 +52597,15 @@
 
 <http://data.knora.org/users/c266a56709> rdf:type knora-base:User ;
 
-                                         knora-base:userid "dokubib01" ;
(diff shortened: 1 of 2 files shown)


subotic:
I tried SALSAH out with the images-demo-data but there seems to be a problem with the data. If you load the data ```images-demo-data``` and search for "3 Flugzeuge" in SALSAH you will get one search results. Clicking on it, the webapi returns a 500 Internal Server Error. The log in webapi contains this:

```
webapi 2016-03-01 22:06:58,751 - ERROR - akka.dispatch.Dispatcher - webapi-akka.actor.default-dispatcher-32 - akka.dispatch.Dispatcher - Subject http://data.knora.org/d208fb9357d5/reps/794c6ca8972f01 has no owner
webapi org.knora.webapi.InconsistentTriplestoreDataException: Subject http://data.knora.org/d208fb9357d5/reps/794c6ca8972f01 has no owner
```

PR #59, opened 2016-03-01
SALSAH Integration

subotic:
I added a script for loading the ```images-demo-data``` as this kind of data is simpler and should be easier to get working first in SALSAH.


subotic:
I tried SALSAH out with the images-demo-data but there seems to be a problem with the data. If you load the ```images-demo-data``` and search for "3 Flugzeuge" in SALSAH you will get one search results. Clicking on it, the ```webapi``` returns a 500 Internal Server Error. The log in webapi contains this:

```
webapi 2016-03-01 22:06:58,751 - ERROR - akka.dispatch.Dispatcher - webapi-akka.actor.default-dispatcher-32 - akka.dispatch.Dispatcher - Subject http://data.knora.org/d208fb9357d5/reps/794c6ca8972f01 has no owner
webapi org.knora.webapi.InconsistentTriplestoreDataException: Subject http://data.knora.org/d208fb9357d5/reps/794c6ca8972f01 has no owner
```

benjamingeer:
The error message is correct: the file values in `images-demo-data` have no `knora-base:attachedToUser` and no permissions. This is because of bugs in the export script, which have since been fixed. Since there are 200 file values with the same problem in `images-demo-data`, changing them manually looks cumbersome. How hard would it be to regenerate `images-demo-data` using the data from the export script?

subotic:
> How hard would it be to regenerate images-demo-data using the data from the export script?

Probably much easier than adding the missing triples by hand :-) I will recreate the ```images-demo-data```.

benjamingeer:
I think you can merge this. I'll try to check `images-demo-onto` and `images-demo-data` with the GraphDB consistency checker and fix any remaining problems on the `wip/consistency-checking` branch.

subotic:
I will delete this branch and then again create a new one, so that I can try to fix the bug in salsah.

PR #61, opened 2016-03-02
Wip/sipi values responder

tobiasschweizer:
Add support for the change of existing file values.

The main difficulty is that file values are stored like ordinary values in Knora, but they were handled quite differently in the old SALSAH - as locations. In consequence, we do not know the Iris of file values from a resource's response.

Because of this, a Put request has to deliver also the Iri of the resource in question, so the current file value objects can be retrieved. Then, this is treated like any other value change by calling `changeValueV1` in values responder.

For more explanation, please read `sipi-and-knora.rst`.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
@@ -1005,7 +1005,7 @@ class ResourcesResponderV1 extends ResponderV1 {
 
                         // check if the file type returned by Sipi corresponds to the expected fileValue property in resourceClassInfo.fileValueProperties.head
                         _ = if (SipiConstants.fileType2FileValueProperty(sipiResponse.file_type) != resourceClassInfo.fileValueProperties.head) {
-                            // TODO: remove the file from SIPI
+                            // TODO: remove the file from SIPI (delete request)
                             throw BadRequestException(s"Type of submitted file (${sipiResponse.file_type}) does not correspond to expected property type ${resourceClassInfo.fileValueProperties.head}")
                         }
 
--- a/webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
@@ -39,6 +39,7 @@ import org.knora.webapi.util.ActorUtil._
 import org.knora.webapi.util._
 
 import scala.concurrent.{Future, Promise}
+import scala.util.Try
 
 /**
   * Responds to requests for information about resources, and returns responses in Knora API v1 format.
@@ -880,6 +881,7 @@ class ResourcesResponderV1 extends ResponderV1 {
       * @return a [[ResourceCreateResponseV1]] informing the client about the new resource.
       */
     private def createNewResource(resourceClassIri: IRI, label: String, values: Map[IRI, Seq[CreateValueV1WithComment]], sipiConversionRequest: Option[SipiResponderConversionRequestV1] = None, projectIri: IRI, userProfile: UserProfileV1, apiRequestID: UUID): Future[ResourceCreateResponseV1] = {
+
         /**
           * Implements a pre-update check to ensure that an [[UpdateValueV1]] has the correct type for the `knora-base:objectClassConstraint` of
           * the property that is supposed to point to it.
@@ -1003,6 +1005,7 @@ class ResourcesResponderV1 extends ResponderV1 {
 
                         // check if the file type returned by Sipi corresponds to the expected fileValue property in resourceClassInfo.fileValueProperties.head
                         _ = if (SipiConstants.fileType2FileValueProperty(sipiResponse.file_type) != resourceClassInfo.fileValueProperties.head) {
+                            // TODO: remove the file from SIPI
                             throw BadRequestException(s"Type of submitted file (${sipiResponse.file_type}) does not correspond to expected property type ${resourceClassInfo.fileValueProperties.head}")
                         }
 
@@ -1013,16 +1016,14 @@ class ResourcesResponderV1 extends ResponderV1 {
                 } else {
                     // resource class requires no binary representation
                     // check if there was no file sent
+                    // TODO: in all cases of an error, the tmp file has to be deleted
                     sipiConversionRequest match {
                         case None => Future(None) // expected behaviour
                         case Some(sipiConversionFileRequest: SipiResponderConversionFileRequestV1) =>
                             throw BadRequestException(s"File params (GUI-case) are given but resource class $resourceClassIri does not allow any representation")
                         case Some(sipiConversionPathRequest: SipiResponderConversionPathRequestV1) =>
-                            // a tmp file has been created by the resources route, delete it (SipiResponder was not called, so do it here)
-                            InputValidation.deleteFileFromTmpLocation(sipiConversionPathRequest.source)
                             throw BadRequestException(s"A binary file was provided (non GUI-case) but resource class $resourceClassIri does not have any binary representation")
                     }
-
                 }
 
                 // Everything looks OK, so we can create an empty resource and add the values to it. This must
@@ -1094,19 +1095,17 @@ class ResourcesResponderV1 extends ResponderV1 {
             } yield apiResponse
         }
 
-        val maybeUserIri: Option[IRI] = userProfile.userData.user_id
-        // TODO: remove dummy user after testing!
-        //val maybeUserIri: Option[IRI] = Some("http://data.knora.org/users/91e19f1e01")
-
-        val namedGraph = settings.projectNamedGraphs(projectIri).data
-        val resourceIri: IRI = knoraIriUtil.makeRandomResourceIri
-
-        for {
+        val resultFuture = for {
         // Don't allow anonymous users to create resources.
-            userIri <- maybeUserIri match {
-                case Some(iri) => Future(iri)
-                case None => Future.failed(ForbiddenException("Anonymous users aren't allowed to create resources"))
+            userIri: IRI <- Future {
+                userProfile.userData.user_id match {
+                    case Some(iri) => iri
+                    case None => throw ForbiddenException("Anonymous users aren't allowed to create resources")
+                }
             }
+
+            namedGraph = settings.projectNamedGraphs(projectIri).data
+            resourceIri: IRI = knoraIriUtil.makeRandomResourceIri
 
             // check if the user has the permissions to create a new resource in the given project
             // get project info that includes the permissions the current user has on the project
@@ -1150,6 +1149,21 @@ class ResourcesResponderV1 extends ResponderV1 {
                 )
             )
         } yield result
+
+        // If a temporary file was created, ensure that it's deleted, regardless of whether the request succeeded or failed.
+        resultFuture.andThen {
+            case _ =>
+                sipiConversionRequest match {
+                    case Some(conversionRequest) =>
+                        conversionRequest match {
+                            case (conversionPathRequest: SipiResponderConversionPathRequestV1) =>
+                                // a tmp file has been created by the resources route (non GUI-case), delete it
+                                InputValidation.deleteFileFromTmpLocation(conversionPathRequest.source, log)
+                            case _ => ()
+                        }
+                    case None => ()
+                }
+        }
     }
 
     /**
--- a/webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
@@ -349,8 +349,7 @@ class ValuesResponderV1 extends ResponderV1 {
     }
 
     /**
-      * Preprocesses a file value change request by calling the Sipi responder to create a new file
-      * and calls [[changeValueV1]] to actually change the file value in Knora.
+      * Adds a new version of an existing file value.
       *
       * @param changeFileValueRequest a [[ChangeFileValueRequestV1]] sent by the values route.
       * @return a [[ChangeFileValueResponseV1]] representing all the changed file values.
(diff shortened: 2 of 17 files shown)


webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
```diff
@@ -1003,6 +1005,7 @@ class ResourcesResponderV1 extends ResponderV1 {
 
                         // check if the file type returned by Sipi corresponds to the expected fileValue property in resourceClassInfo.fileValueProperties.head
                         _ = if (SipiConstants.fileType2FileValueProperty(sipiResponse.file_type) != resourceClassInfo.fileValueProperties.head) {
+                            // TODO: remove the file from SIPI (delete request)
```
benjamingeer: Why is there still a TODO to delete the file from Sipi? I thought this was done by exception handling.

webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
```diff
@@ -345,6 +349,130 @@ class ValuesResponderV1 extends ResponderV1 {
...
+            }
+
+            // the message to be sent to Sipi responder
+            sipiConversionRequest: SipiResponderConversionRequestV1 = changeFileValueRequest.file
+
+            sipiResponse: SipiResponderConversionResponseV1 <- (responderManager ? sipiConversionRequest).mapTo[SipiResponderConversionResponseV1]
+
+            // check if the file type returned by Sipi corresponds to the already existing file value type (e.g., hasStillImageRepresentation)
+            _ = if (SipiConstants.fileType2FileValueProperty(sipiResponse.file_type) != fileValues.head.property) {
+                // TODO: remove the file from SIPI (delete request)
```
benjamingeer: Why is this TODO still here? I thought this was taken care of by exception handling.

webapi/src/main/scala/org/knora/webapi/routing/v1/ValuesRouteV1.scala
```diff
@@ -195,12 +200,44 @@ object ValuesRouteV1 extends Authenticator {
...
+
+    }
+
     def rapierPath(_system: ActorSystem, settings: SettingsImpl, log: LoggingAdapter): Route = {
         implicit val system = _system
         implicit val executionContext = system.dispatcher
         implicit val timeout = settings.defaultTimeout
         val responderManager = system.actorSelection("/user/responderManager")
 
+        // TODO: add documentation
```
benjamingeer: Please add the documentation mentioned in this TODO.

webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
```diff
@@ -345,6 +349,130 @@ class ValuesResponderV1 extends ResponderV1 {
...
     /**
+      * Temporary structure to represent existing file values of a resource.
+      *
+      * @param property       the property Iri (e.g., hasStillImageFileValueRepresentation)
+      * @param valueObjectIri the Iri of the value object.
+      * @param quality        the quality of the file value
+      */
+    private case class CurrentFileValue(property: IRI, valueObjectIri: IRI, quality: Option[Int])
+
+    private def changeFileValueV1(changeFileValueRequest: ChangeFileValueRequestV1): Future[ChangeFileValueResponseV1] = {
```
benjamingeer: What happens if two requests try to update the same file at the same time? In other update scenarios, the client has to send the IRI of the previous version of the value, and if someone else has already changed it, we reject the request. But we can't do that here, because the client don't know the IRI of the file value. This is something we need to fix in version 2 of the API. In the meantime, what does Sipi do if it receives 2 simultaneous requests to change the same file? Is there a risk that the file could be corrupted? To be on the safe side, I think this method should acquire a resource lock.

webapi/src/main/twirl/queries/sparql/v1/addValueVersion.scala.txt
```diff
@@ -200,6 +200,7 @@ DELETE {
             ?newValue <@assertion._1> <@assertion._2> .
         }
 
+        @* TODO: in case of a file value, order is not given *@
```
benjamingeer: Isn't this TODO done already?

benjamingeer:
Could you also add a test to ensure that the temp file is deleted if an error occurs?

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
tobiasschweizer: This case is meant to delete a file after it has already been successfully created by Sipi, because when looking at its type we find out that it is not what Knora expects (e.g., we expect an image but Sipi create an audio file).

I have to implement a special route in Sipi to handle delete requests.

webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
tobiasschweizer: see my comment above

webapi/src/main/scala/org/knora/webapi/routing/v1/ValuesRouteV1.scala
tobiasschweizer: Haha, actually this todo was meant for you because I did not get how this is supposed to work without additional information (the client has to submit three Iris) :-)

webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
tobiasschweizer: Actually I thought that Sipi would just create a new file that would kept track of by Knora (the already existing files would still be referenced by previous versions of the file value). 

Does Sipi actually has to know that files are related to one another as versions? I just stores and serves files. Also, a new version of a file value could be totally different binary.

webapi/src/main/twirl/queries/sparql/v1/addValueVersion.scala.txt
tobiasschweizer: No, the thing we dealt with yesterday was the property's `guiOrder`, not `valueHasOrder`

tobiasschweizer:
Good point with the testing for the deletion of the tmp file. I will do that!

webapi/src/main/scala/org/knora/webapi/routing/v1/ValuesRouteV1.scala
benjamingeer: Are you asking how `Segments` works? I think I just found it in the Spray docs. It represents a collection of URL path segments IIRC.

tobiasschweizer: `makeVersionHistoryRequestMessage` expects a set of Iris. Ii is implicitly documented in the error handling:

```scala
if (iris.length != 3) throw BadRequestException("Link value request requires subject IRI, predicate IRI, and object IRI")
```

webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
tobiasschweizer: By the way, I will move the case class `CurrentFileValue` inside  `changeFileValueV1`, as you suggested yesterday.

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
benjamingeer: OK, can you open an issue for this? It's easy to lose/forget TODOs.

webapi/src/main/scala/org/knora/webapi/routing/v1/ValuesRouteV1.scala
benjamingeer: OK so now you know enough to add a comment explaining it. :)

webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
benjamingeer: OK

webapi/src/main/twirl/queries/sparql/v1/addValueVersion.scala.txt
benjamingeer: OK, but what can we do about it? Can you make an issue for this?

benjamingeer:
> Actually I thought that Sipi would just create a new file that would be kept track of by Knora (the already existing files would still be referenced by previous versions of the file value).

OK. But is it possible for two users to have any sort of edit conflict when adding or changing files in a resource? Please think through all possible scenarios.

> Does Sipi actually has to know that files are related to one another as versions? I just stores and serves files. Also, a new version of a file value could be totally different binary.

We had talked about having Sipi store the version number in the filename, to make it possible to write a script that would copy the most recent version of each file directly from the filesystem, without having to query the triplestore. But I've forgotten what we decided. Can you please confirm with Lukas, and then write something in the RST docs about it?

tobiasschweizer:
Regarding testing ig the tmp files was deleted: I keep track of the path of a tmp file in MockSipiResponder, so I can check in the test.

In `SipiV1E2ESpec` I added a test case that will fail. It is currently commented out because it is not handled correctly by test route's `check`. In `ValuesResponderV1Spec`, you checked for failures.

Do you know how to do that in the test route?

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
tobiasschweizer: https://github.com/dhlab-basel/Knora/issues/67

webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
tobiasschweizer: see issue mentioned above

webapi/src/main/twirl/queries/sparql/v1/addValueVersion.scala.txt
tobiasschweizer: https://github.com/dhlab-basel/Knora/issues/68

tobiasschweizer:
> OK. But is it possible for two users to have any sort of edit conflict when adding or changing files in a resource? Please think through all possible scenarios.

I think if Sipi just adds files and Knora keeps track of the filenames, possible conflicts can be prevented by locking Knora resources as we do now.

If we decide that the internal names Sipi gives to files reflect the actual version history of Knora, the thing may get more complicated. If a user changes the file value of a resource, Sipi must not be asked to add another file for the same resource until the first user's request has completed.

tobiasschweizer:
> We had talked about having Sipi store the version number in the filename, to make it possible to write a script that would copy the most recent version of each file directly from the filesystem, without having to query the triplestore. But I've forgotten what we decided. Can you please confirm with Lukas, and then write something in the RST docs about it?

You are totally right: we talked about it but did not take a decision. Currently, I am working with Lukas on Sipi file handling using Lua scripts. I will keep you posted how it goes.

benjamingeer:
> I think if Sipi just adds files and Knora keeps track of the filenames, possible conflicts can be prevented by locking Knora resources as we do now (it is not different from other requests).

But your `changeFileValue` method doesn't lock the resource before calling Sipi.

tobiasschweizer:
That's correct. However, the resource is not changed in `changeFileValueV1`, but in `changeValueV1` which locks the resource.

The problem could be that the value Iris we got for the resource are not up to date anymore when `changeValueV1` is executed.

So I should use locking for `changeFileValueV1`, right?

benjamingeer:
> So I should use locking for `changeFileValueV1`, right?

That seems safest to me. The locks are reentrant, so it's OK to acquire the same lock twice for the same API request ID and then release it twice.

tobiasschweizer:
Perfect! I was worrying exactly about this since `changeFileValueV1` would call `changeValueV1` with the same UUID.

tobiasschweizer:
If you have some time left, could you look at the test I wrote? Thx.

> Regarding testing if the tmp files was deleted: I keep track of the path of a tmp file in MockSipiResponder, so I can check in the test.

> In SipiV1E2ESpec I added a test case that will fail. It is currently commented out because it is not handled correctly by test route's check. In ValuesResponderV1Spec, you checked for failures.

> Do you know how to do that in the test route?

benjamingeer:
> Do you know how to do that in the test route?

I'll look at this tomorrow.

tobiasschweizer:
- tests fail when I want them to :-)
- I acquire also a lock for `changeFileValueV1` re-using the same Id when calling `changeValueV1`

benjamingeer:
Great! Have you and Lukas decided whether Sipi needs to know about versioning?

tobiasschweizer:
Lukas and I briefly talked about it yesterday.

If we do support versioning information in filenames, Sipi would have to know about concepts we use in Knora in the sense that when we change a file value in Knora, we would reuse the same base name in Sipi for the new file to be created and append an incrementing number. Of course, you would have to tell Sipi which base name to use when you call the route.

I think that this somehow contradicts the separation of responsibilities between Knora and Sipi: Knora cares about metadata and annotations and Sipi takes care of binaries.

But of course, if Knora is not available, you would have no idea how files relate to one another if the filenames do not tell you about versions.

To summarise, reflecting file value versions in filenames would make the Knora-Sipi interaction harder to handle. This makes only sense if we decide that it is important to know about how files relate to each other looking at Sipi only. This **could** be important since we are the DaSCH :-) It is actually one of Luaks's paradigms: never separate files from metadata.

CC @lrosenth

tobiasschweizer:
if you want to support file versions in Sipi, Sipi would have to knwo about resource locks in Knora

benjamingeer:
> To summarise, reflecting file value versions in filenames would make the Knora-Sipi interaction harder to handle. This makes only sense if we decide that it is important to know about how files relate to each other looking at Sipi only. This could be important since we are the DaSCH :-) It is actually one of Luaks's paradigms: never separate files from metadata.

But Knora does separate files from metadata: only the triplestore knows that image `xyz.jpg` is an image of page 236 of *Zeitglöcklein des Lebens*.

Even if Knora is not available, you still have the triplestore and can do SPARQL queries yourself to find out which is the most recent version of the file.

The best use case I can think of is if you wanted to write a script that could mirror the most recent version of all the files. Querying Knora for each file would be slow.

benjamingeer:
> if you want to support file versions in Sipi, Sipi would have to knwo about resource locks in Knora

Why?

tobiasschweizer:
Let's talk about this on Skype this afternoon.

Do you think that this branch can be merged back to `develop`?

benjamingeer:
OK, please go ahead and merge this branch.

Issue #63, opened 2016-03-03
Shall I add .sbtrc to the Gitrep?


tobiasschweizer:
Shall I add `.sbtrc` to the gitrep?

When starting `sbt` from the console, it automatically switches to the project `webapi`.

```bash
alias boot = ;reload ;project webapi ;iflast shell
```

CC: @subotic @benjamingeer

subotic:
Please don't. If someone wants it, he or she can easily do it locally.

tobiasschweizer:
ok :-)

Issue #45, opened 2016-02-12
How will Knora or Sipi find resource class icons?


benjamingeer:
The definition of `knora-base:LinkObj` says:

```
         :resourceIcon "link.gif"^^xsd:string ;
```

How can this work? Should it be changed to something else?

tobiasschweizer:
The icons can all be served directly by Sipi, using a IIIF URL.

We just need to be sure that every icon that is used by Knora exists on the Sipi server. We could use a special prefix for that.

subotic:
Strictly speaking, aren't ressource class icons something that the gui (Salsah) would need to take care of? Salsah can have a default mapping of ressource classes to icons, which then could be changed in the project specific settings.

tobiasschweizer:
I think this is a good way to do it! So then we would place the icon files in the salsah-app?

subotic:
Yes, that's what I was thinking. For now we could create a folder under ```app/icons``` and put all resource class icons inside. Then somewhere in Salsah, we will need to define the mappings between the classes and the icons. Later on, these default icons could then be changed by adding custom project specific themes to Salsah, or something like this.

tobiasschweizer:
Now, I added a function that creates URLs for resource icons: <https://github.com/dhlab-basel/Knora/commit/e2a141b45a9207bbed0865860eb8331eb1330cbd>

It takes the Iri of a resource class and gets out the name of the ontology it was defined in. Then it is expected that the actual icon is in a folder with that name in Salsah's icon project folder: http://localhost:3335/project-icons/incunabula/page.gif

PR #72, opened 2016-03-04
Wipi/generate IIIF urls

tobiasschweizer:
Now, true IIIF URLs are generated

What needs to be changed in the future, is that we do not include the filename in the URL but the file value IRI (we need to pass the file value Iri to the function from resources responder)

Sipi will then ask Knora about the permissions and the filename for a given file value Iri.


benjamingeer:
Can you please change the search responder so it uses these URLs, too?

According to #63, we're not adding `.sbtrc` to the repository. If you did this by mistake, you could add it to `.gitignore` instead.

tobiasschweizer:
Ok, I added the handling for previews returned by the search responder

I have already added `.sbtrc` to git `.gitignore` on this branch

tobiasschweizer:
This si still open: https://github.com/dhlab-basel/Knora/issues/45

benjamingeer:
Sorry, my mistake about `.sbtrc`.

OK, looks good to merge.

tobiasschweizer:
Currently the search only looks for stillimagerepresentations, but maybe also a movie could have a preview, e.g a single frame.

tobiasschweizer:
I created issues #74 and #73 so I do not forget :-)

PR #87, opened 2016-03-15
Wip/sipi knora communication

tobiasschweizer:
I changed the Sipi responder so it receives a request from Sipi containing the name of the file the permissions shall be returned for. Given the file name, Knora looks for the corresponding file value Iri to return the permissions.

In Salsah, I added an extra request that is sent to Sipi when the login to Knora has been made. This will set a second cookie that contains the Knora session id. Like this, Sipi can attach the Knora session id when asking Knora about the user's permission on a file. Without this second cookie, Knora could not find out which user is making request.

For security reasons, we should use SSL in the productive version of Knora and make cookies HTTPOnly (so they cannot be read by running JS `document.cookie`).

### DIFF
--- a/salsah/src/public/index.html
+++ b/salsah/src/public/index.html
@@ -179,6 +179,21 @@
 
             SALSAH.ApiPost('session', {username: $('#user_id').val(), password: $('#password').val()}, authorization, function(data) {
 				if (data.status == ApiErrors.OK) {
+					console.log("session created with id " + data.sid);
+					$.ajax({
+						url: "http://localhost:1024/Knora_login",
+						type: "POST",
+						data: {
+							sid: data.sid
+						},
+						xhrFields: {
+							withCredentials: true
+						},
+						success: function(data) {
+							console.log("from Sipi: " + data.status)
+						}
+					})
+
 					$('#dologin').simpledialog('loginbox', 'close');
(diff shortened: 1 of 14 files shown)


salsah/src/public/index.html
```diff
@@ -179,6 +179,21 @@
 
             SALSAH.ApiPost('session', {username: $('#user_id').val(), password: $('#password').val()}, authorization, function(data) {
 				if (data.status == ApiErrors.OK) {
+					console.log("session created with id " + data.sid);
+					$.ajax({
+						url: "http://localhost:1024/Knora_login",
```
benjamingeer: Can we not hard-code the Sipi URL in SALSAH?

benjamingeer: Also, can you add some comments to the JavaScript here to explain what this code is doing?

benjamingeer:
It looks like there's a new `Proxy` class that isn't used for anything?

tobiasschweizer:
I added a new section to the Sipi Knora Interaction `rst` and also comments to the code.

benjamingeer:
OK, please merge this, but we need unit tests in Sipi as soon as possible, to test things like this.

tobiasschweizer:
Yes, absolutely!

Issue #97, opened 2016-03-29
Correct header for the utf8 encoding of JSON responses


tobiasschweizer:
Currently, the HTTP response header of a Knora API request does not contain an encoding header. It should say `content-type: application/json; charset=utf-8`

tobiasschweizer:
```scala
 HttpResponse(
                        status = StatusCodes.OK,
                        entity = HttpEntity(
                            ContentTypes.`text/plain(UTF-8)`,
                            jsonResponseWithStatus.compactPrint
                        )
```

tobiasschweizer:
https://github.com/spray/spray/blob/master/spray-http/src/main/scala/spray/http/HttpEntity.scala

tobiasschweizer:
ok, there does not seem to be apllication/json with UTF-8

tobiasschweizer:
fixed in https://github.com/dhlab-basel/Knora/commit/ccc1bfd8eaa9aef6e7cac557b32d4e05bce88261

subotic:
Can you please only close an issue when it is merged into develop? This commit is in your private branch. You can write in the commit message ``resolves #issueNr`` (see https://help.github.com/articles/closing-issues-via-commit-messages/) then when this commit is merged, the issue will be closed automatically.

Issue #96, opened 2016-03-29
Fix test data


tobiasschweizer:
fix the test data in `develop`. There are a lot of inconsistencies such as missing labels or empty string assignments that cause problems in Scala.

tobiasschweizer:
added missing labels for resource classes and properties in knora-base

Issue #95, opened 2016-03-22
Parsing extended search URL


tobiasschweizer:
For reasons of legacy (SALSAH-prototype), the URL for an extended search looks like this:

`http://localhost:3333/v1/search/?searchtype=extended&property_id[]=http://www.knora.org/ontology/incunabula#seqnum&property_id[]=http://www.knora.org/ontology/incunabula#partOf&compop[]=GT&compop[]=EQ&searchval[]=1&searchval[]=http://data.knora.org/8be1b7cf7103&show_nrows=25&start_at=0&filter_by_restype=http://www.knora.org/ontology/incunabula#page`

Here, square brackets are used to submit several values with the same key. This is supported by jQuery which is used by the SALSAH plugin `jquery.extsearch.js`:

> Object must be Key/Value pairs. If value is an Array, jQuery serializes multiple values with same key based on the value of the traditional setting (described below).
(<http://api.jquery.com/jquery.ajax/>)

This works for the moment. However, when entering or manipulating the URL in the browser's address bar, parsing errors occur:

```
Illegal request, responding with status '400 Bad Request': Illegal request-target, unexpected character '[' at position 43
```

tobiasschweizer:
Solution: deactivate the serialization using square brackets by setting the `traditional` flag to `true` in `01_salsah_api.js`'s `ApiGet`: 

```javascript
	send_params.traditional = true;

	return $.ajax(send_params);
```
This is the " 'shallow' " serialization" which produces this URL:

` http://localhost:3333/v1/search/?searchtype=extended&property_id=http://www.knora.org/ontology/incunabula#pagenum&property_id=http://www.knora.org/ontology/incunabula#partOf&compop=LIKE&compop=EQ&searchval=recto&searchval=http://data.knora.org/5e77e98d2603&show_nrows=25&start_at=0&filter_by_restype=http://www.knora.org/ontology/incunabula#page`

tobiasschweizer:
http://stackoverflow.com/questions/2540827/jquery-parameter-serialization-without-the-bracket-mess/2540897#2540897

tobiasschweizer:
https://github.com/dhlab-basel/Knora/commit/1a37c9c85abe686fe277b46a5c572b9b3641a08c

PR #100, opened 2016-03-30
Wip/gui integration

tobiasschweizer:
Make extended search work with Knora

The main issues are:
- IRIs are not numeric anymore
- valuetypes have changed

A problem is the value type of linking props: unlike before, they consist of the resource class they point to (before this was a specific valuetype and the attributes contained the resource class that could be referred to)


subotic:
resolves #97

subotic:
resolves #95

benjamingeer:
I simplified `OntologyResponderV1` a bit and took care of a couple of TODOs.

About linking props, I think you can change the API responses to return `knora-base:LinkValue` as the `valuetype_id` of link properties. This would actually be more consistent, since we already return the link value's IRI as the value ID. If the resource type of the link target needs to be included, it could be added somewhere else in the API response. This will involve changing `OntologyResponderV1.getResourceTypeResponseV1`, the methods you added to the ontology responder, and the resources responder.

Since you added labels that were missing from resource classes, could you do the same for missing property labels that are currently showing up as `undefined` in the GUI? Then I could have the GraphDB consistency checker make sure that all required labels are present.

benjamingeer:
If every resource is required to have an `rdfs:label`, you could add this as a cardinality to `knora-base:Resource`.

tobiasschweizer:
Thanks for the comments! I will change the `value type_id` as you suggested and put the resource class in the attributes (as in the old SALSAH).

tobiasschweizer:
I will check  for `isLinkProp` to distinguish between linking props and non linking props

```scala
case class PropertyEntityInfoV1(propertyIri: IRI,
                                isLinkProp: Boolean,
                                predicates: Map[IRI, PredicateInfoV1]) extends EntityInfoV1
```

tobiasschweizer:
I had to remove resource class constraints from the GUI attributes in the ontology, the ids were still the ones from the old SALSAH (see #102 ).

tobiasschweizer:
Ok, I added the missing labels.

Could have a final look?

tobiasschweizer:
argh, I still have to add the labels to the test data ...

tobiasschweizer:
ok, it is ready now :-)

benjamingeer:
I'm not finished reading it yet and don't have much more time today, so it might have to wait till Monday.

tobiasschweizer:
That's ok!

I am moving to the resource viewer now ...

webapi/src/main/scala/org/knora/webapi/routing/v1/SearchRouteV1.scala
```diff
@@ -74,6 +76,9 @@ object SearchRouteV1 extends Authenticator {
             case other => None
         }
 
+        //
+        // the params with multiple values have angle brackets appended: param name[]
+        //
```
benjamingeer: Do these params still have brackets appended? I thought the GUI was changed to remove them. Also, [] are called square brackets. Angle brackets are like this: <>.

webapi/src/main/scala/org/knora/webapi/responders/v1/ValueUtilV1.scala
```diff
@@ -513,6 +531,32 @@ class ValueUtilV1(private val settings: SettingsImpl) {
             referenceCount = predicates(OntologyConstants.KnoraBase.ValueHasRefCount).literals.head.toInt
         )
     }
+
+    /** Creates an attribute segment for the Salsah GUI from the given resource class.
+      * Example: if incunabula:book is given, the function returns "restypeid=incunabula:book".
```
benjamingeer: Should this be `incunabula:book` or `http://www.knora.org/ontology/incunabula#book`?

webapi/src/main/scala/org/knora/webapi/responders/v1/ValueUtilV1.scala
```diff
@@ -120,6 +120,24 @@ class ValueUtilV1(private val settings: SettingsImpl) {
...
+      * To the base path, the icon name is appended. In case of a region with the icon name "region.gif",
+      * "http://salsahapp:port/project-icons-basepath/knora-base/region.gif" is returned.
+      *
+      * @param resourceClassIri the Iri of the resource class in question.
+      * @param iconsSrc the name of the icon file.
+      */
+    def makeResourceClassIconURL(resourceClassIri: IRI, iconsSrc: String): IRI = {
+        // get ontology name, e.g. "knora-base" from "http://www.knora.org/ontology/knora-base#Region"
+        // add +1 to ignore the slash
+        val ontologyName = resourceClassIri.substring(resourceClassIri.lastIndexOf('/') + 1, resourceClassIri.lastIndexOf('#'))
```
benjamingeer: This won't work for all IRIs. The last path component of an IRI doesn't have to be a unique name for the ontology. For example, the IRI of `foaf:Person` is `http://xmlns.com/foaf/0.1/Person`; the last path component is `0.1`, which isn't the name of the ontology. Either we need to impose this as a requirement on all ontologies used with Knora (in which case it needs to be documented), or we need to find another way to associate resource classes with ontology names.

webapi/src/main/scala/org/knora/webapi/responders/v1/OntologyResponderV1.scala
```diff
@@ -462,4 +498,240 @@ class OntologyResponderV1 extends ResponderV1 {
...
+                            // It is restricted to the resource class that is given for knora-base:objectClassConstraint
+                            // for the given property which goes in the attributes that will be read by the GUI.
+
+                            PropertyDefinitionV1(
+                                id = propertyIri,
+                                name = propertyIri,
+                                label = entityInfo.getPredicateObject(OntologyConstants.Rdfs.Label),
+                                description = entityInfo.getPredicateObject(OntologyConstants.Rdfs.Comment),
+                                vocabulary = entityInfo.predicates.values.head.ontologyIri,
+                                occurrence = "",
```
benjamingeer: Why is this an empty string? If it's optional, it would be better to use an `Option` if possible. If not, please add a comment to explain.

webapi/src/main/scala/org/knora/webapi/responders/v1/OntologyResponderV1.scala
```diff
@@ -462,4 +498,240 @@ class OntologyResponderV1 extends ResponderV1 {
...
+
+                resInfosForNamedGraph <- Future.sequence(resInfosForNamedGraphFuture)
+
+                resourceTypes: Vector[ResourceTypeV1] = resInfosForNamedGraph.map {
+                    case (resClassIri, resInfo) =>
+
+                        val properties = resInfo.restype_info.properties.map {
+                            (prop) => PropertyTypeV1(
+                                id = prop.id,
+                                label = prop.label.getOrElse("")
```
benjamingeer: Why default to an empty string here? If it's optional, an `Option` would be better if possible. Otherwise, please add a comment to explain.

benjamingeer:
It seems that there are now some assumptions about ontologies that need to be documented, e.g. that each ontology is stored in its own named graph.

benjamingeer:
The problem with empty strings in code is that they provide a false sense of security: they can make you believe you have something when you actually have nothing. Then you might try to use that nothing as if it was something, and get an unexpected string index out of bounds error. From the point of view of defensive programming, it's better to have a `None`, which the compiler forces you to treat as nothing rather than as something.

benjamingeer:
The problem with empty strings in code is that they provide a false sense of security: they can make you believe you have something when you actually have nothing. Then you might try to use that nothing as if it was something, and get an unexpected string index out of bounds error. From the point of view of defensive programming, it's better to have a `None`, which the compiler forces you to treat as nothing rather than as something.

webapi/src/main/scala/org/knora/webapi/routing/v1/SearchRouteV1.scala
tobiasschweizer: You are right, the square brackets are not needed any longer (jQuery traditional mode: https://github.com/dhlab-basel/Knora/commit/1a37c9c85abe686fe277b46a5c572b9b3641a08c).

webapi/src/main/scala/org/knora/webapi/responders/v1/ValueUtilV1.scala
tobiasschweizer: It should be `http://www.knora.org/ontology/incunabula#book`, I have updated the comment

webapi/src/main/scala/org/knora/webapi/responders/v1/ValueUtilV1.scala
tobiasschweizer: Yes, I see the problem for other ontologies than ours.

I put this restriction in the function's comment. For our test data, this requirement is fulfilled. 
For the case of `FOAF` we could provide an ontology like knora-base:dc.

webapi/src/main/scala/org/knora/webapi/responders/v1/OntologyResponderV1.scala
tobiasschweizer: A property's cardinality cannot be set in the context of a named graph because the cardinality is defined for the resource class, and a property might be used in more than one resource class in one named graph.

tobiasschweizer: This is a kind of a special case because normally properties are shown for a given resource class or instance.

webapi/src/main/scala/org/knora/webapi/responders/v1/OntologyResponderV1.scala
tobiasschweizer: I think a `label` should not be optional, but required in the ontology. Maybe we should better throw an error if label is not given for a resource class or a property.

tobiasschweizer:
> It seems that there are now some assumptions about ontologies that need to be documented, e.g. that each ontology is stored in its own named graph.

I made myself an issue for this

benjamingeer:
It looks OK now, you can merge it when you've finished writing the tests.

webapi/src/main/scala/org/knora/webapi/responders/v1/ValueUtilV1.scala
subotic: We would generally need to rethink this. Here we (the webapi) are generating a Salsah URL. Shouldn't the ``webapi`` be GUI agnostic?

subotic:
>> It seems that there are now some assumptions about ontologies that need to be documented, e.g. that each ontology is stored in its own named graph.

> I made myself an issue for this

Is it #105 ?

tobiasschweizer:
Yes, that's the issue!

subotic:
resolves #96

PR #107, opened 2016-04-05
User, Project, and Group Management

### LINKED ISSUE #54: Add user, group, and project management capabilities
Need to add a route and extend the user responder.

### LINKED ISSUE #79: Redesign default permissions
Currently the permissions for a a new resource come from the default permissions attached to the resource class, and permissions for a new value come from default permissions attached to the property definition. But this won't work for classes and properties that are defined only in `knora-base` and not extended in project-specific ontologies.

@lrosenth and I discussed this today and agreed to get rid of class-specific and property-specific default permissions. Instead, each user will be able to configure a set of default permissions for all new resources, and a set of permissions for all new values. If the user doesn't have any such defaults configured, we'll fall back to defaults attached to the project.

subotic:
### Summary
- f: add ``v1/users`` API endpoint
- f: add ``v1/projects`` API endpoint
- f: add ``v1/groups`` API endpoint
- f: add exemplary admin GUI based on Angular2 using the endpoints.
- t: renamed / moved E2ESpec to R2RSpec:
  * R2R - Route-to-Responder (still only internal testing, can but must not be backed by triplestore)
  * E2E - End-to-End (access only over HTTP backed by triplestore)
- t: Authentication: renamed existing test to R2R and add E2E test
- t: StoreRoute: changed existing test (StoreRouteV1E2ESpec) to only use HTTP
- t: add E2ESpec class, which can be used to run E2E tests. For a full example of how to use it, see 'StoreRouteV1E2ESpec'

### Issues
- Resolves #54 
- Resolves #79


benjamingeer:
Great! I will look at it today and give you feedback today or tomorrow morning.

webapi/src/main/scala/org/knora/webapi/responders/v1/UsersResponderV1.scala
```diff
@@ -20,144 +20,369 @@
...
+      *                     - https://crackstation.net/hashing-security.htm
+      *                     - http://blog.ircmaxell.com/2012/12/seven-ways-to-screw-up-bcrypt.html
+      *
+      * @param newUserData a [[NewUserDataV1]] object containing information about the new user to be created.
+      * @param userProfile a [[UserProfileV1]] object containing information about the requesting user.
+      * @return a future containing the [[UserOperationResponseV1]].
+      */
+    private def createNewUserV1(newUserData: NewUserDataV1, userProfile: UserProfileV1, apiRequestID: UUID): Future[UserOperationResponseV1] = {
+        for {
+            a <- Future("")
```
benjamingeer: I guess `a <- Future("")` is just here because you need a `<-` on the first line of the `for` comprehension? It's a bit ugly like this. Instead you could write:

```
// check if username or password are not empty
_ <- Future {
    if (newUserData.username.isEmpty) throw BadRequestException("Username cannot be empty")
    if (newUserData.password.isEmpty) throw BadRequestException("Password cannot be empty")
}

// check if the supplied username for the new user is unique, i.e. not already registered
sparqlQuery <- queries.sparql.v1.txt.getUserByUsername(newUserData.username).toString()
```

webapi/src/main/scala/org/knora/webapi/responders/v1/UsersResponderV1.scala
```diff
@@ -20,144 +20,369 @@
...
+                            email = newUserData.email,
+                            preferredLanguage = newUserData.lang).toString)
+                        // _ = println(createNewUserSparql)
+                        createResourceResponse <- (storeManager ? SparqlUpdateRequest(apiRequestID, createNewUserSparql)).mapTo[SparqlUpdateResponse]
+                    } yield createResourceResponse
+            }, storeManager)
+
+            // Verify that the user was created.
+
+            sparqlQuery <- Future(queries.sparql.v1.txt.getUser(userIri = userIri).toString())
```
benjamingeer: No need for `Future(...)` here.

webapi/src/main/scala/org/knora/webapi/responders/v1/UsersResponderV1.scala
```diff
@@ -20,144 +20,369 @@
...
+            // create the user operation response
+            userOperationResponseV1 = UserOperationResponseV1(newUserProfile, userProfile.userData)
+
+        } yield userOperationResponseV1
+
+    }
+
+
+    private def updateUserV1(userIri: webapi.IRI, propertyIri: webapi.IRI, newValue: Any, userProfile: UserProfileV1, apiRequestID: UUID): Future[UserOperationResponseV1] = for {
+            a <- Future("")
```
benjamingeer: Same as above.

webapi/src/main/scala/org/knora/webapi/responders/v1/UsersResponderV1.scala
```diff
@@ -20,144 +20,369 @@
...
+                // not the user and not a system admin
+                throw ForbiddenException("User information can only be changed by the user itself or a system administrator")
             }
+            _ = if (propertyIri.equals(OntologyConstants.KnoraBase.IsSystemAdmin) && !userProfile.userData.isSystemAdmin.contains(true)) {
+                // the operation of promoting to system admin is only allowed by another system admin
+                throw ForbiddenException("Giving an user system admin rights can only be performed by another system admin")
+            }
+
+            // get current value.
+            sparqlQuery <- Future(queries.sparql.v1.txt.getUser(userIri = userIri).toString())
```
benjamingeer: No need for `Future(...)` here.

webapi/src/main/scala/org/knora/webapi/responders/v1/UsersResponderV1.scala
subotic: Yes. Thanks.

benjamingeer:
This looks great so far!

The main issue that occurs to me is concurrent updates of users, groups, and projects. For resources, we have `ResourceLocker`, but there's nothing in it that's actually specific to resources, so it could be used to lock anything that has an IRI. You could rename it `RdfObjectLocker` and use it for updates to users, groups, and projects.

In `administration.rst`, I don't understand this paragraph:

> It will be possible to define default permissions for newly created resources / values on the system, project, and user level. The following properties need to be defined to point to a list of IRI’s of instances of ‘knora-base:UserGroup’: knora-base:hasDefaultRestrictedViewPermission, knora-base:hasDefaultViewPermission, knora-base:hasDefaultModifyPermission, knora-base:hasDefaultDeletePermission. These default permissions can then be explicitly submitted by the client with each creation request. 

On the system level, where will these properties be stored, and what will be their subject? How will Knora use these three levels of defaults when creating a new resource or value? Why would the user submit default properties with each creation request?

subotic:
Thanks. Yes, locking would be a good idea. I will take a look at ResourceLocker.

> In `administration.rst`, I don't understand this paragraph:

>> It will be possible to define default permissions for newly created resources / values on the system, project, and user level. The following properties need to be defined to point to a list of IRI’s of instances of ‘knora-base:UserGroup’: knora-base:hasDefaultRestrictedViewPermission, knora-base:hasDefaultViewPermission, knora-base:hasDefaultModifyPermission, knora-base:hasDefaultDeletePermission. These default permissions can then be explicitly submitted by the client with each creation request. 

> On the system level, where will these properties be stored, and what will be their subject? How will Knora use these three levels of defaults when creating a new resource or value? Why would the user submit default properties with each creation request?

Since Lukas wants that the user has the ability to set the permissions on each creation request, the idea arose why doesn't the client simply always send the permissions with each creation request. The 'default' permissions that would be defined somehow/somewhere are simply suggestions that the GUI shows to the user, and whatever the user choses is sent along with the creation request. So at the moment of resource / value creation the server has all the information that it needs, and doesn't need to calculate the 'default permissions'.

benjamingeer:
> Since Lukas wants that the user has the ability to set the permissions on each creation request, the idea arose why doesn't the client simply always send the permissions with each creation request. The 'default' permissions that would be defined somehow/somewhere are simply suggestions that the GUI shows to the user, and whatever the user choses is sent along with the creation request. So at the moment of resource / value creation the server has all the information that it needs, and doesn't need to calculate the 'default permissions'.

OK, that makes sense! But how would the system defaults, project defaults, and user defaults be combined in the suggestions? Would the user's defaults override the project and system defaults somehow?

subotic:
> But how would the system defaults, project defaults, and user defaults be combined in the suggestions? Would the user's defaults override the project and system defaults somehow?

I guess, we supply some sensible defaults on the system level (which can be changed on each deployment). Each project can define other defaults. And the user itself can also define other defaults. The GUI will then first show the ones from the user, then the project, and then the system. There wouldn't be any combining. These are just suggestions, nothing will be enforced. At least this is how I understood Lukas.

benjamingeer:
> The GUI will then first show the ones from the user, then the project, and then the system.

OK.

subotic:
Needed to delete branch, as I have squashed a few commits removing some large files, as we are over the allowed quota for our repository.

PR #109, opened 2016-04-06
Consistency checking, version 2

benjamingeer:
This enables GraphDB consistency checks, with some limitations because of the remaining problems described in #33 (specifically [this](https://github.com/dhlab-basel/Knora/issues/33#issuecomment-203434997) and [this](https://github.com/dhlab-basel/Knora/issues/33#issuecomment-206360261)).

There are now separate checks for:

* system properties (which point to things that aren't Resources or Values, i.e. things that can't be marked as deleted)
* content properties (which point to Resources or Values, i.e. things that can be marked as deleted)

I made this distinction because the `isDeleted` flag on Resources and Values should be taken into account when checking cardinalities (a deleted object should be treated as if it doesn't exist). This also means that `isDeleted` has to be a required property on all resources and values, so I've added `isDeleted false` to everything in the test data. However, checking `isDeleted` in consistency checks doesn't work yet, so it's commented out in the consistency rules. Therefore, the consistency checks won't yet deal correctly with deleted resources or values. This is probably OK for now, because imported data won't contain anything that's marked as deleted, and because deletion isn't fully implemented yet in the API. I'll revisit this once the remaining issues in #33 are resolved.

When GraphDB is accessed over HTTP, it can do only one SPARQL update operation per transaction. Therefore the application-level transaction management introduced in #50 can't work, so I've removed it. The use case that it was intended to handle was creating a new resource with its initial values. Consistency checking is done at the end of each transaction, so we have to create a resource and its initial values in a single transaction. Now this is done in a single SPARQL update operation. The values responder generates snippets of SPARQL for creating values, and the resources responder inserts these snippets into its SPARQL update operation for creating the resource. This way, it's all done in one transaction.

Consistency checks require inference, but inference interferes with many of our queries. Until now we've disabled inference in SELECT queries by sending the `infer=false` parameter in the HTTP request to GraphDB. However, it turns out that this doesn't turn off [the OWL `sameAs` optimisation](http://graphdb.ontotext.com/documentation/standard/sameas-optimisation.html), which interfered with one of our queries. To be on the safe side, I've added something to every query template to insert `FROM <http://www.ontotext.com/explicit>` when GraphDB is used, ensuring that no inference will be performed.

I've added some tests in `GraphDBConsistencyCheckingSpec`. Note that for the tests to pass, the `knora-test-unit` repository has to be built with consistency checks enabled (which is now done by `graphdb-se-ci-prepare.sh`.


docs/rst/knora-api-server/design-documentation/triplestore-updates.rst
```diff
@@ -202,35 +216,62 @@ a time can update a given resource. Before each update, the application
...
+update code, because it relies on permission checks.
+
+
+GraphDB's consistency checker can be turned on to ensure that each update
+transaction respects the consistency constraints, as described in the section
+`Consistency checks`_ of the GraphDB documentation. This makes it possible to
+catch consistency constraint violations caused by bugs in Knora, and it also
+checks data that is uploaded directly into the triplestore without going
+through the Knora API. However, this feature is not yet enabled, because of
+problems described in `issue 33`_.
```
subotic: > However, this feature is not yet enabled, because of problems described in issue 33.

This should be probably updated, to reflect that CC is now turned on.

subotic:
Cool, thanks. Looks good. You can merge if you like.

benjamingeer:
Thanks!

PR #121, opened 2016-04-15
folder and build restructuring

### LINKED ISSUE #115: Changes in the folder structure / build process
I would propose the following changes, which would give us a more modular structure, allowing easier testing / deployment.

Relevant for #3.

1. Get rid of SBT in the root folder. Only ``webapi`` is using scala/sbt and by moving it to here, allows us  to have everything that is needed for running it in production or testing in one place. Also makes it easier to create a docker image containing just the ``webapi`` content.

2. Move ``/webapi/_fuseki`` to ``/triplestores/fuseki``. Hopefully, sometime soon, we can add ``graphdb-free`` (still waiting on the approval from Ontotext) to the repository, and then it would live in ``/triplestores/graphdb-free``. This change will also allow us to isolate only the files that are needed for the docker image.

subotic:
Moved the fuseki folder and divided KnoraBuild.sbt in two separate build files.

Sbt needs to be called separately in each folder (webapi, salsah).

resolves #115


benjamingeer:
Looks fine but could you put something in the RST docs to say how to run SALSAH? :) I guess we need a `docs/rst/salsah` directory.

subotic:
Yes, I suppose we need a ``docs/rst/salsah`` folder with a developer section. It would the contain something like:

To start Salsah, do one of these:
- open index.html in browser, or
- execute ``sbt run``, or
- execute ``sbt reStart``, or
- use ``npm lite-server``, or
- whatever else you like that does the job :-) 

I will add one a salsah dev section, when I write up the documentation for the admin interface.

benjamingeer:
Currently we do this:

```
$ sbt
> project webapi
> reStart
> salsah/reStart
```

Will that change? (Will we have to type `project salsah` now?) I think the docs should tell people exactly what to type, especially if they aren't familiar with SBT.

subotic:
No it's just like this now:
``$sbt reStart``

It's the same also for webapi. The ``project salsah`` or ``project webapi`` was needed to select the subproject. We don't have any subprojects anymore.

For webapi:
```
$cd webapi
$sbt
>reStart
```

For salsah:
```
$cd salsah
$sbt
>reStart
```

subotic:
I have reverted the merged pull request. I forgot to update the documentation. I will create a new merge request when I'm done.

PR #161, opened 2016-06-01
Wip/test ontology

tobiasschweizer:
- make resource creation possible using the GUI (jquery.resadd.js)
- add a test ontology
- enable to create regions and to edit them using the GUI
- enable to create a LinkObj using the GUI 
- enable image upload using the GUI on resource creation


benjamingeer:
I think this is OK to merge. The only annoying thing I see is that we had to add default permissions to `hasStillImageFileValue` in `knora-base`. But I guess there's nothing else we can do until the new permission system is ready.

tobiasschweizer:
Yes, I had to add these permissions because otherwise a new resource would not have any permissions at all

tobiasschweizer:
give me a minute so I can have a look at the changes

are you on skype?

benjamingeer:
yes

benjamingeer:
I cleaned up a few things that IntelliJ IDEA pointed out:

- Change postfix syntax (`180 seconds`) to method call syntax (`180.seconds`).
- Use empty parentheses when calling parameterless methods that have side effects.
- Use `.head` instead of `(0)`.

Please see the [Scala Style Guide](http://docs.scala-lang.org/style/method-invocation.html#arity-0) on the convention for parameterless methods.

Also, in `SearchAndEditSpec`, I find this hard to read:

```
            val firstSel = selections.head

            firstSel.selectByValue("http://data.knora.org/lists/71a1543cce")

            // refresh the selection
            selections = page.getHierarchicalListSelectionsInExtendedSearch(1)

            val secondSel = selections(1)

            secondSel.selectByValue("http://data.knora.org/lists/d1fa87bbe3")
```

It's hard to read because I don't know what these IRIs refer to. Could you these IRIs into constants, so the name of the constant would tell me what it is?

tobiasschweizer:
Ok, I fixed the bug for interval_value. And I had to rename float_value to decimal_value in the GUI. When you change names in the route, please adapt them in the GUI as well, otherwise it breaks.

There is no GUI element for boolean_value yet, so I commented it out from Thing in the test ontology.

For time_value: at the moment the GUI sends the same structure as for interval_value, but the route expects not an array of decimal numbers, but just one. @milchkannen What is time_value supposed to look like?

tobiasschweizer:
We should have a look at the missing GUI element for Boolean and the expected format for time_value together with Lukas on Monday.

I added constants to the tests but did not commit that yet. Will commit that on Monday. :-)

benjamingeer:
Looks OK to me, please merge!

PR #168, opened 2016-06-08
User, Project, and Group Management (deprecated)

### LINKED ISSUE #1: Projects need different sorts of permissions from resources and values
- Decide which sorts of permissions projects can grant.
- Decide whether projects have owners.
- Implement appropriate permission-checking in `PermissionUtilV1`, and use it in `ProjectsResponderV1`.

### LINKED ISSUE #54: Add user, group, and project management capabilities

### LINKED ISSUE #79: Redesign default permissions

subotic:
Deprecated by #299 
### Summary
- f: add `v1/users` API endpoint
- f: add `v1/projects` API endpoint
- f: add `v1/groups` API endpoint
- f: add exemplary admin GUI based on Angular2 using the endpoints.
- t: renamed / moved E2ESpec to R2RSpec:
  - R2R - Route-to-Responder (still only internal testing, can but must not be backed by triplestore)
  - E2E - End-to-End (access only over HTTP backed by triplestore)
- t: Authentication: renamed existing test to R2R and add E2E test
- t: StoreRoute: changed existing test (StoreRouteV1E2ESpec) to only use HTTP
- t: add E2ESpec class, which can be used to run E2E tests. For a full example of how to use it, see 'StoreRouteV1E2ESpec'
### Issues
- Resolves #1 
- Resolves #54 
- Resolves #79
### Remarks

This pull request is new, as I needed to squash a few commits. Github doesn't like when history is rewriten. The old pull request is #107.


mrivoal:
Hi guys, 

We just want to let you know that we are working on a real use case that apply both to lumières.lausanne and to the project Dessins de dieux. We will comme back to you very soon with a admin.ttl, data.ttl and onto.ttl. Hopefully it will be easier for you to undestand the problems we are facing. 

And I guess we can then discuss it in a more concrete way.

subotic:
I had a meeting with @lrosenth yesterday regarding permissions in Knora. We discussed it in depth and came up with a few changes, which should allow us to cover a broad range of use cases. I am fairly confident, that **all** the use cases from Lausanne are going to be covered (e.g., Cheese Fondue, etc.). I will rewrite the documentation to reflect the new design in the coming days, so stay tuned.

Issue #101, opened 2016-03-30
Redesign Standoff


benjamingeer:
We've been talking about redesigning how we store standoff in RDF, to make it flexible enough to meet the needs that different projects have described to us.
### Requirements
- Users can create different standoff perspectives on the same text.
- Users can create their own types of standoff nodes with project-specific semantics and custom properties.
- Users can annotate standoff nodes.
- When a new version of a text is made, any annotations on standoff nodes will also point to the equivalent nodes in the new version, assuming those nodes still exist.
- Users can convert an XML document to standoff and back, and get back an equivalent XML document.
### Design

This could be done by representing standoff nodes as Knora resources of class `StandoffTag` rather than as blank RDF nodes.

The `TextValue` class would not point directly to standoff resources, because this would not make sense if there were multiple standoff perspectives on the same text. However, the `TextValue` could optionally point to a default `StandoffPerspective`. When the API received a text value with standoff, it would generate a `StandoffPerspective` and standoff resources, and create a `TextValue` with the property `hasDefaultStandoffPerspective` pointing to the new standoff perspective. In API version 1, when the text value was returned in response to an API GET request, it could be returned along with its default standoff perspective, to match the current API behaviour.

Projects would be able to define their own subclasses of `StandoffTag` with specific semantics. We would provide a new API route for converting an XML document into text with standoff markup. The text would be uploaded along with parameters specifying how to map XML tags (or combinations of tags and attributes) to standoff resource classes.

Currently, we maintain direct links between a resource containing text and resources that the text's standoff markup refers to. This is useful because these links are easily shown as incoming links to the target resource, and are easily found in searches. This is possible because the resource contains a reference count for each of these links. Maintaining reference counts is possible because we lock each resource when changing its contents. Our resource locks are based on the rule that each API operation locks only one resource at a time, to prevent deadlocks. If standoff is represented by a graph of resources, it would no longer be feasible to maintain these direct standoff links. Instead, we could have direct links from the standoff resources. This could be accomplished by making `StandoffLinkTag` a subclass of `LinkObj`. Each `StandoffLinkTag` would then have a link property pointing to the target resource, as well as a property pointing to the text value itself. As a result, we could eliminate reference counts from `LinkValue` and simplify some code.

[...]

PR #186, opened 2016-07-12
Mark resources and values as deleted, and check permissions on links

### LINKED ISSUE #53: When a TextValue containing standoff links is deleted, decrement the links' reference counts
Currently they're kept.

cc @lrosenth @subotic

### LINKED ISSUE #88: What permissions should a LinkValue for a standoff link have?
When we create a new `LinkValue` for a standoff link to a resource, what permissions should the `LinkValue` have? Currently it gets the same permissions as the `TextValue` containing the standoff. But that doesn't really make sense. Resource R1 could have 3 different properties, each of which points to a `TextValue` containing a link to the same resource R2. This would generate one direct link between the resources (via the property `hasStandoffLinkTo`), and one one `LinkValue` for that link, with a reference count of 3. If each of those `TextValue` objects has different permissions, what permissions should the `LinkValue` have?

cc @subotic @tobiasschweizer

### LINKED ISSUE #89: Permissions on links are not checked correctly
When the resources responder queries the incoming links of a resource, this is how it currently decides whether the requesting user has permission to see those incoming links: if the user has permission to see the source and target of the link, they can see the link. Instead (or in addition to this), we should be checking the permissions on the `LinkValue` of each link. After all, one of the reasons we have `LinkValue` is to have a place to store permissions for a link.

See also #88.

### LINKED ISSUE #132: Delete a resource
In the SALSAH GUI, the user can click on a trash can button to delete a resource.

But we have not defined this method in the resources route yet:

```
Request URL:http://localhost:3333/v1/resources/http%3A%2F%2Fdata.knora.org%2Ffcac0fe61801
Request Method:DELETE
Status Code:405 Method Not Allowed
Remote Address:127.0.0.1:3333
```

Can we just add this method to the route and mark the resource in question as deleted (sufficient permissions provided)?

### LINKED ISSUE #182: Handle Permissions on Properties correctly when searching
At the moment, permissions are only checked on resources when searching. 

However, when displaying a resources some of its properties may not be visible due to restrictions.

Example: Do a simple search for `Berthold` or an extended search for a book having `Berthold` as its creator. In the result, `Berthold` shows up as a creator but when you open the book, the `creator` property is not visible (only when you log in).

berthold

Permission checking on resources has to be added in the following SPARQL templates:
- `searchFulltext.scala.txt`
- `searchExtended.scala.txt`
- `getResourceSearchResult.scala.txt`

### LINKED ISSUE #196: Store permissions in a single triple for more efficient queries
The performance of the extended search is currently limited by the way we store permissions. Because each resource or value can have an arbitrary number of permission assertions, we have to use `GROUP BY` and `GROUP_CONCAT` in SPARQL to concatenate those assertions into a string, so they can be returned in a single table cell in the results. The extended search query has to do this for each matching value. Grouping in SPARQL is fast enough if there are only one or two search criteria, but it gets very slow if there are three criteria (about 10 seconds), and much slower if there are four criteria (over 30 seconds). Grouping in Scala isn't an option, because it would mean returning one row per permission assertion per value, which would result in a huge number of rows being returned.

@lrosenth and I had a discussion about how to deal with this, and we concluded that the best option is to store all of a resource's or value's permission assertions in a string, in a single triple. Basically that string would contain the same thing that `GROUP_CONCAT` produces, but in a more compact format (we could use abbreviated codes instead of IRIs for the permission predicates). This would eliminate the grouping of permissions in SPARQL queries.

The only use case that this wouldn't be adequate for seems to be the case where you want to create a mirror of the publicly-viewable content in a repository. But Lukas had an idea about how to do this: instead of doing it with permissions, we could add a `creativeCommonsLicense` property to resources and values that are publicly viewable, specifying which Creative Commons License applies to the content. Then it would be easy to query for those in SPARQL.

@subotic, does this seem OK to you? If so, I can make a branch off `develop` to work on it.

cc @tobiasschweizer

benjamingeer:
The code for marking resources and values as deleted was incomplete and incorrect, so this is a reimplementation.

Previously, we had to make a new version of a value to mark it as deleted. Now this is necessary only for `LinkValue` (because its reference count has to be decremented to 0 first).

All resources and values that are marked as deleted now have a `deleteDate` as well as `isDeleted true`.

I implemented permission-checking on link values in the resources responder and the search responder.

Permissions are now stored in a single string literal as per #196, using the datatype property `knora-base:hasPermissions`. The format of this literal is documented in `knora-base.tex`.

I've also refactored some things in the resources responder, the values responder, and the routes and tests, and added comments.

Closes #53: When a `TextValue` containing standoff links is deleted, decrement the links' reference counts.

Closes #88: What permissions should a `LinkValue` for a standoff link have?

Closes #89: Permissions on links are not checked correctly.

Closes #132: Delete a resource.

Closes #182: Handle Permissions on Properties correctly when searching.

Closes #196: Store permissions in a single triple for more efficient queries.


tobiasschweizer:
Great, you fixed the bug that on value deletion some strange value IRI was written back in the response. Now it is the same as the one that was sent in the original request.

Hpwever, I do not fully understand the documention for the parameter id for `DeleteValueResponseV1`:

> * @param id       the IRI of the value that was marked as deleted. This may be the same value IRI that was given in
  *                 the [[DeleteValueRequestV1]], or it may be the IRI of a new version of that value.

How can there be a new version of that value be returned when it has been deleted just now? Concurrency?

tobiasschweizer:
I have just found a bug in `ExtendedSearch`: We forgot to the statement:

```
?resource knora-base:isDeleted false .
```

tobiasschweizer:
I added a fix to correctly URL encode IRIs for the delete request

benjamingeer:
> How can there be a new version of that value be returned when it has been deleted just now?

This is explained in the design documentation:

https://github.com/dhlab-basel/Knora/blob/wip/delete-values/docs/rst/knora-api-server/design-documentation/triplestore-updates.rst

> Normally, a value is marked as deleted without creating a new version of it. However, link values must be treated as a special case. Before a `LinkValue` can be marked as deleted, its reference count must be decremented to 0. Therefore, a new version of the `LinkValue` is made, with a reference count of 0, and it is this new version that is marked as deleted.

benjamingeer:
> you fixed the bug that on value deletion some strange value IRI was written back in the response. Now it is the same as the one that was sent in the original request.

This wasn't a bug. Please read the description of this pull request above, paying special attention to this sentence:

> Previously, we had to make a new version of a value to mark it as deleted. Now this is necessary only for `LinkValue` (because its reference count has to be decremented to 0 first).

benjamingeer:
The old behaviour was also documented:

https://github.com/dhlab-basel/Knora/blob/develop/docs/rst/knora-api-server/design-documentation/triplestore-updates.rst

> 'Deleting' a value means creating a new version, marked with `knora-base:isDeleted` and pointing to the previous version.

tobiasschweizer:
Ok, now I understand why in the case of a link a new version has to be created.

Could you please add a sentence to the documentation for `case class DeleteValueResponseV1` referring to the documentation you have mentioned?

knora-ontologies/knora-base.ttl
```diff
@@ -972,9 +972,7 @@
 
             rdfs:subPropertyOf :systemProperty ;
 
-            rdfs:comment "Indicates when a resource was deleted"@en ;
-
-            :subjectClassConstraint :Resource ;
```
tobiasschweizer: Did you remove the constraint because this can apply to a resource and/or a value?

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
```diff
@@ -1334,6 +1330,66 @@ class ResourcesResponderV1 extends ResponderV1 {
...
+                }
+            } yield ResourceDeleteResponseV1(
+                id = resourceDeleteRequest.resourceIri,
+                userdata = resourceDeleteRequest.userProfile.userData
+            )
+        }
+
+        for {
+        // Don't allow anonymous users to delete resources.
+            userIri <- resourceDeleteRequest.userProfile.userData.user_id match {
```
tobiasschweizer: Is this is a Future because it is the first statement in the for-clause?

webapi/src/main/twirl/queries/sparql/v1/deleteResource.scala.txt
```diff
@@ -0,0 +1,91 @@
...
+WHERE {
+    BIND(IRI("@dataNamedGraph") AS ?dataNamedGraph)
+    BIND(IRI("@resourceIri") AS ?resource)
+    BIND(NOW() AS ?currentTime)
+
+    @* Make sure the resource is really a resource and isn't marked as deleted. *@
+
+    ?resource rdf:type ?resourceClass ;
+        knora-base:isDeleted false .
+    ?resourceClass rdfs:subClassOf+ knora-base:Resource .
```
tobiasschweizer: The use of `rdfs:subClassOf+ knora-base:Resource` means that this works for sublasses of knora-base:Resource, but not for a knora-base:Resource itself.

However, from the GUI it is possible to create a knora-base:Resource but this fails when making a get request providing the new resource's IRI.

So we have to decide whether knora-base:Resource can be instantiated or only sublasses of it. If yes, we should use the operator *. And we would have to do that in a consistent manner (in every SPARQL template dealing with resources)

knora-ontologies/knora-base.ttl
benjamingeer: Yes.

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
benjamingeer: Yes.

webapi/src/main/twirl/queries/sparql/v1/deleteResource.scala.txt
benjamingeer: I think it makes no sense to create a `Resource` and should be forbidden.

tobiasschweizer:
I do have a problem of understanding regarding the deletion of LinkValues. The following code ist from ValuesResponderV1.scala:

```scala
case linkValue: LinkValueV1 =>
                    // It's a LinkValue. Make a new version of it with a reference count of 0, and mark the new
                    // version as deleted.

                    val linkPropertyIri = knoraIriUtil.linkValuePropertyIri2LinkPropertyIri(findResourceWithValueResult.propertyIri)

                    for {
                        sparqlTemplateLinkUpdate <- decrementLinkValue(
                            sourceResourceIri = findResourceWithValueResult.resourceIri,
                            linkPropertyIri = linkPropertyIri,
                            removedTargetResourceIri = linkValue.objectIri,
                            userProfile = deleteValueRequest.userProfile
                        )

                        sparqlUpdate = queries.sparql.v1.txt.deleteLink(
                            dataNamedGraph = settings.projectNamedGraphs(findResourceWithValueResult.projectIri).data,
                            triplestore = settings.triplestoreType,
                            linkSourceIri = findResourceWithValueResult.resourceIri,
                            linkUpdate = sparqlTemplateLinkUpdate,
                            maybeComment = deleteValueRequest.deleteComment
                        ).toString()

                    } yield (sparqlUpdate, sparqlTemplateLinkUpdate.newLinkValueIri)
```

What happens if the direct link has a reference count of 2? Then it only should be decremented to 1, but the direct link should not be deleted, right?

This is how I understand this line: `val deleteDirectLink = linkValueQueryResult.directLinkExists && newReferenceCount == 0`

But in the template `deleteLink`, there is the following check taking place:

```
@* Delete the direct link. *@
        @if(linkUpdate.deleteDirectLink) {
            ?linkSource ?linkProperty ?linkTarget .
        } else {
            @{throw SparqlGenerationException(s"linkUpdate.deleteDirectLink must be true in this SPARQL template"); ()}
        }
```

So in our example that would be `false` since the reference count would be 1.

benjamingeer:
> What happens if the direct link has a reference count of 2?

This, too, is explained in `triplestore-updates.rst`:

> For consistency, every `LinkValue` contains a reference count. If the link property
> is not `knora-base:hasStandoffLinkTo`, the reference count will always be either 1
> (if the link exists) or 0 (if it has been deleted, in which case the link value will also be
> marked with `knora-base:isDeleted`).

benjamingeer:
See also the documentation in `knora-base.pdf`:

> **valueHasRefCount (1)** The reference count of the link. This is meaningful when the LinkValue describes resource references in Standoff text markup (see Section 4.1.3). Otherwise, the reference count will always be 1 (if the link exists) or 0 (if it has been deleted).

tobiasschweizer:
ok, so only a TextValue's standofflinks could create a ref count > 1.

in decrementLinkValue, I found the following comment:

```
// There's already a LinkValue for links between these two resources. Decrement its
                    // reference count.
```

Isn't this misleading? Of course there is "already a link". It is the one that you attempt to delete.

benjamingeer:
That's in a match-case expression that looks at the result of a SPARQL query to get the details of the link. The link should exist, but if it doesn't, the query result will be `None`. Look at the comment under `case None`.

tobiasschweizer:
Wouldn't it be better to say that you found what you expected in the first case?

tobiasschweizer:
Ok, I see now what `decrementLinkValue` does in case of a standoff link: the direct link will only be deleted if the ref count reached 0.

tobiasschweizer:
Wouldn't it be better to say `targetResourceIRIToBeRemoved` for decrementLinkValue?

Technically, It has not been removed yet. And possibly it won't when the ref count is still > 0 (deleteDirectLink is false).

benjamingeer:
> Wouldn't it be better to say `targetResourceIRIToBeRemoved` for decrementLinkValue?

You're right, `removedTargetResourceIri` isn't clear, it refers to the case where a resource reference has been removed from standoff markup. I'll rewrite the comments in that method so they also make sense for the case where we're deleting a link.

tobiasschweizer:
Yes I will read the KnoraDocs pdf, but still I think you should add comments to the code (also by referring to the docs) to `decrementLinkValue` and explain that this function can be called in two cases :

- delete a LinkValue (such as page x isPartOf book y), where ref count is 1. So after decrementing it it is 0 and deleteDirectLink must be true. Decrementing it by subtraction seems to be useless in this case because ref count must be 0. I only understood this when I looked at the TextValue case, because the function has to deal with both cases.

- delete a TextValue: then you have to check if standoffLinks have been created. If there are two text values attached to the same resource referring to the same resource, then the ref count is 2 (but when you just have one TextValue the direct link is only created once also when you have several links in the text). And when you remove one TextValue, ref count has to be decremented to one, but the link still exists.


- please clarify the comment on `id` like suggested above
```
/**
  * Represents a response to a [[DeleteValueRequestV1]].
  *
  * @param id       the IRI of the value that was marked as deleted. This may be the same value IRI that was given in
  *                 the [[DeleteValueRequestV1]], or it may be the IRI of a new version of that value.
  * @param userdata information about the user that made the request.
  */
case class DeleteValueResponseV1(id: IRI,
                                 userdata: UserDataV1) extends KnoraResponseV1 {
    def toJsValue = ApiValueV1JsonProtocol.deleteValueResponseV1Format.write(this)
}

```

benjamingeer:
I'll do all this, and also fix the permissions in link values for standoff links, as explained in #88.

tobiasschweizer:
Great!

And please also add the tests we talked about:

> Could you also add tests that delete TextValues that have standoff links (scenarios with different ref counts to make sure that the direct link is not deleted if there is still a TextValue with a standoff link to the same target resource if one got deleted)?

benjamingeer:
Yes, I will.

tobiasschweizer:
thx.

benjamingeer:
@tobiasschweizer, I think I've done everything you asked for, could you review this again?

benjamingeer:
The main things that have changed:

* The resources responder and values responder now check permissions on links as described in [this comment](https://github.com/dhlab-basel/Knora/issues/88#issuecomment-233623350). Link values for standoff links now always belong to `knora-base:SystemUser`.
* I refactored some things in the resources responder, values responder, and routes for clarity, and added comments.
* I added tests in the `ValuesResponderV1Spec` and `ResourcesResponderV1Spec`.

Question: how does the search responder check permissions on links? Does it need to be changed to make it work this way?

benjamingeer:
It looks like the extended search SPARQL template will need some work in order to get permissions on link values.

benjamingeer:
Actually the problem with checking permissions in search is more general, as explained in #182.

benjamingeer:
I'm going to keep working on #182. It also just occurred to me that the resource context query doesn't check permissions on `isPartOf` links, either.

tobiasschweizer:
Ok, I will try to finish the review of this pull request by tomorrow. 

If you have time could you have a look at https://github.com/dhlab-basel/Knora/issues/101#issuecomment-235313637? Thx.

benjamingeer:
No need to review this again yet, because #182 is going to take a lot more work.

benjamingeer:
The main things that have changed since this pull request was last reviewed:

* The resources responder checks permissions on link values for incoming and outgoing links in response to `ResourceFullGetRequestV1`, as described in https://github.com/dhlab-basel/Knora/issues/88#issuecomment-233623350. Link values for standoff links now always belong to `knora-base:SystemUser`.
* I redesigned full-text and extended search to check permissions on values (including link values), as described in https://github.com/dhlab-basel/Knora/issues/182#issuecomment-235586885.
* `knora-base:valueHasString` is required on `knora-base:Value` again, because it's needed in search results.
* I refactored some things in the resources responder, values responder, and routes for clarity, and added comments.
* I added tests in `ValuesResponderV1Spec`, `ResourcesResponderV1Spec`, and `SearchResponderV1Spec`.

The resources responder still doesn't check permissions on link values for `ResourceContextGetRequestV1`. I tried implementing that, but it was much too slow (the query took over a minute). I think for now we can live without this check. Since the context query is actually just a specialised search, and since I don't think it can be implemented efficiently anyway, I think we should aim to eliminate it in API v2.

So I think this pull request is ready for review again.

benjamingeer:
I've implemented #196: the permissions of each resource or value are now stored in a single triple, using the datatype property `knora-base:hasPermissions`. The format of the literal object of this property is documented in `knora-base.tex`.

benjamingeer:
Since permission checking is now much faster, I added permission checking on link values for `ResourceContextGetRequestV1`.

tobiasschweizer:
I have a question regarding permission checking of incoming links. In `ResourcesResponderV1` when incoming links are queried, it is checked if the user has the permission to see the referring resource and if he has permissions to see the link (by looking at the reification).

However, in case of a standoff link, the owner is the SystemUser (hasStandoffLinkToValue). So what would happen in that case? Would that case have to treated specially here?

> When querying links (whether outgoing or incoming), ensure that the user has permission to see the source resource and the target resource. Also ensure that the user has permission to see the LinkValue, unless the link property is hasStandoffLinkTo. We're justified in making this property a special case because it's system-maintained as a side effect of standoff.

If I get this right, standoff links should always be displayed. However, the user might not be able to see the underlying TextValue that gives the link semantics.

tobiasschweizer:
The new way of representing permissions required the test data to be changed. I will have to do that for the BEOL test server. What would be the easiest way to do that? I can make a dump, change the data and load it back in.

benjamingeer:
> However, in case of a standoff link, the owner is the SystemUser (hasStandoffLinkToValue). So what would happen in that case? Would that case have to treated specially here?

Yes, it's treated specially. To see a standoff link, the user only needs permission to see the source and target resources. This is documented in `knora-base.tex` section 5.2 (Permissions):

"To view a link between resources, a user needs permission to view the source and target resources. He also needs permission to view the `LinkValue` representing the link, unless the link property is `hasStandoffLinkTo`."

And in section 4.1.3 (`StandoffLink`):

"Link values created automatically for resource references in standoff are automatically visible to all users, as long as they have permission to see the source and target resources."

This special case is implemented in `PermissionUtilV1.getUserPermissionOnLinkValueV1`.

> The new way of representing permissions required the test data to be changed.

I wrote a program to change the test data. The test data on the `wip/delete-values` branch has already been changed. To update the permissions in any Turtle file, you can use this command in SBT:

```
org.knora.webapi.util.TransformTestData permissions input-file output-file
```

The first argument currently has to be `permissions`, but this is intended to be a general-purpose framework for transforming test data in different ways.

tobiasschweizer:
got it: `if (predicateIri == OntologyConstants.KnoraBase.HasStandoffLinkTo) {
            Some(permissionsToV1PermissionCodes(OntologyConstants.KnoraBase.ViewPermission))
        }`

Regrading permissions: Great, I will try to run the program with he beol data

tobiasschweizer:
You refactored search so only one query is done (before, we had separate queries for paging and the actual results).

Now, search results may be filtered out if the user has no permissions to see them (changing the number of search results).

Does this get us into trouble somehow? I do not remember the reason, but we designed the query so we could do the counting directly on the basis of the results returned by the triplestore.

benjamingeer:
The previous design relied on the triplestore to do counting and paging. The SPARQL had to count and return one row per matching resource so the OFFSET and LIMIT would be correct for paging. The revised design does all the counting and paging in Scala, so it's fine if the SPARQL returns more than one row per matching resource.

PR #188, opened 2016-07-15
Api v1 documentation

tobiasschweizer:
- added rst documents describing API V1
- added typescript interfaces formally describing the request and response formats
- HTML API V1 documentation can be created from typescript
- FIX: added missing comparison operators combinations to extended search
- FIX: added support for interval values in the GUI and in resource creation response 
- FIX: submit project IRI when adding a comment to a value from the GUI


benjamingeer:
When I run `make jsonformat`, I get contradictory messages about whether the generated documentation is:

```
iml-wall-e:docs benjamingeer$ make jsonformat
typedoc --out _format_docu --module commonjs ../salsah/src/typescript_interfaces

Using TypeScript 1.8.10 from /usr/local/lib/node_modules/typedoc/node_modules/typescript/lib
Rendering [========================================] 100%

Documentation generated at /Users/benjamingeer/git/Knora/docs/_format_docu

Format Docs Build finished. The HTML pages are in /html.
```

There is no `/html` directory, but there is a `build/html` directory, however it doesn't contain the HTML documentation generated from the TypeScript. The correct location seems to be `_format_docu`.

benjamingeer:
# Comments on the generated API documentation

## index.html

Why does this page contain the Knora README? I think it would make sense for it to say something like, "This is the documentation for the Knora API JSON request and response formats. Please select an API operation from the list on the right."

## Unclear samples

These names seem contradictory:

```
"sampleRequests/sampleCreateResources"
"sampleRequests/sampleDeleteResponses"
"sampleRequests/sampleResourceResponses"
"sampleRequests/sampleValueResponse"
```

Are they sample requests or sample responses?

I don't understand the samples. What are they samples of? There is no such request as "add values", so why is there a `sampleAddValues`?

It would be helpful to give each of the samples a descriptive name (rather than a variable name like `sampleAddValues`), as well as a description, so people know what they can do with it.

## Request URL format and HTTP method

In some cases you give the request URL and HTTP method (GET, PUT, POST, or DELETE). For example, in *Interface changeRichtextValueRequest*, you note: "HTTP PUT to http://www.knora.org/v1/values/valueIRI".

In other cases, you give the request URL but not the HTTP method, e.g. in *Interface resourceFullResponse*, which just says "http://www.knora.org/v1/resources/resourceIRI" but doesn't say it's an HTTP GET.

In other cases, you don't give the request URL or the HTTP method, e.g. in *Interface createResourceResponse*.

Please make this consistent. I think it would be helpful to have a reminder of the request URL format and the HTTP method for each interface, even if the details are in the RST docs.

benjamingeer:
# Comments on the RST docs

These look good to me.

## Request URLs

Please use a consistent form for documenting request URLs. I don't think it's a good idea to tell people to send requests to `www.knora.org`, since there is currently no Knora server running there. It would be better to use a placeholder for the host, and explain that the user should substitute the host and port where the Knora server is running, which by default is `localhost:3333`. We should also distinguish clearly between actual text to be included in the URL, and placeholders for IRIs or other information to be filled in. For example, placeholders could be in bold or italic, like this:

<pre>
http://<b>host</b>/v1/resources/<b>resourceIRI</b>
</pre>

I don't think there's any advantage to putting `curl` in front of each request URL.

In "Adding Resources", you say, "The request has to be sent to the Knora server using the `resources` path segment." But I think it would be clearer to give a complete URL pattern.

benjamingeer:
Please add `docs/_format_docu` to `.gitignore`.

benjamingeer:
I've been working on the code for the resource context query, and I've just realised that it assumes that if A is part of B, then A must have a seqnum. Is this assumption correct? If so, is it documented?

tobiasschweizer:
I made the following changes:

- `make jsonformat` now prints out the correct path to the documentation's index page
- I removed the README and added a title instead
- I excluded the sample requests and responses from the documentation. Their only purpose was to compare actual requests to and responses from Knora with the typescript interfaces to find out if they are correct.
- I made the descriptions of the HTTP requests consistent both in the interfaces and the RST. I use `host` as a placeholder.
- `.gitignore` already has an entry to exclude `format_docu`

I do not understand what you mean by your last question:

> I've been working on the code for the resource context query, and I've just realised that it assumes that if A is part of B, then A must have a seqnum. Is this assumption correct? If so, is it documented?

Do you want to know if the sequence number of a resource that is part of another one (like a page) has to be returned in a context query? I do not think so. The array structures representing the dependent objects are sorted by the sequence number. So yes, every dependent resource needs a sequence number and this has to be used for sorting when creating the response to a context query. Do you want to know if this is documented in `knora-base.pdf`or RST?

benjamingeer:
> Do you want to know if the sequence number of a resource that is part of another one (like a page) has to be returned in a context query?

No, I want to know if a dependent resource is required to have a `seqnum` property in the triplestore.

> Do you want to know if this is documented in knora-base.pdf or RST?

I'm asking whether it's documented anywhere. I guess the `knora-base` doc would be the best place. But maybe we should also change the ontology design to prevent users from forgetting to include the `seqnum` property in dependent resources. For example, we could use `subjectClassConstraint` to require the subject of `isPartOf` to be a subclass of `DependentResource`, which would have a cardinality requiring `seqnum`.

tobiasschweizer:
in `knora-base.pdf`, section 2.2.1:

> seqnum (0-1) The sequence number of the resource, if it is part of an ordered group of resources, such as the pages in a book.

benjamingeer:
> if it is part of an ordered group of resources

This just says that `seqnum` can be used to indicate an order among dependent resources. It doesn't say that such an order must exist.

I don't see why "part of" implies "ordered". Resources A, B, and C could be part of resource X without having any meaningful order. For example, Basel, Lausanne, and Bern are part of Switzerland, but they are not ordered.

benjamingeer:
Also the thing you quoted above doesn't even mention `isPartOf`, it just talks about an ordered group. A group can exist without being part of anything.

tobiasschweizer:
Well, then we should add a section to the docs.

tobiasschweizer:
Ok, please have a look at the changes I made *once you are back from your holiday* (not earlier :-)).

benjamingeer:
Thanks. :) I'm on my way to St Louis now, will be in the office tomorrow.

tobiasschweizer:
Ah, I thought you would be away for another week.

That's great because you can help me with my merging problem https://github.com/dhlab-basel/Knora/issues/201.

PR #191, opened 2016-07-21
removing relative URL paths for http proxying

loicjaouen:
simple url change as the rest of my problems concerning proxying knora behind apache were not linked to knora, but to apache.

 Here's the commit's comment:

```
removing relative paths to icons in css file: it doesn't survive an front apache proxy like:

<Location "/salsah/">
   ProxyPass "http://localhost:3335/"
   ProxyPassReverse "http://localhost:3335/"
</Location>

in order to have salsah/knora/sipi served as paths ( http://myserver.unil.ch/salsah/ ) instead of URL with their ports ( http://myserver.unil.ch:3335/ ).
```

### DIFF
--- a/salsah/src/public/default.css
+++ b/salsah/src/public/default.css
@@ -758,16 +758,16 @@ input.disp_csv {
   cursor: pointer;
 }
 input.disp_list {
-  background: url('../icons/sets/view/list_bullets_icon&16.png') no-repeat top left;
+  background: url('icons/sets/view/list_bullets_icon&16.png') no-repeat top left;
 }
 input.disp_lightbox {
-  background: url('../icons/sets/view/2x2_grid_icon&16.png') no-repeat top left;
+  background: url('icons/sets/view/2x2_grid_icon&16.png') no-repeat top left;
 }
 input.disp_tableedit {
-  background: url('../icons/sets/view/3x3_grid_icon&16.png') no-repeat top left;
+  background: url('icons/sets/view/3x3_grid_icon&16.png') no-repeat top left;
 }
 input.disp_csv {
-  background: url('../icons/sets/view/csv_icon&16.png') no-repeat top left;
+  background: url('icons/sets/view/csv_icon&16.png') no-repeat top left;
 }
 .dock {
   position: absolute;


benjamingeer:
Something doesn't seem right here. The path `icons` is not equivalent to `../icons`, but to `./icons`.

loicjaouen:
I am assuming that `icons` as an absolute path would mean `URL/icons`.
And it might be equivalent to `../icons` or `./icons`, which being relative paths, depends where you are.

But I don't remember the details right now; I will recheck and give an update.

benjamingeer:
`icons`, `./icons`, and `../icons` are all relative paths.

loicjaouen:
@benjamingeer 
> Something doesn't seem right here. The path icons is not equivalent to ../icons, but to ./icons.

they can be made equal by the browser that resolves URLs before sending requests.

in the URL scheme:
`[//[user:password@]host[:port]][/]path[?query][#fragment]`
any path that goes above root (above `/`) is resolved as root.

server | base url | path | url candidate | resolves as | effective path is
-----|-----|-----|-----|-----|---
`//localhost:3335/` | `//localhost:3335/` | `../icons` | `//localhost:3335/../icons` | `//localhost:3335/icons` | `//localhost:3335/icons`
`//localhost/` | `//localhost/salsah/` | `../icons` | `//localhost/salsah/../icons` | `//localhost/icons` | `//localhost:3335/salsah/icons`
`//localhost:3335/` | `//localhost:3335/` | `icons` | `//localhost:3335/icons` | `//localhost:3335/icons` | `//localhost:3335/icons`
`//localhost/` | `//localhost/salsah/` | `icons` | `//localhost/salsah/icons` | `//localhost/salsah/icons` | `//localhost:3335/salsah/icons`

So when proxying a service in a subdomain, like `//localhost:3335/` to `//localhost/salsah/` suddenly the path `../icons` makes sense to the browser but doesn't match the real location.

benjamingeer:
I'm afraid I don't understand. Why not use absolute paths (starting with `/`) for everything? Then it would be clear, just from looking at the code, where the file really is.

loicjaouen:
Using absolute paths implies explicit paths; we would have to know the final path and add a layer of configuration when deploying the code to put `/icons` here and `/salsah/icons` there.

Now, we let the browser do part of the job; it knows the base url and the relative path and works out the final url.

We should not put configuration where things that can be worked.

benjamingeer:
OK. Does this still work if you just run `webapi` and `salsah` from the SBT console?

loicjaouen:
yes :)

benjamingeer:
OK then!

PR #240, opened 2016-08-25
feature (docker): add docker support for graphdb-se 6 and fuseki 3

subotic:
This pull-request is a first in a series to brake up pull-request #168 

Adding docker support, so that fuseki and graphdb can be run inside a container on a local machine. This makes setting up a developer machine much easier. Also, running graphdb inside a docker container is piece in the puzzle of the Knora deployment ( #3 ).


subotic:
#241 supersedes this pull-request.

PR #246, opened 2016-08-30
Remove redundant valueHasURI property.

benjamingeer:
We have both `valueHasUri` (which is used) and `valueHasURI` (which is not used). This removes the unused one.


tobiasschweizer:
merge when the tests have passed :-)

PR #252, opened 2016-09-07
Redesign the ontology responder for better performance

### LINKED ISSUE #185: Query for all Restypes takes a very long time
This query is very slow: <http://iml-beol.iml.unibas.ch:3333/v1/resourcetypes?vocabulary=0>

Often, I get a timeout (after 20 seconds). 

This query is cached but still it is a problem when we get a timeout the first time. 

The problem is that all existing restypes have to be queried (all vocabularies when para is set to 0).

We should have a look how to improve this query

benjamingeer:
This pull request implements a new loading and caching design in the ontology responder.

The old design was based on the assumption that resource and property definitions could be loaded and cached incrementally, and that since they were cached, they could be loaded via relatively complex (hence relatively slow) SPARQL queries. This proved to be too slow in the case where all definitions must be loaded and returned for a single API request (#185).

The new design loads and caches all ontology information when the application starts. It uses simple, fast SPARQL queries, and does the rest of the processing in Scala, e.g. to calculate cardinality inheritance in resource classes.

To make this work, the ontology responder needs to receive a `LoadOntologiesRequest` when the application starts. `KnoraService` sends this message on startup, and I also changed each test spec to send it after the test data is loaded.

Previously, the data in an `EntityInfoV1` was pre-filtered according to the user's preferred language and the system's default language. Now, an `EntityInfoV1` contains all available translations, and the method `getPredicateObject` can be used to filter by language.

I fixed some bugs in `knora-base.ttl`:

* `hasPermissions` was attached to some resource classes, but it is only for resource instances.
* Some classes had empty definitions in `knora-dc.ttl` as well as full definitions in `knora-base.ttl`, but Knora assumes that each class is defined in only one ontology.

I also fixed some incorrect test data.

Closes #185.


knora-ontologies/knora-base.ttl
```diff
@@ -2283,7 +2284,9 @@
 
         :resourceIcon "region.gif"^^xsd:string ;
 
-        :hasPermissions "V knora-base:UnknownUser,knora-base:KnownUser|M knora-base:ProjectMember" ;
```
tobiasschweizer: This is the state I have on my test server for BEOL. What happens if someone creates a Region now? Will it have any permissions?

benjamingeer: No, it will have no permissions.

tobiasschweizer:
`http://www.knora.org/ontology/knora-base#hasValue` is now returned in the property list: <http://localhost:3333/v1/propertylists?vocabulary=0>

I think this should be excluded since it is a generic property that has to be used via subproperties only.

PR #258, opened 2016-09-13
tests (browser tests): remove beol data from test data

tobiasschweizer:
Exclude beol data from browser test data because we do not need it.

However, beol ontology has still to be loaded because the beol named graph is not ignored when beol ontology is not loaded since this is statically defined in `application.conf`.


benjamingeer:
OK please go ahead and merge.

PR #259, opened 2016-09-13
Remove needless requirements in knora-base.ttl

### LINKED ISSUE #233: Document that seqnum is required for a depedent resource (partOf)
Add a section in `knora-base.pdf` that explains that for a dependent resource (a resource that is part of another resource, e.g. a page belonging to a book) a `seqnum` is required.

> I've been working on the code for the resource context query, and I've just realised that it assumes that if A is part of B, then A must have a seqnum

benjamingeer:
- Remove cardinalities for `knora-base:seqnum` and `knora-base:hasRepresentation` from
  `knora-base:Resource`.
- Support unordered collections, by removing the requirement that a resource that
  has the property `knora-base:isPartOf` must also have the property `knora-base:seqnum`.

Closes #233.


tobiasschweizer:
Do you still need `anything:seqnum` in anything ontology? To my understanding, it was just there because it is a subpoperty of knora-base:seqnum and you wanted to add it since it is made optional for knora-base:Resource.

tobiasschweizer:
Please run the browser times the next time you make changes to the ontology. 

We should find a better way to get property fields in the GUI than by just getting them by their order in the GUI.

PR #262, opened 2016-09-14
Small improvements in consistency checking and TransformData

benjamingeer:
- Add a transformation to `TransformData` to add missing `knora-base:isDeleted` statements.
- Add a cardinality requiring an `rdfs:label` on every `knora-base:Resource`.
- Improve error-handling if an `rdfs:label` is missing in a search result.
- Add a consistency rule that forbids empty strings as objects.
- Fix the Dokubib ontology so it's consistent with exported test data.


tobiasschweizer:
Shouldn't there be a check in the InputValidation in `toSparqlEncodedString` is the string is empty? If so, we should throw an error.

Otherwise when using Knora with fuseki we would end up with empty strings whereas when using graphdb a consistency rule would be violated.

benjamingeer:
OK, how's this?

tobiasschweizer:
It is beautiful!

Please add an E2E test for it.

PR #264, opened 2016-09-16
type (docs):added remark about pyvenv for anaconda users on mac and linux

SepidehAlassi:
- the workaround to get pip and pyvenv work is explained for those who use anaconda on Mac or Linux


tobiasschweizer:
I cannot reproduce this since I do not have that problem :-)

But please got ahead an merge it.

PR #267, opened 2016-09-19
Graph data API operation

### LINKED ISSUE #269: Redesign graph view and graph query
This concerns the API operation to get a graph of resources that are reachable from a certain resource. It has these requirements:
- The depth of the search should be configurable in the request.
- Include any edge that's a subproperty of `knora-base:hasLinkTo`.
- Exclude any edge that's a subproperty of `knora-base:isPartOf`.
- Include links that go in either direction.
- Exclude deleted resources, and any resources that are reachable only via a deleted resource.
- Exclude resources that the user doesn't have permission to see, resources that are reachable only via links that the user doesn't have permission to see, and any other resources reachable only via such paths.

The last time we tried to do this, we attempted to implement exactly what the old API did, but we didn't find a good SPARQL design. Also, it was difficult to make it efficient, because the old API provides an excessive amount of data, including all the properties of every node in the graph.

Thanks to Damyan Ognyanov at Ontotext, we have a proof-of-concept of a better SPARQL design (using a CONSTRUCT query with nested OPTIONAL blocks up to the requested maximum depth). It looks like we can get all the necessary data in one SPARQL query.

@lrosenth and I talked this over, and we're going to redesign the API operation, do a new design in the Knora API server, and change the implementation in the SALSAH GUI accordingly.

To deal with both large and small graphs, the GUI could support two ways of displaying the graph:
1. If it's a small graph, the whole graph could be shown all at once, as in the current GUI.
2. If it's a large graph, we could show just the initial node and its immediate neighbours. The user could then click on a neighbour to get its neighbours.

The API operation will return the minimum information needed to draw the graph, which is probably just the adjacency list, along with the label of each node and each edge, and the label of each node's resource class.

cc @milchkannen

benjamingeer:
This implements the API operation `/v1/graphdata/`, which returns a graph of resources that are reachable from a certain resource.

- The depth of the search is configurable in the request (via the `?depth` query parameter, currently not used in the SALSAH GUI, so it defaults to a depth of 4).
- The graph includes both outbound and inbound links.
- The graph includes any link that's a subproperty of `knora-base:hasLinkTo`, except:
    - If resource R1 `knora-base:isPartOf` resource R2, no link from R1 to R2 is included in the graph.
- The graph excludes deleted resources and links, as well as resources and links that the user doesn't have permission to see.

The old SALSAH GUI used [arbor.js](https://github.com/samizdatco/arbor) to render the graph in JavaScript. Since that library hasn't been updated in 5 years, I've replaced it with the better-maintained and easier-to-use [Cytoscape](http://js.cytoscape.org/) library.

After much experimentation with different SPARQL query designs (see #269), this implementation uses simple recursive SPARQL queries. I haven't tested it with a huge graph, but performance seems OK for graphs that aren't very big. To improve performance, it doesn't return resource properties (as the old SALSAH did). RDFS inference is also used on GraphDB for better performance.

I also did a little refactoring so the responders no longer decide whether to use inference, in the interest of simplifying the responders. Since not every triplestore offers this choice on a per-query basis, I think it's better to assume that either the repository was created with inference enabled or it wasn't. So now the store package enables inference only with GraphDB, and it's up to the templates to disable it if they don't want it (which they currently do anyway).

Closes #269.


benjamingeer:
@tobiasschweizer Could you help me add this to the TypeScript-based API docs?

tobiasschweizer:
Sure thing. Are you in on Monday?

benjamingeer:
Yes, I'll be there at the usual time.

tobiasschweizer:
I have a question regarding permissions. I wanted to open <http://data.knora.org/a-thing> in graph view, but did not see any results. I am logged in as `anything-user`.

These two API requests give me all information about the resources:
<http://localhost:3333/v1/resources/http%3A%2F%2Fdata.knora.org%2Fa-thing-with-text-values>
<http://localhost:3333/v1/resources/http%3A%2F%2Fdata.knora.org%2Fa-thing>

But the request to graph view gives me an empty result
<http://localhost:3333/v1/graphdata/http%3A%2F%2Fdata.knora.org%2Fa-thing-with-text-values> 

```
{
    "nodes": [],
    "edges": [],
    "userdata": {
        "email": "anything.user01@example.org",
        "username": "anything-user",
        "firstname": "Anything",
        "projects_info": [{
            "basepath": null,
            "shortname": "anything",
            "description": null,
            "logo": null,
            "id": "http://data.knora.org/projects/anything",
            "keywords": null,
            "rights": null,
            "longname": "Anything Project"
        }],
        "user_id": "http://data.knora.org/users/9XBCrDV3SRa7kS1WwynB4Q",
        "lastname": "User01",
        "token": null,
        "active_project": null,
        "projects": ["http://data.knora.org/projects/anything"],
        "lang": "de",
        "password": null
    },
    "status": 0
}
```

tobiasschweizer:
Do you plan to include the resources' properties (besides than linking properties) as well?

In the old graph view, we had the option to expand a resource's properties to get more information about the instance.

webapi/_test_data/all_data/anything-data.ttl
```diff
@@ -20,7 +20,7 @@
 	knora-base:isDeleted false ;
 	knora-base:attachedToUser <http://data.knora.org/users/9XBCrDV3SRa7kS1WwynB4Q> ;
 	knora-base:attachedToProject <http://data.knora.org/projects/anything> ;
-	rdfs:label "Ein Ding für wen, der die Dinge liebe" ;
+	rdfs:label "Ein Ding für wen, dem die Dinge gefallen" ;
```
tobiasschweizer: für **jemanden**, dem

the new generation of triplestores will do a grammar check and randomly reject transactions if the found a violation of German grammar

benjamingeer:
> Do you plan to include the resources' properties (besides linking properties) as well?

Lukas and I decided not to do this, for better performance. We can add some JavaScript that opens the resource viewer if you click on a node.

benjamingeer:
> But the request to graph view gives me an empty result

This is because I forgot that link values for standoff links don't have permissions. Fixed.

I also noticed that if you query a node that has no links, you get an empty graph. Fixed this too.

Added tests for both cases.

tobiasschweizer:
Ok, the query for <http://localhost:3333/v1/graphdata/http%3A%2F%2Fdata.knora.org%2Fa-thing-with-text-values> works now.

I encounter a strange behavior when opening some resources of type `Thing` in the GUI. Sometimes the GUI tries to display them as if they were books.

For example in the case of "excluded alpha":

excluded_alpha

Could this be caused by a change in `imagebase.js`?

Another thing that attracted my attention is the mouse wheel event in the graph view. I would expect it to make the graph scale but instead it flickers.

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
```diff
@@ -74,244 +72,291 @@ class ResourcesResponderV1 extends ResponderV1 {
...
+
+                        // Get the edges from the query results.
+                        val edges: Seq[QueryResultEdge] = rows.map {
+                            row =>
+                                val rowMap = row.rowMap
+                                val nodeIri = rowMap("node")
+
+                                QueryResultEdge(
+                                    linkValueIri = rowMap("linkValue"),
+                                    sourceNodeIri = if (outbound) startNode.nodeIri else nodeIri,
```
tobiasschweizer: Could you add a comment here explaining why the Iris are swapped depending on the state of outbound (outgoing or incoming links)?

benjamingeer:
I opened another issue for the `isPartOf` problem: #334, and added comments as requested. OK to merge now?

PR #283, opened 2016-10-12
Wip/beol onto

tobiasschweizer:
Add bibliography and persons to both the knora-base and beol ontologies.


knora-ontologies/knora-base.ttl
```diff
@@ -2703,11 +2920,99 @@ rdfs:label rdfs:subPropertyOf :objectCannotBeMarkedAsDeleted .
...
+                   "Person"@en ,
+                   "Personne"@fr ,
+                   "Persona"@it ;
+
+        rdfs:comment "A resource representing a person"@en .
+
+
+###  http://www.knora.org/ontology/knora-base#Publication
+
+:Publication rdf:type owl:Class ;
```
mrivoal: I don't think I would completely agree with those new additions. 

To me, they are really restrictive and project-specific. What if you want to use a `Book` class, as a subclass of `Publication`, but in this ontology authors are nodes from a list value? What if `Author` is not a single `Person` but a collective or an organisation?  What if `Author` is anonym and consequently wouldn't fullfill the properties required for a `Person`? Then, such project will have to build its own ontology, without possibility to link their data to the broader general ontology.

And don't you think those properties are not so "basic" anymore and not related to "knora-base" grammar but somewhat to a vocabulary that might be common to various projects ? If so, shouldn't it be in another named graph and couldn't it allow a little bit more flexibility (I am just asking)? Imagine doing the same in a more painting-specific perspective, or define what should be a `Location`, or an `Archeological site` in knora-base.ttl... I don't think it would make sense but it might in a separate named graph.

I mean, the old Salsah was all about allowing project to define their own data model. I don't think binding together data from various projects should result in such strong prescriptions. But maybe I am wrong... (and maybe there is no way to do otherwise at the moment)?

tobiasschweizer: We put that in knora-base because we want to reuse existing vocabularies (see <http://link.springer.com/chapter/10.1007/978-3-319-04777-5_5>) and further refine them in BEOL.

The difficulty of doing that lies in Knora's subject and object class constraint (see our discussion about the DC). But I also think that once you want to define things that are used over more than one project, you have to think about certain commonalities which makes you less flexible in your own project.

But yes, nothing speaks against having several general ontologies.

benjamingeer: I understand wanting to use existing vocabularsies, FOAF was designed for personal profile pages, not for humanities research. A `foaf:person` has a birthday (month and day) but no year of birth and no date of death. It doesn't have a way to specify that a person is known by multiple names, and that one of those names is canonical. It actually seems to me that there isn't anything in `foaf:Person` that can be used as-is in the humanities.

And some of the things you want to add to `knora-base` seem rather BEOL-specific. Why create a class called `Publication` instead of something more general like `Work`? Not everything that can be cited in a bibliography is a publication.

My guess is that it will take time to answer questions like this, and it will require input from more than one project, not just BEOL. Meanwhile, I think it would make sense to create a separate bibliography ontology that could be considered less stable than `knora-base`.

tobiasschweizer: > Meanwhile, I think it would make sense to create a separate bibliography ontology that could be considered less stable than knora-base.

This would be fine! Let's do it.

webapi/src/main/resources/application.conf
```diff
@@ -242,7 +242,7 @@ app {
                 name = "http://www.knora.org/data/beol"
             }
             {
-                path = "_test_data/ontologies/biblio.ttl"
+                path = "_test_data/ontologies/biblio-onto.ttl"
```
tobiasschweizer: Do  we need a project for `biblio`? 

`knora-dc` is similar to `biblio`, and it does not have a project, it is just defined like this in `application.conf`: 

```
rdf-data = [
            {
                path = "../knora-ontologies/knora-base.ttl"
                name = "http://www.knora.org/ontology/knora-base"
            }
            {
                path = "../knora-ontologies/knora-dc.ttl"
                name = "http://www.knora.org/ontology/dc"
            } ... ]
```

As far as I understand, resource types and properties from `biblio` can be used among different projects, but there won't be a `biblio` project.

So in my opinion `biblio-onto.ttl` is all we need.

webapi/_test_data/ontologies/beol-onto.ttl
```diff
@@ -290,6 +290,57 @@
...
+                      knora-base:hasDefaultDeletePermission knora-base:Owner .
+
+    ### ###########################################
+    ### beol:hasSonValue
+
+    :hasSonValue rdf:type owl:ObjectProperty ;
+
+                    rdfs:subPropertyOf knora-base:hasLinkToValue ;
+
+                    knora-base:subjectClassConstraint :person ;
```
tobiasschweizer: Maybe we should make this a `knora-base:Person` so we could link to persons outside the BEOL project (yes, I added person ;-))

webapi/_test_data/ontologies/biblio-onto.ttl
```diff
@@ -0,0 +1,603 @@
...
+
+                    rdfs:comment "A resource representing a journal article"@en .
+
+
+
+###  http://www.knora.org/ontology/biblio#Journal
+
+:Journal rdf:type owl:Class ;
+
+                    rdfs:subClassOf :Publication, bibo:Journal,
```
tobiasschweizer: Shouldn't rather `Article` be a subclass of `Publication`?

webapi/_test_data/ontologies/biblio-onto.ttl
```diff
@@ -0,0 +1,603 @@
...
+                    rdfs:comment "A resource representing a publisher"@en .
+
+
+
+
+###  http://www.knora.org/ontology/biblioe#Article
+
+:Article rdf:type owl:Class ;
+
+                    rdfs:subClassOf knora-base:Resource, bibo:Article,
```
tobiasschweizer: see comment below (`Journal`)

webapi/src/main/resources/application.conf
```diff
@@ -302,6 +310,13 @@ app {
                 name = "Bernoulli-Euler Online"
             }
             {
+                project = "http://data.knora.org/projects/DczxPs-sR6aZN91qV92ZmQ"
```
tobiasschweizer: Do we need a `biblio`-project? I think the ontology is enough (like `knora:dc`)

webapi/_test_data/ontologies/beol-onto.ttl
```diff
@@ -386,3 +437,33 @@
 
           knora-base:hasDefaultModifyPermission knora-base:ProjectMember ,
                                                 knora-base:Owner .
+
+    ### ###########################################
+    ### beol:Person
+
+    :Person rdf:type owl:Class ;
```
benjamingeer: > Do we capitalise resource classes by convention?

We haven't been consistent. I think that would be a good convention.

knora-ontologies/knora-base.ttl
```diff
@@ -880,11 +884,101 @@ rdfs:label rdfs:subPropertyOf :objectCannotBeMarkedAsDeleted .
...
+
+			 rdfs:comment """Represents a Given name."""@en ;
+
+             :objectClassConstraint :TextValue ;
+
+             <http://www.knora.org/ontology/salsah-gui#guiElement> <http://www.knora.org/ontology/salsah-gui#SimpleText> ;
+
+             :subjectClassConstraint :Person ;
+
+             rdfs:subPropertyOf :hasValue, foaf:GivenName .
```
benjamingeer: It's `foaf:givenName`.

webapi/_test_data/ontologies/beol-onto.ttl
```diff
@@ -290,6 +290,57 @@
 
                   knora-base:hasDefaultDeletePermission knora-base:Owner .
 
+    ### ###########################################
+    ### beol:hasSon
+
+    :hasSon rdf:type owl:ObjectProperty ;
```
benjamingeer: Could we make this `hasChild`?

webapi/_test_data/ontologies/beol-onto.ttl
```diff
@@ -290,6 +290,57 @@
...
+                                                          knora-base:KnownUser ;
+
+                      knora-base:hasDefaultModifyPermission knora-base:ProjectMember ;
+
+                      knora-base:hasDefaultDeletePermission knora-base:Owner .
+
+    ### ###########################################
+    ### beol:hasSonValue
+
+    :hasSonValue rdf:type owl:ObjectProperty ;
```
benjamingeer: hasChildValue?

benjamingeer:
Please make naming consistent: we have `publicationHasTitle` but `hasName`.

Note: each property can have only one `objectClassConstraint`, which can be an abstract class, although this doesn't yet work in the GUI; we will fix it very soon.

benjamingeer:
Also I suggest merging `Article` with `JournalArticle` and just calling it `Article`.

SepidehAlassi:
naming is now consistent for most of properties, the ones which are properties of multiple resources are untouched. ex: hasName is now property of both Journal and Publisher, so the name is remained so. Same goes for pageStart, and pageEnd which will be used not only by Article, but also by other resources soon.

> Note: each property can have only one objectClassConstraint, which can be an abstract class, although this doesn't yet work in the GUI; we will fix it very soon.

have set the object of isPartOf to be an abstract class, which should work for import but not GUI for now.

benjamingeer:
OK, great. The merge conflict in `OntologyResponderV1Spec` is probably my fault, so I'll fix it tomorrow afternoon. Then I think we can merge this branch.

tobiasschweizer:
Once all the tests pass, could you merge the branch back to `develop`? Thx.

PR #285, opened 2016-10-13
fix (comment on links): returns a comment on a link

tobiasschweizer:
- get comment on a link from the reification statement, i.e. the link value

The comment was always in the triplestore, but was not returned by the API.


tobiasschweizer:
If no comment is given, shouldn't it be rather a `null` than an empty string?

benjamingeer:
> If no comment is given, shouldn't it be rather a `null` than an empty string?

That would make sense to me. Looking at the code, I don't see why `PropertyV1` contains `comments: Seq[String]` instead of `Seq[Option[String]]`. Maybe this is what the GUI expects? Can you check?

tobiasschweizer:
I believe that we did this because of the GUI. I will compare it to the old API.

tobiasschweizer:
API V1 (in SALSAH) returns an empty string  when no comment is defined for a value.

An empty string is passed to `jquery.valcomment.js` an initialization time when no comment is given:

`jquery.propedit.js:138`
```
.valcomment({value_id: propinfo[prop].value_ids[value_index], comment: (propinfo[prop].comments === undefined) ? '' : propinfo[prop].comments[value_index]})
```

The same applies to the default args:

`jquery.valcomment.js:51`
```
localdata.settings = {
                    value_id: -1, // negative means no restype_id known
					comment: ''
                };
```

We should use `null` instead and with `with NullOptions` set in scala we can use `Seq[Option[String]]` in the case classes as suggested by Ben.

tobiasschweizer:
In V2, we could avoid `null` by using the optional statement in typescript. This actually expands a type definition to a union with `undefined` (https://www.typescriptlang.org/docs/release-notes/typescript-2.0.html#optional-parameters-and-properties). `null` should be avoided because it will not be possible to assign it to another type likes string when using typescript with `--strictNullChecks` (https://www.typescriptlang.org/docs/release-notes/typescript-2.0.html).

benjamingeer:
These changes are being made in #261.

PR #287, opened 2016-10-20
fix (sipi error message): display the error message returned by SIPI

tobiasschweizer:
Also return the error message sent by SIPI in case of image conversion problems

resolves https://github.com/dhlab-basel/Knora/issues/266


webapi/src/main/scala/org/knora/webapi/responders/v1/SipiResponderV1.scala
```diff
@@ -160,15 +160,22 @@ class SipiResponderV1 extends ResponderV1 {
             conversionResultResponse <- recoveredConversionResultFuture
 
             // get file type from Sipi response
-            responseAsMap: Map[String, JsValue] = conversionResultResponse.entity.asString.parseJson.asJsObject.fields.toMap
+            responseAsMap: Map[String, JsValue] = conversionResultResponse.entity.asString.parseJson.asJsObject.fields
 
             statusCode: Int = responseAsMap.getOrElse("status", throw SipiException(message = "Sipi did not return a status code")) match {
                 case JsNumber(ftype: BigDecimal) => ftype.toInt
```
tobiasschweizer: I have just realized that `BigDecimal` does not have anything to do with Sipi's behaviour, but with `spray.json`. I (wrongly) assumed that we did this because Sipi used to return floating point numbers for all numeric values although they might be integers in Lua. See: https://github.com/dhlab-basel/Sipi/pull/65

tobiasschweizer:
I am thinking about improving consistency regarding the messages exchanged between Sipi and Knora. In Knora, we have case classes to parse the Sipi messages into (`SipiImageConversionResponse`). But in Sipi, we just create these messages using tables which have no template at all (https://github.com/dhlab-basel/Sipi/blob/develop/scripts/convert_from_binaries.lua). It is very easy to misspell or forget members, resulting in messages Knora cannot handle.

webapi/src/main/scala/org/knora/webapi/responders/v1/SipiResponderV1.scala
```diff
@@ -160,15 +160,22 @@ class SipiResponderV1 extends ResponderV1 {
...
 
             // check if Sipi returned a status code != 0
-            _ = if (statusCode != 0) throw BadRequestException(s"Sipi returned a HTTP 200 status code, but an unsuccessful status code ${statusCode}")
+            _ = if (statusCode != 0) {
+                // Sipi returned an unsuccessful status code, get the error message
+                val sipiErrorMessage: String = responseAsMap.getOrElse("message", throw SipiException(message = "Sipi did not return an error message with key 'message' although status is != 0")) match {
+                    case JsString(msg) => msg
+                    case other => throw SipiException(message = s"Sipi did not return an error message of type string, but ${other.toString()}")
+                }
+                throw BadRequestException(s"Sipi returned an HTTP 200 status code, but an unsuccessful status code ${statusCode} with error message: '${sipiErrorMessage}'")
```
benjamingeer: This doesn't make sense to me:

```
Sipi returned an HTTP 200 status code, but an unsuccessful status code ${statusCode}
```

If the status code was 200, then it was successful, so why do you say it was unsuccessful?

tobiasschweizer: Yes, this is indeed confusing! 

It means that Sipi did return a HTTP 200 status code, but in it JSON message, it sends a status which is not 0.

Instead, it should sent a non 200 status code if there was a problem.
I will have a look at it, I think Lukas made a function for that.

tobiasschweizer: https://github.com/dhlab-basel/Sipi/issues/66

benjamingeer: I don't see why Sipi sends another status code besides the HTTP status code. I think the HTTP status code should be enough. Can you ask him if he can get rid of the other status code, and just use HTTP status codes?

tobiasschweizer: I think it is the same reason why we are having an extra status code in Knora, too ;-)

tobiasschweizer: fyi: I do not know that reason :-)

benjamingeer: We can't get rid of it in Knora because it's part of API v1, but since no one is using Sipi yet, perhaps we can still get rid of it there.

benjamingeer: https://github.com/dhlab-basel/Sipi/issues/67

tobiasschweizer: Since the routes for Sipi are written in Lua, the user can freely define the JSON to be returned. We can just change the Lua scripts related to Knora: https://github.com/dhlab-basel/Sipi/tree/develop/scripts

For the status code, there is a method `sendStatus` which can be called in Lua:

```
-- if check returns false, the user's input is invalid
if not check then
    result = {
        status = 1,
        message = "Mimetypes and/or file extension are inconsistent."
    }

    jsonstr = server.table_to_json(result)

    server.sendStatus(400)
    server.print(jsonstr)

    return
end
```

I will change the Lua scripts using the pull request https://github.com/dhlab-basel/Sipi/pull/65:
- get rid of extra status code in JSON msg
- send an apt HTTP status code in case of an error using `server.sendStatus`

tobiasschweizer:
Note for myself: check GUI case callback handling (status is missing now)

benjamingeer:
Looks OK to me!

tobiasschweizer:
Ok, I am about to merge. Running the tests locally right now.

tobiasschweizer:
Mööööööörtsching now

PR #288, opened 2016-10-20
relative paths

loicjaouen:
related to :
https://github.com/dhlab-basel/Knora/pull/191

Sorry @benjamingeer I only thought later about `less`.


loicjaouen:
sure, here you are

Issue #138, opened 2016-05-18
Which things in knora-base can/must be used as is?


benjamingeer:
Which classes and properties in `knora-base` can/must be used as they are, and which ones can/must be used via subclasses/subproperties? For example:
- `Region`
- `Annotation`
- `LinkObj`
- `ExternalResource`
- `ListNode`
- `isPartOf`
- `isAnnotationOf`

@lrosenth 

Needs #354.

mrivoal:
As this issue came out again last week, I would like to insist on it.

My point is the following: several projects have already expressed their needs to **customize a `Region`**. They would like to define themselves the relevant properties for what would be a `Region` (or a subclass of a `Region`) in their project. Basically, they want to be able to define a property pointing to a hierarchical list for a `Region`, which is currently impossible.

I definitely think that the API v2 and Salsah 2.0 should provide such features in the near future


For the classes `LinkObj`, `Annotation`, it is still not clear to me what they are meant for, so I won't say anything. But regarding `isPartOf`, a project-specific subproperty of `isPartOf` is generally used.

benjamingeer:
Definitely people need to be able to extend these entities. I would be inclined to say that we should require all these entities to be extended rather than used directly. I will talk to @lrosenth about this and see if we can set a clear policy.

benjamingeer:
Closed by mistake.

mrivoal:
What is the status for this issue? Has a policy been defined already? I am asking only because of `Region`, which I would really like to extend. What would it take to you to allow it?

mrivoal:
> It looks to me like the only place in Knora where `knora-base:Region` is really hard-coded is in the `getRegions` SPARQL query in API v1, which is used only in the `reqtype=context` request. I think this could easily be changed to recognise subclasses of `knora-base:Region`.

Do think that we would then have to change the existing data using `knora-base:Region` as well as project ontologies so that they include a project specific region which a subclass of `knora-base:Region`? Or do you plan to support the current schema as well?

> However, there's a lot of hard-coded stuff about regions in the JavaScript code of SALSAH 1.5, and I don't know whether it would be feasible to change it.

No idea. But I guess if we really want to change Salsah 1.5 that will be on us. So, if possible, we will more likely wait for Knora-ui to support regions (including subclasses of `knora-base:Region`).

tobiasschweizer:
> So, if possible, we will more likely wait for Knora-ui to support regions (including subclasses of `knora-base:Region`).

[...]

PR #289, opened 2016-10-21
docs: Convert Knora base ontology doc to RST.

benjamingeer:
This converts the LaTeX file documenting the Knora base ontology to reStructuredText, and integrates it with the rest of the Knora documentation.


docs/rst/knora-ontologies/knora-base.rst
```diff
@@ -0,0 +1,1544 @@
...
+
+.. contents:: :local:
+
+Introduction
+============
+
+Resource Description Framework (RDF)
+------------------------------------
+
+Knora uses a hierarchy of ontologies based on the Resource Description
```
tobiasschweizer: We used to explain what Knora stands for: Knowledge Organization, Representation, and Annotation, http://knora.org

But maybe since this is not a standalone document anymore, we can assume that the reader has learned about that in another section.

docs/rst/knora-ontologies/knora-base.rst
```diff
@@ -0,0 +1,1544 @@
...
+The recommended way to store text with markup in Knora is to use
+Knora's built-in support for “standoff” markup, which is stored
+separately from the text. This has some advantages over embedded markup
+such as XML. [#]_ While XML requires markup to have a hierarchical
+structure, and does not allow overlapping tags, standoff nodes do not
+have these limitations (Schmidt2016_). A standoff tag can be attached to
+any substring in the text by giving its start and end positions. [#]_
+For example, suppose we have the following text:
+
+    This sentence has overlapping visual attributes.
```
tobiasschweizer: This sentence should be formatted (see
screen shot 2016-10-24 at 09 19 11
)

benjamingeer:
I did the initial conversion with pandoc, then converted the diagrams and references manually.

> What happened to the TODOs (including their reference to open issues)?

I disabled the Sphinx TODO plugin several months ago because it makes a mess on the front page of the documentation (for each TODO, it shows the absolute path to the original file on the computer where the documentation was generated).

There were just two TODOs left in `knora-base.tex`:

* Issue #138. I think it's enough to have an issue open for this.
* Give more details on how `ExternalResValue` works. I think first we have to implement this functionality in the Knora API server. We could open an issue for that.

> graphviz has to be installed on the system first

I'll add something about that to the README.

> We used to explain what Knora stands for: Knowledge Organization, Representation, and Annotation, http://knora.org
>
> But maybe since this is not a standalone document anymore, we can assume that the reader has learned about that in another section.

Yes, it's already explained in `what-is-knora.rst`.

> This sentence should be formatted

Unfortunately that's not possible in RST, because [inline markup cannot be nested](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#inline-markup). That's why I changed the following sentence to explain the intended effect.

tobiasschweizer:
Ok, I fixed the formatting using raw HTML.

tobiasschweizer:
In general, the conversion to LaTeX seems to have some problems (TOC only given for the ontology documentation, but with missing page numbers; problems with images; problems with Unicode chars)

benjamingeer:
Maybe we should just write all the documentation directly in HTML. :)

benjamingeer:
I added issue #293 for `ExternalResource` and `ExternalResValue`.

tobiasschweizer:
möörge

PR #300, opened 2016-10-31
Add script to update docs on knora.org

benjamingeer:
This adds a script that you can run manually to update the documentation on knora.org when the RST or generated API docs have changed.

The script assumes that you're on the `develop` branch and have already committed any changes you made there. It builds the docs and updates the `gh-pages` branch with the resulting HTML.

In the future we could automate this to run every time `develop` is updated.

### DIFF
--- a/docs/update-gh-pages.sh
+++ b/docs/update-gh-pages.sh
@@ -9,9 +9,16 @@
 # If you don't want git to ask you for your username and password, use SSH instad of HTTPS
 # to clone the Knora repository.
 
-TMP_HTML="/tmp/knora-html"
-
 set -e
+
+TMP_HTML="/tmp/knora-html" # The temporary directory where we store built HTML.
+CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD) # The current git branch.
+
+if [ "$CURRENT_BRANCH" != "develop" ]
+then
+    echo "Current branch is $CURRENT_BRANCH, not develop"
+    exit 1
+fi
 
 # Build the HTML docs.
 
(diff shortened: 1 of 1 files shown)


docs/update-gh-pages.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# If you don't want git to ask you for your username and password, use SSH instad of HTTPS
+# to clone the Knora repository.
+
+TMP_HTML="/tmp/knora-html"
+
+set -e
+
+# Build the HTML docs.
+
```
tobiasschweizer: add a check that the current branch is `develop`:
```
branch=`git rev-parse --abbrev-ref HEAD`
``` 
http://stackoverflow.com/questions/1593051/how-to-programmatically-determine-the-current-checked-out-git-branch

docs/update-gh-pages.sh
```diff
@@ -0,0 +1,48 @@
...
+rm -rf $TMP_HTML
+mkdir $TMP_HTML
+cp -R _build/html $TMP_HTML/manual
+cp -R _format_docu $TMP_HTML/api
+
+# Switch to the gh-pages branch and remove the existing HTML docs from it.
+
+git checkout gh-pages
+rm -rf ../documentation/manual
```
tobiasschweizer: You are just deleting all the files. However, they remain under version control appearing as missing in git. I think you should use `git rm`. Imagine the case that someone removes a file from the docs. It will appear as missing in `gh-pages` although it is not longer under version control in the `docs` folder.

tobiasschweizer:
If I understand correctly, the branch `gh-pages` contains the built docs as if they were conventional files under version control.

Your script builds the docs, switches the local branch to `gh-pages`, takes the current docs as the new version and pushes everything to `gh-pages`. 

I assume that github pages is configured in a way that it pulls automatically whenever a new commit is pushed.

PR #301, opened 2016-11-01
Use inference to optimise searches on GraphDB

benjamingeer:
When we added support for searching for subclasses and subproperties in #283, there was a performance penalty in some cases. This pull requests optimises those cases by using inference on GraphDB.

For most of the existing search tests, the performance improvement is small, but it's significant in a few cases.

Before:

```
[info] - should return 19 books when we search for all books in the Incunabula test data (321 milliseconds)
```

After:

```
[info] - should return 19 books when we search for all books in the Incunabula test data (9 milliseconds)
```

Before:

```
[info] - should return 79 pages when we search for all pages that have a sequence number greater than 450 in the Incunabula test data (189 milliseconds)
```

After:

```
[info] - should return 79 pages when we search for all pages that have an incunabula:seqnum greater than 450 in the Incunabula test data (57 milliseconds)
```

I also added support for searching for subproperties of link properties, which we forgot to do in #283.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/store/triplestore/embedded/JenaTDBActor.scala
+++ b/webapi/src/main/scala/org/knora/webapi/store/triplestore/embedded/JenaTDBActor.scala
@@ -118,7 +118,7 @@ class JenaTDBActor extends Actor with ActorLogging {
       * method first returns `Failure` to the sender, then throws an exception.
       */
     def receive = {
-        case SparqlSelectRequest(sparqlSelectString) => future2Message(sender(), executeSparqlSelectQuery(sparqlSelectString), log)
+        case SparqlSelectRequest(sparqlSelectString, useInference) => future2Message(sender(), executeSparqlSelectQuery(sparqlSelectString, useInference), log)
         case SparqlUpdateRequest(sparqlUpdateString) => future2Message(sender(), executeSparqlUpdateQuery(sparqlUpdateString), log)
         case ResetTriplestoreContent(rdfDataObjects) => future2Message(sender(), resetTripleStoreContent(rdfDataObjects), log)
         case DropAllTriplestoreContent() => future2Message(sender(), Future(dropAllTriplestoreContent()), log)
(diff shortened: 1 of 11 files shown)


webapi/src/main/twirl/queries/sparql/v1/searchExtended.scala.txt
```diff
@@ -27,14 +27,16 @@
  * Performs an extended search. The number of rows returned per matching resource is the product of the number of
  * values that each search criterion matched in the resource.
  *
+ * If the triplestore type is GraphDB, the generated SPARQL is optimised by taking advantage of forward-chaining
+ * RDFS inference.
+ *
```
tobiasschweizer: Please put more information here:
- whenever the SPARQL is generated depending on the variable `@triplestore` and it is `graphdb`, inference is used
- however, we are only using inference to do certain things (getting subclasses and subproperties of the explicitly given class or property as well), but when we want to know which property actually macthed, we have to turn of inference in order to know ("GRAPH <http://www.ontotext.com/explicit>")
- if inference is not used (fuseki), we still have to simulate inference using `subclassof` or `subpropertyOf` and a property path expression + or *

webapi/src/main/scala/org/knora/webapi/store/triplestore/embedded/JenaTDBActor.scala
```diff
@@ -132,9 +132,10 @@ class JenaTDBActor extends Actor with ActorLogging {
       * Submits a SPARQL query to the embedded Jena TDB store and returns the response as a [[SparqlSelectResponse]].
       *
       * @param queryString the SPARQL request to be submitted.
+      * @param useInference not supported.
```
tobiasschweizer: Does the parameter `useInference` ever take any effect then? If I understand correctly, inference is always activated for graphdb (unless "GRAPH <http://www.ontotext.com/explicit>" is present) but not for fuseki?

I understood that inference is handled in the SPARQL queries themselves, depending on the triplestore used. So I do not understand the usefulness of this parameter sent to the store manager.

webapi/src/main/twirl/queries/sparql/v1/searchExtended.scala.txt
tobiasschweizer: We should put that information in `/searchExtended.scala.txt `, `searchFulltext.scala.txt `, and `getResourceSearchResult.scala.txt `

webapi/src/main/scala/org/knora/webapi/store/triplestore/embedded/JenaTDBActor.scala
benjamingeer: GraphDB accepts an `infer` HTTP parameter to enable or disable inference. The `useInference` parameter is used for that. However, it is not sufficient, because setting `infer=false` does not turn off the [owl:sameAs optimisation](http://graphdb.ontotext.com/documentation/standard/sameas-optimisation.html). Therefore, to completely disable inference for a query in GraphDB, we must also use `FROM <http://www.ontotext.com/explicit>` in the SPARQL.

PR #311, opened 2016-11-07
Make TransformData fix “test” labels on regions

benjamingeer:
Before SALSAH correctly supported labels, labels were hard-coded as `test`, and this affected regions created by the BEOL project. This pull request adds a `regionlabels` transformation to `TransformData`, which looks for regions whose `rdfs:label` is `test`, and replaces each one's label with the `valueHasString` of its `hasComment`.


tobiasschweizer:
Great, please merge!

PR #326, opened 2016-11-16
Return HTML instead of JSON when requested.

### LINKED ISSUE #314: HTML representation of a resource does not work
When requiring an HTML representation of a resource, JSON is served instead of HTML.

Example: <http://localhost:3333/v1/resources.html/http%3A%2F%2Fdata.knora.org%2Fc5058f3a?noresedit=true&reqtype=properties> returns JSON. The R2R tests pass though.

Response header: 
```
HTTP/1.1 200 OK

Server: akka-http/2.4.11

Date: Tue, 08 Nov 2016 15:42:54 GMT

Content-Type: application/json

Content-Length: 9510

```

Please note that the encoding is missing, see #313

benjamingeer:
This implements `RouteUtilV1.runHtmlRoute` to return HTML instead of JSON, and refactors RouteUtilV1 a bit to simplify it.

Closes #314.


tobiasschweizer:
Ok, this fixes the problem I reported!

PR #330, opened 2016-11-21
Fix various inconsistencies in knora-base.ttl and docs.

### LINKED ISSUE #327: JDC or JDN?
We have small inconsistency in our ontology.

In the docs, we speak of Julian Day Numbers (JDN) while we call them Julian Day Count (JDC) in the triplestore.

The difference is that JDNs represent a whole day while JDCs also contain the time of the day as a fraction. However, our JDC props have an object constraint of integer which does not allow for fractions.

### LINKED ISSUE #328: Knora Base Ontology and documentation
- Docu.
	- "Representations": "… resource that has file values must belong to [one] of the subclasses of kb:Representation …"
	- "Standard Resource Classes": proposal to add kb:Representation and kb:Person with a short explanation, and put this section before "Representations".
	- "StandoffLinkTag": "kb:standoffTagHasLink (1)" can be grey, conform.

- kb:hasComment:
	- onto.: Proposal to rephrase the comment: "Represents a comment [as an object (non-literal)]".
	- docu: "Standard Resource Classes": kb:Annotation: "The hasComment property points to the text [object (non-literal)] of the annotation."

- kb:Representation: kb:filename: is this used?

- kb:TextRepresentation: should there be a cardinality restriction of 1-n on property kb:hasTextFileValue?

- kb:DocumentRepresentation: should the restriction be on kb:has[Document]FileValue?

- kb:DocumentFileValue: should there be a cardinality restriction of 0-n on property kb:hasComment?

- kb:MovingImageFileValue: should there be a cardinality restriction of 1 on property kb:qualityLevel?

- Can default permissions also be attributed to a property, e.g. kb:hasStillImageFileValue? If so, should the other similar properties get similar permissions?

- kb:User:
	- Assuming foaf:familyName and foaf:givenName needing to be related to kb:objectCannotBeMarkedAsDeleted (as the rest of the properties with a kb:User subject): a similar solution as for rdfs:label can be applied. A solution can be [kb:familyName rdfs:subPropertyOf foaf:familyName, kb:objectCannotBeMarkedAsDeleted.].
	- kb:isInGroup:
		- rdfs:subPropertyOf foaf:member: impossible since foaf has the inverse meaning of "has member". Can be declared: kb:isMemberOf owl:inverseOf foaf:member. kb:isInGroup rdfs:subPropertyOf kb:isMemberOf.
		- Should the cardinality be 1-n instead of 0-n, assuming a user should be in at least one user group?
	- kb:isInProject:
		- is a project considered a group, because declared as rdfs:subPropertyOf foaf:member, leading to e.g. kb:knoraProject rdfs:subClassOf foaf:Group. For the rest the same issue as for kb:isInGroup.
	- kb:currentproject:
		- rename it to kb:currentProject ?
		- make it rdfs:subPropertyOf kb:isInProject ?
		- cf. comment: why only 1 active project ?
			- if kept -> add cardinality 1 in restriction on it for kb:User ?

- kb:SystemUser
	- is a kb:User named "Knora" -> is kb:SystemUser a foaf:Person ?
	- need to add rdf:type owl:NamedIndividual ?

- kb:UserGroup and kb:Institution: if foaf:name to be related to kb:objectCannotBeMarkedAsDeleted: similar as for foaf:familyName.

- Rename kb:protocol to kb:internetProtocol ?

- kb:Person:
	- kb:hasFamilyName rdfs:subPropertyOf foaf:familyName BUT an object of the former is a non-literal and an object of the latter is a literal, cf. comment of foaf:name. One is entitled to use the subject and object of the sub-property with the property, generating a conflict. Same for kb:hasGivenName and foaf:givenName.
	- kb:hasBirthDate rdfs:subPropertyOf foaf:birthday: also here the first is an object property and the second has a literal object. Additionally foaf only expresses month and day.

- kb:valueHasGeometry: comment: "… [squartes -> squares] …"

- kb:valueHasGeonameCode kb:objectDatatypeConstraint: in ontology xsd:string vs. in docu. xsd:integer

- kb:fps: proposal to add label "Frames per second"

- kb:valueHasListNode:
	- rdf:type [owl:DatatypeProperty -> owl:ObjectProperty]
	- delete [rdfs:subPropertyOf kb:valueHas]

benjamingeer:
Fixes inconsistencies in `knora-base.ttl`:

* Add datatype properties `knora-base:name`, `knora-base:givenName`, and `knora-base:familyName`, as subproperties of the corresponding `foaf` properties and of `knora-base:objectCannotBeMarkedAsDeleted`, and use the `knora-base` properties instead of the `foaf` ones, so consistency rules can check them properly.
* Change the type of `knora-base:valueHas` to `rdf:Property`, so its subproperties can be either object properties or datatype properties.
* Make `valueHasListNode` an object property.
* Remove incorrect cardinalities for `hasComment` on value classes.
* Fix cardinalities on `DocumentRepresentation` and `TextRepresentation`.
* Remove unused properties `currentproject`, `filename`, and `protocol`.
* Don't use `foaf:birthday`, whose object should be a literal.
* Don't use `foaf:member`, because it means the opposite of what was intended.
* Improve comments and labels.
* Use the term Julian Day Number instead of Julian Day Count (#327).

Test data and tests are fixed accordingly. There are also some corrections and clarifications in `knora-base.rst`.

Closes #327.
Closes #328.


knora-ontologies/knora-base.ttl
```diff
@@ -1473,7 +1459,7 @@
 
 ###  http://www.knora.org/ontology/knora-base#valueHas
 
-:valueHas rdf:type owl:DatatypeProperty ;
+:valueHas rdf:type rdf:Property ;
```
tobiasschweizer: Why would a subproperty of `valueHas` be an object property? Wouldn't it rather be a subproperty of `hasLinkTo` then?

knora-ontologies/knora-base.ttl
```diff
@@ -1637,7 +1623,7 @@
 
 ###  http://www.knora.org/ontology/knora-base#valueHasListNode
 
-:valueHasListNode rdf:type owl:DatatypeProperty ;
+:valueHasListNode rdf:type owl:ObjectProperty ;
```
tobiasschweizer: Is this change the answer to my question above?

knora-ontologies/knora-base.ttl
```diff
@@ -2730,11 +2700,7 @@
 
 :TextFileValue rdf:type owl:Class ;
```
tobiasschweizer: Are we going to store a document in the triplestore itself or are we going to store a file path?

I think in case of an XML file or any other simple plain text format we could store it in the triplestore, but not other non trivial formats.

webapi/_test_data/responders.v1.SipiResponderV1Spec/incunabula-data.ttl
```diff
@@ -9,40 +9,110 @@
 @prefix salsah-gui: <http://www.knora.org/ontology/salsah-gui#> .
```
tobiasschweizer: I do not understand what you did in this file.

tobiasschweizer: Or did you transform it using Transform data to change jdc to jdn?

webapi/src/main/scala/org/knora/webapi/util/TransformData.scala
```diff
@@ -254,8 +260,6 @@ object TransformData extends App {
 
     /**
       * Looks for regions that have the label "test" and changes their label to the value of their `knora-base:hasComment`.
-      *
```
tobiasschweizer: Why did you remove the documentation of the required param? Because it is already documented in the abstract super class?

knora-ontologies/knora-base.ttl
```diff
@@ -724,9 +735,7 @@
 
              :subjectClassConstraint :User ;
 
-             :objectClassConstraint :knoraProject ;
-
-             rdfs:subPropertyOf foaf:member .
```
benjamingeer: Yes, as the pull request description says: "Don't use `foaf:member`, because it means the opposite of what was intended."

knora-ontologies/knora-base.ttl
benjamingeer: For example, `valueHasListNode` has to be an object property, because its object is a`ListNode`, not a literal. It can't be a subproperty of `hasLinkTo`, because both the subject and object of `hasLinkTo` must be instances of `Resource`.

knora-ontologies/knora-base.ttl
benjamingeer: Yes.

knora-ontologies/knora-base.ttl
benjamingeer: It would make sense to store it in the triplestore if we could search its contents using full-text search. I think Lucene can be configured to handle XML syntax, but we'd have to look into this. Otherwise, I think it makes more sense to store it in a file.

webapi/_test_data/responders.v1.SipiResponderV1Spec/incunabula-data.ttl
benjamingeer: Yes.

webapi/src/main/scala/org/knora/webapi/util/TransformData.scala
benjamingeer: Yes, but I'll put it back.

PR #331, opened 2016-11-23
Fix language switching and login/logout in SALSAH

### LINKED ISSUE #316: Switch language
When we want to switch language from another, we get the following error after we have been asked for the confirmation:

capture d ecran 2016-11-11 a 12 04 21

The API request answers :
```
    Request URL:http://platec.unil.ch/knora/v1/userdata
    Request Method:POST
    Status Code:404 Not Found
```

benjamingeer:
This fixes language switching in SALSAH. You can click on the language links at the top right corner of the GUI, and the language of GUI prompts changes. This doesn't affect the language of data (such as property labels), which still depend on the user's preferred language (stored in the user's `knora-base:User` in the triplestore).

Closes #316.


salsah/src/public/index.html
```diff
@@ -329,6 +327,34 @@
...
+	
+	$(function(){
+		if (urlparams["lang"] !== undefined) {
+    		SALSAH.userdata = {lang: urlparams["lang"]};
+		}
+		else {
+    		SALSAH.userdata = {lang: 'en'};
+		}
+		$.get('lang/' + SALSAH.userdata.lang + '.json', function(data) {
+    		console.log("Getting language strings: " + SALSAH.userdata.lang)
```
tobiasschweizer: comment out the console output :-)

tobiasschweizer:
When asking the user if he really wants to change the language of the GUI he should be warned that the page is going to be reloaded (open windows and unsaved changes will be lost).

tobiasschweizer:
Have you run the browser tests?

benjamingeer:
I added a browser test for changing the language, and the other tests pass, too.

PR #332, opened 2016-11-24
Allow connections over HTTPS

benjamingeer:
This provides:

* Configuration options for enabling HTTPS connections to the Knora API server, in addition to or instead of HTTP connections.
* Scripts for creating a self-signed certificate for `localhost` for testing.
* Docs.


benjamingeer:
Thanks for looking at this.

> http = true; https = true: it fails

Sorry about that. It's surely because the HTTPS implementation doesn't recognise the fake test CA. I've changed `E2ESpec` so it only uses HTTP.

> http = false; https = true

That was a typo, fixed now.

PR #336, opened 2016-11-29
Optimise queries for GraphDB and refactor query optimisation

benjamingeer:
This optimises more SELECT queries using inference on GraphDB. In particular, `getResourcePropertiesAndValues.sparql.txt` is about five times faster.

I've also refactored query optimisation by making separate templates for standard SPARQL queries and GraphDB-specific queries, so they're easier to read.


tobiasschweizer:
Please have a look at the changes I made in commit: https://github.com/dhlab-basel/Knora/pull/336/commits/ee1fcd6aaf7223890be9e2f85dda0cca6a26b792

webapi/src/main/twirl/queries/sparql/v1/getFileValueStandard.scala.txt
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        BIND(knora-base:attachedToProject AS ?objPred)
+        BIND(COALESCE(?valueProject, ?resourceProject) AS ?objObj)
+    }
+    UNION
+    {
+        @* Return the base path of whichever project was found. *@
+
+        ?prop rdfs:subPropertyOf* knora-base:hasFileValue .
```
tobiasschweizer: Couldn't we use the plus operator here since `hasFileValue` should not be used directly but only subproperties of it? How could we deal with that for GraphDB (so is semantically equivalent because the inferred query seems to be the same as using the * operator)?

webapi/src/main/twirl/queries/sparql/v1/getResourceInfoGraphDB.scala.txt
```diff
@@ -0,0 +1,96 @@
...
+WHERE {
+    BIND(IRI("@resourceIri") AS ?resource)
+
+    ?resource rdf:type knora-base:Resource ;
+        knora-base:isDeleted false .
+
+    {
+        @* Query statements about the resource's file values. *@
+
+        ?resource knora-base:hasFileValue ?obj .
```
tobiasschweizer: this would behave differently than the Sparql in `getResourceInfoStandard.scala.txt` because there the plus operator is used.

webapi/src/main/twirl/queries/sparql/v1/getResourceInfoStandard.scala.txt
```diff
@@ -0,0 +1,86 @@
...
+
+    ?resource rdf:type ?resourceClass ;
+        knora-base:isDeleted false .
+
+    ?resourceClass rdfs:subClassOf+ knora-base:Resource .
+
+    {
+        @* Query statements about the resource's file values. *@
+
+        ?prop rdfs:subPropertyOf+ knora-base:hasFileValue .
```
tobiasschweizer: see comment above

webapi/src/main/twirl/queries/sparql/v1/getResourcePropertiesAndValuesGraphDB.scala.txt
```diff
@@ -0,0 +1,123 @@
...
+WHERE {
+    BIND(IRI("@resourceIri") as ?resource)
+
+    ?resource rdf:type knora-base:Resource ;
+        knora-base:isDeleted false ;
+        knora-base:attachedToProject ?resourceProject .
+
+    {
+        @* Get properties that point to Value objects. *@
+        ?resource knora-base:hasValue ?obj .
```
tobiasschweizer: same question regarding plus operator in statement  `?prop rdfs:subPropertyOf+ knora-base:hasValue`

webapi/src/main/twirl/queries/sparql/v1/findValueInVersions.scala.txt
```diff
@@ -58,7 +58,7 @@ WHERE {
         FILTER(?objPred != knora-base:attachedToProject)
 
         OPTIONAL {
-            ?objPred rdfs:subPropertyOf* knora-base:valueHasStandoff .
+            FILTER(?objPred = knora-base:valueHasStandoff) .
```
benjamingeer: Will do.

webapi/src/main/twirl/queries/sparql/v1/getContextGraphDB.scala.txt
```diff
@@ -0,0 +1,136 @@
...
+ *@
+
+@import org.knora.webapi.IRI
+
+@**
+ * Given the IRI of a resource, gets information about the resources that are part of that resource, using GraphDB.
+ *
+ * This template is used only by getContext.scala.txt.
+ *
+ * Since the triplestore type is GraphDB, we assume that inference is enabled, and we use it to optimise the generated
```
benjamingeer: We can't. Using `KnoraRules.pie`, which sets up inference rules and consistency checks, is simply a requirement when creating a repository with GraphDB. I can document this.

webapi/src/main/twirl/queries/sparql/v1/getFileValueGraphDB.scala.txt
```diff
@@ -0,0 +1,85 @@
...
+prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+prefix knora-base: <http://www.knora.org/ontology/knora-base#>
+
+SELECT ?objPred ?objObj
+WHERE {
+
+    ?fileValue knora-base:internalFilename "@filename" ;
+        knora-base:isDeleted false .
+
```
benjamingeer: OK.

webapi/src/main/twirl/queries/sparql/v1/getFileValueStandard.scala.txt
benjamingeer: I don't think that these SPARQL queries can be responsible for preventing people from using abstract properties directly. Instead, I think we need a mechanism for preventing such properties from being inserted in the first place; that's issue #354. I'll implement it in another pull request.

If someone has managed to explicitly insert `hasFileValue`, I don't think that silently ignoring it is necessarily a good idea. Perhaps it would be better to return it to the API server and let it cause an error. So now I'm actually inclined to replace all the plus operators with asterisks.

tobiasschweizer: > So now I'm actually inclined to replace all the plus operators with asterisks.

At least this would make them equivalent to the queries that use inference.

webapi/src/main/twirl/queries/sparql/v1/findValueInVersions.scala.txt
benjamingeer: Actually, the SPARQL template you're commenting on doesn't have anything to do with `ValueVersionHistoryGetRequestV1`. It's used only to check whether an update succeeded. So the existing tests that create and change values with standoff already cover it.

benjamingeer:
OK now?

webapi/src/main/twirl/queries/sparql/v1/getFileValueGraphDB.scala.txt
tobiasschweizer: Do you want to add the check if the resource is marked as deleted?

webapi/src/main/twirl/queries/sparql/v1/findValueInVersions.scala.txt
tobiasschweizer: ok, fine

PR #343, opened 2016-12-06
Upgrade to Scala 2.12

### LINKED ISSUE #341: Switch from Dispatch to Akka HTTP Client and upgrade to Scala 2.12
I would like to upgrade Knora to Scala 2.12.

The [Dispatch](https://github.com/dispatch/dispatch) HTTP client library seems to be the only library we use that isn't available for Scala 2.12. Dispatch has been unmaintained for a while.

Since we're already using Akka HTTP, I think we could use its [HTTP Client API](http://doc.akka.io/docs/akka-http/current/scala/http/introduction.html#http-client-api) instead.

This only seems to affect three files:

* `HttpTriplestoreConnector.scala`
* `GraphProtocolAccessor.scala`
* `CompareApiResults.scala` (which we should really remove because it is no longer relevant)

benjamingeer:
* Update Scala to version 2.12.1.
* Update SBT to version 0.13.13.
* Update Akka HTTP to version 10.0.0.
* Update other libraries.
* Use Akka HTTP instead of [Dispatch](https://github.com/dispatch/dispatch), which has been unmaintained for some time and isn't available for Scala 2.12.
* Remove `CompareApiResults` (the old utility for comparing the output of the old SALSAH with the output of the Knora API server). It's no longer useful, since we now have end-to-end tests, and since we have been adapting the old SALSAH GUI rather than maintaining the old API in every detail.
* Fix various small things related to changes in the Scala compiler.

Closes #341.


subotic:
Ah, yes. It got deprecated ;-)

benjamingeer:
Yes. :) Thanks!

subotic:
Thank you for the work :-)

PR #350, opened 2016-12-09
Don't redundantly traverse cycles in graph data query

benjamingeer:
This prevents the resources responder from redundantly traversing a cycle of edges in the graph data query.


tobiasschweizer:
This seems fine.

What happens if a resource links to itself? Do we prevent this?

A small problem loading a js file: 
> http://localhost:3335/js/cola.min.js.map Failed to load resource: the server responded with a status of 404 (Not Found)

tobiasschweizer:
in `cola.min.js`, it is commented out:
> //# sourceMappingURL=cola.min.js.map

benjamingeer:
> What happens if a resource links to itself? Do we prevent this?

Opened #356 for this.

I couldn't find `cola.min.js.map` anywhere, so I'm switching to the non-minified version of `cola.js`.

PR #358, opened 2016-12-21
Display a compound resource without images

### LINKED ISSUE #334: SALSAH GUI can't handle isPartOf with no images
If resource A is part of resource B, and A has no image, and you try to view B, the GUI tries to make an image viewer (as if A was a page in book B), but this doesn't work if A doesn't actually have an image:

excluded-alpha

benjamingeer:
If a compound resource lacks images (like "excluded alpha" in the `anything` test data), display it with a default "no image" image.

Fixes #334.


tobiasschweizer:
Could you add a browser test for this?

benjamingeer:
See example no. 3:

http://www.phdcomics.com/comics/archive.php?comicid=581

benjamingeer:
Done, is this OK?

PR #360, opened 2016-12-22
build (webapi): add deployment packaging

subotic:
### Summary

Steps to create and run the deployment package:

```
 $ sbt stage
 $ cd ./target/universal/stage
 $ ./bin/webapi
```

This can be used both for deploying `webapi` and `salsah`.

Further, there are now additional settings in `application.conf` that can be used to configure `salsah` for deployment. Some of these settings are used to rewrite ``js/00_init_javascript.js` at startup.

- add `stage` to `webapi` and `salsah`
- add additional settings to `salsah`'s `application.conf`
- rename `webapi` repository initialization scripts

### Issues

 - resolves #360


subotic:
Thanks :-)

> Please add something to the RST

Will do.

> Why is triplestores/graphdb-se/Dockerfile here? I thought we removed it in #446.

No idea. Shouldn't be. I will remove it.

subotic:
> In graphdb-se-local-init-knora-test.sh, there is a reference to graphdb-se-knora-test-load-data.ttl. I think it should be graphdb-se-knora-test-data.ttl.

Yes, good catch :-)

subotic:
this is again ready for review. thanks :-)

PR #367, opened 2017-01-05
fix (webapi): Fix permission checking on link values

### LINKED ISSUE #366: Error when hovering over the info box of "has standoff link to"
When hovering over the info box 
infobox
  of a "has standoff link to", an error occurs:

hasstandofflinktoinfobox

This is because a `LinkValue` is queried but it (always) belongs to the system user (and not the user that created the reference or something like that):

linkvalue

So Knora says:

```
{

    "status": 3,
    "error": "org.knora.webapi.ForbiddenException: User http://data.knora.org/users/PSGbemdjZi4kQ6GHJVkLGF does not have permission to see value http://data.knora.org/-umMNyKDSKy7c5m67rbLow/values/_xvRS7epQjiigZIB8dGTbQ",
    "access": "NO_ACCESS"

}
```

I think we handle this as follows: if the user has the right to see the target resource, he can see the link, otherwise not.

benjamingeer:
Fixes #366.

The docs say:

> Link values created automatically for resource references in standoff are automatically visible to all users, as long as they have permission to see the source and target resources.

When the values responder loads a link value, it must:

* Check the link value's permissions using `PermissionUtilV1.getUserPermissionOnLinkValueV1WithValueProps`, which takes into account the special case of standoff links.
* Check that the user has permission to view the source and target resources.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
@@ -1230,7 +1230,7 @@ class ValuesResponderV1 extends ResponderV1 {
     @throws(classOf[NotFoundException])
     private def getLinkValue(subjectIri: IRI, predicateIri: IRI, objectIri: IRI, userProfile: UserProfileV1): Future[ValueGetResponseV1] = {
         for {
-            maybeValueQueryResult <- findLinkValueByObject(
+            maybeValueQueryResult <- findLinkValueByLinkTriple(
                 subjectIri = subjectIri,
                 predicateIri = predicateIri,
                 objectIri = objectIri,
@@ -1320,6 +1320,7 @@ class ValuesResponderV1 extends ResponderV1 {
       * @param targetResourceClass          if a direct link exists, contains the OWL class of the target resource.
       */
     case class LinkValueQueryResult(value: LinkValueV1,
+                                    linkValueIri: IRI,
                                     ownerIri: IRI,
                                     creationDate: String,
                                     projectIri: IRI,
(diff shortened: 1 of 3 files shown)


webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
```diff
@@ -1399,11 +1466,20 @@ class ValuesResponderV1 extends ResponderV1 {
 
             response <- (storeManager ? SparqlSelectRequest(sparqlQuery)).mapTo[SparqlSelectResponse]
             rows: Seq[VariableResultsRow] = response.results.bindings
-        } yield sparqlQueryResults2LinkValueQueryResult(rows, userProfile)
+            maybeLinkValueQueryResult = sparqlQueryResults2LinkValueQueryResult(rows, userProfile)
```
tobiasschweizer: If I understand correctly, `sparqlQueryResults2LinkValueQueryResult` calls `PermissionUtilV1.getUserPermissionOnLinkValueV1WithValueProps` which always returns view permissions for standoff links (those are owned by the system user).

And since `PermissionUtilV1.getUserPermissionOnLinkValueV1WithValueProps` was not called before the changes you made here, no permissions could be granted to any user on a standoff link.

tobiasschweizer: However, standoff links were always returned by the resources responder (the error occurred not before hoveing over the information symbol). How does the handling of standoff links done by values responder relate to the that  performed by resources responder?

tobiasschweizer:
I know that you hate this question, but could you add a test for this (querying a LinkValue representing a standoff link)?

webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
benjamingeer: The resources responder already calls the special `PermissionUtilV1` method to check permissions on link values when necessary, so permissions on standoff links are handled correctly there. So does the search responder. I had just forgotten to use that method in one place in the values responder (because I had forgotten that you could get a link value via the normal route for getting values, not just through the special route for getting link values). This shows that having a special method just to check permissions on link values is a bad design, because it's easy to forget to use it. But I'd rather get the release done than refactor this now. Let's design it better for API v2.

benjamingeer:
Yes of course.

webapi/src/main/scala/org/knora/webapi/responders/v1/ValuesResponderV1.scala
tobiasschweizer: Ok, that's fine. I agree that  we have to get V1 out before we do that refactoring.

PR #377, opened 2017-01-18
fix (webapi): Handle request for nonexistent resource class.

### LINKED ISSUE #376: Error on non existing resource types
If I'm looking for a resource type with an IRI which doesn't exist, I get the following error:
{
"status": 4,
"error": "java.util.NoSuchElementException: key not found: http://www.knora.org/ontology/incunabula#image"
}

benjamingeer:
Fixes #376.


tobiasschweizer:
While you are at it, could you check if there are more cases likes this?

benjamingeer:
OK, I think that should take care of it. Normally you should get `InconsistentTriplestoreDataException` (because usually we get resource class IRIs from the triplestore, not from the user), but in André's case, where the user gave a nonexistent resource class IRI to a route, you'll get `NotFoundException`.

tobiasschweizer:
Yes, this is fine!

Issue #378, opened 2017-01-18
'knora-admin' and 'standoff' ontologies


hanscools:
- Would it be considered to create an own namespace for the 'knora-admin' and 'standoff' ontologies, e.g. <http://www.knora.org/ontology/standoff#>
instead of <http://www.knora.org/ontology/knora-base#>?

- knora-base:standoffTagHasTagname: proposal to change to :standoffTagHasName

- Note: some elements in 'standoff' are not in 'knora-base', e.g.
	- knora-base:standoffTagHasXMLNamespace
	- knora-base:StandoffReferenceTag
	- knora-base:StandoffHrefTag

- knora-base:systemProperty is used in the declaration of knora-base:institutionWebsite in the 'knora-admin' ontology and for 11 properties in the 'standoff' ontology, but is not declared itself.

benjamingeer:
`knora-base:systemProperty` was replaced by `:objectCannotBeMarkedAsDeleted`. @tobiasschweizer, could you please fix this in the standoff ontology?

benjamingeer:
`knora-base:systemProperty` was replaced by `knora-base:objectCannotBeMarkedAsDeleted`. @subotic, could you please fix this in the `knora-admin` ontology?

tobiasschweizer:
> Would it be considered to create an own namespace for the 'knora-admin' and 'standoff' ontologies, e.g. http://www.knora.org/ontology/standoff#
instead of http://www.knora.org/ontology/knora-base#?

Please note that standoff is still in development on the branch `StandoffUtil` (https://github.com/dhlab-basel/Knora/pull/319). There is an ontology `http://www.knora.org/ontology/standoff` for all non system relevant standoff definitions. However, system relevant definitions are kept in `knora-base`.

> knora-base:standoffTagHasTagname: proposal to change to :standoffTagHasName

`standoff.ttl` (https://github.com/dhlab-basel/Knora/blob/develop/knora-ontologies/standoff.ttl) on `develop` is deprecated and has been removed on branch `StandoffUtil`. I forgot to remove it from develop, it is a left-over of an earlier attempt to implement the desired standoff design. Sorry.

 > Note: some elements in 'standoff' are not in 'knora-base', e.g.
> knora-base:standoffTagHasXMLNamespace
> knora-base:StandoffReferenceTag
> knora-base:StandoffHrefTag

ditto

[...]

PR #379, opened 2017-01-19
fix (knora-ontologies): remove obsolete property

subotic:
### Summary

  - removing ``knora-base:systemProperty`` from the declaration of ``knora-base:institutionWebsite``

### Issues
  - #378

### DIFF
--- a/knora-ontologies/knora-admin.ttl
+++ b/knora-ontologies/knora-admin.ttl
@@ -244,8 +244,6 @@
 
             rdfs:comment "The URL of a web site"@en ;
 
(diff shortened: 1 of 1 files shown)


subotic:
It is already there :) I also checked, that  every property has this.

knora-ontologies/knora-admin.ttl
```diff
@@ -244,8 +244,6 @@
```
subotic: ``:objectCannotBeMarkedAsDeleted`` is already present.

benjamingeer:
OK please merge!

subotic:
Thanks :)

PR #402, opened 2017-01-31
fix (knora-ontologies): Some ontology fixes

tobiasschweizer:
Fix ontology bugs relating to the new standoff and admin designs.


benjamingeer:
TODO: Fix consistency errors in `_test_data/all_data/permissions-data.ttl`.

tobiasschweizer:
https://github.com/dhlab-basel/Knora/pull/402/commits/697050db080b0d65297f3003553c5cbe40f72e9d fixes a problem that occurred when writing string values to XML attributes without escaping them first

@benjamingeer Is this the only place attributes are handled?

benjamingeer:
> Is this the only place attributes are handled?

For standoff to XML conversion, yes. For the conversion in the other direction, there's the function `xmlAttrs2StandoffAttrs`.

PR #407, opened 2017-02-03
Fix inconsistencies in incunabula-demo-data.ttl.

### LINKED ISSUE #405: Incunabula-demo-data.ttl doesn't pass the consistency checking
We can't have the incunabula-demo-data.ttl pass the consistency checking.
Is it expected?

We have fixed what appear to be several mistakes, but it looks like there are some more...

graphdb 8.0.3
Incunabula-demo-data.ttl 81dfe00dee95485d41ef58419c52994645eea080

benjamingeer:
Fixes #405.

* Ran `incunabula-demo-data.ttl` through `TransformData -t all`.
* Added missing `valueHasCreationDate`.
* Added text to fix empty strings.
* There were many references to sidebands, but no actual sidebands. Removed all references to sidebands.

### DIFF
--- a/docs/rst/introduction/example-project.rst
+++ b/docs/rst/introduction/example-project.rst
@@ -33,7 +33,7 @@ about books, pages, their contents, and relationships between them. At the momen
 only the RDF data is provided in the example project, not the images.
 
 The ``incunabula`` ontology is in the file ``incunabula-onto.ttl``, and its
-data is in the file ``incunabula-data.ttl``. Both these files are in a
+data is in the file ``incunabula-demo-data.ttl``. Both these files are in a
 standard RDF file format called Turtle_. The Knora distribution includes
 sample scripts (in the ``webapi/scripts`` directory) for importing these files
 directly into different triplestores. If you are starting a new project from


subotic:
We created it at the time when we published Knora on Github, so that a potential user would have some demo data that would load in a shorter time, then the full incunabula data.

IMHO we can remove this. We have enough ``ttl`` files that we need to maintain.  We still have ``images-demo.ttl` which could be used for a quick test.

I don't know if Lausanne is using this.

mrivoal:
No, we don't use it. 

I loaded it the last time because I got lost in the .ttl files avaiblable. I wanted to run some tests but I could have done it with incunabula-data.ttl.

benjamingeer:
It's part of the documentation:

http://www.knora.org/documentation/manual/rst/introduction/example-project.html

tobiasschweizer:
Right! Ok, it can be merged then.

loicjaouen:
i don't see a reference to `incunabula-demo-data.ttl` but `incunabula-data.ttl`.

tobiasschweizer:
`store-module.rst`:

```
            {
                path = "_test_data/demo_data/incunabula-demo-data.ttl"
                name = "http://www.knora.org/data/incunabula"
            }
```

benjamingeer:
If you're all happy for the huge `incunabula-data.ttl` to be used in the docs, I'm happy to delete `incunabula-demo-data.ttl`. One less thing to maintain.

loicjaouen:
we are particularly happy with the huge file, but we got side-tracked by the outdated small one.

It is now in sync, fine, let's keep it.
But there is no automated ways to check it, so maintaining it will difficult and chances of lags are high.

PR #418, opened 2017-02-13
Remove knora-base:attachedToProject from values

### LINKED ISSUE #417: Remove knora-base:attachedToProject from values
Currently the subject of `knora-base:attachedToProject` can be a `knora-base:Resource` or a `knora-base:Value`. As a result of #409, we decided that only instances of `knora-base:Resource` will have `knora-base:attachedToProject`. This requires:

* Changing the SPARQL templates and the Scala code that uses them
* Adding a transformation to `TransformData`
* Updating the test data
* Changing tests that check the old behaviour

benjamingeer:
This changes the code that reads and writes values in RDF, so they always get their `knora-base:attachedToProject` from the containing resource. Also:

* Rename `owner` to `creator` everywhere.
* Fix permissions in GraphDB consistency check test.
* Add transformation to `TransformData` to remove projects from values.
* Change `PermissionsResponderV1.defaultObjectAccessPermissionsStringForEntityGetV1` to return a case class message rather than an `Option[String]`.
* Make capitalisation more consistent (`fooIri` instead of `fooIRI`, constants begin with a capital letter but methods don't).

Closes #417.


benjamingeer:
@subotic There is one test failing but I don't know why, could you have a look? It's 

```
[info] - should 2b2. drawings-gods user is not allowed to create a new value inside the drawings-gods project (no permissions for this project are defined which would lead to no permissions attached to the value) *** FAILED *** (130 milliseconds)
[info]   200 OK did not equal 400 Bad Request (ResourceValuePermissionsV1E2ESpec.scala:178)
```

This is because `PermissionsResponderV1.defaultObjectAccessPermissionsStringForEntityGetV1()` is returning this instead of throwing an exception:

```
CR knora-base:ProjectAdmin|D knora-base:ProjectMember|M knora-base:KnownUser|RV knora-base:UnknownUser
```

Also, can I make that method return `String` instead of `Option[String]`?

subotic:
> This is because PermissionsResponderV1.defaultObjectAccessPermissionsStringForEntityGetV1() is returning this instead of throwing an exception

Strange. This is basically the case where someone from a project wants to create something inside his project, but since no default permissions for his project are defined, this should result in an empty permissions string, which would trigger a  ``BadRequest`` exception.

> Also, can I make that method return ``String`` instead of ``Option[String]``?

Yes, certainly. Before it was allowed to return no permissions, but we wanted to change that. Now that an exception is thrown in the ``None`` case, we can change this to a ``String``.

subotic:
>> This is because PermissionsResponderV1.defaultObjectAccessPermissionsStringForEntityGetV1() is returning this instead of throwing an exception

> Strange. This is basically the case where someone from a project wants to create something inside his project, but since no default permissions for his project are defined, this should result in an empty permissions string, which would trigger a  ``BadRequest`` exception.

Actually, now this use case isn't possible anymore (values cannot belong to another project), so we can remove this test altogether.

benjamingeer:
@subotic The Travis build is broken because I renamed some Twirl templates, and it seems that you have to type `clean` in SBT to remove the old compiled Twirl classes, but Travis doesn't do this. Do you know how to fix this?

webapi/src/main/scala/org/knora/webapi/responders/v1/PermissionsResponderV1.scala
```diff
@@ -737,23 +735,23 @@ class PermissionsResponderV1 extends ResponderV1 {
...
-      * @param resourceClassIRI the IRI of the resource class for which the default object access permissions are requested.
-      * @param propertyIRI the IRI of the property for which the default object access permissions are requested.
+      * @param projectIri the IRI of the project.
+      * @param resourceClassIri the IRI of the resource class for which the default object access permissions are requested.
+      * @param propertyIri the IRI of the property for which the default object access permissions are requested.
       * @param permissionData the permission data of the user for which the default object access permissions are requested.
       * @return an optional string with object access permission statements
       */
-    def defaultObjectAccessPermissionsStringForEntityGetV1(projectIRI: IRI, resourceClassIRI: Option[IRI], propertyIRI: Option[IRI], permissionData: PermissionDataV1): Future[Option[String]] = {
+    def defaultObjectAccessPermissionsStringForEntityGetV1(projectIri: IRI, resourceClassIri: Option[IRI], propertyIri: Option[IRI], permissionData: PermissionDataV1): Future[DefaultObjectAccessPermissionsStringResponseV1] = {
```
subotic: Do we need a special return message to just return a string?

Also, I have an informal convention that a request called ``SomethingSomethingGetRequestV1`` is returning a ``SomethingSomethingGetResponseV1``, and that ``SomethingSomethingGetV1`` is returning something else, in this case a ``String``. Before you ask, the whole distinction between ``SomethingSomethingGetRequestV1`` and ``SomethingSomethingGetV1`` is needed because you would like ``NotFound`` exceptions to be thrown inside the responder methods called from routes, which on the other hand makes them unsuitable for internal use.

So, I would suggest that we either keep the distinction between ``SomethingSomethingGetRequestV1 `` and ``SomethingSomethingGetV1 ``, so then ``DefaultObjectAccessPermissionsStringResponseV1`` would need to be renamed in something like ``DefaultObjectAccessPermissionsStringV1``, or we move the throwing of ``NotFound`` error messages to the routes, which would basically entail writing a small match statement.

benjamingeer:
> Do we need a special return message to just return a string?

A string doesn't carry any type information with it. Imagine if all our messages were just strings. You might receive the wrong type of message and never know. If you use a case class, then you get a nice `ClassCastException` if you receive the wrong type of message. So I think all messages should have types.

> So, I would suggest that we either keep the distinction between `SomethingSomethingGetRequestV1 ` and ``SomethingSomethingGetV1`, so then `DefaultObjectAccessPermissionsStringResponseV1` would need to be renamed in something like `DefaultObjectAccessPermissionsStringV1`

I like naming conventions, that seems fine to me. :) Let's mention it in the design docs.

> you would like `NotFound` exceptions to be thrown inside the responder methods called from routes, which on the other hand makes them unsuitable for internal use.

In this case, though, it's OK, because now we've decided that the permissions always have to be there, right? So the responder can return the exception even for internal use.

subotic:
>> Do we need a special return message to just return a string?

> A string doesn't carry any type information with it. Imagine if all our messages were just strings. You might receive the wrong type of message and never know. If you use a case class, then you get a nice `ClassCastException` if you receive the wrong type of message. So I think all messages should have types.

Yes, I agree. I would prefer, if possible, to call it something like `DefaultObjectAccessPermissionsStringV1`.

> I like naming conventions, that seems fine to me. :) Let's mention it in the design docs.

Cool :)

> In this case, though, it's OK, because now we've decided that the permissions always have to be there, right? So the responder can return the exception even for internal use.

Yes. When this exception is thrown, then there is really something wrong.

benjamingeer:
I tried to implement this naming convention, but the problem is that there are already a lot of messages that are only used internally and whose names end in `Response`. These are the ones I found:

* `SparqlSelectResponse`
* `SparqlConstructResponse`
* `SparqlUpdateResponse`
* `EntityInfoGetResponseV1`
* `StandoffEntityInfoGetResponseV1`
* `StandoffClassesWithDataTypeGetResponseV1`
* `StandoffAllPropertyEntitiesGetResponseV1`
* `GenerateSparqlToCreateMultipleValuesResponseV1`
* `CheckSubClassResponseV1`
* `ResourceCheckClassResponseV1`
* `VerifyMultipleValueCreationResponseV1`
* `GetMappingResponseV1`
* `SipiResponderConversionResponseV1`
* `SipiImageConversionResponse`
* `SipiTextResponse`
* `SipiErrorConversionResponse`

We could rename all these, but I don't know what else we could call them. My preference would be to leave them as they are.

> or we move the throwing of `NotFound` error messages to the routes, which would basically entail writing a small match statement.

How would this work? Currently the routes never see the reply message from the responder, only `RouteUtilV1` sees it.

subotic:
> We could rename all these, but I don't know what else we could call them.

What about leaving them but renaming the ones coming from the route to something like ``UserProfileByIRIGetRouteRequestV1`` and ``UserProfileRouteResponseV1``? I like the idea of having a naming scheme that allows us to clearly see what messages are requests/responses from/to the outside and what is only used internally, as in the outside case we need to deal with permissions and making sure we don't leak sensitive information, whereas in the internal case we can be more liberal.

>> or we move the throwing of NotFound error messages to the routes, which would basically entail writing a small match statement.

>How would this work? Currently the routes never see the reply message from the responder, only RouteUtilV1 sees it.

We could move this:
```
knoraResponse <- (responderManager ? requestMessage).map {
                case replyMessage: KnoraResponseV1 => replyMessage

                case other =>
                    // The responder returned an unexpected message type (not an exception). This isn't the client's
                    // fault, so log it and return an error message to the client.
                    throw UnexpectedMessageException(s"Responder sent a reply of type ${other.getClass.getCanonicalName}")
            }
```

from ``RouteUtilV1.runJsonRoute`` to each route, do the match and then call what is left of the ``RouteUtilV1.runJsonRoute``. To make it simpler to use, we could package it inside a trait:

```
trait RouteResponderRequest {
  def makeRequest(requestMessage: KnoraRouteRequestV1): Future[KnoraRouteResponseV1] = {
      for {
          knoraResponse <- (responderManager ? requestMessage).map {
                case replyMessage: KnoraRouteResponseV1 => replyMessage

                case other =>
                    // The responder returned an unexpected message type (not an exception). This isn't the client's
                    // fault, so log it and return an error message to the client.
                    throw UnexpectedMessageException(s"Responder sent a reply of type ${other.getClass.getCanonicalName}")
          }
      } yield knoraResponse
  }
}
```

then in the route:
```
object UsersRouteV1 extends Authenticator with RouteResponderRequest
. . .
RouteUtilV1.runRestOfJsonRoute(
     makeRequest(requestMessage),
     requestContext,
     settings,
     responderManager,
     log
)
```

subotic:
Just a thought, what if we not only have a naming scheme for requests coming/going from/to routes, but also a separate case class hierarchy? Something like:

```
trait KnoraRouteRequestV1 extends KnoraRequestV1
UserProfileByIRIGetRouteRequestV1 extends KnoraRoteRequestV1

trait KnoraRouteResponse extends KnoraResponseV1
UserProfileResponseV1 extends KnoraRouteResponse
```

All the other internally used messages are subclasses of ``KnoraRequestV1``/ ``KnoraResponseV1``.

This would make a nice barrier to the outside world.

benjamingeer:
Those all seem like good ideas to me, but could we put them in a separate issue and save them for after the beta release?

subotic:
Yes, certainly. We should also, pretty soon, start thinking about the ``V2`` implementation, and then do both at the same time.

benjamingeer:
I made #424 for that. Can I merge this now?

subotic:
Yes, please.

Issue #386, opened 2017-01-24
memory leak in test file


SepidehAlassi:
by running webapi via an IDE, intelliJ, I get the following errors

Warning:scalac: java.lang.OutOfMemoryError: Java heap space
Error:scalac: Error while emitting org/knora/webapi/responders/v1/ResourcesResponderV1SpecContextData$
Java heap space

increasing the maximum heap size of JVM by factor of 100 doesn't solve the problem either (setting Maximum heap size  to 76800!!!) 
Probably there is a memory leak in the file ResourcesResponderV1SpecContextData.

tobiasschweizer:
duplicate of https://github.com/dhlab-basel/Knora/issues/349

subotic:
``ResourcesResponderV1SpecContextData`` is very large, as it is a dump of response messages. You need to increase the memory of the JVM. Running everything inside SBT works if SBT is started with 2GB of memory.

``export SBT_OPTS="-Xms2048M -Xmx2048M -Xss6M"``

subotic:
@SepidehAlassi is this solved?

PR #421, opened 2017-02-15
replace bcrypt by scrypt

loicjaouen:
We are hashing passwords with bcrypt.

We use the implementation: http://www.mindrot.org/projects/jBCrypt/

It uses BCrypt *version* '$2a$', most recent is '$2b$', details are here (with links):
http://security.stackexchange.com/questions/20541/insecure-versions-of-crypt-hashes

The alternative libraries do not match this '$2b$' version, even http://bouncycastle.org/ .

An easy workaround is to switch to scrypt, which is better and similar to bcrypt, password-based key derivation function 
- http://www.tarsnap.com/scrypt.html
- https://en.wikipedia.org/wiki/Scrypt
- https://github.com/Tarsnap/scrypt
- java implementation: https://github.com/wg/scrypt


benjamingeer:
Thanks! Travis is reporting that the build is broken on this branch, because of compile errors in the tests. You can type `test:compile` in SBT to compile them. Then please run `graphdb:test` and make sure the tests all pass. I assume this will involve changing the password hashes in `webapi/_test_data/all_data/admin-data.ttl`.

loicjaouen:
Yes, of course, i just wanted to be sure you didn't mind this proposition as it was not exactly what we agreed upon. I'll fix the test and revert.

benjamingeer:
I'm also concerned that the Java implementation at https://github.com/wg/scrypt hasn't been updated for four years. How about using this implementation instead?

http://docs.spring.io/spring-security/site/docs/current/apidocs/org/springframework/security/crypto/scrypt/SCryptPasswordEncoder.html

https://mvnrepository.com/artifact/org.springframework.security/spring-security-core/4.2.1.RELEASE

loicjaouen:
@benjamingeer could you have a look at the code already?
previous solutions were easy plug-in replacements because they were static methods calls.

Now we have a (java):
```java
new SCryptPasswordEncoder().matches(password, hp);
```

is this the scala translation?
```scala
val encoder = new SCryptPasswordEncoder
encoder.matches(password, hp)
```

webapi/WebapiBuild.sbt
```diff
@@ -162,17 +163,17 @@ lazy val webApiLibs = Seq(
 
 lazy val javaRunOptions = Seq(
     // "-showversion",
-    "-Xms2048m",
-    "-Xmx4096m"
+    "-Xms8G",
```
benjamingeer: Are you sure you need this much memory? This seems to be making the Travis build fail, because the JVM can't allocate that much memory on Travis.

webapi/WebapiBuild.sbt
```diff
@@ -162,17 +163,17 @@ lazy val webApiLibs = Seq(
...
     // "-verbose:gc",
     //"-XX:+UseG1GC",
     //"-XX:MaxGCPauseMillis=500"
 )
 
 lazy val javaTestOptions = Seq(
     // "-showversion",
-    "-Xms2048m",
-    "-Xmx4096m"
+    "-Xms8G",
```
benjamingeer: See comment about memory above.

benjamingeer:
> is this the scala translation?

Yes, or you can also write the same thing as in Java.

loicjaouen:
"-Xms8G": I changed that in a desperate move when getting hit by #386

benjamingeer:
But Ivan's comment in #386 suggests setting an environment variable in your shell, rather than changing `WebapiBuild.sbt`. Did that not work for you?

loicjaouen:
no, I am reverting this changes

loicjaouen:
i'm running the test atm

loicjaouen:
how do I do to run only one specific test, let's say `AuthenticatorSpec`?

benjamingeer:
I type `test-only *AuthenticatorSpec` at the `sbt` prompt.

subotic:
@loicjaouen also, if possible, please create branches in the future under ``wip/`` (which stands for work in progress), i.e. ``wip/bcrypt_2a_rep``.

loicjaouen:
@subotic : sorry, i didn't think it would become a `wip`:)
@benjamingeer  : cool, the trick is to run it in the sbt interactive console, when running in batch, it restarts the tests from scratch

loicjaouen:
I don't think that the remaining errors are due to my changes:
```
[info] *** 4 TESTS FAILED ***
[error] Failed tests:
[error] 	org.knora.webapi.e2e.v1.ResourceValuePermissionsV1E2ESpec
```

subotic:
> I don't think that the remaining errors are due to my changes

If you created this branch from ``develop``, then they probably are :-). A prerequisite for merging something into ``develop`` is that all tests are passing.

loicjaouen:
@subotic : you were right ;)

benjamingeer:
Many thanks @loicjaouen !

loicjaouen:
@benjamingeer & @subotic 
thanks for the assistance ;)

PR #429, opened 2017-02-21
fix (ontologies): fix reference for knora-admin

afa-unibas:


afa-unibas:
just checking if PR's work ;)

tobiasschweizer:
be careful changing the prefix because now all the definitions beginning with `:` belong to `knora-admin`

ioolkos:
ah, yes of course, this is only half the story. 
We need to add the knora-base prefix here, and then prefix every reference accordingly. (exactly like you did in the standoff ontology).

I might add this at some later point...need to do other stuff right now ;)

afa-unibas:
ah, yes of course, this is only half the story.
We need to add the knora-base prefix here, and then prefix every reference accordingly. (exactly like you did in the standoff ontology).

I might add this at some later point...need to do other stuff right now ;)

subotic:
Just a few suggestions: 

 - All tests need to pass
 - Most certainly [OntologyConstants.scala](https://github.com/dhlab-basel/Knora/blob/develop/webapi/src/main/scala/org/knora/webapi/OntologyConstants.scala) will need some changes:
    - ``OntologyConstants.KnoraBase`` -> ``OntologyConstants.KnoraAdmin``
  - also every occurence of these constants throughout our codebase.

afa-unibas:
Did I just open Pandora's box? :) 
but jep, I'll take the assignement.

subotic:
yes :) I have the feeling, that from now on, we are going to have only this kind of assignments. Every time I touch something it proves to be more work than anticipated.

afa-unibas:
@subotic Ivan, can you give me a hint on why travis tests are still failing? is it a certain class of tests failing?

subotic:
@afa-unibas This looks to me like you misspelled something in ``OntologyConstants.scala``. Some property (which needs to be there) could not be found.

subotic:
Also, I don't see any changes to ``admin-data.ttl``, which I would expect.

afa-unibas:
Closing, so that I can sync repos.

PR #432, opened 2017-02-27
Require hasPermissions on resources and values

benjamingeer:
This changes the cardinality of `hasPermissions` to `owl:cardinality 1` on `knora-base:Resource` and `knora-base:Value`, and changes SPARQL templates, Scala code, and test data accordingly.

There was a special case, implemented in `PermissionUtilV1`, to allow any user to see a link value representing a standoff link. This has been replaced with explicit permissions on those link values: `CR knora-base:Creator|V knora-base:UnknownUser`.

The `standoff` transformation in `TransformData` now also corrects `knora-base:attachedToUser` and `knora-base:hasPermissions` on link values for standoff links.


benjamingeer:
@subotic I can't figure out why the Travis builds are failing for this PR, could you have a look? All the tests pass on my machine, but on Travis some of them time out.

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
```diff
@@ -97,16 +96,16 @@ object PermissionUtilV1 {
...
-      *                    included, but they will be ignored, so there is no need to filter them before passing them to
-      *                    this function.
+      * @param valueProps     a [[ValueProps]] containing the permission-relevant predicates and objects
+      *                       pertaining to the value, grouped by predicate. The predicates must include
+      *                       [[org.knora.webapi.OntologyConstants.KnoraBase.AttachedToUser]], and should include
+      *                       [[org.knora.webapi.OntologyConstants.KnoraBase.AttachedToProject]]
+      *                       and [[org.knora.webapi.OntologyConstants.KnoraBase.HasPermissions]]. Other predicates may be
+      *                       included, but they will be ignored, so there is no need to filter them before passing them to
+      *                       this function.
       * @param subjectProject if provided, the `knora-base:attachedToProject` of the value. Otherwise, this predicate
```
tobiasschweizer: I thought we got rid of project assignments to values. 

As stated in `TransformData.scala`: 
> '$ValueProjectTransformationOption' (removes 'knora-base:attachedToProject' from values)

see also <https://github.com/dhlab-basel/Knora/blob/ca51f099e50551ab2c692453a1e6b3210099060d/webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala#L119>

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
```diff
@@ -316,106 +315,36 @@ object PermissionUtilV1 {
         // Anything with permissions must have an creator and a project.
         val subjectCreator: IRI = assertionMap.getOrElse(OntologyConstants.KnoraBase.AttachedToUser, throw InconsistentTriplestoreDataException(s"Subject $subjectIri has no creator"))
         val subjectProject: IRI = assertionMap.getOrElse(OntologyConstants.KnoraBase.AttachedToProject, throw InconsistentTriplestoreDataException(s"Subject $subjectIri has no project"))
-        val subjectPermissionLiteral: Option[String] = assertionMap.get(OntologyConstants.KnoraBase.HasPermissions)
+        val subjectPermissionLiteral: String = assertionMap.getOrElse(OntologyConstants.KnoraBase.HasPermissions, throw InconsistentTriplestoreDataException(s"Subject $subjectIri has no knora-base:hasPermissions predicate"))
```
tobiasschweizer: Is this consistent? I think there are a lot of places in the code where there is no check at all because the existence of a permission is assumed (required by the ontology).

Otherwise all the old `get` (resulting in an option of a permission) should consistently be changed to `getOrElse` to throw an apt error message and not getting a Java-exception because the element was not found.

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
```diff
@@ -316,106 +315,36 @@ object PermissionUtilV1 {
...
-                userProfile = userProfile
-            )
-        }
-    }
-
-    /**
       * Parses the literal object of the predicate `knora-base:hasPermissions`.
       *
-      * @param maybePermissionListStr the literal to parse.
+      * @param permissionListStr the literal to parse.
```
tobiasschweizer: In the ontology, `kb:hasPermission` has a `:objectDatatypeConstraint` of `xsd:string`. Could this be made more specific? You do not just assume any string, but a specific string format.

In XML-schema, one can provide a pattern. 

This is an example taken from the *Meditationes* constraining attribute values to a certain pattern:
```
<xs:pattern value="(\s?underline|\s?center|\s?strike|\s?above|\s?marginal|\s?below)+"/>
```

subotic:
> @subotic I can't figure out why the Travis builds are failing for this PR, could you have a look? All the tests pass on my machine, but on Travis some of them time out.

Did you run the tests with Fuseki locally? The same test fails for me also locally, but only with Fuseki. With GraphDB it works.

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
benjamingeer: > I think there are a lot of places in the code where there is no check at all because the existence of a permission is assumed (required by the ontology).

I don't think so. It wasn't required by the ontology until I made it required in this pull request. Everything I saw had an `Option` for permissions. Did you find an example?

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
```diff
@@ -194,7 +193,7 @@ object PermissionUtilV1 {
     def getUserPermissionV1(subjectIri: IRI,
```
benjamingeer: Yes, as the Scaladoc says, `getUserPermissionV1` checks permissions on any subject (which could be a resource or a value), while  `getUserPermissionV1WithValueProps` checks permissions on a value, for which the caller happens to have a `ValueProps`.

If want to check permissions on a value, the choice between these two methods depends on the format that you have the data in.

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
tobiasschweizer: I simply wanted to point out that you replaced a `get` with a `getOrElse` here which would throw an apt error message, while you just removed `get` in other cases without replacing it with a `getOrElse`.

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
benjamingeer: It's the project of the containing resource. Changed comment to clarify.

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
```diff
@@ -218,20 +218,19 @@ class ResourcesResponderV1 extends ResponderV1 {
...
                             // the user must have some permission on the link value and on the source and target
                             // nodes.
                             val hasPermission = visibleNodeIris.contains(edge.sourceNodeIri) && visibleNodeIris.contains(edge.targetNodeIri) &&
-                                PermissionUtilV1.getUserPermissionOnLinkValueV1(
-                                    linkValueIri = edge.linkValueIri,
-                                    predicateIri = edge.linkProp,
-                                    linkValueCreator = edge.linkValueCreator,
-                                    containingResourceProject = edge.sourceNodeProject,
-                                    linkValuePermissionLiteral = edge.linkValuePermissions,
+                                PermissionUtilV1.getUserPermissionV1(
```
benjamingeer: Yes.

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
benjamingeer: GraphDB can't enforce datatype constraints at all, not even to check the type, never mind the contents.

benjamingeer:
@subotic It works for me with Fuseki locally, I've just run it three times with no errors :(

```
[info] - should return 2 books with the title 'Zeitglöcklein des Lebens und Leidens Christi' when we search for book titles containing the word 'Lebens' but not containing the word 'walfart' (using MATCH BOOLEAN) in the Incunabula test data (416 milliseconds)
```

benjamingeer:
@subotic Thanks, I reproduced it on Fuseki, will figure it out.

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
benjamingeer: > you just removed `get` in other cases without replacing it with a `getOrElse`.

I think all those cases are getting values from a `VariableResultsRow`, which uses `ErrorHandlingMap`.

benjamingeer: I've checked again, and all those cases are using `rowMap`, which is an `ErrorHandlingMap`.

benjamingeer:
Can I merge this now?

tobiasschweizer:
I have to check the transformation script first with the Bernoulli Euler test data. OK?

benjamingeer:
Yes, good idea.

tobiasschweizer:
I will do that in the afternoon today.

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilV1.scala
tobiasschweizer: Ok, then it's fine like it is.

tobiasschweizer:
Tests pass both with graphdb and fuseki.

I could also transform the BEOL data with the transformation `standoff`.
If there are no standoff links, this transformation is not needed, right? Would graphdb's consistency checker detect missing permissions for a standoff link value (I would think so because cardiniality checking works quite well)?

webapi/src/main/twirl/queries/sparql/v1/searchExtendedGraphDB.scala.txt
```diff
@@ -465,7 +459,7 @@ WHERE {
                 }
 
                 @if(searchCriterion.matchBooleanNegativeTerms.nonEmpty) {
-                    MINUS {
+                    FILTER NOT EXISTS {
```
tobiasschweizer: Is this intended? I remember that we had some issues with unexpected behaviour with `FILTER NOT EXISTS` and `MINUS`.

webapi/src/main/twirl/queries/sparql/v1/searchExtendedStandard.scala.txt
```diff
@@ -482,7 +476,7 @@ WHERE {
                 }
 
                 @if(searchCriterion.matchBooleanNegativeTerms.nonEmpty) {
-                    MINUS {
+                    FILTER NOT EXISTS {
```
tobiasschweizer: see above

webapi/src/main/twirl/queries/sparql/v1/getVersionHistoryGraphDB.scala.txt
```diff
@@ -72,11 +72,6 @@ WHERE {
         BIND(true as ?isLinkValue)
     }
 
-    OPTIONAL {
-        ?value knora-base:attachedToUser ?valueCreator .
```
tobiasschweizer: How could this be optional? Did the owner on a value use to be optional in knora-base?

webapi/src/main/twirl/queries/sparql/v1/getVersionHistoryStandard.scala.txt
```diff
@@ -70,11 +70,6 @@ WHERE {
         BIND(true as ?isLinkValue)
     }
 
-    OPTIONAL {
-        ?value knora-base:attachedToUser ?valueCreator .
```
tobiasschweizer: see above

webapi/src/main/twirl/queries/sparql/v1/searchExtendedGraphDB.scala.txt
benjamingeer: Yes. I did a little experiment and found that it was a bit faster this way.

webapi/src/main/twirl/queries/sparql/v1/searchExtendedStandard.scala.txt
benjamingeer: This change was necessary to keep Fuseki from hanging in this query.

webapi/src/main/twirl/queries/sparql/v1/getVersionHistoryGraphDB.scala.txt
benjamingeer: No, I think this was a mistake that's been there for a long time.

webapi/src/main/twirl/queries/sparql/v1/getVersionHistoryStandard.scala.txt
benjamingeer: Same.

benjamingeer:
> If there are no standoff links, this transformation is not needed, right?

That's correct.

webapi/src/main/twirl/queries/sparql/v1/searchExtendedStandard.scala.txt
benjamingeer: Fuseki has severe performance problems with `MINUS` and often does much better with `FILTER NOT EXISTS`.

PR #435, opened 2017-02-28
Correctly use and display user's name and projects

### LINKED ISSUE #118: set SALSAH.userdata correctly after login or refresh with an active session
SALSAH.userdata only contains the userdata when you do a login, not when you refresh the page with an active session

Also, the user's active project is not set

we need the user's active project when creating a resource

### LINKED ISSUE #415: After logging into Salsah, user is shown as undefined
undefined-user

benjamingeer:
User-visible changes in SALSAH:

* Display the user's first and last name instead of their username.
* Allow the user to select the active project.

Internal changes:

* Include details of the user's projects in `userProfile`, instead of just the project IRIs.
* Have SALSAH get `userProfile` on login, rather than just `userData`, and store it.
* Remove `userData` from API responses, since SALSAH no longer needs it.

Note: To get the GUI tests to work, I had to change them so the user is logged in at the beginning of each test, and logged out at the end of each test. Otherwise reloading the page seemed to make SALSAH forget that the user was logged in. I had this problem only with Selenium, not in real life.

Fixes #118.
Fixes #415.


benjamingeer:
I found a nice service that generates TypeScript interfaces from sample data: http://json2ts.com/

salsah/src/typescript_interfaces/basicMessageComponents.ts
```diff
@@ -122,11 +122,6 @@ export module basicMessageComponents {
          * Knora status code
          */
         status:KnoraStatusCode;
-
-        /**
-         * The current user's data
-         */
-        userdata:userdata;
```
benjamingeer: Now I've moved it to `sessionResponseFormats.ts`, which has the whole login response format. There's a sample in `sampleLoginResponse.ts`.

benjamingeer:
OK I changed `ResourceCreationSpec` so the multi-project user creates two resources in different projects. Could you check this again now?

salsah/src/typescript_interfaces/sessionResponseFormats.ts
```diff
@@ -0,0 +1,229 @@
...
+         * The user's profile.
+         */
+        userProfile: userProfile;
+    }
+
+	export interface userProfile {
+		/**
+		 * The IRIs of the groups that the user belongs to.
+		 */
+        groups: Array<string>;
```
tobiasschweizer: You could use the KnoraIri pseudo type here just to make it more clear that is is an Knora IRI: https://github.com/dhlab-basel/Knora/blob/wip/salsah-username/salsah/src/typescript_interfaces/basicMessageComponents.ts#L245

tobiasschweizer:
go on, genius

PR #437, opened 2017-02-28
Make it easier to find the documentation on the date string format

### LINKED ISSUE #431: Documentation of DateValue type missing
Example: instead of 1974 , date_value should be GREGORIAN:1974 . This is not documented.

benjamingeer:
Closes #431.


tobiasschweizer:
This is fine.

Could you check if we also need to add `|| true` after `git commit` in `update-gh-pages.sh`?
I think if this is not added and nothing was actually changed to be committed, the script would terminate with -1, not switching back to `develop`.

tobiasschweizer:
This PR can be merged!

PR #453, opened 2017-03-15
Generate dereferenceable IRIs using the rdfh.ch hostname

benjamingeer:
This changes the function `KnoraIdUtil.makeRandomResourceIri()`:

* Include the project's shortname (which will be a unique ID) in the IRI.
* Use `rdfh.ch` as the IRI domain name. This will point to a server that redirects requests to Knora API servers depending on the project ID in the IRI.

I also removed some unused methods, and clarified and reformatted some code.


tobiasschweizer:
Can this branch be deleted?

PR #461, opened 2017-03-16
Ensure validity of a date

tobiasschweizer:
So far, we have only checked the formal validity of a date string that was about to be converted to a Java `GregorianDate` (and then further to a JDN). We use a regex in order to do so.

However, nothing stops the user from submitting formally valid, but non existing dates like "GREGORIAN:2017-02-29". Internally, Java just takes the first day of the following month. When you submit `00` for a day, it just takes the last day of the preceding month.

This might be arithmetically correct, but it is also a possible source for creating wrong dates. It happened to me that I made a mistake (No, I am not the pope :-)) when parsing the date string "1712-00-00". Instead of creating a date string with year precision (i.e. "GREGORIAN:1712") I mistakenly submitted "GREGORIAN:1712-00-00". Knora accepted the date string and it was only later that I found out that this resulted in "30th of November 1711" which is clearly not what we want to have.

Disabling leniency for `GregorianDate` allows for checking that the date actually exists:

> Leniency
>
> Calendar has two modes for interpreting the calendar fields, lenient and non-lenient. When a Calendar is in lenient mode, it accepts a wider range of calendar field values than it produces. When a Calendar recomputes calendar field values for return by get(), all of the calendar fields are normalized. For example, a lenient GregorianCalendar interprets MONTH == JANUARY, DAY_OF_MONTH == 32 as February 1.
>
> When a Calendar is in non-lenient mode, it throws an exception if there is any inconsistency in its calendar fields. For example, a GregorianCalendar always produces DAY_OF_MONTH values between 1 and the length of the month. A non-lenient GregorianCalendar throws an exception upon calculating its time or calendar field values if any out-of-range field value has been set.

(https://docs.oracle.com/javase/8/docs/api/java/util/Calendar.html)

However, as we deal with a lot of historical dates I am not sure if we should rely on Java to do such checks. The docs also state:

> public int get(int field)
> Returns the value of the given calendar field. In lenient mode, all calendar fields are normalized. In non-lenient mode, all calendar fields are validated and this method throws an exception if any calendar fields have out-of-range values. The normalization and validation are handled by the complete() method, which process is calendar system dependent.

(https://docs.oracle.com/javase/8/docs/api/java/util/Calendar.html#get-int-)

I added some tests that contain dates that are rejected with the new behaviour.


tobiasschweizer:
This is interesting:

> Julian calendar in GregorianCalendar class always adds leap day when year is divisible by 4 and it does not deal with leap year error before 8 AD. GregorianCalendar class can compute dates by extrapolating current rules indefinitely far backward and forward in time. However, dates obtained using GregorianCalendar class are historically accurate only from March 1, 4 AD onward due to the leap year error.

(http://www.zhuwu.me/blog/posts/handle-historical-date-in-java)

webapi/src/main/scala/org/knora/webapi/util/DateUtilV1.scala
```diff
@@ -102,33 +102,69 @@ object DateUtilV1 {
...
 
-                val intervalEnd = new GregorianCalendar
-                intervalEnd.setGregorianChange(changeDate)
-                intervalEnd.set(dateSegments(0).toInt, 11, 31, 12, 0, 0) // December 31st of the given year
+                try {
+                    val intervalStart = new GregorianCalendar
+                    intervalStart.setLenient(false)
+                    intervalStart.setGregorianChange(changeDate)
+                    intervalStart.set(dateSegments(0).toInt, 0, 1, 12, 0, 0) // January 1st of the given year. Attention: in java.util.Calendar, month count starts with 0
+                    intervalStart.get(0)
```
benjamingeer: Please add a comment here to explain that you're calling `get(0)` because this causes all calendar fields to be validated, and throws an exception if any fields have out-of-range values.

webapi/src/main/scala/org/knora/webapi/util/DateUtilV1.scala
```diff
@@ -102,33 +102,69 @@ object DateUtilV1 {
...
+                    val intervalEnd = new GregorianCalendar
+                    intervalEnd.setLenient(false)
+                    intervalEnd.setGregorianChange(changeDate)
+                    intervalEnd.set(dateSegments(0).toInt, 11, 31, 12, 0, 0) // December 31st of the given year. Attention: in java.util.Calendar, month count starts with 0
+                    intervalEnd.get(0)
+
+                    DateRange(intervalStart, intervalEnd, KnoraPrecisionV1.YEAR)
+
+                } catch {
+                    case e: Exception => throw BadRequestException(s"The provided date $dateString could not be handled correctly ${e.getMessage}")
```
benjamingeer: Would it be possible to catch a more specific exception here? What exception is actually thrown if the calendar is invalid?

webapi/src/main/scala/org/knora/webapi/util/DateUtilV1.scala
tobiasschweizer: Ok, done in https://github.com/dhlab-basel/Knora/pull/461/commits/945fc68cf7518abd71568028b6322fb312dd6673

webapi/src/main/scala/org/knora/webapi/util/DateUtilV1.scala
tobiasschweizer: it is an `IllegalArgumentException`.

done in https://github.com/dhlab-basel/Knora/pull/461/commits/945fc68cf7518abd71568028b6322fb312dd6673

tobiasschweizer:
Ok?

benjamingeer:
Great, thanks!

Issue #470, opened 2017-03-24
Search does not work properly for Knora documentation on knora.org


tobiasschweizer:
When I do a search on <http://www.knora.org/documentation/manual/search.html> I get very few or no results for things that occur often. Locally, I get the proper set of results.

Example: <http://www.knora.org/documentation/manual/search.html?q=Ontology+Checking>

Is there some problem with absolute URLs etc. when deploying the HTML build of the docs?

Have you changed something recently, the design looks different than before.

benjamingeer:
I have no idea. I switched from the ugly `haiku` theme to the better-looking `agogo` theme, which is working fine on the Sipi web site. But I don't have time to figure out Sphinx's JavaScript right now, so I made PR #471 to revert back to `haiku`.

benjamingeer:
This seems to have fixed itself.

tobiasschweizer:
Wow. Self-patching code.

PR #471, opened 2017-03-24
fix (docs): Revert to haiku theme to fix search (#470).

benjamingeer:
This attempts to fix search on knora.org by switching back from the `agogo` theme (which looks much better) to the `haiku` theme (which is ugly IMHO but was working before).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -111,7 +111,7 @@ todo_include_todos = False
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'agogo'
+html_theme = 'haiku'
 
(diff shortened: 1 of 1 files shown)


benjamingeer:
> I assume that you have already updated the gh-pages branch.

No, I haven't, I was waiting to merge this first so I could run the update script.

> The search works again on knora.org! Thx.

I didn't do anything, so I have no idea why it works now.

Since it works now, perhaps I should just delete this branch without merging it?

tobiasschweizer:
strange ...

maybe it is one of these magic gh-pages things like the `nojekyll`. References to CSS files are broken first and then after some time it suddenly works.

tobiasschweizer:
yes, I guess you can delete this branch if you prefer the other design.

Issue #227, opened 2016-08-19
(Search panel : extended search) Redseign and extend functionalities


mrivoal:
I think « Extended search » requires a lot more brainstorming of the whole team. 
Functionalities should be extended, in order to allow users to cross-search, for example : Given 2 objects, « Book » and « Author », users should be able to search for all the Books whose « title » (property:title) includes « king » and written by Authors born between « 1700 » and « 1850 » (property : birthDate) in « England » (property:countryOfBirth).

An operator OR should be added as well (Books whose « title » (property:title) includes « king » OR « queen »).

Given the fact that RDF modelling allows (and even encourage) to decompose objects into separate entities, allowing user to cross-search through several objects is a key feature.

I think we should try to figure out what would be the most convenient way to deal with this in Salsah from a UX point of view, as well as check with backend developers such as Ben and Tobias, what are the options in Knora to implement such a functionality according to the backend.

benjamingeer:
Two options considered in the last UX meeting:

- Enable project admins to create search templates representing complex searches with blanks that users can fill in to customise them.
- Allow more technicially proficient users to write search criteria in a simple expression-based language (with parentheses and operators like AND and OR) that would be more powerful than the GUI-based extended search form.

mrivoal:
I think we should probably discuss these options all together again because we can't reasonnably expect every project admins to write SPARQL queries templates for their team (if this is SPARQL queries you are involving).

And I don't think either a query as "basic" as the one I suggested should require more technically proficients users to write it. A graphic interface could probably be investigated here. We will try to make some suggestions and I suggest we discuss this together.

benjamingeer:
I was thinking of templates written in the search language suggested above.

I think the "basic" query you suggested above would actually be quite difficult to represent clearly in a graphical interface, but I would love to be proven wrong.

loicjaouen:
Isn't it possible to use the editor with wildcards?
If you consider an editor like described in #231, a simple search would use the same form but leave some fields empty:
![search](https://cloud.githubusercontent.com/assets/3933654/17851503/9b908082-6862-11e6-830f-b3cbae829ee1.png)
This would look for _Resource_ matching or linked to a person named _Barbeyrac_ in _Lausanne_.

This is a naïve search that we can decide to leave simple (therefore limited) on purpose.

[...]

PR #478, opened 2017-03-27
Create V2 for Search

tobiasschweizer:
This makes a start at implementing #530. The rest will be done in subsequent pull requests.


mrivoal:
I am not sure whether this is relevant at this stage, but as for the design of V2 API, I just want to be sure that the requirements to run queries throughout various (and customed) properties of 2 resource classes will be taken into account (such as expressed here: #227).

This is just a reminder that @loicjaouen has begun to work on an graphic interface (as a proposition for Salsah 1.5/V1 API but with Salsah 2.0 and V2 API in mind), but unfortunately, he hasn't had much time to dedicate to it yet.

tobiasschweizer:
We have to do that together, absolutely.

As for now, I am trying to have a prove of concept at the end of the week for the fulltext search v2. The Response format should be JSON-LD.

I was fighting with responder manager conf ...

tobiasschweizer:
@geer @subotic Hey masters of disaster, does this setup for v2 look reasonable to you?

tobiasschweizer:
```
CONSTRUCT { 
	?matchingSubject ?matchingProperty ?literal .
    ?matchingSubject knora-base:attachedToUser ?valueCreatorMatchingSubject ;
            knora-base:hasPermissions ?valuePermissionsMatchingSubject .
    ?matchingSubject a ?typeMatchingSubject .
    
    ?resource ?prop ?matchingSubject .
    ?resource rdfs:label ?label .
    ?resource a ?typeResource .
    ?resource knora-base:attachedToUser ?valueCreator ;
            knora-base:hasPermissions ?valuePermissions .
}

WHERE {
    
    ?matchingSubject ?matchingProperty ?literal .
    ?literal <http://www.ontotext.com/owlim/lucene#fullTextSearchIndex> 'Zeitglöcklein' .
    
        ?matchingSubject ?matchingProperty ?literal .
        ?matchingSubject a ?typeMatchingSubject .
    }
    
    ?matchingSubject knora-base:isDeleted false .
    
    ?matchingSubject knora-base:attachedToUser ?valueCreatorMatchingSubject ;
    	knora-base:hasPermissions ?valuePermissionsMatchingSubject .
    
    
    OPTIONAL {
    	?resource knora-base:hasValue ?matchingSubject .
    
        	?resource ?prop ?matchingSubject .
            ?resource a ?typeResource .
    	}
    
    	?resource knora-base:isDeleted false .
    
        ?resource rdfs:label ?label .
        
        
        
        ?resource knora-base:attachedToUser ?valueCreatorResource ;
        	knora-base:hasPermissions ?valuePermissionsResource .
    }
}
```

loicjaouen:
For info: here is the first mockup to make a search form we did some time ago (it doesn't include suggested improvements that we collected):
https://platec.unil.ch/test/xtdsrch/

You can click on `Persons` and `Works` to list the items you can play with.  
The simple search, select `Person` or `Work` depending what you look for, type `ez` press `search` and you get a list of results.  
The extended search allows to do search by exploring the graph, like saying `works from authors who published in 1912` (change it to 1916 and the final result changes quite a bit):
extended search clickable demo

tobiasschweizer:
```
object ConstructResponseUtilV2 {

    case class ValueObject(valueObjectIri: IRI, assertions: Seq[(IRI, String)], standoff: Map[IRI, Seq[(IRI, String)]])

    case class ResourceWithValues(resourceAssertions: Map[IRI, Seq[(IRI, String)]], valuePropertyAssertions: Map[IRI, Seq[ValueObject]], linkPropertyAssertions: Map[IRI, Seq[IRI]])

    /**
      * A [[SparqlConstructResponse]] may contain both resources and value objects.
      * This method splits the results by their type and returns them as separate structures.
      *
      * @param constructQueryResults the results of a SPARQL construct query.
      * @return a Map[resource Iri -> [[ResourceWithValues]]].
      */
    def splitResourcesAndValueObjects(constructQueryResults: SparqlConstructResponse) = {

        /*
        // 2 tuple: value objects (._1) and resources/standoff (._2)
        val (valueObjects: Map[IRI, Seq[(IRI, String)]], resourcesAndStandoff: Map[IRI, Seq[(IRI, String)]]) = constructQueryResults.statements.partition {
            case (subjectIri: IRI, assertions: Seq[(IRI, String)]) =>

                val predicateMap: Map[IRI, String] = new ErrorHandlingMap(assertions.toMap, { key: IRI => s"Predicate $key not found for $subjectIri" })

                val rdfType = predicateMap(OntologyConstants.Rdf.Type)

                // returns true if it is a valueObject, false in case of a resource or a standoff node
                OntologyConstants.KnoraBase.ValueClasses.contains(rdfType)

        }

        // get standoff node Iris from value objects
        val standoffNodeIris: Vector[IRI] = valueObjects.flatMap {
            case (subjectIri: IRI, assertions: Seq[(IRI, String)]) =>

                // a value object may have more than one standoff node, so we cannot use a map here (making knora-base:valueHasStandoff the predicate)
                assertions.filter {
                    case (pred: IRI, obj: String) =>
                        pred == OntologyConstants.KnoraBase.ValueHasStandoff
                }.map {
                    case (pred: IRI, obj: String) =>
                        // we are only interested in the standoff node Iri
                        obj
                }

        }.toVector

        // separate resources from standoff nodes
        val (standoffNodes: Map[IRI, Seq[(IRI, String)]], resources: Map[IRI, Seq[(IRI, String)]]) = resourcesAndStandoff.partition {
            case (subjectIri: IRI, assertions: Seq[(IRI, String)]) =>

                standoffNodeIris.contains(subjectIri)

        }

        // determine value and linking properties
        val valuePropObjectIris = resources.flatMap {
            case (subjectIri: IRI, assertions: Seq[(IRI, String)]) =>

                assertions.filter {
                    case (pred: IRI, obj: String) =>
                        pred == OntologyConstants.KnoraBase.HasValue
                }.map {
                    case (pred: IRI, obj: String) =>
                        // we are only interested in the value objectIri
                        obj
                }
        }.toVector*/

        // spilt statements abozt resources and other statements (value objects and standoff)
        val (resourceStatements: Map[IRI, Seq[(IRI, String)]], otherStatements: Map[IRI, Seq[(IRI, String)]]) = constructQueryResults.statements.partition {

            case (subjectIri: IRI, assertions: Seq[(IRI, String)]) =>

                assertions.contains((OntologyConstants.Rdf.Type, OntologyConstants.KnoraBase.Resource))

        }

        println(resourceStatements)

        println("+++++++++++")

        resourceStatements.map {
            case (resourceIri: IRI, assertions: Seq[(IRI, String)]) =>

                // remove inferred statements (non explicit)
                val assertionsFiltered = assertions.filterNot {
                    case (pred, obj) =>
                        (pred == OntologyConstants.Rdf.Type && obj == OntologyConstants.KnoraBase.Resource) || pred == OntologyConstants.KnoraBase.HasValue || pred == OntologyConstants.KnoraBase.HasLinkTo
                }

                val objMap: ErrorHandlingMap[String, IRI] = new ErrorHandlingMap(assertionsFiltered.map {
                    case (pred, obj) =>
                        (obj, pred)
                }.toMap, { key: IRI => s"object $key not found for $resourceIri" })

                // get a map from property Iris to value object Iris
                val valuePropertyToObjectIris: Map[IRI, Seq[String]] = assertions.filter {
                    case (pred, obj) =>
                        pred == OntologyConstants.KnoraBase.HasValue
                }.foldLeft(Map.empty[IRI, Seq[String]]) {
                    case (acc, (hasValue: IRI, valObj: String)) =>
                        val propIri = objMap(valObj)

                        if (acc.keySet.contains(propIri)) {
                            val existingValsForProp = acc(propIri)

                            acc + (propIri -> (valObj +: existingValsForProp))

                        } else {
                            acc + (propIri -> Vector(valObj))
                        }

                }

                println("+++++")

                val valuePropertyToValueObject: Map[IRI, Seq[ValueObject]] = valuePropertyToObjectIris.map {
                    case (property: IRI, valObjs: Seq[String]) =>

                        (property, valObjs.map {
                            case valObjIri =>

                                // get all the standoff nodes possibly belonging to this value object
                                val standoffNodeIris: Seq[String] = otherStatements(valObjIri).filter {
                                    case (pred: IRI, obj: String) =>
                                        pred == OntologyConstants.KnoraBase.ValueHasStandoff
                                }.map {
                                    case (pred: IRI, obj: String) =>
                                        // we are only interested in the standoff node Iri
                                        obj
                                }

                                val standoffAssertions: Map[IRI, Seq[(IRI, String)]] = otherStatements.filter {
                                    case (subjIri: IRI, assertions: Seq[(IRI, String)]) =>
                                        standoffNodeIris.contains(subjIri)

                                }

                                ValueObject(valueObjectIri = valObjIri, assertions = otherStatements(valObjIri), standoff = standoffAssertions)

                        })

                }

                println(valuePropertyToValueObject)

                // map property to value objects


                val linkPropTargets: Seq[IRI] = assertions.filter {
                    case (pred, obj) =>
                        pred == OntologyConstants.KnoraBase.HasLinkTo
                }.map {
                    case (pred, obj) => obj
                }

                ""


        }


    }
```

tobiasschweizer:
A small experiment for BEOL data: get all the letters exchanged between two persons:

```
PREFIX beol: <http://www.knora.org/ontology/beol#>
CONSTRUCT {
    
 ?resource rdfs:label ?label .   
 
 ?resource rdfs:label ?label ;
           a ?resourceClass .   
    
} where { 
 ?resource a knora-base:Resource ;
           knora-base:isDeleted false .
    
 ?resource knora-base:hasLinkTo ?referredResource .
 
 	?resource ?linkProperty ?referredResource .
 }   
    
 ?referredResource rdfs:label "Bernoulli,  Johann 1667-1748" ; 
 	a beol:person .
    
 ?resource knora-base:hasLinkTo ?referredResource2 .
 
 	?resource ?linkProperty2 ?referredResource2 .
 }   
    
 ?referredResource2 rdfs:label "Verzaglia,  Giuseppe 1669-1728" ; 
 	a beol:person .
 
    FILTER((?linkProperty = beol:hasAuthor || ?linkProperty = beol:hasRecipient) && (?linkProperty2 = beol:hasAuthor || ?linkProperty2 = beol:hasRecipient))   
    
 ?resource rdfs:label ?label .
        
 
          ?resource a ?resourceClass .
    }   
} 
```

benjamingeer:
Instead of using `?resource knora-base:hasLinkTo ?referredResource .` and `FILTER`, I think you could use a `UNION` of `?resource beol:hasAuthor ?referredResource .` and `?resource beol:hasRecipient ?referredResource .`

tobiasschweizer:
That's right :-) Thx.

tobiasschweizer:
both persons can be either author or recipient of the letter (it selects the correspondence as a whole, regardless of the "direction")

I will be happy to work with you on the v2 of the extended search :-)

tobiasschweizer:
but now: enjoy your days off!

benjamingeer:
OK, OK. :)

tobiasschweizer:
I have some difficulties with `LIMIT`: we cannot just limit the numbers of triples returned because some required information could be missing (e.g., a resource's label) resulting in a scala error. Instead, the number of resources would have to be limited. Probably we also need to limit the values, because there may be values with a lot of text. 

If the result of a SPARQL query gets too big, graphdb cannot write the result and send it back to Knora.

benjamingeer:
> I have some difficulties with `LIMIT`: we cannot just limit the numbers of triples returned because some required information could be missing (e.g., a resource's label) resulting in a scala error. 

I would email `graphdb-support@ontotext.com` to ask them about this. Also about limiting the amount of time a query takes.

tobiasschweizer:
Ok!

I was just refactoring a little bit (replacing OPTIONAL with UNION).

We can reuse a lot of our optimization we did in the WHERE clauses.

Standoff seems to be quite a bottle neck, maybe we do not eturn it when doing a search (unless one does a standoff search), but only when the whole resource is requested

benjamingeer:
> Standoff seems to be quite a bottle neck

More than in version 1?

> maybe we do not return it when doing a search

Sounds OK to me.

tobiasschweizer:
I think you will find a lot to be optimized in the V2 queries based on your v1 experience

benjamingeer:
> Standoff seems to be quite a bottle neck, maybe we do not return it when doing a search

Wovon man nicht sprechen kann, darüber muss man schweigen.

tobiasschweizer:
@benjamingeer I moved the handling of subclasses from ontology responder v1 to v2

However, I think just dispatching messages from V1 responders to V2 is not a good idea. When you send a V1 message to a V1 responder you expect a V1 response (the same is true for V2). This means we would have to duplicate the messages.

Therefore I think it would be better to directly call V2 wherever possible. We would only leave those things in V1 that are not needed in V2.

Example `OntologyResponderV1`:
```
    private def checkSubClass(checkSubClassRequest: CheckSubClassRequestV1): Future[CheckSubClassResponseV1] = {
        for {
            response <- (responderManager2 ? CheckSubClassRequestV2(subClassIri = checkSubClassRequest.subClassIri, superClassIri = checkSubClassRequest.superClassIri, checkSubClassRequest.userProfile)).mapTo[CheckSubClassResponseV1]

        } yield response
    }
```

It just dispatches the message to ontology responder V2 which would have to send back a V2 message (which then would be transformed into V1 to make it consistent).

I think we should eliminate this in V1 responder (including the v1 messages) and only use V2.

V2 would be self contained then and V1 would use V2 wherever possible, but the V1 specific stuff would be in V1 only and not obfuscate V2. The drawback of course is that V1 is harder to understand when it involves v2 calls.

I hope you get what I mean ;-)
Practically this means that we would refactor V1 while working on V2: everything we want to keep from v1 would become v2.   

I would prefer to work on that with you. I think my time is better invested in designing the extended search today and tomorrow.

benjamingeer:
If I understand you correctly, this is exactly what I said yesterday, just rename the case classes:

* `CheckSubClassRequestV1` becomes `CheckSubClassRequestV2` and is used in both V1 and V2 code
* `CheckSubClassResponseV1` becomes `CheckSubClassResponseV2` and is used in both V1 and V2 code

But then the V1 responders will use V2 messages, so we cannot do this:

> When you send a V1 message to a V1 responder you expect a V1 response

But I don't think it matters. I think it's OK to get a V2 response from a V1 responder, as long as it's not going back to the API client.

tobiasschweizer:
> But then the V1 responders will use V2 messages, so we cannot do this:

> But I don't think it matters. I think it's OK to get a V2 response from a V1 responder, as long as it's not going back to the API client.

I think that this will be confusing and we should not do it.
 Also we the case classes to make the request would be duplicated. 
So why not call v2 directly?

tobiasschweizer:
but probably we should talk about that next week, that seems easier to me

benjamingeer:
> the case classes to make the request would be duplicated.

Why would they be duplicated?

> So why not call v2 directly?

This is exactly what I'm saying. V1 code uses V2 case classes directly.

benjamingeer:
Ah now I think I understand what you mean: Just don't have methods in `OntologyResponderV1` that duplicate methods in `OntologyResponderV2`. If `ResourcesResponderV1` needs to call those methods, it can call `OntologyResponderV2` directly. OK, yes, that makes sense to me!

tobiasschweizer:
>  If ResourcesResponderV1 needs to call those methods, it can call OntologyResponderV2 directly.

Yes! This means that V1 code has to be refactored as you work on V2 (in order to avoid duplicates).

I think I will proceed working on the V2 ontology responder next week.

I will try to figure out how the extended search could work, supporting multiple levels. I am quite sure that the V1 SPARQL dealing with value types could be partially reused (as in the fulltext query). I guess recursive template calls could allow for querying multiple levels.

tobiasschweizer:
Working on something else, I had the idea of unifying responder manager for v1 and v2. Since the messages for v1 and v2 are derived from different traits, we exactly know which responders they should be forwarded to.

Unifying them would have the advantage that there would be only one instance of a responderManager all the messages could be sent to. Otherwise when refactoring V1 messages to V2, the responderManager would have to be adapted too. I think this could hardly be done automatically (using refactoring in intellij for instance) and can not be checked at compile time. 

@benjamingeer @subotic 
Is there a design / architectural reason why v1 and v2 responder managers should not be unified?

benjamingeer:
> I had the idea of unifying responder manager for v1 and v2

Sounds like a good idea to me.

> Is there a design / architectural reason why v1 and v2 responder managers should not be unified?

Not that I can think of.

tobiasschweizer:
great, will do that next week then!

benjamingeer:
Proposed API ontology entity naming scheme:

* `http://api.knora.org/ontology/incunabula/v2#book` is a subclass of `http://api.knora.org/ontology/knora-api/v2#Resource`
* `http://api.knora.org/ontology/knora-api/simple/v2#book` is a subclass of `http://api.knora.org/ontology/incunabula/simple/v2#Resource`

benjamingeer:
You can take `OntologyConstants` and `InputValidation` from the `wip/bulk-xml-import-datatypes` brach for a start.

benjamingeer:
About the question of what characters should be allowed in a resource class or property name: since we have to use these names as XML tag labels in the XML import, they will have to conform to the definition of an XML NCName, which means that we can validate them with `InputValidation.OntolologyPrefixLabelPattern`. We could rename it `XmlNCName`.

benjamingeer:
Following your suggestion, I removed the `#` character from ontology IRIs in `InputValidation`.

tobiasschweizer:
@benjamingeer Next week I will be in Riga (Archiving Conf) and won't be able to work on the code at all. If you manage to finish the bulk import and merge it back to `develop`, I could merge `develop` into my branch, resolve conflicts and continue with the work you have already done.

tobiasschweizer:
@benjamingeer I applied your changes to this branch and tomorrow I will add the functions to InputValidation to convert to and from knora-api entity names.

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
```diff
@@ -1176,66 +1176,6 @@ class ResourcesResponderV1 extends Responder {
 
 
     /**
-      * Implements a pre-update check to ensure that an [[UpdateValueV1]] has the correct type for the `knora-base:objectClassConstraint` of
```
tobiasschweizer: @benjamingeer I assume that you did mean to delete these methods. By what have they been replaced?

tobiasschweizer: I would be grateful if you could have a look at this merge https://github.com/dhlab-basel/Knora/pull/478/commits/a9b6780d2fcad45123342e65b2771565d3c4ee0f

benjamingeer: They've been replaced by additional code in `checkResource()`.

I can look at your merge tomorrow.

benjamingeer:
`org.knora.webapi.util.search.v2.SearchParserV2` can now parse simple SPARQL CONSTRUCT queries into a hopefully convenient data structure. Now we can work on converting that data structure into the actual CONSTRUCT query that will be sent to the triplestore for the extended search.

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
tobiasschweizer: Does it look ok to you? :-)

webapi/src/main/scala/org/knora/webapi/util/InputValidation.scala
```diff
@@ -600,6 +614,91 @@ object InputValidation {
...
+      *
+      * @param internalEntityIri the Iri of the internal ontology entity.
+      * @param errorFun a function that throws an exception. It will be called if the form of the string is not
+      *                 valid for an internal ontology entity IRI.
+      * @return the corresponding simplified knora-api v2.
+      */
+    def internalEntityIriToSimpleApiV2EntityIri(internalEntityIri: IRI, errorFun: () => Nothing): IRI = {
+        internalEntityIri match {
+            case ProjectSpecificInternalOntologyEntityRegex(prefixLabel, entityName) =>
+                val apiPrefixLabel = if (prefixLabel == "knora-base") "knora-api" else prefixLabel
```
benjamingeer: Since the prefix label can be `knora-base`, the regex shouldn't be called `ProjectSpecificInternalOntologyEntityRegex`. Can we just rename it to `InternalOntologyEntityRegex`?

webapi/src/main/scala/org/knora/webapi/responders/v1/ResourcesResponderV1.scala
benjamingeer: I just had a quick look and it just looks like you took everything unchanged from `develop`. What changes did you make in `ResourcesResponderV1`?

tobiasschweizer: I added user profile to some of the case classes while refactoring ontology responder, see `ResourcesResponderV1.scala` here https://github.com/dhlab-basel/Knora/pull/478/files#diff-13eb43a5343d8a8f59b7bd0a75fa4350

benjamingeer:
> Shall I do right away on this branch?

OK!

webapi/src/main/scala/org/knora/webapi/util/InputValidation.scala
tobiasschweizer: fixed in https://github.com/dhlab-basel/Knora/pull/478/commits/8d11387eeff0946de95f036cf5badc37c51780ee

benjamingeer:
Why does `KnoraResponseV2` contain a `SettingsImpl`? This is already available in routes and responders, why would we need to send it with every response?

tobiasschweizer:
> Why does KnoraResponseV2 contain a SettingsImpl? This is already available in routes and responders, why would we need to send it with every response?

I use it to get the fallback language and the Sipi base path when turning the case classes into JSON-LD. And this happens in `KnoraResponseV2`.

tobiasschweizer:
Do you have a suggestion how to improve this?

benjamingeer:
> Do you have a suggestion how to improve this?

Pass it to `KnoraResponseV2.toJSONLDWithValueObject`.

tobiasschweizer:
> Pass it to KnoraResponseV2.toJSONLDWithValueObject.

fixed in https://github.com/dhlab-basel/Knora/commit/393ff19ced41e2f20074b187748d6ae6298401a4

benjamingeer:
To avoid generating redundant extra triples about variables representing resources, you could generate them just based on the type inspection result, before you look at the statements in the WHERE clause.

tobiasschweizer:
Having just taken a walk outside, I figured out the same 👍😀


Sent from my iPad

On 27 Jun 2017, at 21:52, Benjamin Geer <notifications@github.com<mailto:notifications@github.com>> wrote:


To avoid generating redundant extra triples about variables representing resources, you could generate them just based on the type inspection result, before you look at the statements in the WHERE clause.

benjamingeer:
Great minds think alike. 😊

tobiasschweizer:
sample queries:

```
// search for all the letters exchanged between two persons
                    val sparql_tmp =
                        """
                          |PREFIX beol: <http://api.knora.org/ontology/beol/simple/v2#>
                          |PREFIX knora-api: <http://api.knora.org/ontology/knora-api/simple/v2#>
                          |
                          |CONSTRUCT {
                          |    ?letter knora-api:isMainResource true .
                          |
                          |    ?letter ?linkingProp1  <http://rdfh.ch/beol/oU8fMNDJQ9SGblfBl5JamA> .
                          |
                          |    ?letter ?linkingProp2  <http://rdfh.ch/beol/6edJwtTSR8yjAWnYmt6AtA> .
                          |
                          |} WHERE {
                          |    ?letter a knora-api:Resource .
                          |    ?letter a beol:letter .
                          |
                          |    # Scheuchzer, Johann Jacob 1672-1733
                          |    ?letter ?linkingProp1  <http://rdfh.ch/beol/oU8fMNDJQ9SGblfBl5JamA> .
                          |    ?linkingProp1 knora-api:objectType knora-api:Resource .
                          |    FILTER(?linkingProp1 = beol:hasAuthor || ?linkingProp1 = beol:hasRecipient )
                          |
                          |    <http://rdfh.ch/beol/oU8fMNDJQ9SGblfBl5JamA> a knora-api:Resource .
                          |
                          |    # Hermann, Jacob 1678-1733
                          |    ?letter ?linkingProp2 <http://rdfh.ch/beol/6edJwtTSR8yjAWnYmt6AtA> .
                          |    ?linkingProp2 knora-api:objectType knora-api:Resource .
                          |
                          |    FILTER(?linkingProp2 = beol:hasAuthor || ?linkingProp2 = beol:hasRecipient )
                          |
                          |    <http://rdfh.ch/beol/6edJwtTSR8yjAWnYmt6AtA> a knora-api:Resource .
                          |}
                        """.stripMargin


                    // search for a letter that has the given title and mentions Isaac Newton
                    var sparql_tmp2 =
                        """
                          |PREFIX beol: <http://api.knora.org/ontology/beol/simple/v2#>
                          |PREFIX knora-api: <http://api.knora.org/ontology/knora-api/simple/v2#>
                          |
                          |CONSTRUCT {
                          |    ?letter knora-api:isMainResource true .
                          |
                          |    ?letter a beol:letter .
                          |
                          |    ?letter beol:title "1707-05-18_2_Hermann_Jacob-Scheuchzer_Johann_Jakob" .
                          |
                          |    ?letter beol:mentionsPerson <http://rdfh.ch/beol/NUkE4PxyT1uEm3K9db63wQ> .
                          |
                          |} WHERE {
                          |    ?letter a knora-api:Resource .
                          |    ?letter a beol:letter .
                          |
                          |    ?letter beol:title ?title .
                          |    beol:title knora-api:objectType xsd:string .
                          |
                          |    ?title a xsd:string .
                          |    FILTER(?title = "1707-05-18_2_Hermann_Jacob-Scheuchzer_Johann_Jakob")
                          |
                          |    # Newton,  Isaac 1643-1727
                          |    ?letter beol:mentionsPerson <http://rdfh.ch/beol/NUkE4PxyT1uEm3K9db63wQ> .
                          |    beol:mentionsPerson  knora-api:objectType knora-api:Resource .
                          |
                          |    <http://rdfh.ch/beol/NUkE4PxyT1uEm3K9db63wQ> a knora-api:Resource .
                          |
                          |}
                        """.stripMargin

                    // search for letters that link to another letter via standoff that is authored by a person with IAF id "120379260"
                    var sparql_tmp3 =
                        """
                          |PREFIX beol: <http://api.knora.org/ontology/beol/simple/v2#>
                          |PREFIX knora-api: <http://api.knora.org/ontology/knora-api/simple/v2#>
                          |
                          |CONSTRUCT {
                          |    ?letter knora-api:isMainResource true .
                          |
                          |    ?letter a beol:letter .
                          |
                          |    ?letter knora-api:hasStandoffLinkTo ?anotherLetter .
                          |
                          |    ?anotherLetter beol:hasAuthor ?author .
                          |
                          |    ?author beol:hasIAFIdentifier "120379260" .
                          |} WHERE {
                          |
                          |    ?letter a beol:letter .
                          |    ?letter a knora-api:Resource .
                          |
                          |    ?letter knora-api:hasStandoffLinkTo ?anotherLetter .
                          |    knora-api:hasStandoffLinkTo knora-api:objectType knora-api:Resource .
                          |    ?anotherLetter a knora-api:Resource .
                          |
                          |    ?anotherLetter beol:hasAuthor ?author .
                          |    beol:hasAuthor knora-api:objectType knora-api:Resource .
                          |
                          |    # Scheuchzer, Johann 1684-1738
                          |    ?author a beol:person .
                          |    ?author a knora-api:Resource .
                          |
                          |    ?author beol:hasIAFIdentifier "120379260" .
                          |    beol:hasIAFIdentifier knora-api:objectType xsd:string .
                          |
                          |}
                        """.stripMargin

```

benjamingeer:
Some notes on how to generate V2 search SPARQL for Fuseki or for another triplestore where we can't use inference:

Here's a comparison of what we do for V1 extended search on GraphDB vs. Fuseki.

## Matching a property of a resource

We want to match the property and its subproperties.

### GraphDB

```
?resource <@searchCriterion.propertyIri> ?valueObject@index .
```

### Fuseki

```
?p@index rdfs:subPropertyOf* <@searchCriterion.propertyIri> .
?resource ?p@index ?valueObject@index .
```

## Matching a link to another resource

We match:

* The user-supplied link property and its subproperties.
* The corresponding link value property and its subproperties.

We find out the actual link property, and we use it to match the `knora-base:LinkValue`.

### GraphDB

```
?resource <@searchCriterion.propertyIri> ?targetResource@index .
?resource <@{searchCriterion.propertyIri}Value> ?valueObject@index .

    ?resource ?linkProperty@index ?targetResource@index .
}

?targetResource@index rdfs:label ?literal@index ;
    knora-base:isDeleted false ;
    knora-base:attachedToUser ?targetResourceCreator@index ;
    knora-base:attachedToProject ?targetResourceProject@index ;
    knora-base:hasPermissions ?targetResourcePermissions@index .

?valueObject@index rdf:type knora-base:LinkValue ;
    rdf:subject ?resource ;
    rdf:predicate ?linkProperty@index ;
    rdf:object ?targetResource@index .
```

### Fuseki

```
?linkProperty@index rdfs:subPropertyOf* <@searchCriterion.propertyIri> .
?resource ?linkProperty@index ?targetResource@index .
?linkValueProperty@index rdfs:subPropertyOf* <@{searchCriterion.propertyIri}Value> .
?resource ?linkValueProperty@index ?valueObject@index .

?targetResource@index rdfs:label ?literal@index ;
    knora-base:isDeleted false ;
    knora-base:attachedToUser ?targetResourceCreator@index ;
    knora-base:attachedToProject ?targetResourceProject@index ;
    knora-base:hasPermissions ?targetResourcePermissions@index .

?valueObject@index rdf:type knora-base:LinkValue ;
    rdf:subject ?resource ;
    rdf:predicate ?linkProperty@index ;
    rdf:object ?targetResource@index .
```

## Matching the class of a resource

### GraphDB

```
?resource a <@restypeIri> .
```

### Fuseki

```
?resClass rdfs:subClassOf* <@restypeIri> .
?resource a ?resClass .
```

benjamingeer:
We're going to merge this PR as is, because the basic API v2 infrastructure is now needed for other PRs. We'll make another PR to finish the v2 search functionality.

PR #489, opened 2017-04-21
Release IRI lock if a task throws an exception instead of returning a Future

benjamingeer:
If the task passed to `IriLocker.runWithIriLock` handled an error by throwing an exception instead of by returning a failed `Future`, the lock would not be released. This fixes `IriLocker` so it releases the lock in that case, and adds some tests.

Please try this `IriLocker` on your branch and let me know if it fixes the problem you saw.


subotic:
Thanks. You are to fast. I was in the process of writing the test ;-)

tobiasschweizer:
I know, I am a genius. This is what the told me in the special school I went to.

PR #490, opened 2017-04-24
doc (webapi): add graphdb initialization for unit testing

subotic:


SepidehAlassi:
looks good , but there is also a part about GraphDB in knora-api-server/deployment/webapi-getting-started which should be updated.

subotic:
Thanks, I missed it.

subotic:
Thanks :-)

PR #491, opened 2017-04-25
Improve bulk import

### LINKED ISSUE #434: Generate XML schema from ontology
- there should be a route which extracts xsd from ontology
- this xsd should be used by user and by Knora to validate xml

### LINKED ISSUE #454: In XML import, require the type of each value to be specified
Like this:

```
<biblio:publicationHasDate type="date">GREGORIAN:1974</biblio:publicationHasDate>
```

It will also be easier to parse it this way.

### LINKED ISSUE #455: Support file values in XML import
This will work as follows:

1. The user will submit each file to Sipi and get back the filename that Sipi generated for it.
2. The user will use the Sipi-generated filenames to generate the XML for the import, using them as the values of properties such as `knora-base:hasStillImageFileValue`.
3. During the import, Knora will insert the appropriate file metadata into the triplestore, without needing to contact Sipi.

This means that:

* All the files for the whole import are already in Sipi when the XML upload starts. So if the XML upload fails (e.g. because of an error in the XML), the images in Sipi remain in a consistent state.
* The data in the triplestore always refers to images that are already available in Sipi, so there are no broken links.
* Knora doesn't need to communicate with Sipi during the import, making the import simpler and more efficient.

### LINKED ISSUE #456: Support text with markup in XML import
It should be possible to include XML markup in text values submitted to the `/v1/resources/xml` route, just as in text values submitted to the `/v1/resources` route.

benjamingeer:
This PR makes a number of improvements to the XML import functionality.

* A GET to `/v1/resources/xmlimportschemas/ontologyIri` returns a Zip file of XML schemas for validating an XML import based on the specified ontology. The schemas are generated in `ResourcesRouteV1` from the relevant ontologies.
* If your XML is valid according to those schemas, you can POST it to `/v1/resources/xmlimport/projectIri` to create resources. The import is validated with the same schemas (in addition to the usual validation performed by the API server).
* You can include the filesystem paths of files on the server that should be uploaded to Sipi and attached to resources.
* You can include links to existing resources as well as links to other resources described in the import.
* You can include text values with XML markup to be stored as standoff. The XML markup can also include links to existing resources as well as links to other resources described in the import.

Docs are in `docs/rst/knora-api-server/api_v1/adding-resources.rst`.

Tests are in `ResourcesV1R2RSpec` and `SipiV1ITSpec`.

Fixes #434.
Fixes #454.
Fixes #455.
Fixes #456.


benjamingeer:
I opened #509 for any remaining SALSAH->Knora data conversion tasks.

docs/rst/knora-api-server/api_v1/adding-resources.rst
```diff
@@ -132,70 +126,232 @@ The resource's Iri has to be provided in the URL (as its last segment). The new
...
-The resources to be created must be described in an XML file. The XML file
-containing the resource descriptions can be imported directly to Knora by a
-POST request. The request has to be sent to the Knora server using the
-``resources/xml`` path segment.
+The procedure for using this feature is as follows:
 
-::
+1. Make a request to the Knora API server to get XML schemas describing the XML to be provided
+   for the import.
+2. If the project's data includes digital representations to be stored in Sipi, upload those files to
```
tobiasschweizer: Alternatively, we could allow the user to upload files to a place Knora has access to and indicate the file name and mime type (e.g., via FTP). Then we could use the logic as in the Multipart upload (the binaries sent to Knora are also saved to a temporary location: `InputValidation.createTempFile`).

Knora could then use the Sipi route `convertFromBinaries`

Making the thumbnails and copying the data returned into the XML does not seem very convenient to me.

docs/rst/knora-api-server/api_v1/adding-resources.rst
```diff
@@ -132,70 +126,232 @@ The resource's Iri has to be provided in the URL (as its last segment). The new
...
+- If a property has mutliple values, these are represented as multiple adjacent property elements.
+- The type of each value must be specified using the attribute ``knoraType``.
+- A link to another resource described in the XML import is represented as a child element of a property element,
+  with attributes ``knoraType="link_value"`` and ``linkType="ref"``, and a ``target`` attribute containing
+  the ID of the target resource.
+- There is a specfic syntax for referring to properties from other ontologies. In the example, ``beol:comment``
+  is defined in the ontology ``http://www.knora.org/ontology/beol``. In the XML, we refer to it as
+  ``biblio:beol__comment``.
+- A text value can contain XML markup. In this case:
+    - The text value element must have the attribute ``mapping_id``, specifying a mapping from XML to standoff markup (see :ref:`XML-to-standoff-mapping`).
```
tobiasschweizer: The mapping_id is only needed in case some markup is provided, right? If there is just plain text, no mapping needs to be indicated.

webapi/src/test/scala/org/knora/webapi/e2e/v1/ResourcesV1R2RSpec.scala
```diff
@@ -1250,45 +1177,181 @@ class ResourcesV1R2RSpec extends R2RSpec {
                 }
 
                 assert(utf8str == "another simple text")
+            }
+        }
 
+        "create resources from an XML import" in {
```
tobiasschweizer: Could you provide a test using the anything ontology with all the possible value types occurring?

webapi/src/test/scala/org/knora/webapi/e2e/v1/ResourcesV1R2RSpec.scala
```diff
@@ -1250,45 +1177,181 @@ class ResourcesV1R2RSpec extends R2RSpec {
...
+            val xmlImport =
+                s"""<?xml version="1.0" encoding="UTF-8"?>
+                   |<knoraXmlImport:resources xmlns="http://api.knora.org/ontology/biblio/xml-import/v1#"
+                   |    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   |    xsi:schemaLocation="http://api.knora.org/ontology/biblio/xml-import/v1# biblio.xsd"
+                   |    xmlns:biblio="http://api.knora.org/ontology/biblio/xml-import/v1#"
```
benjamingeer: Both `http://ecommerce.example.org/schema` and `http://ecommerce.example.org/schema#` are valid XML namespaces, [and they are considered different XML namespaces](https://stackoverflow.com/a/431315).

In RDF, a namespace and a local part are concatenated, so you have to decide whether the namespace should end with `#` or `/`. Thus in RDF you have [either a 'hash namespace' or a 'slash namespace'](https://www.w3.org/2001/sw/BestPractices/VM/http-examples/2006-01-18/#naming).

But in XML, the namespace and the local part are not concatenated. The namespace and the local part are two separate things, which together uniquely identify an element. There is no concept in XML of concatenating them together to make a single URI, as there is in RDF. See [this question and the answers](http://lists.xml.org/archives/xml-dev/201207/msg00308.html).

However, we would like our API to be based on dereferenceable ontology entities, at least in version 2. So I think we should use RDF namespace conventions in the XML import. Then if someone tries to dereference `http://api.knora.org/ontology/biblio/xml-import/v2#JournalArticle`, they should get some information about that class. I don't want to implement this in v1, but for consistency I'd like to use the same naming convention there, too.

docs/rst/knora-api-server/api_v1/adding-resources.rst
benjamingeer: In any case, all the files have to be uploaded to Sipi before any data is added to the triplestore. So the import really has two distinct parts: image upload and triplestore update. The image upload part is going to be a long, error-prone process in which Sipi may crash. Then the user will have to submit a bug report to Sipi, the bug will have to be fixed, and the user will have to try again. I think we should keep this process as simple as possible, to reduce the number of things that can go wrong. Therefore, in the image upload part of the import, I think it's better if the user only has to deal with Sipi, rather than Sipi and Knora.

I don't know whether we actually need to generate thumbnails in an import. Maybe we can just use a Sipi route that directly stores the image, without making a thumbnail. From Knora's point of view, all that matters is that the necessary image metadata is provided.

I don't see this as terribly inconvenient, and I'm going to have to do it myself. I'll just write a little script that sends all the images to Sipi and writes the metadata to a CSV file. Then I'll read this CSV file when generating the XML for the import.

docs/rst/knora-api-server/api_v1/adding-resources.rst
benjamingeer: Yes, as shown in the example (most of the text values don't have mappings).

webapi/src/test/scala/org/knora/webapi/e2e/v1/ResourcesV1R2RSpec.scala
benjamingeer: OK.

docs/rst/knora-api-server/api_v1/adding-resources.rst
```diff
@@ -132,70 +126,232 @@ The resource's Iri has to be provided in the URL (as its last segment). The new
...
+2. If the project's data includes digital representations to be stored in Sipi, upload those files to
+   Sipi, and store the metadata that Sipi returns for each file.
+3. Convert your data into XML, including any file metadata.
+4. Use an XML schema validator such as `Apache Xerces`_ or Saxon_, or an XML development environment
+   such as Oxygen_, to check that your XML is valid according to the schemas you got from the Knora
+   API server.
+5. Submit your XML to the Knora API server.
 
-     HTTP POST to http://host/v1/resources/xml
+In this procedure, the person responsible for generating the XML import data need not be familiar
```
benjamingeer: 'need' is correct: https://www.englishgrammar.org/auxiliary-verb-ordinary-verb/

docs/rst/knora-api-server/api_v1/adding-resources.rst
benjamingeer: After discussion with Tobias and Lukas, we will do it like this: The XML will specify the path for each file. Knora will send the file to Sipi. This way, we ensure that the metadata (including original filename and original MIME type) is correct, and that Sipi can read the file.

loicjaouen: So we need to upload files to `knora` outside of `knora` so `knora? can find them locally and be able to upload them to `sipi`?

benjamingeer: The idea is that you have all the files in a directory somewhere on disk, and the XML import data just specifies the path of each file on disk. As Knora is processing the XML, it reads each file and sends it to Sipi.

webapi/src/test/scala/org/knora/webapi/e2e/v1/ResourcesV1R2RSpec.scala
benjamingeer: This is done.

docs/rst/knora-api-server/api_v1/adding-resources.rst
benjamingeer: This is done.

benjamingeer:
@tobiasschweizer can you please review again?

webapi/src/test/scala/org/knora/webapi/e2e/v1/ResourcesV1R2RSpec.scala
tobiasschweizer: So in short this means, in XML an element or attribute name has solely to be uniquely identified but not dereferenceable?

We absolutely want our Iris used in RDF to be dereferenceable, I totally agree.

docs/rst/knora-api-server/api_v1/adding-resources.rst
tobiasschweizer: ok

docs/rst/knora-api-server/api_v1/adding-resources.rst
tobiasschweizer: ok, learned something about English grammar :-)

docs/rst/knora-api-server/api_v1/adding-resources.rst
tobiasschweizer: ok

webapi/src/test/scala/org/knora/webapi/e2e/v1/ResourcesV1R2RSpec.scala
tobiasschweizer: ok

webapi/src/test/scala/org/knora/webapi/e2e/v1/ResourcesV1R2RSpec.scala
benjamingeer: It looks like the XML specs encourage you to make namespaces dereferencable, but there is no standard way to attach an element name to its namespace and get an IRI (dereferenceable or not).

PR #514, opened 2017-05-29
docs (recover Futures)

### LINKED ISSUE #502: Make a design pattern for chaining stack traces between actors
If actor A sends a message to actor B, actor B then sends a message to actor C, and actor C returns a failed `Future`, we should report the whole context of the error, i.e. what A was doing when it sent its message to B, what B was doing when it sent its message to C, and what C was doing when it threw the exception. This should be reported in the form of chained exceptions. Let's make a design pattern that accomplishes this as painlessly as possible.

tobiasschweizer:
Provides some documentation about how to use  `recover` on  `Future`.

Fixes #502.


subotic:
Cool thanks :-)

PR #527, opened 2017-06-01
update ckeditor to 4.7.0

### LINKED ISSUE #526: update ckeditor
We are on [4.6](http://ckeditor.com/release/CKEditor-4.6.2), [4.7](http://ckeditor.com/release/CKEditor-4.7.0) is out.

Why do we want to update ckEditor to a newer version?
- [best practices](http://docs.ckeditor.com/#!/guide/dev_best_practices) recommendations
- it works better with word import
- at least, I was more successful when installing [footnotes](http://ckeditor.com/addon/footnotes) addon

loicjaouen:
fixes #526


tobiasschweizer:
Thx.

I will have to do that next week, I am not working tomorrow.

loicjaouen:
no problem, have fun!

salsah/src/public/js/jquery.htmleditor.js
```diff
@@ -165,8 +166,25 @@
 							
 						}
 					},
-					toolbar: [ ['Source', 'Bold', 'Italic', 'Underline', 'Strike', 'Subscript', 'Superscript', '-','RemoveFormat', 'Link', 'Unlink'] ] // configuration for toolbar buttons, must conform to `STANDARD_MAPPING`
-				};
+
+					// configuration for toolbar buttons, must conform to `STANDARD_MAPPING`,
```
tobiasschweizer: Are you sure that this conforms to the standard mapping?

salsah/src/public/js/jquery.htmleditor.js
```diff
@@ -141,7 +141,8 @@
                 var filter = ' p em strong strike u sub sup; a[!href](salsah-link)';
```
tobiasschweizer: The configuration of CKeditor, the filter and the standard mapping have to conform with one another.

tobiasschweizer:
I think for the thing you have in mind, project specific configurations would be well suited. However, the old SALSAH GUI does not support that. So you would have to find a way to implement that.

This would involve:
- a specific mapping
- a specific configuration for CKeditor (toolbar -> controls the markup produced by the editor)
- a specific filter for CKeditor (markup validation)

You could assign each project a configuration. When creating a new text value, CKeditor should be loaded with the project's configuration. When editing an existing text, the configuration used before would have to be reloaded.

salsah/src/public/js/jquery.htmleditor.js
loicjaouen: yes, 



that's respectively `StandoffBoldTag`, `StandoffItalicTag`, `StandoffUnderlineTag`, `StandoffStrikethroughTag`, `StandoffSubscriptTag`, `StandoffSuperscriptTag` and `StandoffUriTag`

salsah/src/public/js/jquery.htmleditor.js
loicjaouen: and they do

loicjaouen:
@tobiasschweizer : thanks for the review
if you don't mind, let's split issues, 
- the upgrade of ckeditor (4.6.2 to 4.7.0), here
- the extension of ckEditor use #528

tobiasschweizer:
Yes, that's a good idea!

Could you please restore the original toolbar configuration and test it? Also it would be good to have browser tests for CKeditor.

loicjaouen:
on this branch, branch the only addition to the original toolbar is the "in app" versionning:



> Also it would be good to have browser tests for CKeditor.

sure, but let's put that in another issue too.

tobiasschweizer:
@loicjaouen Would you like to revert the changes that do not include ckeditor files? So the new version would used with the existing configuration.

Of course you would have to make sure that the new version is backwards compatible witht he old config options :-)

loicjaouen:
@tobiasschweizer : no ;)  
 this branch is only an upgrade of ckEditor with the required changes on the config to strictly do what the old one was doing.   
The changes on the config are just to maintain the same feature set as the previous version, while following the best practices listed in ckEditor's documentation.  
(I tested it even though I did not add the browser test yet)

tobiasschweizer:
it seems to me that now a lot of features are enabled in the config that were not before: salsah/src/public/js/jquery.htmleditor.js

The browser tests for ckeditor are skipped

salsah/src/public/js/jquery.htmleditor.js
```diff
@@ -165,8 +166,25 @@
...
+						{ name: 'paragraph', groups: [ 'list', 'indent', 'blocks', 'align', 'bidi', 'paragraph' ] },
+						{ name: 'links', groups: [ 'links' ] },
+						{ name: 'insert', groups: [ 'insert' ] },
+						{ name: 'styles', groups: [ 'styles' ] },
+						{ name: 'colors', groups: [ 'colors' ] },
+						{ name: 'tools', groups: [ 'tools' ] },
+						{ name: 'others', groups: [ 'others' ] },
+						{ name: 'about', groups: [ 'about' ] }
+					],
+					removeButtons: 'Cut,Copy,Paste,Anchor,NumberedList,BulletedList,Outdent,Indent,About'
```
loicjaouen: this removes un-needed features.

loicjaouen:
It looks like it, but check on the screenshot, it is just plain boring as it use to be.  

What the configurator does is that it takes a set of features, and then removes some:  
`jquery.htmleditor.js:186`
```javascript
removeButtons: 'Cut,Copy,Paste,Anchor,NumberedList,BulletedList,Outdent,Indent,About'
```

tobiasschweizer:
I get it but why can't we stick to the old config that just added what we need: `toolbar: [ ['Source', 'Bold', 'Italic', 'Underline', 'Strike', 'Subscript', 'Superscript', '-','RemoveFormat', 'Link', 'Unlink'] ]`

Do you think that this is not the recommended way to do it?

salsah/src/public/js/jquery.htmleditor.js
```diff
@@ -141,7 +141,7 @@
                 var filter = ' p em strong strike u sub sup; a[!href](salsah-link)';
 
 				var config = {
-					language: 'de', // customize language
+					language: SALSAH.userprofile.userData.lang === undefined ? 'de' : SALSAH.userprofile.userData.lang,
```
loicjaouen: by the way, should the default language remain German? 
(by default the rest of the page is in English)

tobiasschweizer: I think you can set the default ton English

tobiasschweizer:
Should I have a look at this again?

loicjaouen:
yes, please :)

salsah/src/public/js/jquery.htmleditor.js
```diff
@@ -165,7 +165,8 @@
 							
 						}
 					},
-					toolbar: [ ['Source', 'Bold', 'Italic', 'Underline', 'Strike', 'Subscript', 'Superscript', '-','RemoveFormat', 'Link', 'Unlink'] ] // configuration for toolbar buttons, must conform to `STANDARD_MAPPING`
+					toolbar: [ ['Source', 'Bold', 'Italic', 'Underline', 'Strike', 'Subscript', 'Superscript', '-','RemoveFormat', 'Link', 'Unlink'] ], // configuration for toolbar buttons, must conform to `STANDARD_MAPPING`
+					removeButtons: ''
```
tobiasschweizer: Why is this line needed?

loicjaouen: this overwrites the default config, which comes from the file [config.js](https://github.com/dhlab-basel/Knora/blob/wip/update-ckeditor/salsah/src/public/vendor/ckeditor_4.7.0/config.js)

tobiasschweizer: got it, thx.

loicjaouen: I see now that on the previous ckEditor, you (?) emptied that file.  
We can also do that.

tobiasschweizer: not that I would be aware of, no ;-)

This seems to work, so let's use it like that.

loicjaouen: concerning the log that shows up ("TypeError: can't access dead object" `htmlPanel.js(1082,17)`) .

As we don't load that library directly (no `<script src="htmlPanel.js">`), do you know where it is loaded from?

From this [MDN page](https://developer.mozilla.org/ca/docs/Web/JavaScript/Reference/Errors/Dead_object), it looks like it comes from firefox add-ons.

I don't get this error, but maybe running the same on firefox with add-ons disabled or on another firefox-based browser (like the [developer edition](https://www.mozilla.org/en-US/firefox/developer/)) would help to decide if this message is linked to our code or not.

tobiasschweizer: I think it comes from Firebug since I use the console. I have the impression it is caused by switching between the source code and the rendered text (toolbar "Source"). The whole editor instance is kept in an iframe which has its own DOM. When switching views, some DOM may be exchanged by another, breaking references. 

I think we should disable source view and merge this pull request.

loicjaouen:
It is an error that only shows up in FireBug's console when the options [show chrome errors/messages](https://getfirebug.com/wiki/index.php/Console_Panel) are enabled and I could not see any side-effects.

It looks to me a bit of an overkill to axe the source-view button... except if we have an otherwise reason to do so (less risk to feed knora with unexpected tags).

But discussing it might also be an overkill :)

I remove it and merge this PR.

(sometimes it shows strange things)
capture d ecran 2017-06-15 a 14 46 50

tobiasschweizer:
I only added it to look at the markup that is created, The user shouldn't have to know about that (WYSIWYG).

The "strange thing" above is the select box for a referred resource overlying CKeditor. I think this is a CSS problem.

tobiasschweizer:
great, thx.

loicjaouen:
:) I merge as soon as the build ends

PR #541, opened 2017-06-16
cosmetics on the documentation

loicjaouen:


loicjaouen:
thanks for the good documentation and work, everything we need is there!

tobiasschweizer:
Great to hear :-) Ben and I are going to work on the child to parent relations of standoff elements (https://github.com/dhlab-basel/Knora/pull/438) some day ;-)

At the moment, we assume that the client submits XML that as between validated using a schema. But also in standoff we should validate certain child to parent relations (wherever they make sense).

PR #548, opened 2017-06-20
Add an untyped key-value store backed by the triplestore

### LINKED ISSUE #524: Add an untyped key-value store backed by the triplestore
The new SALSAH GUI would like to be able to persist some state, such as draft ontologies that the user is developing, but not permanently. We can provide a simple key-value store for this, backed by the triplestore, with data stored in a separate named graph.

benjamingeer:
Basic ideas:

* `knora-base:Map` and `knora-base:MapEntry` are the OWL classes.
* They're stored in a separate named graph.
* The IRI of a map can contain a path representing some kind of hierarchy. So can the key of a map entry.
* Whenever you change a map entry, you first check whether the map exists, and create it if it doesn't.
* Both maps and map entries have a `knora-base:lastModificationDate`.

Resolves #524.

### DIFF
--- a/knora-ontologies/knora-admin.ttl
+++ b/knora-ontologies/knora-admin.ttl
@@ -436,11 +436,11 @@
 
 :isInMap rdf:type owl:ObjectProperty ;
 
-    rdfs:subPropertyOf knora-base:objectCannotBeMarkedAsDeleted ;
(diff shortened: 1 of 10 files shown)


benjamingeer:
Moving this into #478.

PR #552, opened 2017-06-23
Wip/salsah searchbox

### LINKED ISSUE #546: evanescent searchbox
When the searchbox pops down, it doesn't hide the buttons below.

We not only see them, they remain clickable which is problematic as they might be clicked instead of the menu item.

ckeditor

ckeditor-not

### LINKED ISSUE #547: volatile searchbox choices
Choices made in a searchbox are a bit volatile; a click and they disappear but when the resource is saved, the choice that just vanished is still remembered.

link-reclick-disappear

loicjaouen:
fixes #546 
fixes #547 

screencast before/after is in #547 

There is still no test...


webapi/_test_data/ontologies/anything-onto.ttl
```diff
@@ -70,7 +70,9 @@
 
                salsah-gui:guiOrder "1"^^xsd:integer ;
 
-               salsah-gui:guiElement salsah-gui:Searchbox .
+               salsah-gui:guiElement salsah-gui:Searchbox ;
+	       
+	       salsah-gui:guiAttribute "size=45".
```
tobiasschweizer: Would this have to be adapted for all ontologies using linking properties?

salsah/src/public/js/jquery.searchbox.js
```diff
@@ -206,7 +208,12 @@
 							left: pos.left,
 							top: pos.top
 						}).show();
-						$this.val('');
+						// if the input already has a value and it is valid, back it up
```
tobiasschweizer: If I understand correctly, `$this.val()` is the text displayed in the input box.

However, the main thing we are interested in is the Iri of the resource referred to: `$this.data('value', $(this).data('id'));`

salsah/src/public/js/jquery.searchbox.js
```diff
@@ -185,6 +185,8 @@
 //						position: 'absolute',
 						left: pos.left,
 						top: pos.top +$this.outerHeight(),
+						background: 'white',
+						'z-index': 2, /* :( i don't like black magic numbers or const */
```
tobiasschweizer: What do you mean by that? `z-index` defines if an element overlays another element (or the other way round)

webapi/_test_data/ontologies/anything-onto.ttl
loicjaouen: it is optional, so there is no need to change any ontologies.
(plus, who knows how salsah v2 will make use of this)

salsah/src/public/js/jquery.searchbox.js
loicjaouen: The underlying buttons remains on top of the menu (not only visible, also selectable, which is a problem to select the "drop down" search list items), see animated gif on #546 .  

So we have to set a `z-index`to bring the menu items on top.  

And here I set `2` which is a _black magic number_ (like "white" [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) but more dangerous and error prone because of the esoteric, arbitrary and non documented picking).

At other places in the code, other black magic values where chosen (55 in `jquery.editvalue.js` or `jquery.tableedit.js`, 1000 in `jquery.valcomment.js`or `jquery.geonames.js`).

webapi/_test_data/ontologies/anything-onto.ttl
tobiasschweizer: Right! 

For V2, I haven't looked at the gui elements yet.

salsah/src/public/js/jquery.searchbox.js
tobiasschweizer: CSS z-index is indeed difficult (and different every browser) ...

if it works, I guess it's fine

salsah/src/public/js/jquery.searchbox.js
loicjaouen: exactly.

Do you mean that instead of using cached values, I should go and fetch the source of this value? Yes, that would be smarter.

But when leaving the scope on `jquery.searchbox.js:212` finding that value was not obvious to me.

tobiasschweizer: > finding that value was not obvious to me

neither for me :-)

I think as long as the user does not select a new resource (click), the Iri remains saved so it should be ok.

loicjaouen: yes, the iri remains, I used a cache because finding it was too difficult

salsah/src/public/js/jquery.resadd.js
```diff
@@ -517,6 +517,7 @@
 												}
 											}
 											tmpele.attr('placeholder', placeholderText);
+											tmpele.attr('autocomplete', 'off');
```
loicjaouen: oh man, I'm sorry to push this one late, but it is a little refinement that helps; turning autocomplete off makes that there is no drop down list of previous attempts that hides the "official list":

capture d ecran 2017-06-26 a 11 13 40

loicjaouen: here "thi" that I previously entered hides the "New entry..." below

tobiasschweizer: sorry, but I do not understand what the attribute `autocomplete` is. Was is already there or did you add it?

loicjaouen: I did add it.  

As you know it, autocomplete is a feature of the browser, it does the same as the searchbox does: suggesting results as the user is typing.  

And therefore, they are concurrent by behaving the same way.  

Disabling `autocomplete` is a directive to the browser not to store and propose previous choices.

With `autocomplete` enabled, the browser keeps a list of previous values and suggests the matching ones. So if I searched for "thi" before, it will pop-down a list with "thi", hiding the results of the `searchbox`.

I added it, I don't know how, but salsah v1.0 didn't try to autocomplete.

tobiasschweizer:
@loicjaouen Shall I look at this again now?

loicjaouen:
yes please :)

salsah/src/public/js/jquery.searchbox.js
tobiasschweizer: So this means the caching is only needed to restore the text the user has typed if he gave focus to another element and then switches back to the text box?

salsah/src/public/js/jquery.resadd.js
tobiasschweizer: Ok, now I understand. I think it's good to deactivate the standard autocomplete because this could confuse the user.

salsah/src/public/js/jquery.searchbox.js
```diff
@@ -217,6 +224,15 @@
 								$this.css({cursor: 'text'});
 							}
 							localdata.ele.sel.hide();
+							// when leaving this input, if the value is empty, but we had a previous valid input
+							// set back the previous input
+							if ($this.attr("isValid") && $this.attr("isValid") == "false") {
```
tobiasschweizer: In general, I prefer to write `$this.attr("isValid") !== undefined` (in case the requested attribute isn't set, `undefined` would be returned).

tobiasschweizer: I simply try to avoid relying on type conversions because this could have unexpected effects. Whenever possibly, I use `===` or `!==` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators).

tobiasschweizer: But I guess in this case you could also write: `$this.attr("isValid") === "false"`: he condition will only be true, if the string "false" is returned.

Probably I am opening Pandora's box here because this kind of discussion could totally get out of hand ;-)

loicjaouen: not at all, you are completely right as stated on page 109, I will _make [myself] a better programmer_ and avoid `==` from now on.

tobiasschweizer: To be honest: the old GUI is full of JS bugs and this is also my fault (I also blame @lrosenth) ;-)

loicjaouen: change comitted for that

salsah/src/public/js/jquery.searchbox.js
loicjaouen: if the input box never had focus, it holds no value, there is then no need for a cache.  
The problem is that we want to clear an existing text to let room for the user to type in a new text to search, but this new text search might not end up in a valid answer.  
Still the input value is the text the user typed in so far.

That's why I added a `isValid`, if we find a value and click on its line in the drop-down list, then we set the input's value with a new text and we set the `isValid` attribute to true.

But by default we have to set it to false.

As it is a html tag, it gets turned into a string... it might help clarity if I put all `isValid` assignment to string?

loicjaouen: I did so. ("stringified" `true`/`false` values I set `isValid`) in 5beebbd8e6c83f1d0d70af56eafdb5f32a332a17

Thanks for the "javascript, the good parts" reference!

tobiasschweizer: great, thx.

> Thanks for the "javascript, the good parts" reference!

It contains some useful information, but some really academic cases too. 

check out the JS pope: https://www.youtube.com/watch?v=NPB34lDZj3E

loicjaouen:
@tobiasschweizer : I didn't expect the knora CI build to fail as I touched only salsah (and a bit of `salsah-gui.ttl`), I don't fancy how this can have a side-effect on the CI build?

But the other fix I just committed passed the build just fine... (scratching my head)

subotic:
It's failing since `b2dffe0` according to this PR's history.

tobiasschweizer:
This test fails: `[info] - should show outgoing standoff links if the user has view permission on both resources, but show other outgoing links only if the user also has view permission on the link *** FAILED ***
[info]   false did not equal true (ResourcesResponderV1Spec.scala:1392)`

webapi/src/test/scala/org/knora/webapi/responders/v1/ResourcesResponderV1Spec.scala
```diff
@@ -206,7 +206,7 @@ object ResourcesResponderV1Spec {
             targetResourceIri = "http://data.knora.org/project-thing-2"
         )),
         occurrence = Some("0-n"),
-        attributes = "restypeid=http://www.knora.org/ontology/anything#Thing",
+        attributes = "size=45;restypeid=http://www.knora.org/ontology/anything#Thing",
```
tobiasschweizer: Since you defined a size for the property `anything:hasOtherThing`, this attribute's value has to be included in the expected test results.

tobiasschweizer: I should have noticed this when I did the review.

loicjaouen: damned... I should not change anything, at least `anything`.  
Sorry about that.

tobiasschweizer: Don't worry, I think now it should be fine. I running all the tests with graphdb locally right now.

loicjaouen:
the official build is successful too :)
shall I merge?

tobiasschweizer:
From my side we are good to go. If everything you wanted works as expected now, please merge.

loicjaouen:
many thanks for your time and insight Tobias

PR #557, opened 2017-06-30
sort props by guiorder

### LINKED ISSUE #555: salsah-gui:guiOrder works for the "add resource" icon but not when you look at a resource already created
`salsah-gui:guiOrder` should also define the way the properties of a resource are displayed after the resource is created.

capture d ecran 2017-06-29 a 11 26 03

loicjaouen:
fixes #555

we take the Object:

```json
{ 
  propname: propdata,
  propname: propdata,
  ...
}
```
that looks like:

```json
"props": {
        "http://www.knora.org/ontology/anything#hasDecimal": {
            "regular_property": 1,
            "guiorder": 5,
            "is_annotation": "0",
            "valuetype_id": "http://www.knora.org/ontology/knora-base#DecimalValue",
            "label": "Dezimalzahl",
            "guielement": "text",
            "attributes": "size=80;maxlength=255",
            "occurrence": "0-n",
            "pid": "http://www.knora.org/ontology/anything#hasDecimal"
        },
```

turn it into an array:

```json
[
  [propname, propdata], [propname, propdata], ...
]
```

it makes that we find `propname` and `propdata` as:
```json
propname = sortedprops[i][0];
propdata = sortedprops[i][1]; 
```

```json
[
  [ "http://www.knora.org/ontology/anything#hasDecimal", 
    {
            "regular_property": 1,
            "guiorder": 5,
            "is_annotation": "0",
            "valuetype_id": "http://www.knora.org/ontology/knora-base#DecimalValue",
            "label": "Dezimalzahl",
            "guielement": "text",
            "attributes": "size=80;maxlength=255",
            "occurrence": "0-n",
            "pid": "http://www.knora.org/ontology/anything#hasDecimal"
        }
  ],
  ...
]
   ```

we sort it according to `propdata.guiorder` with the method: `prop1[1].guiorder - prop2[1].guiorder`.

### CODE REVIEW

**Comment by loicjaouen** [comment]
i tried to optimize with the assumption that `propname` is found in `propdata.pid`, which made teh code simpler, but I had to revert because along the way we add to the original data the such elements as `propname="__label__"` which has a different propdata.

**Comment by loicjaouen** [comment]
@tobiasschweizer : can I bother you again with a review request?

**Comment by loicjaouen** [comment]
@tobiasschweizer : do you think you will have some time to eventually drop an eye on this?

**Comment by tobiasschweizer** [comment]
Sure! Is it ok if I do it next week?

**Comment by loicjaouen** [comment]
that would be perfect :)

**Comment by tobiasschweizer** [comment]
If I understand correctly, you turned the object representing a resources properties into a new data structure, that is sorted by `guiorder`. This data structure is a 2-dimensional array, whose items have the length two: the first item represents the propname, the second item represents the property value.

The 2-dimensional array can be sorted according to `guiorder` by passing a anonymous function to `sort` that compares the `guiorder`. 

But shouldn't this happen already on the server side? `ResourcesResponderV1` has a method called `queryResults2PropertyV1s` that is responsible for creating the properties. @benjamingeer I can see that value objects are ordered by `ValueHasOrder`. Should we add `sortBy` guiorder?

**Comment by loicjaouen** [comment]
@tobiasschweizer, in the issue's (#555) comments, @subotic was also suggesting to do the sort server side as it would be a one liner fix.  
As I already coded it (and released it for demo), I pulled this request.  
But it can be dropped if the feature is provided in another way.

**Comment by tobiasschweizer** [comment]
I talked to @benjamingeer: the problem is that we create a Map before writing the information to JSON. because of this, we cannot sort the results.

We would have to use another data structure (an Array instead of an object).

So your approach is good!

File: salsah/src/public/js/imagebase.js [hunk 1]
```diff
@@ -90,6 +90,10 @@ $(function() {
 			//***************
 
 
+			var sortedprops = Object.keys(resource.props).map(function(propname) {return [propname, resource.props[propname]];});
```
tobiasschweizer: Could you add some documentation describing the data structure you are creating here?

Probably you could also create an Array of objects instead of a 2-dimensional Array. You could use keys `propname` and `value`.

tobiasschweizer:
```
sortedprops.sort(function(prop1, prop2){
				return prop1.value.guiorder - prop2.value.guiorder;
			});
```

benjamingeer:
> the problem is that we create a Map before writing the information to JSON. because of this, we cannot sort the results.

It's not because we use a Map, it's because a JSON object is by definition unordered:

> An object is an unordered set of name/value pairs.

http://www.json.org/

salsah/src/public/js/imagebase.js
loicjaouen: @tobiasschweizer : yes, it is a bit tricky, that's why I described it at length at the beginning of the issue #557. Would like a shorter version in the code?

tobiasschweizer: Yes, that would be good!

loicjaouen:
Hello @tobiasschweizer , I added some implementation notes, and you are right the Array of Array is not as nice as an array of Object with named properties: 
```js
propname = sortedprops[i][0]
propdata = sortedprops[i][1]
```
vs.:
```js
propname = sortedprops[i].name
propdata = sortedprops[i].data
```

My habit of strongly typed language still rings big _warning_ bell: who knows that there is a `.name` property? I do, because I filled the Array, but who tells me that there is a second element in the array (`sortedprops[i][1]`)? Only me, again.

loicjaouen:
Please have a look whenever :)

loicjaouen:
Thanks :)

PR #559, opened 2017-07-04
Wip/beol_onto Era added to KnoraDate + ontology changes 

SepidehAlassi:
features:
-Era (BC/AD) added to KnoraDate with corresponding tests 
-Tests for InputValidation.KnoraDateRegex added
-beol and biblio ontologies are extended wrt needs of beol project


benjamingeer:
It would be nice to also be able to call them BCE and CE instead of BD and AD. This is important to some people:

https://en.m.wikipedia.org/wiki/Common_Era

mrivoal:
Exactly! That is just what we were talking about ;-)

SepidehAlassi:
@benjamingeer @mrivoal in beol we have just BC, and AD. I will look into adding BCE, CE.

PR #560, opened 2017-07-04
feature (Wip/beol date) Era added to KnoraDate + changes on ontology

### LINKED ISSUE #184: Should RDF timestamps use dateTime or dateTimeStamp?
Currently the cardinalities in `knora-base` specify `xsd:dateTimeStamp` (e.g. for `knora-base:lastModificationDate`), but SPARQL's `NOW()` function returns an `xsd:dateTime`, so we end up with some of each.

### LINKED ISSUE #563: SPARQL update for bulk import causes GraphDB stack overflow
The SPARQL updates generated by the bulk import code are triggering a `java.lang.StackOverflowError` in GraphDB SE 8.1. Here’s an example that attempts to create 56 resources:

https://drive.switch.ch/index.php/s/zykh729Wx2sPXLt

I asked Ontotext about it and they replied:

> The SPARQL query parser implements a recursive algorithm for the query model parsing. The only exception is INSERT queries without a WHERE clause. The only effective workaround is to increase your stack size so that it can work with your bigger queries. Here are detailed instructions:
>
> Please check what's the current stack size with:
>
> `java -XX:+PrintFlagsFinal -version | grep ThreadStackSize`
>
> If you get 1024, I suggest increasing it to 4MB with:
>
> `-Xss4m`
>
> This change will increase the application's memory footprint by the formula: (number of parallel threads) * (new bigger value - old smaller value)
>
> If you have a production system with 70 concurrent threads, this will eat up around 210MB extra.

I'm uneasy about this sort of workaround, for two reasons:

1. It requires additional configuration that people could easily forget to do, or that could easily be lost when GraphDB is upgraded.
2. The point of the bulk import feature is to accept large amounts of data, so it seems likely that at some point, no matter what stack size we choose, it won't be enough.

So maybe the best way to fix this is to eliminate WHERE clauses in the SPARQL updates for bulk imports, turning the INSERT operations into INSERT DATA operations.

SepidehAlassi:
- new branch to substitute the beol_onto branch which has commit history problem
- Era (BC,AD,BCE,CE) added to KnoraDate
- beol and biblio ontologies extended  
- Use INSERT DATA instead of INSERT to avoid stack overflow in bulk import with GraphDB
- Use GraphDB Free instead of Fuseki as default triplestore

Fixes #184.
Fixes #563.


webapi/src/main/scala/org/knora/webapi/util/InputValidation.scala
```diff
@@ -54,20 +54,21 @@ object InputValidation {
...
+    // Calendar:YYYY[-MM[-DD]][ EE][:YYYY[-MM[-DD]][ EE]]
+    // EE being the era: one of BC or AD
     private val KnoraDateRegex: Regex = ("""^(GREGORIAN|JULIAN)""" +
         CalendarSeparator + // calendar name
         """\d{1,4}(""" + // year
         PrecisionSeparator +
         """\d{1,2}(""" + // month
         PrecisionSeparator +
-        """\d{1,2})?)?(""" + // day
+        """\d{1,2})?)?( BC| AD| BCE| CE)?(""" + // day
```
tobiasschweizer: You could put `BC| AD| BCE| CE` in a string constant (just to avoid the risk of inconsistency).

webapi/src/main/scala/org/knora/webapi/util/InputValidation.scala
```diff
@@ -54,20 +54,21 @@ object InputValidation {
...
         CalendarSeparator + // separator if a period is given
         """\d{1,4}(""" + // year 2
         PrecisionSeparator +
         """\d{1,2}(""" + // month 2
         PrecisionSeparator +
-        """\d{1,2})?)?)?$""").r // day 2
+        """\d{1,2})?)?( BC| AD| BCE| CE)?)?$""").r // day 2
```
tobiasschweizer: see comment above

webapi/src/main/scala/org/knora/webapi/util/InputValidation.scala
```diff
@@ -333,10 +334,6 @@ object InputValidation {
       * @return the same string.
       */
     def toDate(s: String, errorFun: () => Nothing): String = {
-        // TODO: how to deal with dates BC -> ERA
-
-        // TODO: import calendars instead of hardcoding them
```
tobiasschweizer: > TODO: import calendars instead of hardcoding them

Probably I wrote this TODO. I do not remember what it was supposed to mean. Maybe that we hardcoded JULIAN and GREGORIAN calendars? How would include an additional calendar (could we make this configurable)?

webapi/src/main/scala/org/knora/webapi/util/DateUtilV1.scala
```diff
@@ -96,7 +96,24 @@ object DateUtilV1 {
         val daysInMonth = Calendar.DAY_OF_MONTH // will be used to determine the number of days in the given month
         // val monthsInYear = Calendar.MONTH // will be used to determine the number of months in the given year (generic for other calendars)
 
-        val dateSegments = dateString.split(InputValidation.PrecisionSeparator)
+        var era: Option[Int] = None
```
tobiasschweizer: the use of `var` should be avoided. Can't this be rewritten?

For instance you could make a `match` statement return a value and assign that to a variable declared with `val`.

webapi/src/main/scala/org/knora/webapi/util/DateUtilV1.scala
```diff
@@ -96,7 +96,24 @@ object DateUtilV1 {
...
+
+        var date: String =dateString
+        if (dateString.contains(" ")) {
```
tobiasschweizer: If there is no era indicated, just set it to `AD` so the subsequent handling will be no different.

webapi/src/main/scala/org/knora/webapi/util/DateUtilV1.scala
tobiasschweizer: Proposal:

```
        val dateStringSplitByEra: Array[String] = dateString.split(" ")

        // if there is an era indicated, size is 2
        val era: Int = dateStringSplitByEra.length match {

            case 1 =>
                // no era indicated, assume AD/CE
                GregorianCalendar.AD

            case 2 =>
                dateStringSplitByEra(1) match {
                    case "BC" => GregorianCalendar.BC
                    case "AD" => GregorianCalendar.AD
                    // java Gregorian calendar had just BC and AD as public fields
                    case "BCE" => GregorianCalendar.BC  //BCE = BC
                    case "CE" => GregorianCalendar.AD   //CE = AD
                    case _ => throw BadRequestException(s"Could not handle era in $dateString") // this should never happen due to KnoraDateRegex
                }

            case _ => throw BadRequestException(s"Could not handle era in $dateString") // this should never happen due to KnoraDateRegex
        }


        val dateSegments: Array[String] = dateStringSplitByEra(0).split(InputValidation.PrecisionSeparator)
```

and then subsequently:

```
intervalStart.set(Calendar.ERA, era)
```

tobiasschweizer: However, I would propose to make `" "` a constant. And  `BC` `AD` as well.

webapi/src/test/scala/org/knora/webapi/util/DateUtilSpec.scala
```diff
@@ -52,6 +54,25 @@ class DateUtilSpec extends WordSpec with Matchers {
             reverseConvertedBundesbriefDateValueV1 should be(bundesbriefDateValueV1)
         }
 
+        "convert an era date string with year precision to a Java GregorianCalendar BC" in {
+            val dateRange: DateRange = DateUtilV1.dateString2DateRange("50-02-28 BC", KnoraCalendarV1.GREGORIAN)
+            dateRange.start.get(Calendar.ERA) should be(GregorianCalendar.BC)
```
tobiasschweizer: could you add more extensive checks (day, month, year)?

webapi/src/test/scala/org/knora/webapi/util/InputValidationSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+        "accept JULIAN:2017-05-10:2017-05-12" in {
+            val dateString = "JULIAN:2017-05-10:2017-05-12"
+            InputValidation.toDate(dateString, () => throw new IllegalArgumentException(s"Not accepted ${dateString}"))
+        }
+
+        "accept JULIAN:2017:2017-05-12" in {
+            val dateString = "JULIAN:2017:2017-05-12"
+            InputValidation.toDate(dateString, () => throw new IllegalArgumentException(s"Not accepted ${dateString}"))
+        }
+    }
```
tobiasschweizer: Could you add `BCE` checks here?

webapi/src/main/scala/org/knora/webapi/util/InputValidation.scala
tobiasschweizer: Would it be clearer to put the blank separator after the constant?

webapi/_test_data/ontologies/beol-onto.ttl
```diff
@@ -483,7 +536,134 @@
...
 
+    ### ###########################################
+    ### beol:hasFloruitDate
+
+    :hasFloruitDate rdf:type owl:ObjectProperty ;
+
+                  rdfs:label "Blühtedataum"@de ,
+                             "Floruit Date"@en ;
+
+                  rdfs:comment """Repräsentiert ein Blühtedataum."""@de,
```
SepidehAlassi: no according to martin, Floruit date is a date somebody was seen, or had a publication. in bibliography items this is indicated as (fl. 2.\,Hälfte 20.\,Jh.) This is used mostly for cases that no birth or death date of someone is available.

tobiasschweizer:
Could you please add the era handling to the docs and the typescript interfaces in `salsah/src/typescript_interfaces`?

webapi/_test_data/ontologies/beol-onto.ttl
```diff
@@ -425,6 +442,42 @@
                    rdfs:subPropertyOf knora-base:hasValue .
 
     ### ###########################################
+    ### beol:hasTitle
+
+    :hasTitle rdf:type owl:ObjectProperty ;
```
SepidehAlassi: well, your beol:title is for title of a letter/manuscript, beol:hasTitle is for title of a person. 

I can change the beol:hasTite to beol:personHasTitle and change beol:title for letter and manuscript to beol:hasTitle. How does that sound?

webapi/src/main/scala/org/knora/webapi/util/InputValidation.scala
SepidehAlassi: The convention for writing date with era is date+whitespace+BC/AD/BCE/CE I would say  let's leave the whitespace before the era

tobiasschweizer:
Could add something to regex to prevent people from adding a date with the year 0?

I think we would need to adapt the old SALSAH GUI to BC dates, @rosenth might be able to help

webapi/_test_data/ontologies/beol-onto.ttl
tobiasschweizer: The change to `beol:personHasTitle` would be fine.

I would change `beol:title` myself because I have to coordinate it with my import scripts.

webapi/_test_data/ontologies/beol-onto.ttl
tobiasschweizer: Ok, could you then just choose another label (especially in German) or explain it a little bit in the `rdfs:comment`?

In German, this sounds like the person is a flower :-)

webapi/src/main/scala/org/knora/webapi/util/InputValidation.scala
tobiasschweizer: That's fine for me!

tobiasschweizer:
@SepidehAlassi I will be away next week. Is it ok for you when I do the final review once I am back?

webapi/src/test/scala/org/knora/webapi/util/DateUtilSpec.scala
tobiasschweizer: Could you add a test involving a BC date and explicitly checking for the Julian Day Numbers?

webapi/src/main/scala/org/knora/webapi/util/InputValidation.scala
SepidehAlassi: the same as comment above

webapi/src/main/scala/org/knora/webapi/util/DateUtilV1.scala
SepidehAlassi: done!

webapi/src/main/scala/org/knora/webapi/util/DateUtilV1.scala
SepidehAlassi: done!

webapi/_test_data/ontologies/beol-onto.ttl
SepidehAlassi: changed the beol:hasTitle to beol:personHasTitle

webapi/_test_data/ontologies/beol-onto.ttl
SepidehAlassi: changed to Floruit-Ort and Floruit-datum

webapi/src/test/scala/org/knora/webapi/util/DateUtilSpec.scala
SepidehAlassi: added to DateUtilSpec

webapi/src/test/scala/org/knora/webapi/util/DateUtilSpec.scala
```diff
@@ -52,6 +54,54 @@ class DateUtilSpec extends WordSpec with Matchers {
...
+            dateRange.start.get(Calendar.ERA) should be(GregorianCalendar.AD)
+        }
+        "convert an era date string with julian calendar with year precision to a Java GregorianCalendar BC" in {
+            val dateRange: DateRange = DateUtilV1.dateString2DateRange("50-02-28 AD", KnoraCalendarV1.JULIAN)
+            dateRange.start.get(Calendar.ERA) should be(GregorianCalendar.AD)
+        }
+
+        "convert a date in YYYY-MM-DD Era format, in the Julian calendar, into a Julian day count" in {
+            val bundesbriefDateValueV1 = DateValueV1(
+                dateval1 = "4713-01-01 BC",
```
benjamingeer: Swiss people might not be happy if you say that the Bundesbrief was in 4713 BC. :) https://de.wikipedia.org/wiki/Bundesbrief_von_1291

tobiasschweizer:
> The era still needs to be included when we convert a DateValueV1 to JSON in an API response.

Thx. for the input! 

`salsah/src/public/js/jquery.dateobj.js` would have to be adapted and the ts interfaces in `salsah/src/typescript_interfaces`.

benjamingeer:
I'm getting out-of-memory errors when running the tests with Fuskeki:

```
java.lang.OutOfMemoryError: Direct buffer memory
	at java.nio.Bits.reserveMemory(Bits.java:658)
	at java.nio.DirectByteBuffer.<init>(DirectByteBuffer.java:123)
	at java.nio.ByteBuffer.allocateDirect(ByteBuffer.java:311)
	at org.eclipse.jetty.util.BufferUtil.allocateDirect(BufferUtil.java:128)
	at org.eclipse.jetty.io.ArrayByteBufferPool.acquire(ArrayByteBufferPool.java:69)
	at org.eclipse.jetty.server.HttpOutput.write(HttpOutput.java:396)
	at sun.nio.cs.StreamEncoder.writeBytes(StreamEncoder.java:221)
	at sun.nio.cs.StreamEncoder.implFlushBuffer(StreamEncoder.java:291)
	at sun.nio.cs.StreamEncoder.implFlush(StreamEncoder.java:295)
	at sun.nio.cs.StreamEncoder.flush(StreamEncoder.java:141)
	at java.io.OutputStreamWriter.flush(OutputStreamWriter.java:229)
	at org.apache.jena.atlas.io.IO.flush(IO.java:233)
```

This happens on my machine but not on Travis. I suspect it's because of this issue:

https://issues.apache.org/jira/browse/JENA-181

I assume that it works on Travis because Travis is slower.

benjamingeer:
I tried adding a delay after each HTTP request to Fuseki, and reducing Akka's concurrency, but this didn't help. I also don't understand why this doesn't happen on the `develop` branch. I thought it might be because we switched from "update via POST with URL-encoded parameters" to "update using POST directly", but I tried switching back, and that didn't help, either.

Further options:

* Try a newer version of Fuseki
* Try an alternative to Fuseki:
    * [RDF4J](http://rdf4j.org/)
    * [Virtuoso](http://vos.openlinksw.com/owiki/wiki/VOS)

benjamingeer:
Tried upgrading to Jena 3.4.0, got the same error.

benjamingeer:
I've been trying RDF4J Server, using [this Docker image](https://hub.docker.com/r/yyz1989/rdf4j/). Problems:

* With the disk-based storage (`openrdf:NativeStore`), loading the Knora test data was so slow that I gave up waiting for it.
* With the memory-based storage (`openrdf:MemoryStore`), loading the test data is fast, but some queries in our tests are very slow, many others return unexpected results, and many updates don't work properly.
* There's no documentation on how to configure Lucene. I followed the example [here](https://groups.google.com/forum/#!topic/rdf4j-users/epF4Af1jXGU), and now creating a resource times out after a minute.

I'm going to try Virtuoso. They provide a [Docker image](http://vos.openlinksw.com/owiki/wiki/VOS/VirtuosoDocker).

subotic:
> I'm going to try Virtuoso. They provide a Docker image.

If everything else becomes too complicated, maybe GraphDB Free can become an option? They don't have a docker image for the free edition, but provide [instruction](https://github.com/Ontotext-AD/graphdb-docker) on how to create it.

benjamingeer:
The OOM error occurs on my iMac and on Tobias's MacBook Air, but not on his iMac or on Sepideh's iMac.

loicjaouen:
there was a bug on some intel cpu related to hyperthreading: http://gallium.inria.fr/blog/intel-skylake-bug/ but for the vast majority of cases, the cpu is out of the culprit list.

benjamingeer:
Here's the plan:

1. @subotic is going to make a Docker image for GraphDB Free.
2. In this PR, we'll change the default triplestore to GraphDB Free, including for tests on Travis.
3. We'll merge this PR.
4. I'll add support for some other triplestore in a subsequent PR.

benjamingeer:
@subotic it just occurred to me that this probably won't work on Travis yet, because after we start the GraphDB Free docker image, I think we don't create any repositories for it to use.

subotic:
Yes, you are right. We need to initialize the repository each time we create a new container.

webapi/src/it/resources/graphdb-free.conf
```diff
@@ -4,9 +4,11 @@ app {
     triplestore {
         dbtype = "graphdb-free"
 
-        graphdb {
+        graphdb-free {
```
subotic: here was the error. now `graphdb-free-it` should run with `knora-test-unit`.

tobiasschweizer:
Please tell the people from the NIE project what you are about to to. I think they use Fuseki.

benjamingeer:
> Please tell the people from the NIE project what you are about to to. I think they use Fuseki.

Fuseki is still there, they can still use it if it works for them.

tobiasschweizer:
Ok, but they could have the same issue you had. And we do not know what causes the problem.

benjamingeer:
@subotic Now the tests are failing on Travis because Akka is unable to connect to the GraphDB Free Docker image ("connection refused", "connection reset by peer"). Do you know any way we could diagnose this?

benjamingeer:
@tobiasschweizer I've sent them an email to explain.

tobiasschweizer:
That's great, thank you!

subotic:
@benjamingeer graphdb inside docker needs too much memory and crashes. I will try to run it directly inside the Travis VM and see if I can get it to work this way.

subotic:
Unfortunately, running `graphdb-free` locally doesn't solve the problem. The free version uses to much memory with our data. We only have 7.5 GB. The only thing left to try is using `graphdb-se`. For this to work, we would need to add the encrypted license to the repository, the same way we have done it with the `kakadu` library in the `sipi` repository.

I suppose, that you cannot wait after I'm back from vacation, so I will do it on Monday or so. The next two days we are in the car heading south and no real internet.

subotic:
Alternatively and much easier would be to simply skip the problematic tests when running on Travis. Then when I'm back, we could look for a solution.

benjamingeer:
> Alternatively and much easier would be to simply skip the problematic tests when running on Travis. 

That would be fine with me, but how can a test know that it's running on Travis?

loicjaouen:
Couldn't we host ourselves a graphdb server, dedicated to tests, that the travis jobs would target?

subotic:
we could check if a travis environment variable is set.

benjamingeer:
> we could check if a travis environment variable is set.

Is there one set already that I could use?

subotic:
yes, for example `TRAVIS_BRANCH`

benjamingeer:
I tried disabling what seemed to be the most memory-intensive tests on Travis, but even the search tests run out of memory on Travis.

```
[info] SearchV1R2RSpec:
[2017-09-01 14:30:23,332] INFO  - HttpTriplestoreConnector - Connection OK: 2359
/home/travis/.travis/job_stages: line 57:  5567 Killed                  ./bin/graphdb > /dev/null 2>&1
[2017-09-01 14:30:41,331] ERROR - Dispatcher - Failed to execute insert into triplestore
```

If we're not going to run the search tests on Travis, we might as well disable all tests on Travis.

benjamingeer:
Hmm, actually that test loads some data it doesn't use, let me try again.

benjamingeer:
Still doesn't work:

```
[info] AuthenticationV1R2RSpec:
[2017-09-01 14:50:56,364] INFO  - HttpTriplestoreConnector - Connection OK: 2359
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

subotic:
The triplestore crashes when the incunabula data gets loaded. Probably any test which loads it would need to be skipped.

at least the integration tests run through, as those are a pain to run locally.

PR #574, opened 2017-07-18
chain of trust

loicjaouen:
"a chain is only as strong as its weakest link" and this `http` link will stop you from being called `https` (in modern browsers)


benjamingeer:
Thanks!

Issue #575, opened 2017-07-19
Fulltext Search and Extended Search documentations are not up to date


SepidehAlassi:
The parameters filter_by_project , filter_by_restype, and filter_by_owner of documentation are renamed in code to filterByProject, filterByRestype and filterByOwner. 
It causes confusion for readers of documentation.

tobiasschweizer:
Your are surely right pointing out the inconsistency in the naming of these parameters and the case classes that are used within Knora. We did this in order to be compliant with the old SALSAH Api (the GET URL parameters used by the old GUI). 

The documentation refers to the GET URL parameters that are processed in `SearchRouteV1.scala`:

```
        params.get("searchtype") match {
            case Some(List("extended")) => ()
            case other => throw BadRequestException(s"Unexpected searchtype param for extended search")
        }

        // only one value is expected
        val restypeIri: Option[IRI] = params.get("filter_by_restype") match {
            case Some(List(restype: IRI)) => Some(InputValidation.toIri(restype, () => throw BadRequestException(s"Unexpected param 'filter_by_restype' for extended search: $restype")))
            case other => None
        }

        // only one value is expected
        val projectIri: Option[IRI] = params.get("filter_by_project") match {
            case Some(List(project: IRI)) => Some(InputValidation.toIri(project, () => throw BadRequestException(s"Unexpected param 'filter_by_project' for extended search: $project")))
            case other => None
        }

        // only one value is expected
        val ownerIri: Option[IRI] = params.get("filter_by_owner") match {
            case Some(List(owner: IRI)) => Some(InputValidation.toIri(owner, () => throw BadRequestException(s"Unexpected param 'filter_by_owner' for extended search: $owner")))
            case other => None
        }

        // here, also multiple values can be given
        val propertyIri: Seq[IRI] = params.get("property_id") match {
            case Some(propertyList: Seq[IRI]) => propertyList.map(
                prop => InputValidation.toIri(prop, () => throw BadRequestException(s"Unexpected param 'property_id' for extended search: $prop"))
            )
            case other => Nil
        }
```

I think the user does not have to know how they are called internally. But for V2 we should stick to a consistent naming scheme whenever using GET parameters (although I would like to avoid using them as far as possible).

tobiasschweizer:
We had a look at the problem and found out that the problem was that the parameters' values (Iris) were URL encoded twice. 

I will add a message to the `BadRequestException` in case of an invalid Iri that gives the client a hint that this is a common source of problem.

PR #576, opened 2017-07-19
fix (extended search date LT_EQ GT_EQ): fix wrong conditions for date comparison

tobiasschweizer:
The conditions for LT_EQ and GT_EQ were wrong. Logically, they are simply the negations of GT and LT, respectively.

closes  https://github.com/dhlab-basel/Knora/issues/575

### DIFF
--- a/docs/rst/knora-api-server/api_v1/reading-and-searching-resources.rst
+++ b/docs/rst/knora-api-server/api_v1/reading-and-searching-resources.rst
@@ -252,9 +252,9 @@ Explanation of the comparison operators:
     In case of a date value type, inequality is given if the dates do not overlap in any way, meaning that a date starts after the end of the defined period or ends before the beginning of the defined period
     (dates are internally always treated as periods, see above).
   - ``GT``: checks if a resource's value is *greater than* the search value. In case of a date value type, it assures that a period begins after the indicated period's end.
-  - ``GT_EQ``: checks if a resource's value *equals or is greater than* the search value. In case of a date value type, it assures that a period's start equals the end of the indicated period or the period begins after the indicated period's end.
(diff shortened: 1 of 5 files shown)


tobiasschweizer:
should fix https://github.com/dhlab-basel/Knora/issues/575

aeneaskaiser:
I can verify that the changes to `searchExtendedGraphDB.scala.txt `and `searchExtendedStandard.scala.txt` produce logically consistent search results, so that the sets of results for LT and GT_EQ form a partition of the whole set of date values, and also the sets of results for GT and LT_EQ form a partition of the whole set of date values.
This seems to be a desirable property of the semantics of the comparison operators on date values to me. Especially since the previous semantics would have left some date values unfindable by combining searches for e.g. LT and GT_EQ.

However now the documentation for LT_EQ and GT_EQ in `docs/rst/knora-api-server/api_v1/reading-and-searching-resources.rst` is out of date. It still describes the behaviour prior to this PR. Maybe this should be updated in the same PR?

aeneaskaiser:
As for the changes introduced in `SearchRouteV1.scala` i can verify that no new bugs were introduced, since only error messages were changed. But to judge if the semantics of the error messages make sense, and to judge if this change closes/answers issue #575, i feel i don’t know the Knora code base well enough. Care to ask someone else for a review? Or maybe we could look at the context of the changes together?

tobiasschweizer:
> However now the documentation for LT_EQ and GT_EQ in docs/rst/knora-api-server/api_v1/reading-and-searching-resources.rst is out of date. It still describes the behaviour prior to this PR. Maybe this should be updated in the same PR?

Absolutely! I will do it right away!

> But to judge if the semantics of the error messages make sense, and to judge if this change closes/answers issue #575, i feel i don’t know the Knora code base well enough.

The error messages just indicate that a URL GET parameter's value is not a valid Iri (although it should be).  I added a comment about correct URL encoding because we had a case where a library URL encoded an already URL encoded string (so then it was URL encoded twice and hence the Iri validation failed).

PR #589, opened 2017-08-03
refactor (beol-onto): add missing labels and comments

tobiasschweizer:
added some missing labels and comments in beol-onto


tobiasschweizer:
there is one little commit more, there were some connectivity issues before ...

PR #608, opened 2017-09-18
build (travis): add using graphdb-se

subotic:
### Summary

- add encrypted `graphdb-se` developer single core license
- use Ontotext's official docker image to run `graphdb-se` 
- use the new way of specifying conditional stages, making the publish stage easier to read.


subotic:
@benjamingeer All test are running through with `graphdb-se` on travis.

Issue #612, opened 2017-09-25
Documentation of Graphview


tobiasschweizer:
In the docs about how to use the graphview (<http://www.knora.org/documentation/manual/rst/knora-api-server/api_v1/reading-and-searching-resources.html#get-a-graph-of-resources>), the path contains the segment `search` which is wrong. 

@janCstoffregen Thanks for pointing this out. About your problem with the depth parameter: Could you try this query with the `anything` test data: <http://localhost:3333/v1/graphdata/http%3A%2F%2Fdata.knora.org%2F0C-0L1kORryKzJAJxxRyRQ?depth=4>

tobiasschweizer:
Are you sure that there are enough link-levels in the data so it makes actually a difference?

janCstoffregen:
Sorry, now it's working, maybe there are more links now than yesterday, we updated the data again today, thank you very very much for your help. I will come back if I can reproduce the error.

tobiasschweizer:
No problem! Just as a hint for the next time you encounter such a problem: could you try the test data first to rule out a problem in the api-code (the easiest thing is to look at the automated tests, they give you an idea what you could try out)?

Then we would already have an idea that it could be about the data.

janCstoffregen:
Yes, I will refer to the test data next time, I understand that it's much more usefull and reproducable that way, thank you for the hint!

tobiasschweizer:
Great! The issue will be closed when the PR is merged. Thx.

janCstoffregen:
Ok, thank you!

PR #614, opened 2017-09-25
fix (remove wrong 'search' segment from docs for graphdata)

tobiasschweizer:
closes https://github.com/dhlab-basel/Knora/issues/612


tobiasschweizer:
@benjamingeer Thanks for the approval!

Issue #22, opened 2016-02-03
Handle valueHasComment in Search


tobiasschweizer:
Additionally to `valueHasString` and `rdfs:label`, `valueHasComment` was added to the full text index for the triplestore(s). 

Do we have to adapt the fulltext search templates? 

-> jena/fuseki:

``` sparql
@if(triplestore == "embedded-jena-tdb" || triplestore == "fuseki") {
            ?matchingSubject knora-base:valueHasString ?literal .
            #?matchingSubject ?p ?literal .
            #FILTER(?p = knora-base:valueHasString || ?matchingProperty = rdfs:label)
        }
```

the literal could also be a comment!

Issue #10, opened 2016-01-28
Improve use of Lucene query syntax in extended search


benjamingeer:
In particular, use the search index whenever possible rather than the less efficient regex feature.

benjamingeer:
Is this issue still valid, or can it be closed?

tobiasschweizer:
This must have been created for v1. I think it is not relevant.

PR #620, opened 2017-09-29
KnarQL Route and Documentation

### LINKED ISSUE #530: Implement a DSL for API v2 extended search
Since API v2 extended search needs to be implemented ASAP as part of #478, I'm going to try the approach I outlined [here](https://github.com/dhlab-basel/Knora/issues/227#issuecomment-242228298), i.e. implementing a simple DSL that the GUI can submit to the API, and that will be straightforward to translate to SPARQL. Tobias and I looked for something standard we could use, but didn't find anything that seemed suitable.

It should parse things like the following examples.

Find all the pages with sequence numbers less than or equal to 100, in books whose name contains either "Zeitglöcklein" and "Lebens" or "Itinerarius" and "peregrinarius":

```
B a incunabula:book .
B incunabula:pubdate ge(date("GREGORIAN:1490")) .
B incunabula:pubdate lt(date("GREGORIAN:1500")) .
P a incunabula:page .
P incunabula:isPartOf B .
P incunabula:pagenum le(int(100)) .

{
    B incunabula:title contains("Zeitglöcklein") .
    B incunabula:title contains("Lebens") .
} UNION {
    B incunabula:title contains("Itinerarius") .
    B incunabula:title contains("peregrinarius") .
}
```

Find a letter that has, either as its author or as its recipient, someone whose IAF identifier is either 123 or 456:

```
L a beol:letter .

{
    L beol:hasAuthor P1 .
} UNION {
    L beol:hasRecipient P1 .
} UNION {
    L beol:hasAuthor P2 .
} UNION {
    L beol:hasRecipient P2 .
}

P1 beol:hasIAFIdentifier "123" .
P1 beol:hasIAFIdentifier "456" .
```

Perhaps I can adjust this syntax a bit and get a SPARQL parser to parse it. Otherwise, I'll look into using a parser generator such as [ANTLR](http://www.antlr.org/).

### LINKED ISSUE #598: Refactor extended search v2
Refactor the extended search v2 so it supports paging:

1. do a prequery possibly involving ORDER BY criteria that returns an ordered list of the Iris of the main resource. If no ORDER BY criteria are given, sort by the Iri of the main resource. The prequery is a SELECT query with a WHERE clause containing all the user provided restrictions. The SELECT query has an OFFSET and a LIMIT. The prequery does not get information about the value objects where this is not needed for making the restrictions.

2.  For a certain amount of Iris returned by the prequery, do a CONSTRUCT query (add Filters that restrict the variable representing the main resources to these Iris). Its WHERE clause contains all the user provided restrictions and gets information about the value object. Information about the value objects are obtained in a generic way: 
    ```
    ?mainRes knora-base:hasValue ?valueObject .
    ?valueObject a knora-base:Value .            
         ?valueObject knora-base:isDeleted false .

         ?valueObject a ?valueObjectType ;
         ?valueObjectProperty ?valueObjectValue .
    } 
    ``` 
    Check all resource and value object permissions in Scala: **The user must have permissions on all values in the WHERE clause, otherwise the main resource cannot be returned.**

    Look a the information in the user-provided CONSTRUCT clause and ignore the properties that are not mentioned. Since all the values of a resource are requested in a generic way, the CONSTRUCT query may return more information than the  user wants to see.

tobiasschweizer:
- refactor extended search (now called KnarQL): 
     - permission checking for the full query path: suppress whole resource in case of insufficient permissions for at least one resource or value asked for in the WHERE clause
     - if necessary, suppress resource in answer by showing a proxy resource instead 
     - specify property values in the CONSTRUCT clause that should be returned in the answer: not all values contained in the WHERE clause have to be returned
- add `ApacheLuceneSupport`: Utility that pre-processes a search string so that it supports Apache Lucene Parser Syntax. Different use cases: search as you type (search for `rdf:label`),  fulltext search (use Boolean AND)
- add documentation (RST and TypeScript)
- add tests

Attention: this PR may bore you.

TODO:
- add support for phrases: #641
- check that ORDER BY criterion has cardinality 1:  #642

closes #598 
closes #530

relates to #10
relates to #22


tobiasschweizer:
```
SELECT (count(distinct ?letter) as ?count)
 WHERE {
 ?letter <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     ?letter <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?letter <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/beol#letter> .
 ?letter <http://www.knora.org/ontology/beol#creationDate> ?date .
     ?date <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?letter ?linkingProp1 <http://rdfh.ch/beol/ZkJcQg9yTmyMY_J6nnubxA> .
 <http://rdfh.ch/beol/ZkJcQg9yTmyMY_J6nnubxA> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     <http://rdfh.ch/beol/ZkJcQg9yTmyMY_J6nnubxA> <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?letter ?linkingProp2 <http://rdfh.ch/beol/_yblAQMwT2un_xN7UaVWrg> .
 <http://rdfh.ch/beol/_yblAQMwT2un_xN7UaVWrg> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     <http://rdfh.ch/beol/_yblAQMwT2un_xN7UaVWrg> <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 FILTER(((?linkingProp1 = <http://www.knora.org/ontology/beol#hasAuthor>) || (?linkingProp1 = <http://www.knora.org/ontology/beol#hasRecipient>)))
 FILTER(((?linkingProp2 = <http://www.knora.org/ontology/beol#hasAuthor>) || (?linkingProp2 = <http://www.knora.org/ontology/beol#hasRecipient>)))
     ?date <http://www.knora.org/ontology/knora-base#valueHasStartJDN> ?date__valueHasStartJDN .
 }
 } 
 
```

tobiasschweizer:
If a link property has more than one instance and a variable is used as an object, the row is repeated with he current design. Consider the following example:

```
    PREFIX incunabula: <http://api.knora.org/ontology/incunabula/simple/v2#>
    
    CONSTRUCT {
        ?linkObj knora-api:isMainResource true .
        
    } WHERE {
        ?linkObj a knora-api:Resource .
        ?linkObj a knora-api:LinkObj .
        
        ?linkObj knora-api:hasLinkTo ?book .
        knora-api:hasLinkTo knora-api:objectType knora-api:Resource .
        
        ?book a knora-api:Resource .
        ?book a incunabula:book . 
        
    }
``` 

In Knora, there are three link objects that point to incunabula:books. However, the answer returned by Knora looks like this:

```
{
  "@type" : "schema:ItemList",
  "schema:itemListElement" : [ {
    "@id" : "http://data.knora.org/881405205304",
    "@type" : "knora-api:LinkObj",
    "schema:name" : "Lateinische Übersetzung"
  }, {
    "@id" : "http://data.knora.org/881405205304",
    "@type" : "knora-api:LinkObj",
    "schema:name" : "Lateinische Übersetzung"
  }, {
    "@id" : "http://data.knora.org/ab79ffa43935",
    "@type" : "knora-api:LinkObj",
    "schema:name" : "DEMO_"
  }, {
    "@id" : "http://data.knora.org/cb1a74e3e2f6",
    "@type" : "knora-api:LinkObj",
    "schema:name" : "Diese drei Texte sind in einem Band zusammengebunden."
  }, {
    "@id" : "http://data.knora.org/cb1a74e3e2f6",
    "@type" : "knora-api:LinkObj",
    "schema:name" : "Diese drei Texte sind in einem Band zusammengebunden."
  }, {
    "@id" : "http://data.knora.org/cb1a74e3e2f6",
    "@type" : "knora-api:LinkObj",
    "schema:name" : "Diese drei Texte sind in einem Band zusammengebunden."
  } ],
  "schema:numberOfItems" : 3,
  "@context" : {
    "schema" : "http://schema.org/",
    "knora-api" : "http://api.knora.org/ontology/knora-api/v2#",
    "rdf" : "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
    "rdfs" : "http://www.w3.org/2000/01/rdf-schema#"
  }
}
```

tobiasschweizer:
An incoming query for `StillImageRepresentations` would look like this:

```
 SELECT DISTINCT ?page
 WHERE {
 ?page <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     ?page <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?page <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#StillImageRepresentation> .
 ?page <http://www.knora.org/ontology/knora-base#isPartOf> <http://data.knora.org/8be1b7cf7103> .
 <http://data.knora.org/8be1b7cf7103> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     <http://data.knora.org/8be1b7cf7103> <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?page <http://www.knora.org/ontology/knora-base#seqnum> ?seqnum .
     ?seqnum <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?page <http://www.knora.org/ontology/knora-base#hasStillImageFileValue> ?file .
     ?file <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
     ?seqnum <http://www.knora.org/ontology/knora-base#valueHasInteger> ?seqnum__valueHasInteger .
 }
 }
 
 GROUP BY ?page ?seqnum__valueHasInteger
 ORDER BY ASC(?seqnum__valueHasInteger) ASC(?page)
 OFFSET 0
 LIMIT 25
```

benjamingeer:
Could we support HTTP POST for search queries, in addition to GET? SPARQL supports this:

https://www.w3.org/TR/sparql11-protocol/#select-longpost-direct

tobiasschweizer:
> Could we support HTTP POST for search queries, in addition to GET? SPARQL supports this:

Yes, absolutely. This would just be a change in the route.

docs/rst/knora-api-server/api_v2/reading-and-searching-resources.rst
```diff
@@ -0,0 +1,174 @@
...
+********************
+Search for Resources
+********************
+
+Search for a Resource by its ``rdfs:label``
+-------------------------------------------
+
+Knora offers the possibility to search for resources by their ``rdfs:label``. The use case for this search is to find a specific resource as you type.
+E.g., the user wants to get a list of resources whose ``rdfs:label`` contain some search terms separated by a whitespace character:
+  - Zeit
```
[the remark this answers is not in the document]
tobiasschweizer: fixed in https://github.com/dhlab-basel/Knora/pull/620/commits/690c66f707dbb80d4dfd38e3b841f4371973dc2b

benjamingeer:
We should also explain what a Knora property is.

mrivoal:
May I ask the reason for supporting schema.org more than any other ontology? Referencing through research engines?
Otherwise is it so widespread within the research data or humanities communities? Because to me, it looks like it is something huge (nearly 600 classes and 850 properties!!) but does it mean it is *the* most accurate ontology for us? Why schema.org over DBpedia ontology, for example? 

And are you not afraid that it will bring (unnecessary?) complexity to the API?

tobiasschweizer:
This is surely a legitimate question. My idea was just that it would be nice to re-use some existing semantics, so we would not have to make up our own response format for a resource request or a query. Of course, we can discuss about whether schema.org makes sense in our case or not.

We basically need something to represent a sequence of resources. Later, we are going to need something for values too (including the history of a value).

tobiasschweizer:
> We need to explain that rdfs:label is represented by schema:namein the API, and that we may, in the future, provide other options.

done in https://github.com/dhlab-basel/Knora/pull/620/commits/d6d8a964cd3e043a78109c7cfd9e7261599c5883

mrivoal:
> This is surely a legitimate question. My idea was just that it would be nice to re-use some existing semantics, so we would not have to make up our own response format for a resource request or a query. 

I see, now. And I can only agree.

> Of course, we can discuss about whether schema.org makes sense in our case or not.

What would bother me with schema.org (I don't know it well at all, I must confess) is that it is a very generic ontology and it covers a lot of things. 

I am not sure of what we would need exactly, but I guess I would be more confortable with ontologies dedicated to data and datasets description, such as [DCAT](https://www.w3.org/TR/vocab-dcat/) or [Void](https://www.w3.org/TR/void/). But I am not sure one of these ontologies would fit all our needs and I am not quite up-to-date on the subject, so I have no better suggestion to make. Besides I am sure you have already delved further into schema.org.

benjamingeer:
`ForbiddenResource` needs to be added to the API ontologies.

benjamingeer:
@mrivoal I'm not sure that `schema.org` is the best option either, but in any case we're only using a little bit of it. The idea is that this could facilitate making Knora query results show up in search engines, which is something I could imagine a project wanting to do. For now we are just using two things from `schema.org`. Let's see how it goes and revisit this issue later.

webapi/src/main/scala/org/knora/webapi/responders/ResponderV2.scala
```diff
@@ -0,0 +1,90 @@
...
+import org.knora.webapi.messages.v1.responder.usermessages.UserProfileV1
+import org.knora.webapi.util.ConstructResponseUtilV2
+import org.knora.webapi.util.ConstructResponseUtilV2.{MappingAndXSLTransformation, ResourceWithValueRdfData}
+
+import scala.concurrent.Future
+
+/**
+  * An abstract class with utility methods for v2 responders.
+  */
+abstract class ResponderV2 extends Responder {
```
benjamingeer: Please rename `ResponderV2` to something more specific, like `ResourceTextResponder`, and have only the responders that really need `getMappingsFromQueryResultsSeparated` subclass it.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -1159,15 +1939,12 @@ class SearchResponderV2 extends Responder {
                 transformer = triplestoreSpecificQueryPatternTransformerSelect
             )
 
-            // _ = println(triplestoreSpecificPrequery.toSparql)
+            _ = println(triplestoreSpecificPrequery.toSparql)
```
benjamingeer: Please comment this out, I forgot to do so.

subotic:
@mrivoal `schema.org` (only a small fraction) is used as a description of the `container` holding the API V2 responses. The actual response content is using another ontology.

We could easily create our own ontology, by only taking the parts we need from `schema.org` or any other ontology for that matter.

> The idea is that this could facilitate making Knora query results show up in search engines, which is something I could imagine a project wanting to do.

The question is if Google and co are going to actually use this data since according to the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool) our responses are not valid:

screenshot 2017-10-24 17 47 05

mrivoal:
Thanks @benjamingeer and @subotic for the explanations.

I totally understand the point of reusing parts of an ontology that would fit our needs instead of creating a new one. 

>>The idea is that this could facilitate making Knora query results show up in search engines, which is something I could imagine a project wanting to do.

Sure!

>The question is if Google and co are going to actually use this data since according to the Google Structured Data Testing Tool our responses are not valid:

As for this point, I pass :)

webapi/src/main/scala/org/knora/webapi/responders/ResponderV2.scala
tobiasschweizer: fixed in https://github.com/dhlab-basel/Knora/pull/620/commits/0c13561911483771d29e52f822d66d117c03a4b4

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
tobiasschweizer: also fixed in https://github.com/dhlab-basel/Knora/pull/620/commits/0c13561911483771d29e52f822d66d117c03a4b4

musicEnfanthen:
Regarding `schema.org`: From my experiences with other DH projects, it could become the next "big thing" in terms of standardisation (even in musicology projects it is discussed already, have a look at the [IncipitSearch project](https://incipitsearch.adwmainz.net/participation/) of the Akademie der Wissenschaften Mainz). I agree with @mrivoal that it is very flat and not very specific. Another point to be aware of is surely, that it is Google, Amazon and the like, who are the drivers behind `schema.org`. But on the other side this is also its strength because of a wide dissemination and reuse of the same ontology structures. So, being conform (or at least being able to map) to schema.org wouldn't be the worst thing one can do.

@subotic : The response in Google Structured Data Testing Tool fails only in those parts where the response does not follow the schema.org specification. As far as I understand, the failing properties will be ignored in search algorithms. If you want to get rid of the red crosses, this could be a possible way (but I don't know if this would still be conform with Knora?):

screenshot-2017-10-25 test-tool fur strukturierte daten

There is also a [`DataCatalog`](http://schema.org/DataCatalog) and [`DataSet`](http://schema.org/Dataset) in the schema specification. The acknowledgement section says: 

> This class is based upon W3C DCAT work, and benefits from collaboration around the DCAT, ADMS and VoID vocabularies. See http://www.w3.org/wiki/WebSchemas/Datasets for full details and mappings.

Maybe one could use `schema:DataCatalog` or `schema:DataSet` as generic _entry point_ for the metadata description which can then be filled up in depth with the structures from `DCAT` or `VoID`.

benjamingeer:
@musicEnfanthen Thank you very much for these suggestions, we will have a look.

subotic:
@musicEnfanthen Yes, I'm afraid that for Google and Co. to actually use the data, it would need to completely conform to `schema.org`. This can probably be done, but not for the regular API we use Salsah to communicate with. We could have our routes return our data completely in `schema.org` or our own schemas, depending on the client. We already have `ApiV2WithValueObjects` and `ApiV2Simple` as response schemas. Now we could add a third one `ApiV2SchemaOrg` (or something like this).

subotic:
There is currently an issue with Travis and pulling of images from Docker Hub: "We’re investigating reports of timeouts in builds while pulling images from Docker Hub."

That's why the Travis is failing at the moment.

tobiasschweizer:
@subotic I would like to merge this PR. How can I do this (it is blocked).

tobiasschweizer:
@subotic Thanks!

PR #624, opened 2017-10-04
test (webapi): add resource creation by different users

### LINKED ISSUE #610: Shouldn't a knora-base:SystemAdmin be allowed to create resources in every project ?
Just wondering... According to Knora documentation, 

> The knora-base:SystemAdmin group receives implicitly the following permissions:
> 
> - receives implicitly ProjectAllAdminPermission for all projects.
> - receives implicitly ProjectResourceCreateAllPermission for all projects.
> - receives implicitly CR on all objects from all projects.
> 
> Theses permissions are baked into the system, and cannot be changed.

By reading 
> receives implicitly ProjectResourceCreateAllPermission for all projects.

I thought that there was no need to indicate for a `knora-base:SystemAdmin` that he was in a project or member of a group within a project. But, as **I got the following error in attempting to create a new resource as knora-base:SystemAdmin**: 
```
error
Bad request
The request content was malformed:
Object is missing required member 'project_id' 
```
I guess I should precise something. But what is it? The projects IRIs? The projects IRIs and their group IRIs respectively? or should I use the property `knora-base:isInProjectAdminGroup` and have it point to all the project IRIs ? I tried the last option and I still had the same error message as above.

Isn't the whole point of being a `knora-base:SystemAdmin` not being forced to define every project and/or group IRIs for test purpose? However, what should I do to be able to access several projects as a member of their `knora-base:admin group` (I think it should be possible, shouldn't it)?

subotic:
### Summary

- [x] Add test creating a resource with different users.
- [x] Move drawings-gods files from PR #625 to this branch
- [x] Add `CR knora-base:Creator` to `defaultObjectAccessPermissionsStringForEntityGetV1` as the fallback permissions if none could be found.
- [x] Update documentation describing the new behavior.

### Issues

- resolves #610


subotic:
@mrivoal @loicjaouen I have merged #625 into this branch. I had to change the ontology for `http://www.knora.org/ontology/0105#Person` where I have set the minimum cardinality to 0 for `hasDrawing` and `hasDrawingValue`. Maybe it would be good to have one of each in the example data.

Also, I have done some renaming to the IRIs so that they now conform to the new spec.

There is still a problem with `root` not able to create a resource. It works in other tests but somewhere there is a problem with the `drawings-gods` tests data. I'm investigating.

mrivoal:
> Maybe it would be good to have one of each in the example data.
Yes, we will provide it as soon as possible (hopefully next week).

subotic:
I have found the problem. Here the relevant documentation excerpt:

> **Default Object Access Permissions**: When a user creates a resource or value, then **only** the default object permissions from the **highest level** are applied. If a user is a member of more than one group on the same level (only possible for custom groups) then the defined permissions are summed up and the most permissive are applied.
>
>In the case of users belonging to the **SystemAdmin** group, but which are not members of a project and thus no group belonging to the project, the *default object access permissions* from the **highest defined level** will apply.

In the case of the `drawings-gods` project, only permissions on custom groups are defined. 
The question is what to do? I could implement it, so that all the permissions on all custom groups are found, sum them up, and apply the most permissive of them. I don't know if you would be happy with such a behavior because the result would be a bit difficult to comprehend (you have a lot of different groups). This would be in line with the behavior of the same user being in two or more custom groups.

Alternatively, I could implement something, so that you could define a default object access permission for a `SystemAdmin` who **is not part** of your project. Though this wouldn't solve the question, what default object access permissions should a `SystemAdmin` have when he **is part** of the project?

The easiest and fastest fix would be, if you would define some basic default object access permissions on the `ProjectAdmin` group, then those would be also used for the `SystemAdmin`.

mrivoal:
Thanks for the investigation!

> In the case of the drawings-gods project, only permissions on custom groups are defined.
The question is what to do? I could implement it, so that all the permissions on all custom groups are found, sum them up, and apply the most permissive of them. I don't know if you would be happy with such a behavior because the result would be a bit difficult to comprehend (you have a lot of different groups). This would be in line with the behavior of the same user being in two or more custom groups.

I am not sure I am able to see the whole picture right now concerning this option.

>The easiest and fastest fix would be, if you would define some basic default object access permissions on the ProjectAdmin group, then those would be also used for the SystemAdmin.

I think I like this idea. If I understand well, that would imply that even for projects where only custom groups are used, we could add `Default Object Access Permission` on the `ProjectAdmin` group so that even if **not part** of the project, `SystemAdmin` could create, modify resources within the given project (according to the level of `Default Object Access Permission`). 
To me, this point is important for testing purpose. As `SystemAdmin`, you want to be able to sometimes overwrite the permissions set for `ProjectMember` or even the `ProjectAdmin` group. 
But if the project is sensitive and if they don't want `SystemAdmin` to access their data, we could also restrict the access to `SystemAdmin` via the DOAP for `ProjectAdmin`, as long as the project uses a custom admin group. 
I may not be very clear... But I think your idea is fine!

subotic:
I think I'm going to implement it a bit differently:

> In the case of users belonging to the `SystemAdmin` group, but which are not members of a project and thus not a member of any group belonging to the project, the default object access permissions from the `ProjectAdmin` or `ProjectMember` group will apply. If no permissions are defined, then the resulting permission will be `CR knora-base:Creator`.

We need to keep in mind, that the default object access permissions are just the first half of the solution. There will be a GUI to change the permissions at creation time or afterward.

Also, please keep in mind, that the `SystemAdmin` and `ProjectAdmin` (with `ProjectAdminAllPermission`) will always be able to read and change any resource. There is no way to restrict access for the `SystemAdmin`. This would only give a false sense of security, as the `SystemAdmin` will have access to the Triplestore, and could simply look at the data this way.


At some point in time (probably next year), I would like to revisit the whole permission implementation. When we have a few projects using the current system, then we will have a much better sense of what is missing.

mrivoal:
> I think I'm going to implement it a bit differently:

>>    In the case of users belonging to the `SystemAdmin` group, but which are not members of a project and thus not a member of any group belonging to the project, the default object access permissions from the `ProjectAdmin` or `ProjectMember` group will apply. If no permissions are defined, then the resulting permission will be CR `knora-base:Creator`.

Ok. So, I wouldn't have to create a default object access permissions on `ProjectAdmin` group only for `SystemAdmin`commodity.

> We need to keep in mind, that the default object access permissions are just the first half of the solution. There will be a GUI to change the permissions at creation time or afterward.

> Also, please keep in mind, that the `SystemAdmin` and `ProjectAdmin` (with `ProjectAdminAllPermission`) will always be able to read and change any resource. There is no way to restrict access for the `SystemAdmin`. This would only give a false sense of security, as the `SystemAdmin` will have access to the Triplestore, and could simply look at the data this way.

You are right.

> At some point in time (probably next year), I would like to revisit the whole permission implementation. When we have a few projects using the current system, then we will have a much better sense of what is missing.

Aouch. Difficult news to read on a monday morning. I can see myself having nightmares already by the very idea of it.

subotic:
> Ok. So, I wouldn't have to create a default object access permissions on ProjectAdmin group only for SystemAdmincommodity.

Correct.

> Aouch. Difficult news to read on a monday morning. I can see myself having nightmares already by the very idea of it.

LOL. Hopefully not. The idea is to make life simpler :-)

subotic:
@mrivoal @loicjaouen The tests are now passing.

mrivoal:
Great news! Thanks. We will have a look at it.

docs/rst/knora-api-server/design-documentation/administration.rst
```diff
@@ -302,6 +302,14 @@ permissions from the **highest level** are applied. If a user is a member of mor
 In the case of users belonging to the **SystemAdmin** group, but which are not members of a project and thus no group
 belonging to the project, the *default object access permissions* from the **highest defined level** will apply.
 
+In the case of users belonging to the **SystemAdmin group**, but which are not members of a project and thus not a
```
loicjaouen: can I suggest?
- and thus not a member
+ and thus not members

loicjaouen:
@subotic, thanks for the work, the tests and our usage are showing the expected behaviour. I am not best qualified to review the scala though, do expect that from me?

subotic:
@loicjaouen Thanks for the feedback. There is not much new scala code (maybe 10 lines) to review. The rest are all tests that changed. No problem. I will ask @benjamingeer to review it, as soon as I have removed some logging and added your suggestion.

PR #625, opened 2017-10-05
Updating drawings-gods files (issue #610)

mrivoal:
An update of drawings-gods files, as requested for issue #610.


subotic:
will move this into PR #624

subotic:
data moved to #624

Issue #648, opened 2017-10-25
Serve salsah-gui and standoff ontologies


benjamingeer:
```
{"status":9,"error":"org.knora.webapi.NotFoundException: Named graph not found: http://www.knora.org/ontology/standoff"}
```

The ontology responder doesn't have these ontologies, so we may need to encode them in Scala.

benjamingeer:
Actually it does have the standoff ontology, but in a completely separate data structure.

benjamingeer:
@tobiasschweizer We have to serve the standoff classes that are defined in `knora-base`, as part of serving the `knora-api` ontologies. But the problem is that, because these classes are derived from the `ValueBase` classes, they have predicates like `valueHasStartJDN`. The same thing could actually happen with any custom standoff tag, because you could always make a custom subclass of `StandoffDateTag`.

Maybe the solution is to treat the standoff datatype tags specially, and transform their properties depending on which ontology schema we're using.

tobiasschweizer:
Couldn't we re-use the exact same logic we have already developed for the different value types?

benjamingeer:
I was thinking that we could hard-code all the standoff entities in `knora-api`. For each entity, there could also be a set of `knora-base` predicates to be replaced. So then, when serving a tag class T:

1. Does T have a standoff datatype?
2. If so, look up the datatype definition D.
    a. Get the predicates to be replaced from D, and remove them from T.
    b. Add all of D’s predicates to T.

Actually we could do the same thing for everything in `knora-base`, then we would only have to store the labels and comments once. :)

benjamingeer:
I was just proposing having a table of properties in standoff classes that should be replaced with other properties when we serve those classes via the API.

> re-use existing logic

The existing logic is for serving instances, but here we are talking about serving class definitions. How could we re-use the existing logic for that?

tobiasschweizer:
When you serve a definition for a `knora-api:DateValue` in the ontology route, how would a standoff tag of type date be different form that? They both inherit from `DateBase`.

Maybe I do not understand what you are referring to.

benjamingeer:
Ah, I thought you were talking about the existing logic for serving values (not value classes).

When I serve a definition for `knora-api:DateValue`, I just serve a hard-coded definition. But I can’t do that for standoff classes, because people can make their own standoff classes. If you make a standoff class called `MyDateTag`, we can’t serve a hard-coded definition of that class, because we don’t have one.

[...]

Issue #632, opened 2017-10-13
ontology prefix convention


loicjaouen:
we should decide and document the ontology prefix to be used for:

- resource classes:
  as in [anything-onto.ttl](https://github.com/dhlab-basel/Knora/blob/8c1c0405e1616aa0b5a59cb4f79d81b787c2c6d4/webapi/_test_data/ontologies/anything-onto.ttl):

  ```
  http://www.knora.org/ontology/anything#Thing
  ```

- property declaration:
  as in [anything-onto.ttl](https://github.com/dhlab-basel/Knora/blob/8c1c0405e1616aa0b5a59cb4f79d81b787c2c6d4/webapi/_test_data/ontologies/anything-onto.ttl):

  ```
  http://www.knora.org/ontology/anything#hasBlueThingValue
  ```

- resource instances
  as in [anything-data.ttl](https://github.com/dhlab-basel/Knora/blob/8c1c0405e1616aa0b5a59cb4f79d81b787c2c6d4/webapi/_test_data/all_data/anything-data.ttl):

  ```
  <http://data.knora.org/0C-0L1kORryKzJAJxxRyRQ> a anything:Thing ;
  ```

- values
  as in [anything-data.ttl](https://github.com/dhlab-basel/Knora/blob/8c1c0405e1616aa0b5a59cb4f79d81b787c2c6d4/webapi/_test_data/all_data/anything-data.ttl):

  ```
  <http://data.knora.org/a-thing-with-text-values/values/1> a knora-base:TextValue ;
  ```

- lists
  as in [anything-data.ttl](https://github.com/dhlab-basel/Knora/blob/8c1c0405e1616aa0b5a59cb4f79d81b787c2c6d4/webapi/_test_data/all_data/anything-data.ttl):

  ```
  <http://data.knora.org/anything/treeList> a knora-base:ListNode ;
  ```

- user
  as in [admin-data.ttl](https://github.com/dhlab-basel/Knora/blob/8c1c0405e1616aa0b5a59cb4f79d81b787c2c6d4/webapi/_test_data/all_data/admin-data.ttl):

  ```
  <http://data.knora.org/users/BhkfBc3hTeS_IDo-JgXRbQ> a knora-base:User ;
  ```

- group
  as in [drawings-gods_admin-data.ttl](https://github.com/dhlab-basel/Knora/blob/develop/webapi/_test_data/other.v1.DrawingsGodsV1Spec/drawings-gods_admin-data.ttl):

  ```
[...]

Issue #595, opened 2017-08-21
[Salsah 1.5] Knora's IRIs syntax results in a cryptic projects' names for users


mrivoal:
After our discussion in Basel in march and @lrosenth's mail on 16.03.2017 (DaSCH-Mise en place), it was decided that Knora's IRIs would comply to this syntax: http://rdfh/projectId/resourceId, "projectId" being a number of 4 digits.

As far as we understand, the "projectId" is to be set in the admin named graph like this :
```
<http://data.knora.org/projects/0001> a knora-base:knoraProject ;
        knora-base:projectBasepath "/data/SALSAH/parole-religieuse"^^xsd:string ;
        knora-base:projectLongname "Embarras de la parole religieuse"^^xsd:string ;
>>>        knora-base:projectShortname "0001"^^xsd:string ;
	knora-base:projectDescription "blablabla."^^xsd:string ;
	knora-base:projectKeywords "religion, genre, médias, sociologie des controverses, problèmes publics, démocratie libérale, approche quali-quantitative, réseaux"^^xsd:string ;
	knora-base:projectOntologyGraph "http://www.knora.org/ontology/parole-religieuse"^^xsd:string ;
        knora-base:projectDataGraph "http://www.knora.org/data/parole-religieuse"^^xsd:string ;
        knora-base:status "true"^^xsd:boolean ;
        knora-base:hasSelfJoinEnabled "false"^^xsd:boolean .
```

But `knora-base:projectShortname` is also the place where Salsah 1.5 looks for the name of a project to display it to the user. And this results in non-transparent projects names, such as here: 

![capture d ecran 2017-08-21 a 14 41 29](https://user-images.githubusercontent.com/17294309/29520414-37596e2e-8681-11e7-8b9f-c385d9b81309.png)

It is not very convenient and as the number of projects grows, it might become very obscure. I don't know the way it is supposed to work in Salsah 2.0 (will things be different there?) but I think there should be another way to display the name of the projects to the users.

At the moment, relying on `knora-base:projectLongname` is not an option because the lenght of some project's title will raise display issues... unless the long/official name of the project is recorded in the project metadata as discussed in #566? 

Any suggestion on how we could deal with it and provide greater clarity to the user?

loicjaouen:
#472 "remove short names in Salsah from voc and project"  
It was done for vocabularies, not for projects.  
For sure, having long names twice in the top bar (next to salsah's logo) won't fit.  
[...]

Issue #523, opened 2017-05-31
Create a process for standardising ontologies


benjamingeer:
We will have two categories of ontologies:

* Project-specific ontologies, which the project admins may change. So we should try to prevent other projects from using them, to avoid data consistency problems. (E.g. if you try to change a cardinality that someone else is using, GraphDB won't allow you to do so.)
* Standardised ontologies, which are available for reuse and will not be changed.

We need a process for standardising ontologies. This would involve a public forum where proposals could be posted and discussed. We could create a GitHub project for this.

Ideally, before an ontology can be standardised, it should be discussed and agreed by at least two projects that want to use it. Those projects need not put everything they require into the standard, but only the requirements they have in common. Other features can be added by subclassing in project-specific ontologies.

The IRI of a standardised ontology will contain something indicating that it has been standardised, as well as a version number (as many ontologies do, like `http://xmlns.com/foaf/0.1/`). So if a new version is made, all the entities in the new version will have different IRIs. Anyone who wants to migrate their data to the new version will have to do so deliberately.

So when you create a new resource, and the GUI asks the API server what classes are available, the response will contain only the standardised classes and the classes in your project.

SepidehAlassi:
@danielasubotic & @subotic I believe this is relevant to the metadata ontology. We have actually many resource classes that are repeatedly defined in various project ontologies and also in the project metadata ontology. Maybe it is now time to actually consider standardizing the ontologies.

PR #649, opened 2017-10-25
Ontology update infrastructure part 4

benjamingeer:
* Treat built-in ontologies other than `knora-api` and `knora-base` (e.g. `dc`) as built-in ontologies, not as project-specific ones.
* Serve standoff classes and properties, including the ones in `standoff`, but not yet the ones in `knora-base` or standoff datatype tags (see #648).
* Refactor representation of OWL classes and properties to be more suitable for updates as well as read operations.

Fixes https://github.com/dhlab-basel/Salsah/issues/95.


benjamingeer:
TODO: refactor ontology case classes with read and write wrappers. Put inherited cardinalities in the read wrapper, and mark them with an annotation property in JSON-LD, so the ontology editor can distinguish them from the others.

benjamingeer:
TODO: Figure out how to serve standoff classes. There needs to be a way to restrict this to standoff classes that the user would be interested in.

docs/rst/knora-api-server/api_v2/knora-iris.rst
```diff
@@ -18,7 +18,96 @@
...
+
+- ``http://api.knora.org/ontology/knora-api/v2`` (for the default schema)
+- ``http://api.knora.org/ontology/knora-api/simple/v2`` (for the simple schema)
+
+Ontology Entity IRIs
+^^^^^^^^^^^^^^^^^^^^
+
+Knora ontologies use 'hash namespaces' (see `URI Namespaces`_). This means that the IRI of an ontology entity (a class or property definition) is constructed by adding a hash character (``#``) to the ontology IRI, followed by the name of the entity. In Knora, an entity name must be a valid XML NCName_. Thus, the class ``incunabula:book`` has the following IRIs:
+
+- ``http://www.knora.org/ontology/incunabula#book`` (for the internal entity)
```
mrivoal: Shouldn't we also use projectUUID in these IRIs? Because if we don't, chances are that we will have to face duplicate in the names of the project with the (hopefully) growing number of DaSCH satellites.

But we should keep the idea of the `knora-base:projectShortName` property to be used by the GUI as discussed in #595 and add a `knora-base:projectCode`corresponding the the projectUUID.

docs/rst/knora-api-server/api_v2/knora-iris.rst
```diff
@@ -18,7 +18,96 @@
...
+-------------
+
+Knora generates IRIs for data that it creates in the triplestore. Each generated data IRI contains one or more UUID_ identifiers to make it unique. To keep data IRIs relatively short, each UUID is Base64_ encoded, using the 'URL and Filename safe Base64 Alphabet' specified in Table 2 of RFC 4648, without padding; thus each UUID is a 22-character string.
+
+Data IRIs are not currently intended to be dereferenced as URLs. Instead, each dereferenceable data IRI will have a corresponding permalink URL. However, every generated data IRI begins with ``http://rdfh.ch``. This domain is owned by the Data and Service Center for the Humanities (DaSCH_), which develops Knora, so it would be possible to make data IRIs directly dereferenceable in the future.
+
+The formats of generated data IRIs for different types of objects are as follows:
+
+- Project: ``http://rdfh.ch/projects/PROJECT_UUID``
+- Resource: ``http://rdfh.ch/PROJECT_ID/RESOURCE_UUID`` (where ``PROJECT_ID`` is not a random UUID, but the project's unique ID, which can be registered with the DaSCH)
```
mrivoal: One the line above, ``PROJECT_UUID`` is mentionned whereas here, it is refered to as ``PROJECT_ID``. It should probably be unified.

benjamingeer: The resource ID has the project ID (should be project code, but is currently project shortname) in case someday we want resource IRIs to be directly dereferenceable. But again, how do we guarantee that project codes are unique? Will people have to register a project code with us even to create a test project? And if they don’t, suppose they then want to turn their test project into a real project, and they have already created resources with a test project code that isn’t unique?

mrivoal: > The resource ID has the project ID (should be project code, but is currently project shortname) in case someday we want resource IRIs to be directly dereferenceable. 

Our resources IRIs (in Lausanne) are currently based on the project code (or project-ID, this four number chain of characters that we have discussed earlier), as I think was decided earlier this spring.

> But again, how do we guarantee that project codes are unique? Will people have to register a project code with us even to create a test project? And if they don’t, suppose they then want to turn their test project into a real project, and they have already created resources with a test project code that isn’t unique?

Being in charge, the DaSCH would/could allocate project-IDs range to an institution, like we did already between Basel and Lausanne. Until another system like a registry of projects is set up, as mentionned by @lrosenth this spring as well.

docs/rst/knora-api-server/api_v2/knora-iris.rst
mrivoal: By `projectUUID`, I actually meant `projectcode`, as until now we haven't mentionned projectUUID (at least I didn't understand it that way).

>An ontology can exist independently of a project. There will be ontologies shared by multiple projects. So I don’t think we can include any sort of project ID (whether UUID, shortname, or project code) in an ontology IRI. 

Yes, but if you are talking about the kind of standard ontologies shared by various projects, my guess is that these ontologies will undergo a review process. Their status will also change as they proceed from 'project-specific ontologies' to 'standard built-in ontologies'. In the process, aside from cleaning and improving these ontology, can't we imagine that the initial project-ID attached to the IRIs will be changed into a more generic name (such as bibliography, for example)?

subotic: > I could imagine a problem occurring if someday we do federated searches in which triplestores on different servers are talking to each other.

I don't know if we are ever going to support federated searches on the triplestore level. More likely on the federated `webapi` server level. In those cases, the different hostnames will give us unique names.

>  Or if two ontologies created with the same name on different servers are later moved to the same server.

Maybe it would be easier to require a name change in those rare cases?

docs/rst/knora-api-server/api_v2/knora-iris.rst
subotic: Mass renaming?

lrosenth:
I would suggest that each location will get a unique prefix which has to be used to form the project id. With this concept we would not have problems. Federated queries on RDF-level is on my wish list :-)

Sent from my iPad

On 27 Oct 2017, at 17:29, Ivan Subotic <notifications@github.com<mailto:notifications@github.com>> wrote:


@subotic commented on this pull request.

________________________________

In docs/rst/knora-api-server/api_v2/knora-iris.rst<https://github.com/dhlab-basel/Knora/pull/649#discussion_r147442213>:

> +For example, suppose a Knora API server is running at ``knora.example.org<http://knora.example.org>`` and hosts the ``incunabula`` ontology. That ontology can then be requested using either of these IRIs:
+
+- ``http://knora.example.org/ontology/incunabula/v2`` (for the default schema)
+- ``http://knora.example.org/ontology/incunabula/simple/v2`` (for the simple schema)
+
+While the internal ``incunabula`` ontology refers to definitions in ``knora-base``, the external ``incunabula`` ontology that is served by the API refers instead to a ``knora-api`` ontology, whose IRI depends on the schema being used:
+
+- ``http://api.knora.org/ontology/knora-api/v2`` (for the default schema)
+- ``http://api.knora.org/ontology/knora-api/simple/v2`` (for the simple schema)
+
+Ontology Entity IRIs
+^^^^^^^^^^^^^^^^^^^^
+
+Knora ontologies use 'hash namespaces' (see `URI Namespaces`_). This means that the IRI of an ontology entity (a class or property definition) is constructed by adding a hash character (``#``) to the ontology IRI, followed by the name of the entity. In Knora, an entity name must be a valid XML NCName_. Thus, the class ``incunabula:book`` has the following IRIs:
+
+- ``http://www.knora.org/ontology/incunabula#book`` (for the internal entity)


I could imagine a problem occurring if someday we do federated searches in which triplestores on different servers are talking to each other.

I don't know if we are ever going to support federated searches on the triplestore level. More likely on the federated webapi server level. In those cases, the different hostnames will give us unique names.

Or if two ontologies created with the same name on different servers are later moved to the same server.

Maybe it would be easier to require a name change in those rare cases?

docs/rst/knora-api-server/api_v2/knora-iris.rst
benjamingeer: OK but this is open source software. Anyone can download and run it even if they aren’t (yet) a DaSCH satellite and don’t gave an allocated ID range.

docs/rst/knora-api-server/api_v2/knora-iris.rst
benjamingeer: @mrivoal A UUID is not something assigned manually, it’s a very long random number, which the Knora API server encodes as a 22-character string of letters and digits.

The problem with changing the IRI of an ontology is that you also have to change all the data that uses the ontology. In the case of a project-specific ontology that became standardised, I think ideally there would be a project-specific ‘bridge’ ontology so the data wouldn’t have to change.

So... maybe including a short project ID (not a UUID) in the ontology IRI wouldn’t be such a bad idea. But it would mean that all existing ontologies and data would have to be updated.

benjamingeer:
@lrosenth OK but the question is whether we include the project ID in the ontology IRI. This would mean changing all existing ontologies and data.

loicjaouen:
So, if I start writing an ontology for `incunabula` at UNIL today, I would:
- add the project's definition in the admin data graph under the IRI  
  `http://rdfh.ch/projects/abcdefdhijklmnopqrstuv` (and we should for now make a UUID generated for us as we write that by hand, no problem)
- it's internal ontology would be `http://www.knora.org/ontology/unil-incunabula`
- it's internal ontology's entities would be `http://www.knora.org/ontology/unil-incunabula#book`
- and its resources would be like `http://rdfh.ch/0110/abcdefdhijklmnopqrstuv`

knowing that `0110` is in the range of ID that is allocated to UNIL

Does that sum up the conversation so far?

docs/rst/knora-api-server/api_v2/knora-iris.rst
mrivoal: > @mrivoal A UUID is not something assigned manually, it’s a very long random number, which the Knora API server encodes as a 22-character string of letters and digits.

Yes, I figured it out :)

> The problem with changing the IRI of an ontology is that you also have to change all the data that uses the ontology. 
I am very aware of that, we have already faced the problem when we changed the syntax of  our IRIs to include the projectcode instead of the projetshortname.

> So... maybe including a short project ID (not a UUID) in the ontology IRI wouldn’t be such a bad idea. But it would mean that all existing ontologies and data would have to be updated.

Honestly, I thought this was the road we all agreed on several months ago.

benjamingeer: > Honestly, I thought this was the road we all agreed on several months ago.

I thought we had only agreed that project IDs would be used in resource IRIs. I don’t remember anything about including them in ontology IRIs. But I’m not known for having the best memory. Anyway, we can certainly do it, but it will take some work to convert everything.

loicjaouen: We decided that there would be project ID, and that IRIs would be `http://rdfh.ch/PROJECT_ID`, we didn't specify clearly "resource" IRI and "project" IRI (c.f. minutes by mail from Lukas dated 2017-03-16).  
So we apparently left room for misunderstandings.  
As this was to avoid collisions, we thought it should be in the ontology too.

docs/rst/knora-api-server/api_v2/knora-iris.rst
loicjaouen: We could do the same as for IPs and DNS: have a [domestic range](https://en.wikipedia.org/wiki/IP_address#Private_addresses) of project IDs.

benjamingeer: Good idea!

docs/rst/knora-api-server/api_v2/knora-iris.rst
benjamingeer: I’d like to see if I can make this work in a backwards-compatible way, so existing data doesn’t have to be changed.

docs/rst/knora-api-server/api_v2/knora-iris.rst
loicjaouen: the problem is that there is no software code, it is only convention that we set amongst ourselves, we can't allocate a range and put it as a default value in a config file.  
Of course, we can explain it in the doc... but human understanding is error prone :)

benjamingeer:
I think it will be possible to have optional project IDs in ontology IRIs, so existing data doesn't have to be changed. Thus either of these would work as an internal ontology IRI:

* `http://www.knora.org/ontology/incunabula`
* `http://www.knora.org/ontology/0010/incunabula`

I need to test this more to confirm.

mrivoal:
@benjamingeer, @lrosenth, @loicjaouen, @subotic, I think the discussion is getting really confused and we are not going anywhere. Could we consider a Skype to go through the problem together (again) and to set those IRIs once and for all? 

I mean, we have been asking for this and discussing it since march, I think it is high time we come to a conclusion and stick to it.

But @benjamingeer, what you are suggesting right now doesn't fit with what we understood:
Instead of `http://www.knora.org/ontology/incunabula#book`, with projectID within ontology IRIs, we would expect `http://www.knora.org/ontology/0010#book` and to-date, this is the way our ontologies are defined.

benjamingeer:
@mrivoal `http://www.knora.org/ontology/0010#book` isn’t possible, because it wouldn’t allow a project to have more than one ontology.

mrivoal:
You are right... But we did it anyway.

benjamingeer:
Hmm, that will be a problem, because an ontology name is also used as a prefix label, which must be a valid XML NCName

benjamingeer:
Hmm, that will be a problem, because an ontology name is also used as a prefix label, which must be a valid XML NCName, which cannot start with a number.

mrivoal:
But we have defined corresponding prefixes like this:
`@prefix h-steiner: <http://www.knora.org/ontology/0110#> .`

If we need to change things, we will. The most important is that we can agree on some schema and stick to it. Hence the suggestion to talk about it. Again.

benjamingeer:
In API v2 (and even in API v1 for bulk import), the API server needs to be able to generate a prefix automatically from an ontology IRI. If we use project IDs as namespaces for ontology names, then both the project ID and the ontology name will need to be in the prefix. For example, given this ontology IRI:

```
http://www.knora.org/ontology/0010/incunabula
```

The prefix could be `p0010-incunabula` (that's what I implemented yesterday on this branch, for now).

benjamingeer:
@mrivoal @loicjaouen Lukas, Ivan, and I talked about this briefly. Lukas decided that in order to be able to progress more quickly right now, we will go ahead with the [optional project ID in the ontology IRI](#issuecomment-340526567). If your ontologies are working for you now with API v1, you don't need to change them now. When the time comes (but not before January), we will help you make any necessary changes to keep your data working with API v2. So, no meeting is necessary. :)

mrivoal:
Ok, then. Fine.

But as we are working on new ontologies, we still need a detailled list of the correct expected syntaxes for 
- ontologies IRIs
- admin data stuff
- permissions data 
- and for data (lists, instances of resources, values, etc.)

@benjamingeer, could you please review, update and correct @loicjaouen's summary in issue #632 (@loicjaouen did not include the `rdfh` prefixes for instances) so that we can move forward and so that everything is clear for everyone? At the moment, I am sorry but everything is way too confused for me to proceed.

benjamingeer:
@mrivoal More documentation is on its way.

loicjaouen:
I have an additional couple of questions, @benjamingeer you previously said:

> An ontology can exist independently of a project. There will be ontologies shared by multiple projects. So I don’t think we can include any sort of project ID (whether UUID, shortname, or project code) in an ontology IRI.

The nearly syllogism of the first and last sentences is: "an ontology can exist without a project so no ontology should be linked to a project".

But there is a link between a project and a specific ontology, isn't it? In the [admin configuration](https://github.com/dhlab-basel/Knora/blob/develop/webapi/_test_data/all_data/admin-data.ttl) defining a project, here `incunabula`, we have:
```
 knora-base:projectOntologyGraph "http://www.knora.org/ontology/incunabula"^^xsd:string ;
```
And we load this incunabula in that named graph.

We can still use this ontology in other projects.

What does it change to have an `projectOntologyGraph`? 

Actually, wouldn't be usefull for projects to namely declare ontologies a project use, so the GUI can restrict the choices of resource types for that project and the user is not drowned under similar choices? As a user, if I want to add a book from a bibliographical ontology, I might have troubles to select which of the potentially many bibliographical ontology to choose from, and most likely, I might have trouble to consistently do the same choice.  

And later:
> http://www.knora.org/ontology/0010#book isn’t possible, because it wouldn’t allow a project to have more than one ontology.

I don't get this one, as we can use any existing ontologies.

subotic:
- I'm working on updating the project information (#654). The property `projectOntologyGraph` is now `projectOntology`, where one project can have `0-n` ontologies. Ontologies are stored in a named graph with the same IRI as the ontology. 
- As far as I understand it, nothing is preventing a project to use any kind of ontology inside the triplestore. We artificially restrict a project to only use the ones defined in its own project (via `projectOntology`), because project ontologies are allowed to change. If project A uses an ontology from project B, and then project B decides to change the ontology, what should then happen? Either let the data from project A become inconsistent, or not allow project B to change its own ontology?
- Ontologies shared by multiple projects will 'live' in a special place. They could, for example, belong to the `SystemProject`. Since they are stable, users will be able to use them directly or base their project ontologies on them. Thus I would suggest that every ontology is part of a project (user's or system).

loicjaouen:
thanks @subotic for pointing at #654, I read to quickly and missed to fully understand the implications of that issue :( .  Got it now!

So, either the ontologies are _public_ or project specific?

benjamingeer:
> So, either the ontologies are _public_ or project specific?

That's the goal of #523. I think the API server should enforce this, so this scenario can't happen:

> If project A uses an ontology from project B, and then project B decides to change the ontology, what should then happen?

subotic:
> I think the API server should enforce this, so this scenario can't happen:

>> If project A uses an ontology from project B, and then project B decides to change the ontology, what should then happen?

You are right, the server should enforce this. It would need to check every:
 
- project ontology creation/update
 - resource creation/change

and make sure that only the project's ontologies or system ontologies are involved.

loicjaouen:
Then there is no problem in having projects IDs or name in project specific ontologies.  
It might even be a necessity.

But that's not the right issue to discuss it.

I think we just got caught in misunderstandings due to difference of contexts; us thinking in the "here and now" (#632) and you living already in the next iteration (#654) ;)

Clearing these contexts is a matter of release management, something we should now formalize as we are in production and the next release will require a serious step to port our existing data.

PR #659, opened 2017-11-06
feature (webapi): add project short code

### LINKED ISSUE #654: Store an optional short code and multiple ontologies per project
Each project needs to be able to have an optional short code, which is a hexadecimal number of at least four digits. Both the short code and the project's short name must be unique. When the projects responder returns information about a project, it should return the short code if present.

Although the short code is optional for now, we could make it required later.

Each project also needs to have one or more ontologies associated with it. In the API, we need to be able to get a list of a project's ontologies.

The IRI of the named graph of each ontology should simply be the IRI of the ontology, rather than being configurable. Therefore the projects responder doesn't need to store the IRI of the ontology's named graph separately.

The IRI of the named graph of the project's data should be predictable from the project IRI, rather than being configurable.

If a short code is provided, the project IRI could contain it instead of a UUID.

See also #595 and #472.

### LINKED ISSUE #660: Refactor admin and permission test-data IRIs to conform to our new standard
old: `http://data.knora.org/projects/[UUID]`
new: `http://rdfh.ch/projects/[UUID]`

- projects
- groups
- users
- permissions

### LINKED ISSUE #663: Namespace collisions between projects and standoff
There seems to be a namespace collision between standoff mappings:

`http://data.knora.org/projects/standoff/mappings/StandardMapping/elements/4b-Cx4qYQOyy9mUswmvW7`

and projects:

`http://data.knora.org/projects/[shortcode/][shortname]` or now `http://rdfh.ch/projects/[shortcode/][shortname]`

Maybe you could use something like this:

``http://data.knora.org/standoff/mappings/StandardMapping/elements/4b-Cx4qYQOyy9mUswmvW7`?

subotic:
### Summary

- [x] Add `projectShortcode`
- [x] Remove `projectDataGraph`
- [x] Rename `projectOntologyGraph` to `projectOntology` with a minimum cardinality of 0.
- [x] Add the creation of a project with and without shortcode. If shortcode is present, then it will be used instead of a UUID for the project's IRI.
- [x] The project is created without a `projectOntology`.
- [x] Added the shortcode `00FF` to the `images` test project and renamed the project IRI to `http://rdfh.ch/projects/00FF`. Also changed the ontology IRI, project data, all tests, and load test data scripts accordingly.
- [x] Add messages to the `ProjectsResponderV1` for adding and removing a `projectOntology`. The idea is that the `OntologyResponder` should send this messages.
- [x] Return named graphs for each `projectOntology`.
- [x] Update `knoraIdUtil.makeRandomResourceIri` to take a `ProjectInfoV1` and construct the resource IRI including the `shortcode` if available.
- [x] Update admin and permissions test data IRIs (#660)
- [x] Update standoff ontology and data IRIs (#663)
- [x] Update documentation (typescript interfaces) with the new property + add short description of how to migrate existing data
- [x] Update `OntologyResponderV2` to send a `ProjectOntologyAddV1` message when a new ontology is created.
- [x] Fix integration tests.
- [x] Fix/Run  `Salsah1` browser tests
- [x] Added project shortcode validation. The validation step transforms the shortcode string to uppercase. This way `00ff` will become `00FF`. This is also in line with the project shortcode specification from @lrosenth.

### Reminders

- [ ] Update `Salsah2` GUI with changes to `ProjectInfo` (https://github.com/dhlab-basel/Salsah/issues/97)

### Issues

- resolves #654
- resolves #660
- resolves #663


loicjaouen:
@subotic : there is a project code collision as `0101` is in Unil's range and we already allocated it.

According to @lrosenth :
> The projects „0000“ to „00FF“ are used for test purposes are not redicrected by the central IRI-resolver! Thus the first useful project will be „0100“.
> In the beginnings Unil uses the project numbers „0100“-"07FF“ , unibas uses „0800“ - „08FF"

subotic:
@loicjaouen Thanks. I wasn't sure about this and wanted to ask @benjamingeer anyway. I will change to `00FF`.

benjamingeer:
@loicjaouen Thanks, I didn't know about that either. :)

subotic:
@benjamingeer The PR is now ready for review. Thanks.

webapi/src/main/scala/org/knora/webapi/util/StringFormatter.scala
```diff
@@ -1500,4 +1502,23 @@ class StringFormatter private(knoraApiHttpBaseUrl: String) {
...
+    def projectDataNamedGraph(projectInfo: ProjectInfoV1): IRI = {
+        if (projectInfo.shortcode.isDefined) {
+            OntologyConstants.NamedGraphs.DataNamedGraphStart + "/" + projectInfo.shortcode.get + "/" + projectInfo.shortname
+        } else {
+            OntologyConstants.NamedGraphs.DataNamedGraphStart + "/" + projectInfo.shortname
+        }
+    }
+
+    def isValidShortcode(shortcode: String): Boolean = {
+        // ToDo: do some checking
```
benjamingeer: Could you implement this TODO? There's already a regex sub-pattern for it, `ProjectIDPattern`, which you could use to make a regex for validating shortcodes.

subotic:
Thanks. I forgot about it. Is added now.

subotic:
And thanks for the review :-)

webapi/src/main/scala/org/knora/webapi/util/StringFormatter.scala
```diff
@@ -232,12 +232,15 @@ class StringFormatter private(knoraApiHttpBaseUrl: String) {
...
+    private val NCNameRegex: Regex = ("^" + NCNamePattern + "$").r
+
+    // A regex sub-pattern for project IDs, which consist of at least 4 hexadecimal digits.
     private val ProjectIDPattern: String =
         """\p{XDigit}{4,}"""
 
-    // A regex for matching a string containing only an ontology prefix label or a local entity name.
-    private val NCNameRegex: Regex = ("^" + NCNamePattern + "$").r
+    // A regex for matching a string containing the project ID.
+    private val ProjectIDRegex: Regex = (ProjectIDPattern).r
```
benjamingeer: This regex matches hex digits anywhere in a string:

```
scala> import scala.util.matching.Regex
import scala.util.matching.Regex

scala> val ProjectIDPattern: String = """\p{XDigit}{4,}"""
ProjectIDPattern: String = \p{XDigit}{4,}

scala> val ProjectIDRegex: Regex = (ProjectIDPattern).r
ProjectIDRegex: scala.util.matching.Regex = \p{XDigit}{4,}

scala> ProjectIDRegex.findFirstIn("****1234****")
res0: Option[String] = Some(1234)
```

I think you need this:

```
scala> val ProjectIDRegex: Regex = ("^" + ProjectIDPattern + "$").r
ProjectIDRegex: scala.util.matching.Regex = ^\p{XDigit}{4,}$

scala> ProjectIDRegex.findFirstIn("****1234****")
res1: Option[String] = None

scala> ProjectIDRegex.findFirstIn("1234")
res2: Option[String] = Some(1234)
```

subotic:
Just as a reminder of what was specified, so that I can find it again:

>Each project gets a number in the range of 0 - 65535 (unsigned short) which is then encoded as 4-decimal hexnumber. Thus we have the range from 0000 to FFFF (=65535 possible projects). The signs used for
>10 = A
>11 = B
>12 = C
>13 = D
>14 = E
>15 = F
> The A-F must be upper case.
>
> Remarks: I thing we should use simple HEX-encoding and not Base64 - it’s much simpler and easier to understand to the human reader!
>
>The projects „0000“ to „00FF“ are used for test purposes are not redicrected by the central IRI-resolver! Thus the first useful project will be „0100“.
>
>In the beginnings Unil uses the project numbers „0100“-"07FF“ , unibas uses „0800“ - „08FF"

webapi/src/main/scala/org/knora/webapi/util/StringFormatter.scala
subotic: Thanks. I have also tightened it up (I think), to require 4 digits.

benjamingeer:
I'm not sure 4 digits is going to be enough in the long run, but OK. :)

subotic:
Thanks. Shall I make it 5 digits? This would be over 1 million.

subotic:
But I think that Lausanne is already using 4 digit project codes. I will leave it at 4 and we can always make it larger later.

benjamingeer:
Yes, it would be easy to increase it.

loicjaouen:
we will have to make changes anyway, if there are changes to be made, let's do then now

benjamingeer:
If Lausanne gets `0100-07FF`, that gives them 1792 projects (in base 10). If every institution gets 1792 projects, then we can have a maximum of 36 institutions. That doesn't seem like enough to me. Could you ask Lukas about it?

loicjaouen:
the key is to find a short enough number for human memory but still long enough for the duty :)
we can propose 6 digits?

benjamingeer:
Moved the shortcode length issue to #664.

PR #662, opened 2017-11-07
Implement smart IRIs

benjamingeer:
- [x] Add a `SmartIri` class to simplify IRI inspection, validation, and conversion.
- [x] Rename methods in `StringFormatter` to clarify which ones inspect (`is*`), which ones validate (`validate*`), and which ones convert (`to*`).
- [x] Use `SmartIri` for the IRIs of ontologies and ontology entities in API v2 code.
- [x] Use `SmartIri` for all IRIs in KnarQL.
- [x] Don't use v2 ontology info classes in v1 code, because that would mean changing a lot of v1 code to use `SmartIri`. Instead, add v1 wrappers around v2 ontology info classes. In the wrappers, convert smart IRIs to string IRIs. Use only the wrappers in v1 code.
- [x] Refactor v2 ontology info classes so they can support both read and write operations using the wrapper pattern. The read wrappers can convert themselves from internal to external, and the wrapped classes can convert themselves from internal to external and vice versa.
- [x] Separate schema conversion from JSON-LD generation. When serving ontology info, first convert IRIs from internal to external as needed, generating a new `ReadEntityDefinitionsV2` instance. Then convert that instance to JSON-LD.
- [x] For better performance, cache `SmartIri` instances representing Knora ontology entities..


benjamingeer:
TODO: Instead of using stringFormatter.toSmartIri(iri) in existing code, try using an implicit class:

```
implicit class ConvertibleIri(val self: String) extends AnyVal {
    def toSmartIri(implicit stringFormatter: StringFormatter): SmartIri = stringFormatter.toSmartIri(self)
}
```

Then:

```
val stringFormatter = StringFormatter.getInstance
val iri: IRI = // ...
val smartIri = iri.toSmartIri
```

benjamingeer:
`StringFormatter` could be renamed `ValueFormats`.

benjamingeer:
## TODO

### IRIs as predicate objects

`PredicateInfoV2` needs to know which of its objects are IRIs, so it can convert their schemas when needed. Instead of having it store them as strings, make types like this:

```
trait PredicateObjectV2
case class IriPredicateObjectV2(iri: SmartIri) extends PredicateObjectV2
case class LiteralPredicateObjectV2(literal: String, language: Option[String]) extends PredicateObjectV2
```

Make the `objects` member of `PredicateInfoV2` be a `Set[PredicateObjectV2]`, and get rid of `objectsWithLangs`. `OntologyResponderV2` can determine which objects are IRIs by calling `stringFormatter.isIri()`. In the future, it can get this information from the the types indicated in the SPARQL responses.

Then when predicates are converted to another ontology schema, there won't be any need for hard-coded lists of the predicates that have IRI objects.

### Equality of SmartIri objects

Give the `SmartIri` trait a protected abstract method that returns the underlying RDF4J IRI object. Then use that object in `equals` and `hashCode`, instead of `toString`.

### StringFormatter method naming

Rename `getInstance` to `getGeneralInstance`.

### StringFormatter and SmartIri design docs

In order to parse project-specific API v2 ontology IRIs, the `StringFormatter` class contains regexes that are calculated based on the Knora API server's hostname, which is set in `application.conf`, which is not read until the Akka `ActorSystem` starts. Therefore `StringFormatter` has to be a class, not an object.

There are two instances of `StringFormatter`. One (`StringFormatter.getGeneralInstance`) is initialised after the `ActorSystem` starts, and can handle project-specific API v2 ontology IRIs. This instance is used almost everywhere in the API server. The other instance (`StringFormatter.getInstanceForConstantOntologies`) can be used as soon as the JVM starts, but can't handle project-specific API v2 ontology IRIs. This instance is used only to initialise the hard-coded API v2 ontologies.

The smart IRI implementation, `SmartIriImpl`, is nested in the `StringFormatter` class because it needs access to private members of that class (especially the regexes that are computed when the general `StringFormatter` instance is constructed). However, this means that the type of a `SmartIriImpl` instance is dependent on the instance of `StringFormatter` that constructed it. Therefore, you can't compare two instances of `SmartIriImpl` created by two different instances of `StringFormatter`.

To make it possible to compare smart IRI objects, the publicly visible smart IRI type is the `SmartIri` trait, which is defined in the `StringFormatter` object (and is extended by `SmartIriImpl`). Since `SmartIri` is defined in an object rather than in a class, two instances of `SmartIri` can be compared, even if they were made by different instances of `StringFormatter`, as long as `SmartIri` provides its own `equals` and `hashCode` methods.

The `StringFormatter` class has factory methods for constructing smart IRI objects, but calling these methods is too verbose. To make code more concise, the object `IriConversions` provides an implicit class that in effect allows you to call `toSmartIri` on any string.

benjamingeer:
Instead of `UnknownSchema` make the schema an option in `IriInfo`, and use `None` for Knora data IRIs.

benjamingeer:
@tobiasschweizer `SearchRouteV2R2RSpec` still has errors, and I need your help to understand why:

```
org.knora.webapi.NotImplementedException: Value type http://www.knora.org/ontology/knora-base#DateValue not supported in FilterExpression
	at org.knora.webapi.responders.v2.SearchResponderV2$AbstractExtendedSearchTransformer.transformFilterExpression(SearchResponderV2.scala:832)
```

Looking at that part of the code in `SearchResponderV2`, it's looking for things like `OntologyConstants.KnoraBase.Date`, which doesn't seem like it should exist. There are comments like this in `OntologyConstants`:

```
val Date: IRI = KnoraBasePrefixExpansion + "Date" // TODO: find a better solution for v2 simple
```

Could you explain? I'm guessing that you did this because it would have been more work to convert the IRI `knora-api:Date` to the IRI `knora-base:DateValue`. But this is done automatically now by `SmartIri`. So should this code in `SearchResponderV2` now look for `knora-base:DateValue`?

tobiasschweizer:
The reason is that in a KnarQL query the user would use simple Iris like a `knora-api:Date` that will be converted to a `knora-base:DateValue`. The main rationale is hiding the value object complexity from the user (to her it is simply a date, not a value object). However, we could surely discuss if omitting `Value` makes a difference at all.

In `SearchResponderV2` there are many places where Iris have to converted from external to internal (when looking at type annotations) or from internal to external (when comparing an already converted Iri to the given type annotations). It would make the could much clearer if smart Iris would handle that automatically. We can also look at that later today.

tobiasschweizer:
> Looking at that part of the code in SearchResponderV2, it's looking for things like OntologyConstants.KnoraBase.Date, which doesn't seem like it should exist.

Yes, it was kind of a hack ;-)

benjamingeer:
Maybe my question wasn’t clear. I’m asking why the search responder refers to an entity called `http://www.knora.org/ontology/knora-base/Date` instead of `http://www.knora.org/ontology/knora-api/simple/v2/Date`. There is no `Date` in `knora-base`, only in `knora-api/simple/v2`. Why does the search responder refer to a nonexistent `Date` entity in `knora-base`. Why doesn’t it use the `Date` entity that does exist in `knora-api/simple/v2`?

`SmartIri` can convert `http://www.knora.org/ontology/knora-base/DateValue` to `http://www.knora.org/ontology/knora-api/simple/v2/Date`. Perhaps that is what you need?

tobiasschweizer:
Simply because the external Iri (type annotation) is converted to an internal one before looking at it, using a regex. And the regex just replaces a part of the Iri, but does not handle the missing `Value`:

https://github.com/dhlab-basel/Knora/blob/9236b0fca594dc035e39b31ae56dbb292cd2058f/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala#L602

However, this is not really consistent behavior since there are also non Knora types like `xsd:string` that cannot be converted. 

So I would be happy to change this.

benjamingeer:
> So I would be happy to change this.

OK, could you please do it on this branch? Then I think `SearchRouteV2R2RSpec` should pass.

I already changed the KnarQL parser and type inspector so they use `SmartIri`. Once you have a `SmartIri`, you can just call `toOntologySchema` on it to convert it from external to internal and vice versa.

tobiasschweizer:
> OK, could you please do it on this branch? Then I think SearchRouteV2R2RSpec should pass.

I will look at it :-)

benjamingeer:
TODO: Add a test for the route `/v2/ontologies/namedgraphs`.

Also, if we're not going to call them named graphs, we should rename the routes.

benjamingeer:
@tobiasschweizer In `SearchResponderV2`, the variable `orderByDependentResVars` is not used:

https://github.com/dhlab-basel/Knora/blob/8876feb9432a62767c23daefd22547bc72670991/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala#L1643

tobiasschweizer:
> @tobiasschweizer In SearchResponderV2, the variable orderByDependentResVars is not used:

removed in https://github.com/dhlab-basel/Knora/pull/662/commits/9c2a697996fbb86f0cde6287ba75def97835525a

The reason this was there is that in an earlier version sorting of the results was done using the main resource variable followed by all the dependent resource variables to make the sorting deterministic. However, later I had to change the logic because you do not want to have more than one result row for the same main resource. Now, we group by the main resource.

benjamingeer:
@tobiasschweizer I renamed the `/ontologies/namedgraphs` route to `/ontologies/allentities`. The route that just returns ontology metadata is now called `/ontologies/metadata`.

tobiasschweizer:
@benjamingeer Thank you very much for the notification. I will make a new branch in SALSAH that will reflect this change and other changes on this branch that effect the routes or the messages. 

Actually, we need to do coordinated merges: basically you need to open both pull requests and hit the merge buttons at the same time. I guess we need tentacles for this.

benjamingeer:
@tobiasschweizer I started added some API v design docs, including a discussion of `SmartIri`, in `docs/rst/knora-api-server/design-documentation/api-v2.rst`. See what you think.

tobiasschweizer:
Ok, I look at it now. Just making myself a cappuccino ...

docs/rst/knora-api-server/design-documentation/api-v2.rst
```diff
@@ -0,0 +1,189 @@
...
+Implementation
+--------------
+
+JSON-LD Parsing and Formatting
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Each API response is represented by a class that extends ``KnoraResponseV2``, which
+has a method ``toJsonLDDocument`` that specifies the target schema. It is currently
+up to each route to determine what the appropriate response schema should be. Some routes
+will support only one respnse schema. Others will allow the client to choose, and there will
```
tobiasschweizer: respnse -> response

docs/rst/knora-api-server/design-documentation/api-v2.rst
```diff
@@ -0,0 +1,189 @@
...
+
+Whenever possible, the same data structures are used for input and output. Often more data is
+available in output than in input. For example, when a value is read from the triplestore, its IRI
+is available, but when it is being created, it does not yet have an IRI. In such cases, there is a
+class like ``ValueContentV2``, which represents the data that is used both for input and for output.
+When a value is read, a ``ValueContentV2`` is wrapped in a ``ReadValueV2``, which additionally
+contains the value's IRI. When a value is created, it is wrapped in a ``CreateValueV2``, which has
+the resource IRI and the property IRI, but not the value IRI.
+
+A ``Read*`` wrapper can be wrapped in anotehr ``Read*`` wrapper; for example, a ``ReadResourceV2``
```
tobiasschweizer: anotehr -> another

tobiasschweizer:
I had a look at it and I think that it is very useful!

tobiasschweizer:
I am adding more KnarQL R2R tests now to make sure that the changes on this branch do not have unintended effects.

tobiasschweizer:
haha, found a bug in my code!!!!!!!!

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -231,13 +230,13 @@ class SearchResponderV2 extends ResponderWithStandoffV2 {
                 case iriRef: IriRef =>
                     // type info keys are external api v2 simple Iris,
                     // so convert this internal Iri to an external api v2 simple if possible
```
tobiasschweizer: we should get rid of this comment because we do not convert to internal anymore, right?

benjamingeer: Do you mean just get rid of the comment, or also get rid of the conversion from internal to external?

tobiasschweizer:
only the comment since the conversion does not happen anymore

PR #666, opened 2017-11-15
must be empty

### LINKED ISSUE #665: extended search: condition exists fails
Conditions:
- with a fresh Salsah 1.5 session, 
- when making an extended search on 
- a resource
- that has a link to another resource
- and the search condition is that this linked resource exists

Salsah then complains about: `Bad request error`
api-error

loicjaouen:
but not undefined

fixes #665

### DIFF
--- a/salsah/src/public/js/jquery.extsearch.js
+++ b/salsah/src/public/js/jquery.extsearch.js
@@ -750,7 +750,7 @@
 								searchval.push(""); // must be!!!
 							}
 						}
-						if (!searchval[0]) {
+						if (searchval[0] === undefined) {
 							searchval[0] = "";
 						}
 					});
--- a/salsah/src/public/js/jquery.extsearch.js
+++ b/salsah/src/public/js/jquery.extsearch.js
@@ -747,8 +747,11 @@
 								break;
 							}
 							default: {
-								searchval.push("empty"); // must be!!!
+								searchval.push(""); // must be!!!
 							}
+						}
+						if (!searchval[0]) {
+							searchval[0] = "";
 						}
 					});
 
--- a/salsah/src/public/js/jquery.extsearch.js
+++ b/salsah/src/public/js/jquery.extsearch.js
@@ -747,7 +747,7 @@
 								break;
 							}
 							default: {
-								searchval.push(undefined); // must be!!!
+								searchval.push("empty"); // must be!!!
 							}
 						}
 					});


salsah/src/public/js/jquery.extsearch.js
```diff
@@ -747,7 +747,7 @@
 								break;
 							}
 							default: {
-								searchval.push(undefined); // must be!!!
+								searchval.push("empty"); // must be!!!
```
tobiasschweizer: Could you also just pass an empty string `""`?

loicjaouen: checking...

salsah/src/public/js/jquery.extsearch.js
```diff
@@ -747,9 +747,12 @@
 								break;
 							}
 							default: {
-								searchval.push(undefined); // must be!!!
+								searchval.push(""); // must be!!!
 							}
 						}
+						if (!searchval[0]) {
```
tobiasschweizer: I do not understand why `searchval[0]`would be undefined after the switch-case since the default case statement adds an empty string.

loicjaouen: because in the case described here, we have a linked resource, so a `searchbox` case in the switch, thus the executed code is:
```
case 'searchbox': {
    searchval.push($(searchval_ele[index]).searchbox('value'));
``` 

where `searchval` gets pushed a undefined value (`searchval[0] = undefined`).

As a side note:  

It is interesting to know that if you happen to make a search, then the input searchbox gets a value, and remembers it.
Then you switch the search operator from `=` to `exists`, the searchbox disappear from the user's view, but `searchval` still receives it.

It is not quite a bug because it is harmless :)

tobiasschweizer: So couldn't you check `($(searchval_ele[index]).searchbox('value')` for `undefined` and add a `""` if so?

tobiasschweizer: you could actual use a ternary operator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

loicjaouen: *Check in the switch case?*
A test at the end makes that we don't have to add this on each of the switch cases (I didn't test if `searchbox` was the only problematic case).

*Use ternary operator?*
Sure, that would be possible even though style-wise, I don't like them that much when the value is the test case:
```js
searchval.push($(searchval_ele[index]).searchbox('value') ? $(searchval_ele[index]).searchbox('value') : "");
```
(I find it hard to read thus hard to maintain)

Don't you think so?

tobiasschweizer: Whatever you prefer :-)

However, you could also write: 

```
var givenSearchVal = $(searchval_ele[index]).searchbox('value');

var searchVal = givenSearchVal !== undefined ? givenSearchVal : "";

searchval.push(searchVal);
```

You could also make this more generic by handling only `givenSearchVal` in the cases and then doing the assignment after that.

```
var givenSearchVal;
switch (gui_element) {
    case 'text': {
        givenSearchVal  = $(searchval_ele[index]).val();
	break;
    }
   case 'date': {
      var dateObj = $(searchval_ele[index]).dateobj('value');
      givenSearchVal = SALSAH_API_LEGACY.make_date_string(dateObj);
     break;
    }
   ...
}

var searchVal = givenSearchVal !== undefined ? givenSearchVal : "";

searchval.push(searchVal);
```

loicjaouen: if it's my call, I prefer to be the least intrusive possible when fixing legacy code.  
I buy your "undefined" test as less error prone as we might have a boolean `false` some day.

loicjaouen: if `value` was not to return `undefined`m we might also change [jquery.searchbox.js#L385](https://github.com/dhlab-basel/Knora/blob/fix/search-exists/salsah/src/public/js/jquery.searchbox.js#L385) to do the test for undefined and return the empty string.

PR #670, opened 2017-11-20
Update ontology metadata and entities (part 1)

### LINKED ISSUE #698: Make input wrappers for processing ontology JSON-LD from both API input and test output
In #670, `CreatePropertyRequestV2` does the opposite of `ReadOntologiesV2`, i.e. it takes as input the JSON-LD that `ReadOntologiesV2` generates as output. There will be other input messages that need to process this same JSON-LD structure.

Also, it would be easier to write tests if we could take the JSON-LD output of `ReadOntologiesV2` and transform it into a similar case class instance for checking. For example, `OntologyV2R2RSpec` submits JSON-LD to create a property, and gets back essentially the same output. It would then like to compare the input with the output.

It wouldn't be possible to do all this in `ReadOntologiesV2`, which contains information from the ontology responder that isn't directly represented in JSON-LD, as well as `Content` objects.

The current I/O wrapper design calls for wrappers like `Read`, `Create`, and `Change`. But in the test use case, these names wouldn't apply, so we need something a bit more abstract. This could be called `InputOntologiesV2`, and would just contain the `Content` objects. Its companion object would have the `fromJsonLD` method.

The create property route could use `InputOntologiesV2.fromJsonLD` to process the input, and `CreatePropertyRequestV2` could then contain the resulting `InputOntologiesV2` instance. Additional error-checking would have to be done on the contents of the `InputOntologiesV2`, probably in the constructor of `CreatePropertyRequestV2`, e.g. to check that it's in the right schema, that contains exactly one property, that the property's type is `owl:ObjectProperty`, etc.

`OntologyV2R2RSpec` could convert its own JSON-LD request string into an `InputOntologiesV2`, and also convert the API server's response into an `InputOntologiesV2`. The relevant part of the contents of the two objects (an `InputPropertyV2`) could then be compared with `==`.

benjamingeer:
- [x] When creating an ontology, give it some metadata (currently just an `rdfs:label` and a timestamp).
- [x] Return ontology metadata from the triplestore.
- [x] Change an ontology's metadata.
- [x] Add a property to an ontology.
- [x] Change type of error functions in `StringFormatter` from `() => Nothing` to `=> Nothing`.
- [x] In API v2 with value objects, rename `IntegerValue` to `IntValue` for consistency with `knora-base`.
- [x] Make input wrappers for processing ontology JSON-LD from both API input and test output (closes #698).
- [x] Coordinate with SALSAH GUI (https://github.com/dhlab-basel/Salsah/pull/111)


tobiasschweizer:
coordination is needed with https://github.com/dhlab-basel/Salsah/tree/wip/coordination_knora_ontology-updates-6

-> ontologycache.service.ts

benjamingeer:
TODO:

* Fix remaining tests.
* In test that adds property, tell the ontology responder to reload, and check the results again.
* Add e2e tests.

tobiasschweizer:
Have you considered adding tests for `org/knora/webapi/util/jsonld/JsonLDUtil.scala` (parsing JSON-LD)?

benjamingeer:
> Have you considered adding tests for `org/knora/webapi/util/jsonld/JsonLDUtil.scala` (parsing JSON-LD)?

I considered it, but I didn't do it, because I wanted you to have the satisfaction of asking me to do it. :)

webapi/src/main/scala/org/knora/webapi/responders/v2/OntologyResponderV2.scala
```diff
@@ -1108,30 +1275,563 @@ class OntologyResponderV2 extends Responder {
...
+                    maybeLinkValuePropertyDef = maybeLinkValuePropDef,
+                    lastModificationDate = createPropertyRequest.lastModificationDate,
+                    currentTime = currentTime
+                ).toString()
+
+                _ <- (storeManager ? SparqlUpdateRequest(updateSparql)).mapTo[SparqlUpdateResponse]
+
+                // Check that the update was successful.
+
+                _ <- checkOntologyLastModificationDateAfterUpdate(internalOntologyIri = internalOntologyIri, expectedLastModificationDate = currentTime)
```
tobiasschweizer: Is this check for the last modification date sufficient? Or would you have to check in a more extensive manner?

I guess the whole transaction (`src/main/twirl/queries/sparql/v2/createProperty.scala.txt`) would have been rejected if there had been any error.

tobiasschweizer:
I think it would be good to have tests for that since this will be re-used for data. 

I have already added tests to the Angular GUI including mocked http service data. So I already had my dose today ;-)

webapi/src/main/scala/org/knora/webapi/responders/v2/OntologyResponderV2.scala
benjamingeer: Ideally, the property would be read back from the triplestore to check that it was stored correctly. But this would require refactoring the SPARQL that queries ontology data, as well as the code in `OntologyResponderV2` that reads the query results. It was never designed to read just one property or one class. I would like to do this later, changing the SPARQL queries to use CONSTRUCT instead of SELECT, and refactoring the Scala code to handle just one property or all the properties in the triplestore, similarly to what you did for reading resources in v2.

benjamingeer:
> I think it would be good to have tests for that since this will be re-used for data.

Yes, yes, I'll add them. :)

webapi/src/main/scala/org/knora/webapi/responders/v2/OntologyResponderV2.scala
tobiasschweizer: ok, that's fine for me

we can put that in an issue and do it later

https://de.wiktionary.org/wiki/morgen,_morgen,_nur_nicht_heute,_sagen_alle_faulen_Leute :-)

tobiasschweizer:
> Yes, yes, I'll add them. :)

great, thx. The related PR is done in the SALSAG GUI.

benjamingeer:
I added a test for `JsonLDUtil`. Is this OK now?

tobiasschweizer:
I am listening to very loud music, I cannot hear you

benjamingeer:
It would be sad if you went deaf, then I wouldn't be able to annoy you so easily anymore.

tobiasschweizer:
when the tests are finished, you can merge this PR. I will then merge the GUI PR.

Issue #597, opened 2017-08-30
When clicking on "save" to create a resource, there is no sign that Salsah 1.5 is processing the request 


mrivoal:
When clicking on "save" to create a resource, there is no sign that Salsah is processing the request. 

As the processing can take some times (up to several seconds and sometime several tens of seconds... Performance issue that we need to talk about), **the user doesn't know wether Salsah is processing his request or not** and of course it is very confusing for him.

Something has to be done to let him know that something is happening.

loicjaouen:
also happens when adding a sub-resource while creating the main resource or editing an existing one:
<img width="279" alt="Capture d’écran 2019-11-12 à 11 36 35" src="https://user-images.githubusercontent.com/3933654/68664573-d783b400-0540-11ea-9ce4-1cd291d626d1.png">

this "New entry..." pops-up a window with a save button that does not prevent multiple clicks

loicjaouen:
and with multiple clicks come duplicate entries :(

PR #674, opened 2017-11-23
Fix/salsah 1.5 processing sign

loicjaouen:
fixes: #597


loicjaouen:
@tobiasschweizer : would you mind reviewing this, please? it is short and quite simple, reusing `simpledialog` to add a turning wheel telling the user that the system is processing when they call `save` because it causes users to worry, freak out and click 10 times or reload.

loicjaouen:
exactly

loicjaouen:
thanks :)

PR #607, opened 2017-09-12
Echo filenames as test data is being loaded

benjamingeer:
Because of a recent update to the RDF4J console, when a script such as `graphdb-se-local-init-knora-test.sh` is loading Knora test data, the name of each TTL file is no longer echoed to the terminal. This changes those scripts to use an [expect](https://linux.die.net/man/1/expecthttps://linux.die.net/man/1/expect) script that echoes each command that it sends to the RDF4J console, thus imitating the old behaviour of the console.

PR #682, opened 2017-12-08
feature (webapi): add basic Webern test data

musicEnfanthen:
This small PR adds some basic Webern test data to the webapi (`graphdb-knora-test.expect`) after @subotic was so kind to explain where the necessary steps have to be taken.

It includes
- creation of `webern-onto.ttl`, 
- creation of `webern-data.ttl`,
- additions of project and test user to `admin-data.ttl` and 
- additions of permissions to `permissions-data.ttl`

Everything is loaded to graphdb without errors here.

Also, I added a new folder `queries` inside `webapi` to store SPARQL query files that can be used to test the data. For now, there are two very basic queries for the Webern data.

PR #683, opened 2017-12-08
feature (webapi): allow loading data into graphdb on windows

musicEnfanthen:
In order to get data loaded into graphdb on windows there are some small changes necessary to the load files in `webapi/scripts`: 

Because getting `Expect` run as executable under `usr/bin/expect` on windows is kind of hell, instead of `.expect` file I adapted the load script reusing the former `cat xxx.ttl | $CONSOLE` command that was removed in #607  . (Thanks to @subotic for pointing me to this solution.)
In accordance to this I also recreated the former `graphdb-knora-test-data.ttl` file. 

Also a small adaption to classpath in console.sh was necessary due to windows use of another file seperator (;) instead of (:).

I named the new files with `win-...`, so none of the existing POSIX files are touched.


musicEnfanthen:
depends on #682

musicEnfanthen:
is superseded by PR #814

PR #704, opened 2018-01-12
Add support for LIKE and MATCH in KnarQL Text Search

tobiasschweizer:
- [x] MATCH: exact match of a word or expression using Lucene. However, in KnarQL we need to use a virtual property that gets converted to whatever the triplestore needs.
- [x] LIKE: use of REGEX 
- [x] add tests

This PR adds support for the regex function. 

As a temporary solution, it also adds support for `contains`. However, this is turned into a statement   involving the special Lucene property for graphdb (the FILTER does not end up in the SPARQL executed by the triplestore). 

I will refactor that later (see #706). #705 introduces a more generic handling of function calls and can be integrated too when doing the refactoring (however, we should have a list of functions that we actually support).


tobiasschweizer:
Consider defining a custom function for "match": http://docs.rdf4j.org/custom-sparql-functions/

otherwise we would need to use a special property like shown her:

```

    
        CONSTRUCT {
        
            ?mainRes knora-api:isMainResource true .
            
            ?mainRes <http://0.0.0.0:3333/ontology/incunabula/simple/v2#title> ?propVal0 .
        
        } WHERE { 
        
            ?mainRes a knora-api:Resource .
            
            ?mainRes a <http://0.0.0.0:3333/ontology/incunabula/simple/v2#book> .
            
            
                        ?mainRes <http://0.0.0.0:3333/ontology/incunabula/simple/v2#title> ?propVal0 .
                        <http://0.0.0.0:3333/ontology/incunabula/simple/v2#title> knora-api:objectType <http://www.w3.org/2001/XMLSchema#string> .
                        ?propVal0 a <http://www.w3.org/2001/XMLSchema#string> .
                    
                        ?propVal0 knora-api:textMatches "Land" . 
                
                        knora-api:textMatches knora-api:objectType <http://www.w3.org/2001/XMLSchema#string> .
            
        }
        
       
``` 

This is not consistent with the idea of the simple api since `?propVal0` is conceptually a literal and not an object

benjamingeer:
Is there a test for `contains`?

tobiasschweizer:
yeah, you got me this time ;-)

tobiasschweizer:
here you go: 785d24d7c1da77e7800699d82b485935c71b7dbf

I added an e2e test for the contains function. I chose "Zeitglöcklein" to make you even happier.

I did not add a test for the parser since this will be changed in the close future (#705).

tobiasschweizer:
I have tears in my eyes

Zeitglöcklein der Seele: https://books.google.ch/books?id=qW_LYgDINo4C&pg=PA153&lpg=PA153&dq=inkunabel+zeitglöcklein+der+seele&source=bl&ots=NBBumG5t9G&sig=QU8Ci6_LeySRoQ4ySHsALQmMPXU&hl=de&sa=X&ved=0ahUKEwjU0Zi-8oHZAhXSCuwKHRfzCAoQ6AEIKDAA#v=onepage&q=inkunabel%20zeitglöcklein%20der%20seele&f=false

PR #705, opened 2018-01-15
Accept function calls in filters in KnarQL

benjamingeer:
This allows `SearchParserV2` to accept function calls in filters in KnarQL. A function call is represented as a `FunctionCallExpression`, which has an IRI identifying the function and a sequence of `Entity` arguments.

### DIFF
--- a/webapi/WebapiBuild.sbt
+++ b/webapi/WebapiBuild.sbt
@@ -91,7 +91,7 @@ lazy val webApiCommonSettings = Seq(
     name := "webapi",
     version := "0.1.0-beta",
     ivyScala := ivyScala.value map { _.copy(overrideScalaVersion = true) },
-    scalaVersion := "2.12.1"
+    scalaVersion := "2.12.3"
 )
 
 lazy val webApiLibs = Seq(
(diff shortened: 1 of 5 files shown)


tobiasschweizer:
We should have a list of functions that we actually support, so we can check. However, I do not think that the parser has to do that. It may happen in the responder.

benjamingeer:
@tobiasschweizer Do you still want this PR?

tobiasschweizer:
Yes, I do. I would like to consider it once I can do the refactoring  #706

tobiasschweizer:
I think we need to consider the changes I have done for the `contains` function so KnarQL still works as it is (so we do not need to change SALSAH). But I guess that the way you did it for this PR is more generic and preferable to what I did to support contains (`develop`).

I can surely work on it, but at the moment I am stuck with the standoff refactoring. I could probably do it next week (before my holiday). @benjamingeer @SepidehAlassi Would that be ok for you?

benjamingeer:
It's OK with me if it's OK with @SepidehAlassi .

SepidehAlassi:
Fine with me!

PR #707, opened 2018-01-16
Update ontology entities (part 2)

### LINKED ISSUE #104: How to handle Consistency Checking when the Ontology is changed
Assuming we have an admin interface that allows for changing the ontology, how would we handle consistency checking?

> Example: A user defines an ontology using the admin interface and creates data that conforms to that ontology. He creates several instances of a resource class and now recognizes that he forgot to add a required property both to the ontology and the data. If he adds the new property to the ontology and marks it as required (e.g. cardinality = 1), the consistency of the data would be violated. Instead, he would have to add the property as an option to the resource class, add the property to all of the instances of the resource class and can then mark it as required in the ontology.

To which degree could/should this be automated?

### LINKED ISSUE #157: Ontology responder should give clearer error message when cardinality cannot be parsed
When a cardinality contains a malformed integer value, like this (the space after the `1` makes it invalid):

```
owl:minCardinality "1 "^^xsd:nonNegativeInteger
```

The ontology responder should throw a helpful exception instead of `java.lang.NumberFormatException`.

### LINKED ISSUE #242: Make sure that a property is defined in the ontology
When you define a cardinality for a property on a resource class, we do not make sure that this property is actually defined.

Can we make the consistency checker check that every reference to a property can be resolved?

### LINKED ISSUE #286: Check ontologies for validity
We need an ontology checker. It's too easy to make small mistakes in ontologies that produce strange results and are difficult to diagnose. Common cases:
- You use an incorrect prefix definition.
- You forget to make your property a subproperty of `knora-base:hasValue` or `knora-base:hasLinkTo`.
- You write `rdfs:subClassOf :Resource` instead of `rdfs:subClassOf knora-base:Resource`.
- Class A is a subclass of B, which is a subclass of A: this causes an infinite recursion in the ontology responder.
- A property has a `knora-base:objectClassConstraint` of `xsd:string`.
- A resource class has a cardinality for a property whose `knora-base:subjectClassConstraint` is incompatible with that class.
- You give a class and a property the same name.

This check needs to be done:
- In a command-line program, so people can check their ontologies while developing them, without having to run the Knora API server.
- In the ontology responder, to verify all ontologies when the Knora API server starts.

Perhaps this could be done in Scala using the [OWL API](https://github.com/owlcs/owlapi).

See also:
- #242
- #157
- #104

### LINKED ISSUE #579: Updating of (project) ontologies
For the new Salsah we need a post method for the changes users make when they edit resourceTypes.

### LINKED ISSUE #603: Implement ontology updates
This needs to:

* Create an empty ontology when a project is created
* Add a property definition to an ontology
* Add a class definition to an ontology
* Change or delete a property definition if it's not used in classes or in data
* Change or remove a cardinality if it's not used in data
* Change a property or class label

This is for API v2. It will use the `knora-api` ontology and JSON-LD.

Class definitions can specify the order of properties (see #601).

### LINKED ISSUE #710: Two warnings while compiling webapi
I get the following warnings with the newest version from develop branch (2018-01-25):

`[warn] Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[warn]
[warn] 	* com.github.jsonld-java:jsonld-java:0.10.0 is selected over 0.5.1
[warn] 	    +- org.apache.jena:jena-arq:3.0.0                     (depends on 0.5.1)
[warn] 	    +- org.knora:webapi_2.12:0.1.0-beta                   (depends on 0.5.1)
[warn]
[warn] Run 'evicted' to see detailed eviction warnings
[info] Compiling 294 Scala sources to [...]/knora/webapi/target/scala-2.12/classes...
[warn] [...]/knora/webapi/src/main/scala/org/knora/webapi/responders/admin/GroupsResponderADM.scala:88: non-variable type argument org.knora.webapi.messages.store.triplestoremessages.LiteralV2 in type pattern scala.collection.immutable.Map[String,org.knora.webapi.messages.store.triplestoremessages.LiteralV2] (the underlying of Map[String,org.knora.webapi.messages.store.triplestoremessages.LiteralV2]) is unchecked since it is eliminated by erasure
[warn]                 case (groupIri: IRI, propsMap: Map[String, LiteralV2]) =>
[warn]                                                ^
[warn] [...]/knora/webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala:110: non-variable type argument org.knora.webapi.messages.store.triplestoremessages.LiteralV2 in type pattern scala.collection.immutable.Map[org.knora.webapi.IRI,org.knora.webapi.messages.store.triplestoremessages.LiteralV2] (the underlying of Map[org.knora.webapi.IRI,org.knora.webapi.messages.store.triplestoremessages.LiteralV2]) is unchecked since it is eliminated by erasure
[warn]                 case (userIri: IRI, propsMap: Map[IRI, LiteralV2]) =>
[warn]                                               ^
[warn] two warnings found`

benjamingeer:
- [x] Change a property's label or comment.
- [x] Create a class.
- [x] Change a class's label or comment.
- [x] Add cardinalities to a class if it is not used in data.
- [x] Replace a class's cardinalities with new ones if it is not used in data.
- [x] Delete a property if it is not used in data or ontologies.
- [x] Delete a class if it is not used in data or ontologies.
- [x] After updating an ontology entity, reload it from the triplestore to check that it was saved correctly.
- [x] Document the ontology API.

If an ontology is created using this API, the ontology inconsistencies discussed in #286 are prevented.

I had to:

* Refactor the loading and caching of ontologies in `OntologyResponderV2`.
* Add support for blank nodes as subjects in `SparqlExtendedConstructResponse`, because we need them to read cardinalities from a class definition using `CONSTRUCT`.
* Undo the SPARQL-escaping of the input after an update, so the input can be compared to the data that was saved in the triplestore.

Resolves #603.
Resolves #579.
Resolves #286.
Resolves #242.
Resolves #157.
Resolves #104.
Resolves #710.


benjamingeer:
TODO: prevent these from #286:

* Class A is a subclass of B, which is a subclass of A: this causes an infinite recursion in the ontology responder.
* You give a class and a property the same name.

benjamingeer:
TODO: allow adding an optional cardinality to a class even if it's used in data.

benjamingeer:
@kilchenmann I think this is ready to use now. Could you please read the documentation in `docs/rst/knora-api-server/api_v2/ontology-information.rst` and see if it's clear and complete enough?

benjamingeer:
@kilchenmann Bon voyage!

benjamingeer:
@tobiasschweizer There is a failing test in `SearchRouteV2R2RSpec`, and it doesn't make sense to me, could you please have a look?

The failing test is `should search for an anything:Thing that may have a Boolean value that is true`. It expects 9 results, but gets 10 on this branch. The KnarQL says:

```
PREFIX anything: <http://0.0.0.0:3333/ontology/anything/simple/v2#>

CONSTRUCT {

     ?thing a anything:Thing .

     ?thing anything:hasBoolean ?boolean .
} WHERE {

     ?thing a anything:Thing .
     ?thing a knora-api:Resource .

     OPTIONAL {

         ?thing anything:hasBoolean ?boolean .
         anything:hasBoolean knora-api:objectType xsd:boolean .

         ?boolean a xsd:boolean .

         FILTER(?boolean = true)
     }
} OFFSET 1
```

If I compare the actual response to the expected one, the extra result is:

```
{
    "@id" : "http://data.knora.org/nResNuvARcWYUdWyo0GWGw",
    "@type" : "anything:Thing",
    "schema:name" : "X-ray"
}
```

But this doesn't make sense to me, because that resource doesn't have the property `anything:hasBoolean`. In fact, most of the resources in the search result don't have that property. Is this a bug in the KnarQL implementation?

webapi/src/test/resources/test-data/searchR2RV2/ThingWithBooleanOptional.jsonld
```diff
@@ -1,6 +1,10 @@
 {
   "@type" : "schema:ItemList",
   "schema:itemListElement" : [ {
+    "@id" : "http://data.knora.org/nResNuvARcWYUdWyo0GWGw",
+    "@type" : "anything:Thing",
```
tobiasschweizer: I like things.

docs/rst/knora-api-server/api_v2/ontology-information.rst
```diff
@@ -18,10 +18,1077 @@
...
+
+A successful response will be a JSON-LD document containing ``knora-api:hasOntologies``,
+providing only the ontology's metadata.
+
+Changing an Ontology's Metadata
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Currently, the only modifiable ontology metadata is the ontology's ``rdfs:label``.
+
+::
```
tobiasschweizer: ``knora-api:hasOntologies`` is not used here either (see "Creating a New Ontology")

benjamingeer:
Review tasks from @tobiasschweizer:

* When creating an entity, make sure its local name is validated
* Use `hasOntologies` to update ontology metadata
* In tests, if one ontology is expected, check that there's only one
* make a separate test for `InputOntologiesV2.fromJsonLD`
* When updating an entity, check that it's in the same ontology as the one specified in `hasOntologies`
* In `addCardinalitiesToClass.scala.txt`, check that the class exists
* Add a boolean `isValueClass` to value classes in `knora-api` classes in the default schema

benjamingeer:
@tobiasschweizer Can I merge this now?

tobiasschweizer:
I will have another look at it just after lunch

tobiasschweizer:
@benjamingeer  Can this branch be deleted?

benjamingeer:
Yes thanks!

tobiasschweizer:
I like deleting things. I believe you do too.

PR #720, opened 2018-02-02
build (docker): add utf-8 environment variable

### LINKED ISSUE #717: [SALSAH] Icon name with unmappable characters
```
java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters: /src/src/public/project-icons/Lumi??res.Lausanne
```

Could we please get rid of the special characters? Java doesn't like it.

subotic:
- resolves #717


subotic:
thanks 😄

PR #732, opened 2018-02-12
feature (webapi): version specific exception handlers

### LINKED ISSUE #638: Make separate exception handlers for API v1 and v2
Currently there is just one `KnoraExceptionHandler` for v1 and v2, which returns a plain JSON response containing a v1 API status code. For v2, it would be better to get rid of the status code and return JSON-LD.

subotic:
### Summary

- update some libraries
- [x] extend knora exception handler to answer `v1` and `v2` request exceptions differently 
- [x] url.startsWith("/v1") to make it safer
- [x] separate exception handlers for `v1`, `v2`, and `admin`

### Issues

- resolves #638


subotic:
@benjamingeer you can review whenever you like :-)

webapi/src/main/scala/org/knora/webapi/KnoraExceptionHandler.scala
```diff
@@ -20,12 +20,47 @@ object KnoraExceptionHandler {
...
         /* TODO: Find out which response format should be generated, by looking at what the client is requesting / accepting (issue #292) */
 
-        case rre: RequestRejectedException => complete(exceptionToJsonHttpResponse(rre, settingsImpl))
+        case rre: RequestRejectedException =>
+            extractRequest { request =>
+                val url = request.uri.path.toString
+
+                // println(s"KnoraExceptionHandler - case: rre - url: $url")
+
+                if (url.contains("v1")) {
```
benjamingeer: What if the URL is `/v2/resources/http://foo.bar/a3dbfv1dedc`? I think you could use `url.startsWith("/v1")` to make it safer.

benjamingeer: Also, should the admin routes have their own exception handler?

webapi/src/main/scala/org/knora/webapi/KnoraExceptionHandler.scala
```diff
@@ -20,12 +20,47 @@ object KnoraExceptionHandler {
...
+
+        case ise: InternalServerException =>
+            extractRequest { request =>
+                val uri = request.uri
+                val url = uri.path.toString
+
+                // println(s"KnoraExceptionHandler - case: ise - url: $url")
+                log.error(ise, s"Unable to run route $url")
+
+                if (url.contains("v1")) {
```
benjamingeer: Same as above.

webapi/src/main/scala/org/knora/webapi/KnoraExceptionHandler.scala
subotic: > should the admin routes have their own exception handler

Probably not. At least, as long both are only returning normal JSON.

subotic:
@benjamingeer this is again ready for review. thanks.

subotic:
Thanks :-)

> Someday we'll have to change this when we have /v12 or v21, but I guess it's OK for now.

Your grandchildren can do it then :-)

PR #737, opened 2018-02-13
Use salsah-gui:guiOrder in cardinalities instead of in property definitions

### LINKED ISSUE #601: Replace salsa-gui:guiOrder with an annotation on resource classes
The `salsah-gui:guiOrder` property predicate affects all uses of the property it's attached to, but it would make more sense to specify property order per resource class. We can do this by attaching an annotation to the resource class, containing a comma-separated list of property IRIs.

cc @kilchenmann @sofiag

### LINKED ISSUE #744: salsah-gui:guiAttributeDefinition is undocumented and inconsistent
As part of #737, I'm writing documentation on `salsah-gui`, and I've noticed some things that seem inconsistent.

### guiAttributeDefinition

It looks like `guiAttributeDefinition` defines the types of attributes that are accepted by each `Guielement`. But:

* `Pulldown` has `guiAttributeDefinition "selection=api:lists"`, and `List` has no `guiAttributeDefinition`. In the test ontologies, both `Pulldown` and `List` have `salsah-gui:guiAttribute "hlist=..."`. What is `api:lists`? What is the difference between `Pulldown` and `List`?
* `Radio` has `guiElement "selection=api:list"`. I guess this should be `guiAttributeDefinition`? Is `api:list` different from `api:lists` (which is used in `Pulldown`), or is it a typo?
* `Searchbox` has `guiAttributeDefinition` `"resource=api:resources"`. What is `api:resources`?
  * In `images-onto.ttl`, there are properties that use `Searchbox` and have `guiAttribute` values like `"restypeid=33"`. That can't be right.
  * In `dokubib-ontol.ttl`, there's no `restypeid`, but there is `"numprops=1"`. What does that mean?

### guiElement and guiAttribute in the ontology update API

The ontology update API should allow the user to attach `guiElement` and `guiAttribute` to property definitions, right? In that case, it should check that the `guiAttribute` values provided are valid for the `guiElement` chosen. This means it will have to parse the objects of `guiAttributeDefinition`. And this means that those objects need a clear syntax that's consistent with existing ontologies.

benjamingeer:
Overview:

* Attach `salsah-gui:guiOrder` to cardinalities in class definitions, instead of attaching it to property definitions. This allows the same property to be first in one class and last in another class.
* Validate the objects of `salsah-gui:guiAttribute`.

Tasks:

- [x] Manually update existing test ontologies (except for `drawings-gods_ontology.ttl`) so they use `salsah-gui:guiOrder` on cardinalities rather than on properties.
- [x] Provide a command-line program that updates `salsah-gui:guiOrder` in an existing ontology, and use it to update `drawings-gods_ontology.ttl`, which is too large to update manually.
- [x] Make existing functionality work in API v1 and v2.
- [x] Show `salsah-gui:guiOrder` in cardinalities in class definitions returned by the API v2 ontology routes.
- [x] Allow the client to specify `salsah-gui:guiOrder` when updating cardinalities via API v2.
- [x] Redesign `salsah-gui:guiAttribute` and `salsah-gui:guiAttributeDefinition` as a DSL parsed by `StringFormatter`, and use the attribute definitions to validate the attributes when loading ontologies from the triplestore.
- [x] Fix incorrect uses of `salsah-gui:guiAttribute` in test ontologies.
- [x] Document the `salsah-gui` ontology (reorganising and updating the Knora ontology docs a bit) and the rest of the above.
- [x] Update release notes.

TODO in a subsequent pull request:

- [ ] Show `salsah-gui:guiElement` and `salsah-gui:guiAttribute` when serving ontologies in API v2.
- [ ] Allow `salsah-gui:guiElement` and `salsah-gui:guiAttribute` to be updated via API v2.

Resolves #601.
Resolves #744.


musicEnfanthen:
This is a move highly appreciated, thanks for that! 

Would it make sense to move also the `salsah-gui:guiElement` to the cardinality restrictions in class definitions, so the property definitions are "free" of gui specific values? (Having a certain gui representation should’nt be considered a _semantic_ characteristic of a property.) Thus, one could also have a property to be `richtext` in one class or just `simpletext` in another class, if desired.

mrivoal:
@musicEnfanthen I can see your point, but I think it is a bit less relevant than having the `salsah-gui:gui-order` defined in class definition. And it would make the process of defining a resource very time-consuming if for a property used in multiple resource you have to specify for each resource the `salsah-gui:guiElement` of a given property. 

I would rather define once and for all that a `title` in a given ontology is a `salsah-gui:guiElement salsah-gui:SimpleText`.

benjamingeer:
If we support `salsah-gui:guiElement` in cardinalities, we'd have to support `salsah-gui:guiAttribute` there, too. I could imagine supporting both options. Perhaps SALSAH 2 could use the values on the cardinality if provided, otherwise it would look for ones on the property itself.

But putting more stuff in cardinalities is awkward with the current design in the ontology responder. If I do that, I'd like to do it later as part of #738.

benjamingeer:
@subotic you can review whenever you like :-)

docs/rst/02-knora-ontologies/knora-base.rst
```diff
@@ -443,7 +336,7 @@ Properties of Value
     A resource may have several properties of the same type with
     different values (which will be of the same class), and it may be
     necessary to indicate an order in which these values occur. For
-    example, a book may have several authors which should appear in a
+    example, a book may have several authors which should appear in af
```
subotic: typo at `af`?

docs/rst/02-knora-ontologies/knora-base.rst
```diff
@@ -970,7 +863,7 @@ Knora's built-in support for “standoff” markup, which is stored
 separately from the text. This has some advantages over embedded markup
 such as XML. [#]_ While XML requires markup to have a hierarchical
 structure, and does not allow overlapping tags, standoff nodes do not
-have these limitations (Schmidt2016_). A standoff tag can be attached to
+have these limitations. [Schmidt2016]_ A standoff tag can be attached to
```
subotic: Shouldn't the link be before the period?

benjamingeer:
> While reading the salsah-gui.rst it seems to me that this ontology is very Salsah-V1 specific.

How so? Won't SALSAH 2 also use HTML elements for input?

> Wouldn't something like this :guiAttributeDefinition "hlist(required):iri" be better broken down into individual properties?

What do you mean by individual properties? Can you give some examples?

All I tried to do here was document and enforce the conventions that are already used in project-specific ontologies. Keep in mind that if we change those conventions, we also have to change all project-specific ontologies.

docs/rst/02-knora-ontologies/knora-base.rst
benjamingeer: Yep.

docs/rst/02-knora-ontologies/knora-base.rst
benjamingeer: It's a footnote, so I just tried to format it like the other footnotes in that document. Apparently it was syntactically invalid before, but older versions of Sphinx never complained. When I upgraded Sphinx, it told me there was an error on that line. That's the only reason I changed it.

subotic:
I was just thinking out loud. Nothing that we should change now.

> What do you mean by individual properties? Can you give some examples?

Something like this:
```
:hasListItem rdf:type owl:ObjectProperty ;
   salsah-gui:guiElement salsah-gui:List ;
   salsah-gui:guiHlistAttribute <http://rdfh.ch/lists/xyz> .
```

This would be somewhat safer, but would also require specifying a lot more properties in a lot more detail. Also, as you said, we would need to change a lot of existing data. This is probably to much.

benjamingeer:
> This would be somewhat safer

I agree, and I thought about this too, but I think it just isn't worth going to the trouble to change everyone's ontologies now. At least with this PR, the attributes are checked for validity.

webapi/src/test/resources/test-data/ontologyR2RV2/bildSimple.json
```diff
@@ -12,79 +12,93 @@
         "rdfs:subClassOf" : [ "http://api.knora.org/ontology/knora-api/simple/v2#StillImageRepresentation", {
           "@type" : "owl:Restriction",
           "owl:cardinality" : 1,
-          "owl:onProperty" : "http://0.0.0.0:3333/ontology/00FF/images/simple/v2#bearbeiter"
+          "owl:onProperty" : "http://api.knora.org/ontology/knora-api/simple/v2#creationDate"
```
subotic: Just a general question, where is this URL defined? I didn't find it in `application.conf`.  I don't think that this a good idea, as this would mean, that we would need to have something running or serving at this address? Why not have it like this: `http://0.0.0.0:3333/ontology/knora-api/simple/v2#creationDate"?

benjamingeer: This is documented here:

http://www.knora.org/documentation/manual/rst/03-knora-api-server/api_v2/knora-iris.html#external-ontology-iris

The idea is that we will have a Knora instance serving the built-in ontologies at that address.

If it was `http://0.0.0.0:3333/ontology/knora-api/simple/v2#creationDate`, then the `knora-api` ontology running on project-specific server A would not appear to be the same as the `knora-api` ontology running on project-specific server B. From a client's point of view, they would not be the same ontology so would not be interoperable.

benjamingeer:
I would love to run the SALSAH GUI tests but for that I need a working Sipi. I seem to remember there was a nice Docker image for it, but I can't seem to find it now, is there some documentation I could use?

webapi/src/test/resources/test-data/ontologyR2RV2/bildSimple.json
subotic: Would it be generally possible to "dump" the `knora-api` ontologies to files? I'm thinking of an easy way to serve all our ontologies. If all the ontologies would be available as files, then it wouldn't be difficult to serve them.

subotic:
I personally always go to `.travis.yml` for the documentation:

```
# Salsah browser tests
    #- stage: test
    #  script:
    #    # prepare needed graphdb-se files
    #    - mkdir -p $TRAVIS_BUILD_DIR/data/conf
    #    - cp $TRAVIS_BUILD_DIR/private/graphdb.license $TRAVIS_BUILD_DIR/data/conf/graphdb.license
    #    - mkdir -p $TRAVIS_BUILD_DIR/graphdb
    #    - cp $TRAVIS_BUILD_DIR/webapi/scripts/KnoraRules.pie $TRAVIS_BUILD_DIR/graphdb
    #    # start and initialize graphdb-se
    #    - docker run -d -p 127.0.0.1:7200:7200 -v $TRAVIS_BUILD_DIR/data:/opt/graphdb/home -v $TRAVIS_BUILD_DIR/graphdb:/graphdb ontotext/graphdb:8.2.0-se
    #    - sleep 15
    #    - cd $TRAVIS_BUILD_DIR/webapi/scripts && ./graphdb-se-docker-init-knora-test.sh
    #    # start the sipi container (with test config)
    #    - docker run -d --add-host webapihost:$DOCKERHOST -v /tmp:/tmp -v $HOME:$HOME -p 1024:1024 $SIPI_REPO:develop /sipi/local/bin/sipi --config=/sipi/config/sipi.knora-test-docker-config.lua
    #    # start the webapi and salsah server
    #    - cd $TRAVIS_BUILD_DIR/webapi/ && sbt "run allowReloadOverHTTP" &
    #    - cd $TRAVIS_BUILD_DIR/salsah/ && sbt run &
    #    # wait for the webapi server to start before we proceed with running the salsah tests
    #    - until netstat -an 2>/dev/null | grep '3333.*LISTEN'; do true; done
    #    # unzip chromedriver
    #    - cd $TRAVIS_BUILD_DIR/salsah/lib/chromedriver/ && unzip chromedriver_linux64.zip
    #    # start built-in salsah browser tests
    #    - cd $TRAVIS_BUILD_DIR/salsah/ && sbt headless:test
```
 I should probably add this to the Knora documentation.

subotic:
for sipi, this is the relevant line:
```
$ docker run -d --add-host webapihost:$DOCKERHOST -v /tmp:/tmp -v $HOME:$HOME -p 1024:1024 $SIPI_REPO:develop /sipi/local/bin/sipi --config=/sipi/config/sipi.knora-test-docker-config.lua
```

`$DOCKERHOST` would be the IP address of your machine.

webapi/src/test/resources/test-data/ontologyR2RV2/bildSimple.json
benjamingeer: The `knora-api` ontologies are already served by the Knora API server. You can request them from the API, download them, save them in files. That's what I did to make the test data for `OntologyV2R2RSpec`. It uses these files:

* The `knora-api` v2 ontology in the simple schema: https://github.com/dhlab-basel/Knora/blob/develop/webapi/src/test/resources/test-data/ontologyR2RV2/knoraApiOntologySimple.json
* The `knora-api` v2 ontology in the default schema: https://github.com/dhlab-basel/Knora/blob/develop/webapi/src/test/resources/test-data/ontologyR2RV2/knoraApiWithValueObjects.json

benjamingeer:
Cool, thanks!

> I should probably add this to the Knora documentation.

That would be great. :)

webapi/src/test/resources/test-data/ontologyR2RV2/bildSimple.json
benjamingeer: Currently the API server only serves ontologies in JSON-LD. What I've been planning to do is get the API server to take an HTTP `Accept` header and then be able to convert the JSON-LD to Turtle, XML, or HTML.

subotic: ok, great.

The reasoning behind me asking, is that (as far as I understand) all our ontologies aren't allowed to change (once they are published). So no reason to require a full running Knora stack, just to serve basically static content. Makes the life on the operations side much easier. Just keep in mind, that we don't (and probably never will) have a dedicated operations team. The less dependency there is on a running Knora stack, the less will break when the server is down.

Maybe then changing it to something like `http://www.knora.org/ontology/knora-api/simple/v2#creationDate` would allow them to be published together with the other Knora ontologies.

benjamingeer: We could change it to `www.knora.org`. Or we could just have something like `nginx` serve flat files at `api.knora.org/ontology`.

subotic: I had the idea to put them on the `www.knora.org` site served from Github. With Nginx (or whatever), we would need to put them on our servers. We will see. I have opened an issue ##753 for this. I need to think this a bit trough. Anyway, we have some time until V2 is officially released.

Thanks for bearing with me :-)

subotic:
Did you manage to get `sipi` working?

subotic:
When I finish PR #742 it should be easy to run the whole stack.

webapi/src/test/resources/test-data/ontologyR2RV2/bildSimple.json
benjamingeer: With the current design, we can't use `www.knora.org` for external (API v2) ontologies, because it's reserved for internal ontologies (the ones used in the triplestore):

http://www.knora.org/documentation/manual/rst/03-knora-api-server/api_v2/knora-iris.html#internal-ontology-iris

`SmartIriImpl` uses the hostname to determine whether the IRI is internal or external.

We could redesign all this, but I'd really rather not. :)

> Thanks for bearing with me :-)

No problem. :)

benjamingeer:
I ran out of time to try Sipi this week. Will try on Monday.

webapi/src/test/resources/test-data/ontologyR2RV2/bildSimple.json
benjamingeer: Also, I don't think we should rely on GitHub to serve Knora ontologies. That would mean that if GitHub is down, the DaSCH is down. Also, my understanding is that as critical research infrastructure for Switzerland, the DaSCH is supposed to be hosted in Switzerland.

benjamingeer: > The less dependency there is on a running Knora stack

The DaSCH can't exist without a running Knora stack.

subotic: > We could redesign all this, but I'd really rather not. :)

ok, we should then leave it as it is.

>> The less dependency there is on a running Knora stack

> The DaSCH can't exist without a running Knora stack.

What I meant by this is, if our Knora stack is down for a few hours (which is acceptable by DaSCH's / Lukas's standard) then this shouldn't have a negative impact on all other running Knora instances around Switzerland / World because they all depend on our served ontologies.

That's why I would like to see a separation of the serving of ontologies from the Knora stack and implement the serving of instances in a lightweight and highly available (redundant) way. Nginx serving flat files is a much simpler and lightweight (read costs less) setup, then the whole Knora stack.

We will host the DaSCH Knora stack with Switch, but not completely in a high availability way. The data will have a redundancy of 6 with data in Zürich and Lausanne (losing data is not acceptable), but the machine running the stack will not be redundant (few hours of downtime acceptable).

I would love to have everything highly redundant so that we don't have a minute of downtime, but the budget constraints are what they are. Well, maybe someday this will change.

benjamingeer: OK! I like flat files.

benjamingeer:
I fixed the SALSAH GUI tests by adding a half-second delay to the login method. :)

benjamingeer:
@subotic thanks for the review!

PR #745, opened 2018-02-18
feature (webapi): add monitoring settings (part 1)

subotic:
### Summary

- [x] Add all necessary files to be able to start the monitoring stack using docker-compose.

#### Usage:

 1. Start `webapi` with the necessary `-p` option (e.g., from inside sbt: `run -p`
 2. Start the monitoring stack by executing the following line inside the `monitoring` folder:
```
$ WEBAPIHOST=<YourLocalIP> ADMIN_USER=admin ADMIN_PASSWORD=admin docker-compose up -d
```
 3. Head over to `localhost:3000`, log in using the admin username and password, and open the kamon akka dashboard.


benjamingeer:
I tried this and opened the Grafan [Webapi Akka Actor System](http://localhost:3000/dashboard/db/webapi-akka-actor-system?refresh=1s&orgId=1) dashboard. I'm making requests to webapi in another browser tab, but all the graphs just say "no data points", and everything is at 0, as if nothing was happening. The Grafana dashboard doesn't seem to notice that the application is actually processing messages. Am I doing something wrong?

Also, how do I stop the application?

subotic:
No, you are not doing anything wrong. I didn't manage to get it working either. When you go to `localhost:9095` (the port of the Prometheus reporter), then you can see that data is generated. That's the reason for the part 1 designation. I still need to customize the Grafana Dashboard, so that only the metric interesting to us is shown.

If you run the app with `run`, then you will need to `ctrl-c` out of SBT. If you used `reStart`, then you can use `reStop`.

subotic:
ah, the stack, yes.

subotic:
added `docker-compose down` to the docs.

subotic:
thank you :-)

PR #755, opened 2018-02-27
Include guiElement and guiAttribute in API v2

### LINKED ISSUE #746: Include guiElement and guiAttribute in API v2
* Show `salsah-gui:guiElement` and `salsah-gui:guiAttribute` when serving ontologies in API v2.
* Allow `salsah-gui:guiElement` and `salsah-gui:guiAttribute` to be updated via API v2.

### LINKED ISSUE #760: Does the distinction between hasProperties and hasStandoffProperties make sense?
In the JSON-LD returned by the ontology routes, some properties go in `hasProperties` (if they’re used on resources or values), and others go in `hasStandoffProperties` (if they’re used on standoff tags). But what if a property is used both on values and on standoff tags (like `valueHasUri`)? Where should it go? Should we even have this distinction, or could we get rid of it?

benjamingeer:
- [x] Rewrite ontology loading and caching in `OntologyResponderV2` to support serving built-in ontologies like `salsah-gui` from the triplestore, including OWL named individuals.
- [x] Change ``salsah-gui`` so that GraphDB's consistency checker can check the use of ``guiElement`` and ``guiAttribute``.
- [x] Serve the `salsah-gui` ontology in API v2 in the default schema.
- [x] Make existing functionality work.
- [x] Simplify the `hasOntologies` format, eliminating `hasStandoffClasses` and `hasStandoffProperties`.
- [x] Show `salsah-gui:guiElement` and `salsah-gui:guiAttribute` when serving ontologies in API v2 in the default schema.
- [x] Allow `salsah-gui:guiElement` and `salsah-gui:guiAttribute` to be included in new property definitions created via API v2.
- [x] Update docs.
- [x] Update release notes.

The existing design of ontology loading and caching in `OntologyResponderV2` was made for API v1, which only needed to serve project-specific ontologies. It was based on several different SPARQL SELECT queries for different sorts of entities. These queries knew quite a lot about the structure of Knora ontologies. Only certain kinds of entities and cardinalities were loaded and cached, and the rest of the code relied on that. Now that we have to serve classes like `salsah-gui:Guielement` and properties like `salsah-gui:guiAttribute`, that’s no longer workable. Rather than add more SELECT queries and special cases, it seems better to redesign the whole thing.

The revised design in this PR is based on a CONSTRUCT query that just reads all statements from each ontology graph. This makes the code simpler and probably easier to read. All entities are then processed in the same way into `Content` objects. As a result, code using these objects must now assume that everything that’s in the ontologies in the triplestore is also in the ontology cache. The `Content` objects are then wrapped in `Read` objects, which have more knowledge about the different types of Knora entities and how to represent them in the API.

Also, I've restructured the cache so each ontology, including `knora-base`, is now stored in a `ReadOntologyV2` object. This should make it easier to do #762.

As an extra bonus, IntelliJ IDEA no longer uses 100% CPU when you work on `OntologyResponderV2`.

Resolves #746.
Resolves #760.


benjamingeer:
TODO:

We should serve the `salsah-gui` ontology entirely from the triplestore. But it can't be read using the existing `OntologyResponderV2.loadOntologies` method. Moreover, that method should be able to read the `standoff` ontology, too. So it's time to refactor it. It could work like this:

* Determine the named graphs containing ontologies, e.g. by doing a SPARQL request to find the graphs that have `?ontologyIri rdf:type owl:Ontology` in them.
* Read everything from all those named graphs, making one `CONSTRUCT` request per graph (because Turtle doesn't support named graphs, and I don't know how widespread [TriG](https://www.w3.org/TR/trig/) support is).
* For each named graph result, look at the `rdf:type` of each subject to get the IRIs of the classes, properties, and individuals.
* Construct an entity info content for each one, adapting the methods like `constructResponseToClassDefinition` to take the entity IRI and the whole `CONSTRUCT` response as parameters.
* Then proceed more or less as in the existing `loadOntologies` method.

benjamingeer:
@tobiasschweizer this PR removes `hasStandoffClasses` and `hasStandoffProperties` from the ontology responses, as discussed in #760.

tobiasschweizer:
> this PR removes hasStandoffClasses and hasStandoffProperties from the ontology responses, as discussed in #760.

Could you please check if that requires changes in the SALSAH 2 GUI?

benjamingeer:
> did you maybe run the Salsah 1 tests?

Good catch! Had to fix a bug for that. They pass now.

subotic:
great, thanks :-)

subotic:
Some general thoughts about the release notes. Maybe we should add something like a migration guide. There we could move the stuff from `Required changes to existing data:`. This would be one document, and then we wouldn't need to repeat things.

mrivoal:
@benjamingeer just a question: didn't you say that you would would come with a script to make existing ontologies comply with the new requirements of `salsah-gui:gui-element`?

mrivoal:
Ok, I have just been explained by @loicjaouen  and @gfoo where I should look for this information. Just ignore my question/comment.

benjamingeer:
@mrivoal Existing ontologies should not be affected by this PR; it just affects the way ontologies are represented in API v2.

mrivoal:
Sorry. I mixed up this issue with #737, which requires us to update our existing ontologies according to v.1.2.0 Release Notes.

benjamingeer:
@mrivoal No worries!

PR #758, opened 2018-03-01
build (travis): add execution of publish stage when on tag

subotic:
### Summary

When a commit is tagged, like for our releases, then the publish stage, during which the docker images are built, should also run for these tags, so that the docker images are also tagged with the release tag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,8 @@ before_install:
 
 stages:
   - test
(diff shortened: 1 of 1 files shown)


subotic:
@benjamingeer Should also be a fast one :-)

subotic:
thanks :-)

PR #767, opened 2018-03-08
Use types from the triplestore in predicate objects in ontologies

### LINKED ISSUE #738: Refactor OntologyResponderV2 to use type information from the triplestore
`PredicateInfoV2` needs to know which of its objects are IRIs, so it can convert their schemas when needed. Now that #668 is done, it could use `LiteralV2` instead of strings. But this means refactoring `OntologyResponderV2.loadOntologies` and `getResourceClassDefinitions.scala.txt` to use `CONSTRUCT` instead of `SELECT`.

benjamingeer:
The case class `PredicateInfoV2` stores the objects of predicates of ontology entities. Currently, `PredicateInfoV2` stores all its objects as strings. When ontology information is converted from one schema to another, objects that are actually IRIs have to be converted. Since `PredicateInfoV2` doesn't know the types of its objects, something else has to keep track of that. So there are hard-coded sets of predicates that are known to have IRIs as objects, but this is cumbersome. It also means that to convert an IRI, `PredicateInfoV2` needs to convert a string to a `SmartIri` and back again.

This PR takes advantage of the type information that comes with `SparqlExtendedConstructResponse` to enable `PredicateInfoV2` to store store IRIs as `SmartIri` objects instead of strings.

- [x] Add a trait `OntologyLiteralV2`, which is like `LiteralV2` but has just the implementations needed by `PredicateInfoV2`. (`OntologyLiteralV2` shares some of its implementations with `LiteralV2`.)
- [x] Refactor `PredicateInfoV2` to use `OntologyLiteralV2`.
- [x] Use a `Seq` instead of a `Set` for the objects in `PredicateInfoV2`, so a `Seq[StringLiteralV2]` can be accepted as a `Seq[OntologyLiteralV2]` (this doesn't work with `Set`, because `Set` is [not covariant in its type parameter](https://stackoverflow.com/questions/676615/why-is-scalas-immutable-set-not-covariant-in-its-type)).
- [x] Fix bug: when API v2 served `knora-api` (default schema), `salsah-gui:guiElement` and `salsah-gui:guiAttribute` were not shown in properties in that ontology.
- [x] Fix bug: `salsah-gui:guiOrder` was not accepted when creating a property via API v2.

As a result, `PredicateInfoV2.toOntologySchema` becomes nice and simple:

```scala
def toOntologySchema(targetSchema: OntologySchema): PredicateInfoV2 = {
    copy(
        predicateIri = predicateIri.toOntologySchema(targetSchema),
        objects = objects.map {
            case smartIriLiteral: SmartIriLiteralV2 => smartIriLiteral.toOntologySchema(targetSchema)
            case other => other
        }
    )
}
```

Resolves #738.

### DIFF
--- a/docs/rst/00-release-notes/v1.3.0.rst
+++ b/docs/rst/00-release-notes/v1.3.0.rst
@@ -45,5 +45,8 @@ New features:
 Bugfixes:
 ---------
 
+- When API v2 served ``knora-api`` (default schema), ``salsah-gui:guiElement`` and ``salsah-gui:guiAttribute`` were not shown in properties in that ontology.
+- The predicate ``salsah-gui:guiOrder`` was not accepted when creating a property via API v2.
+
 .. _release: https://github.com/dhlab-basel/Knora/releases/tag/v1.3.0
 .. _v1.3.0 milestone: https://github.com/dhlab-basel/Knora/milestone/7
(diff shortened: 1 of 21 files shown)


benjamingeer:
Thanks for the review!

subotic:
you shouldn't work on a Saturday ;-)

benjamingeer:
Neither should you. :)

PR #775, opened 2018-03-12
Generate knora-api ontologies, and convert standoff classes to external schemas

### LINKED ISSUE #648: Serve salsah-gui and standoff ontologies

### LINKED ISSUE #774: Generate knora-api ontologies, and convert standoff classes to external schemas
Following on from #648:

When we serve ontologies containing standoff classes, those classes need to be converted to the appropriate external schema. For example, `StandoffDateTag` needs to look something like `DateValue` in the external schema. This needs to work both for the `knora-api` ontologies and for project-specific ontologies. It would be possible to hard-code the standoff tags in `knora-api`, but not in project-specific ontologies. For project-specific ontologies, standoff classes will have to be converted when they're served. So we might as well do this for `knora-api` as well, converting from the classes in `knora-base`. And if we're going to do this for standoff classes, it makes sense to do it for the rest of `knora-api`, too (e.g. for `DateValue` itself). So `knora-api` should be generated from `knora-base` as far as possible.

benjamingeer:
## Conversion of `knora-base` entities to external schemas

Currently the `knora-api` ontology is completely hard-coded in Scala, in both schemas. This leads to inconsistencies between `knora-api` and equivalent parts of `knora-base`, and leaves us with no way to convert project-specific standoff datatype classes to external schemas. This PR instead generates `knora-api` from `knora-base`, hard-coding only the differences. Basically, to make `knora-api`:

1. Some entities are removed. When a property is removed, cardinalities for that property are also removed (as with `valueHasStartJDN`).
3. Some entities are added (e.g. `dateValueHasStartYear`).
2. Some cardinalities are added (e.g. `dateValueHasStartYear` to `DateBase`).

This conversion is done on the fly in `ReadOntologyV2` and its components, when converting from the internal schema to an external schema for API output. The conversion rules are in `KnoraApiV2Simple.KnoraBaseTransformationRules` and `KnoraApiV2WithValueObjects.KnoraBaseTransformationRules`.

As a result, Knora value classes, built-in standoff datatype classes, and project-specific standoff datatype classes are all served correctly, since they all get their cardinalities from `ValueBase` classes, and those cardinalities are modified on the fly for the output schema.

### Changes in `knora-api` with value objects

* Rename `hierarchicalListValueAsListNode` -> `listValueAsListNode` (for consistency with `ListValue`)
* Rename `hierarchicalListValueAsListNodeLabel` -> `listValueAsListNodeLabel`
* Rename `hasIncomingLinks` -> `hasIncomingLink` (for consistency with `hasLinkTo`)

### Changes in `knora-api` simple

* Remove these types, because they are identical to `File`:
  * `StillImageFile`
  * `MovingImageFile`
  * `AudioFile`
  * `DDDFile`
  * `TextFile`
  * `DocumentFile`
* Rename properties:
  * `hasIncomingLinks` -> `hasIncomingLink`
  * `hasStillImageFileValue` -> `hasStillImageFile` (was correct in `OntologyConstants` but not in `SearchRouteV2R2RSpec`, I don't understand why it worked before)
  * `hasMovingImageFileValue` -> `hasMovingImageFile`
  * `hasAudioFileValue` -> `hasAudioFile`
  * `hasDDDFileValue` -> `hasDDDFile`
  * `hasTextFileValue` -> `hasTextFile`
  * `hasDocumentFileValue` -> `hasDocumentFile`

### Changes in `knora-base`

* Remove unused properties:
  * `name`
  * `hasRegionValue`
  * `description`
  * `website`
* Add properties:
  * `canBeInstantiated`: indicates whether users can instantiate a class (e.g. false for `Resource`, true for `Region`).
  * `isEditable`: indicates whether users can edit the value of a property (e.g. false for `hasStandoffLinkTo`, true for `hasColor`).

Resolves #648.
Resolves #774.


benjamingeer:
@tobiasschweizer I think this PR will need some small changes in SALSAH 2.

tobiasschweizer:
> @tobiasschweizer I think this PR will need some small changes in SALSAH 2.

Ok, we can make a PR in SALSAH. What would the differences be exactly?

benjamingeer:
> What would the differences be exactly?

The differences are in the description of this PR.

tobiasschweizer:
> The differences are in the description of this PR.

I see. I was just too lazy to read it. 

I guess it would require renaming of some constants: https://github.com/dhlab-basel/Salsah/blob/develop/src/app/app.config.ts

tobiasschweizer:
and some tests would have to be updated

tobiasschweizer:
@benjamingeer I can review this PR next week (probably on Tuesday sine the moving takes place on Monday). I could also create a PR in SALSAH and do the necessary changes there.

benjamingeer:
@tobiasschweizer After you review it, I can click on the merge button if I'm not feeling too lazy. :)

tobiasschweizer:
> After you review it, I can click on the merge button if I'm not feeling too lazy. :)

This is totally up to you :-)

tobiasschweizer:
@benjamingeer Before I do the actual code review, I am trying to make all the necessary changes to SALSAH 2 (see https://github.com/dhlab-basel/Salsah/pull/215).

tobiasschweizer:
> hasStillImageFileValue -> hasStillImageFile (was correct in OntologyConstants but not in SearchRouteV2R2RSpec, I don't understand why it worked before)

You mean that in `OntologyConstants` it was always `hasStillImageFile` and you do not understand how it could be correctly mapped to the internal schema (since this would have been `hasStillImageFileValue`)?

tobiasschweizer:
I have the suspicion that the conversion of the IRI did the job (exchange of its base).

webapi/src/main/scala/org/knora/webapi/OntologyConstants.scala
```diff
@@ -117,23 +117,17 @@ object OntologyConstants {
...
         val KnoraBasePrefixExpansion: IRI = KnoraBaseOntologyIri + "#"
 
+        val CanBeInstantiated: IRI = KnoraBasePrefixExpansion + "canBeInstantiated"
+        val IsEditable: IRI = KnoraBasePrefixExpansion + "isEditable"
+
         val Resource: IRI = KnoraBasePrefixExpansion + "Resource"
-        val ExternalResource: IRI = KnoraBasePrefixExpansion + "ExternalResource"
         val Representation: IRI = KnoraBasePrefixExpansion + "Representation"
         val AudioRepresentation: IRI = KnoraBasePrefixExpansion + "AudioRepresentation"
         val DDDRepresentation: IRI = KnoraBasePrefixExpansion + "DDDrepresentation"
```
tobiasschweizer: This should be `DDDRepresentation`

benjamingeer:
> You mean that in `OntologyConstants` it was always `hasStillImageFile` and you do not understand how it could be correctly mapped to the internal schema (since this would have been `hasStillImageFileValue`)?

In `OntologyConstants` it was always `hasStillImageFile`, but the KnarQL in the tests used `hasStillImageFileValue`. I guess the value in `OntologyConstants` was never used. This PR adds some stuff to `SmartIri` and to `OntologyConstants` so that `hasStillImageFile` is mapped to `hasStillImageFileValue` and back when IRIs are converted.

tobiasschweizer:
> I guess the value in OntologyConstants was never used.

This is what I thought too.

tobiasschweizer:
> This PR instead generates knora-api from knora-base, hard-coding only the differences.

I think this is a good approach. 

Why would we still need extra `knora-api` definitions for `ForbiddenResource`? It is already in `knora-base`.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/ontologymessages/KnoraApiV2WithValueObjects.scala
```diff
@@ -171,1436 +144,490 @@ object KnoraApiV2WithValueObjects {
...
-    val HasIncomingLinks: ReadPropertyInfoV2 = makeProperty(
-        propertyIri = OntologyConstants.KnoraApiV2WithValueObjects.HasIncomingLinks,
-        isResourceProp = true,
-        propertyType = OntologyConstants.Owl.ObjectProperty,
-        subjectType = Some(OntologyConstants.KnoraApiV2WithValueObjects.Resource),
-        objectType = Some(OntologyConstants.KnoraApiV2WithValueObjects.LinkValue),
-        subPropertyOf = Set(OntologyConstants.KnoraApiV2WithValueObjects.HasLinkToValue),
-        isEditable = true,
-        isLinkValueProp = true,
+    private val TextValueHasStandoff: ReadPropertyInfoV2 = makeProperty(
```
tobiasschweizer: Currently we never serve standoff with a `TextValue`, but we send either XML or HTML (if an XSL transformation is attached to the mapping used) to the client. But we could surely serve the plain text and the standoff separately. 

We would just need to come up with a good idea how the client would request the format (as a segment in the URL or in the HTTP accept header).

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/ontologymessages/KnoraApiV2Simple.scala
```diff
@@ -757,11 +285,11 @@ object KnoraApiV2Simple {
         )
     )
 
-    val HasDDDFile: ReadPropertyInfoV2 = makeProperty(
+    private val HasDDDFile: ReadPropertyInfoV2 = makeProperty(
```
tobiasschweizer: Couldn't this not even be made more compact? The systematic difference is that we expose a `knora-base` property pointing to  a `FileValue` as a `knora-api` property with an object type `File` to the outside. This applies to all internal properties pointing to file values. 

Do you think we could reuse the internal labels and comments? Then they would not have to be repeated. Of course if we say in the comment that the property points to a value then this does not work.

benjamingeer:
Changes in last commit:

* `ForbiddenResource` now comes from `knora-base`.
* `SmarIri` now converts `FileValue`, `hasFileValue`, `hasStillImageFileValue`, etc., in `knora-base`, are now converted to `File`, `hasFile`, `hasStillImageFile`, etc., in `knora-api/simple/v2`. It uses the map in `OntologyConstants.CorrespondingIris`.

> Currently we never serve standoff with a `TextValue`, but we send either XML or HTML (if an XSL transformation is attached to the mapping used) to the client. But we could surely serve the plain text and the standoff separately.

I'd like to do that, but not right now. For now, I think we can just show that the standoff is attached to the `TextValue`, just as documentation, to clarify how standoff fits into the ontology.

webapi/src/main/scala/org/knora/webapi/OntologyConstants.scala
tobiasschweizer: It was also wrong in knora-base. This should fix it https://github.com/dhlab-basel/Knora/pull/775/commits/d1446090b09d96988031aca6601f840b4d1d8f04

tobiasschweizer:
@benjamingeer This looks really good now, with much less redundant information!

benjamingeer:
@tobiasschweizer You still have to approve it before I can merge. :)

PR #776, opened 2018-03-13
fix (beol): add beol project shortcode

tobiasschweizer:


tobiasschweizer:
@SepidehAlassi thanks!

PR #782, opened 2018-03-15
build (salsah1): add bash to final image

subotic:
### Summary

- add bash package to final alpine linux based image which does not come with bash but we need it and so on and on goes the description of this pr :-)


subotic:
@benjamingeer this time the description is longer than the change :-)

benjamingeer:
OK but I read it very slowly: "baaaaaassssssssshhhhhhhhhhhh". :)

subotic:
thanks :-)

PR #786, opened 2018-03-16
Nie 328 migrate ontologies

subotic:


subotic:
Please always create a pull-request early, so that we can see what is going on.


- [ ] There are now to many projects in `webapi/_test_data/all_data/admin-data.ttl`. Please move them in a separate file inside your test data folder. You can load them inside your test.

subotic:
The same for `webapi/_test_data/all_data/permissions-data.ttl`. Soon, these files will only include the minimum that is needed to initialize an empty Knora repository.

janCstoffregen:
ok, I will do that, thanks for this information, we always test all out ontologies locally first before we publish them, this is why the files are so big, we will change this when copying it to Knora.

subotic:
Also, please put all your ontologies and test data in a folder which is named like the test, e.g., if your test lives in `org.knora.webapi.other.v1.NieIneV1Spec`, then put all your ontologies and data inside `_test_data/other.v1.NieIneV1Spec`

janCstoffregen:
Lieber Ivan,
derzeit prüfen wir, ob der Fehler bei unseren Files liegt. Anderenfalls würde ich auf euch zukommen mit Folgendem: https://github.com/nie-ine/Ontologies/issues/2

Ich melde mich innerhalb dieser Woche, wenn wir den Fehler nicht finden können.

Vielen Dank und viele schöne Grüsse,
Jan

PR #796, opened 2018-03-27
Delete an ontology via API v2

benjamingeer:
- [x] Provide an HTTP DELETE route for deleting an ontology.
- [x] Have `OntologyResponderV2` maintain an ontology's connection to a project, by including `knora-base:attachedToProject` in each ontology's metadata.
- [x] Remove ontology maintenance code from ADM.
- [x] Refactor API v2 code to remove some dependencies on API v1, especially by using `UserADM` instead of `UserProfileV1` (also in V1 code that depends on V2), and by refactoring `PermissionUtilADM`.
- [x] Get tests working again.
- [x] Update docs.
- [x] Update release notes.


benjamingeer:
@subotic This test is failing but I don't know why:

https://github.com/dhlab-basel/Knora/blob/wip/delete-ontology/webapi/src/test/scala/org/knora/webapi/other/v1/DrawingsGodsV1Spec.scala

Could you have a look with me tomorrow?

lrosenth:
Wow!!

Lukas

Sent from my iPad

On 27 Mar 2018, at 22:51, Benjamin Geer <notifications@github.com<mailto:notifications@github.com>> wrote:


@subotic<https://github.com/subotic> This test is failing but I don't know why:

https://github.com/dhlab-basel/Knora/blob/wip/delete-ontology/webapi/src/test/scala/org/knora/webapi/other/v1/DrawingsGodsV1Spec.scala

Could you have a look with me tomorrow?

loicjaouen:
_Wow!!_ indeed... This comes to our great surprise.

I thought that ontologies were pure abstract models, and that their abstract nature made them independent of any project.

It seems to me that both the nature and the direction of the link is wrong: `project --implements--> ontology` is fine but not `project <--depends on-- ontology`

When semantic is twisted, problems eventually creep in.

benjamingeer:
> I thought that ontologies were pure abstract models, and that their abstract nature made them independent of any project.

Nope. There are two kinds of ontologies in Knora: built-in ontologies like `knora-base`, and project-specific ontologies.

We agreed that:

* Project-specific ontologies should not be used by other projects, because they need to be able to change at any time.
* We will create a process for standardising ontologies that can be shared by multiple projects: #523.

loicjaouen:
I don't see how that opposes to my remark.

Of course, it might ease the technical ways of checking, but otherwise, to me it is like declaring a class "abstract" and at the same time making it "concrete" by saying that it is to be used by this or that class... I also understand that another analogy is that you design a project ontology as a class meant to have a singleton instance.

Nevertheless, I wanted to let you know that it triggered a warning light to me.

loicjaouen:
(and it will make another delicate prod release)

benjamingeer:
> I don't see how that opposes to my remark.

You said you thought ontologies were “independent of any project”. That has never been true.

An ontology has always been attached to a project. The only thing that this PR changes is where the connection between the ontology and the project is stored. Currently it is stored in the project. With this PR, it is stored in the ontology. The meaning is the same.

> Of course, it might ease the technical ways of checking, but otherwise, to me it is like declaring a class "abstract" and at the same time making it "concrete" by saying that it is to be used by this or that class...

I don’t understand what you mean here.

loicjaouen:
- about the relative independence of ontologies and projects
> that has never been true

I mean that ontologies exist outside of a project.

When writing an ontology, can't I use a property of another ontology (used in another project)?

- about the easing technical ways of checking

I meant: "of checking that an ontology is used by a project"

Putting the information in the ontology is straightforward, rather that checking projects to find the one listing a dependency on that ontology.

benjamingeer:
> When writing an ontology, can't I use a property of another ontology (used in another project)?

Currently there is nothing to keep you from doing so, and this PR doesn’t change that. However, we will have to forbid this, to prevent this scenario:

1. Project A creates a class C.
2. Project B makes instances of class C.
3. Project A wants to add a new required property to class C, but can’t, because it would violate consistency rules (because of the existing instances).

This is the reason for #523.

benjamingeer:
Or this scenario:

1. Project A creates a property P.
2. Project B uses the property (in its own ontology or directly in data).
3. Project A wants to change the property (e.g. to change its `objectClassConstraint`), but can’t, because this would violate consistency rules.

benjamingeer:
In other words, we will have to implement restrictions to ensure that project B cannot use properties or classes defined in the ontologies of project A, either in B’s ontologies or in B’s data.

benjamingeer:
@subotic @tobiasschweizer Any chance one of you could review this?

subotic:
yes, I was actually doing it right now :-)

benjamingeer:
OK thank you!

tobiasschweizer:
@benjamingeer Are there any changes that would require adaptions in the productive BEOL data?

benjamingeer:
> Are there any changes that would require adaptions in the productive BEOL data?

I don't think so. This PR shouldn't change any existing functionality.

benjamingeer:
> Are there any changes that would require adaptions in the productive BEOL data?

(Sorry, posted incorrect reply.)

Yes, see the release notes:

https://github.com/dhlab-basel/Knora/blob/8198f88c0e077ca07ffa18bc34508451d1786925/docs/rst/00-release-notes/v1.4.0.rst

tobiasschweizer:
> (Sorry, posted incorrect reply.)

Sorry, now I am confused. 

Since the **ontologies** for beol and biblio are version controlled and built in, they will be updated automatically. This shouldn't affect the productive **data** (`biblio-data.ttl` and `beol-data.ttl`).  The data simply needs to be loaded again into the triplestore but it does not have to be changed.

@subotic `admin-data.ttl` will change too since `knora-base:projectOntology` will be removed. I guess you had to manually adapt it on the productive server.

benjamingeer:
> Since the ontologies for beol and biblio are version controlled and built in, they will be updated automatically. This shouldn't affect the productive data (biblio-data.ttl and beol-data.ttl). The data simply needs to be loaded again into the triplestore but it does not have to be changed.

Yes, I'm just saying that to use this release, you need replace `beol-onto.ttl`, `biblio-onto.ttl`, and `admin-data.ttl` in the triplestore on your server.

tobiasschweizer:
> Yes, I'm just saying that to use this release, you need replace beol-onto.ttl, biblio-onto.ttl, and admin-data.ttl in the triplestore on your server.

Got it :-) Thanks!

subotic:
>  I guess you had to manually adapt it on the productive server.

Yes, I would need to apply all the changes that are described in the release notes by hand.

webapi/src/main/scala/org/knora/webapi/util/PermissionUtilADM.scala
```diff
@@ -22,120 +22,103 @@ package org.knora.webapi.util
...
 
-import scala.collection.immutable.Iterable
 
 /**
-  * A utility that responder actors use to determine a user's permissions on an RDF subject in the triplestore.
+  * A utility that responder actors use to determine a user's permissions on an RDF entity in the triplestore.
   */
 object PermissionUtilADM {
 
-    val log = Logger(LoggerFactory.getLogger(this.getClass))
```
subotic: Any particular reason why you would remove the logger? If we wont to filter logging output for this class, then it needs to be defined here.

benjamingeer: It was unused. I can put it back if you like.

subotic: ah, found it. wasn't removed. just moved.

webapi/src/test/scala/org/knora/webapi/responders/v2/OntologyResponderV2Spec.scala
```diff
@@ -131,6 +130,52 @@ class OntologyResponderV2Spec extends CoreSpec() with ImplicitSender {
...
 
+        "delete the 'foo' ontology" in {
+            actorUnderTest ! DeleteOntologyRequestV2(
+                ontologyIri = fooIri.get.toSmartIri,
+                lastModificationDate = fooLastModDate,
+                apiRequestID = UUID.randomUUID,
+                requestingUser = imagesUser
+            )
+
+            expectMsgType[SuccessResponseV2](timeout)
```
subotic: Could you maybe add another request that checks that the ontology is really not there anymore? This test would also pass if the `DeleteOntologyRequestV2` doesn't do anything besides returning a positive result.

subotic: - maybe also a quick test that tries to delete a non-existing ontology?
- maybe also test with a project-admin user and a non-project-admin user? I suppose that only a project admin should be able to delete an ontology?

subotic:
Also, did you maybe run the Salsah GUI tests?

benjamingeer:
Will do those things, thanks!

subotic:
no, thank you :-)

benjamingeer:
It turns out that I accidentally broke the SALSAH 1 GUI. Trying to sort this out now, wandering through a maze of twisty little dependencies in API v1...

benjamingeer:
@subotic I added the tests you asked for, plus other tests to check that no ontology updates can be done without the right permissions.

Thanks to the integration tests, I realised that I had to put back some functionality needed by SALSAH 1, and fix some other things:

* Include ontology IRIs in `ProjectADM` and `ProjectInfoV1`.
* Have `ProjectsResponderADM` and `ProjectsResponderV1` get ontology IRIs from ontology responders.
* Remove redundant code from `OntologyResponderV2` that caused an infinite loop.
* Fix bugs in SALSAH 1 integration tests.

Is this OK now?

PR #809, opened 2018-04-06
build (webapi): fix path to salsah1

subotic:
### Summary

- build (webapi): fix path to salsah1


subotic:
This is unbelievable. I think that any branch with the string `gh-pages` in it is excluded by Travis.

subotic:
thanks :-)

Issue #803, opened 2018-03-28
Remove remaining dependencies on API v1 code from API v2 code


benjamingeer:
I did most of this in #796. What's left:

* `ConstructResponseUtilV2` uses:
  * `StandoffEntityInfoGetResponseV1`
  * `KnoraCalendarV1`
  * `KnoraPrecisionV1`
  * `MappingXMLtoStandoff`
* `TextValueContentV2` uses `StandoffTagUtilV1`
* `StandoffAndMapping` uses:
  * `StandoffTagV1`
  * `MappingXMLtoStandoff`
* `SearchResponderV2` uses `JulianDayNumberValueV1`
* `OntologyResponderV2` uses `StandoffDataTypeClasses`
* `ReadClassInfoV2` uses `StandoffDataTypeClasses`

tobiasschweizer:
@benjamingeer Do you think refactoring v1 standoff responder and messages should also be part of #699?

benjamingeer:
Which do you think would be easier? Doing a simpler refactoring first to remove these dependencies (e.g. by copying and pasting code into V2), and then doing #699 separately? Or doing both at the same time?

tobiasschweizer:
I guess adding new stuff to the existing v1 version wouldn't make any sense at all, so a simple refactoring at the very beginning seems reasonable to me.

benjamingeer:
OK, would you mind doing this while I work on #800? I'd really like to have a V2 codebase free of V1 code before adding more new features.

tobiasschweizer:
> OK, would you mind doing this while I work on #800? I'd really like to have a V2 codebase free of V1 code before adding more new features.

I will have a look and make an outline of the required refactoring and estimate the time I need.

tobiasschweizer:
I will try to refactor `StandoffResponderV1` and `StandoffMessagesV1` as well as `StandoffTagUtilV1` so that V1 calls V2 but V2 is self contained. We did this for `OntologyResponderV1` and `OntologyResponderV2`.

Also I want to do #396.

tobiasschweizer:
I guess this refactoring will take a couple of days, probably most of next week.

tobiasschweizer:
@benjamingeer I have just realized that I am about to dive into V2 JSON-LD handling :-) 

this time also turning JSON-LD into case classes for the creation of a mapping, but the logic is already there thanks to the ontology writing and updating parts you did :-)

Issue #396, opened 2017-01-30
Handling of Mapping


tobiasschweizer:
- refactor caching of mapping so that a flat representation of the mapping is stored (a sequence of mapping elements like in the XML representation of a mapping)
- transform this flat structure into MappingXMLtoStandoff (XML elements are keys) or the other way (standoff classes become keys) when needed
- add a route to return an existing mapping to the user as XML

PR #811, opened 2018-04-06
Add V2 responder and messages for standoff

### LINKED ISSUE #818: Ontology Information about XSLTransformation
http://0.0.0.0:3333/v2/ontologies/allentities/http%3A%2F%2Fapi.knora.org%2Fontology%2Fknora-api%2Fv2

does not include `XSLTransformation` which is defined in `knora-base`:

```

### http://www.knora.org/ontology/knora-base#XSLTransformation

:XSLTransformation  rdf:type owl:Class ;

                    rdfs:subClassOf :TextRepresentation ,
                                    [ rdf:type owl:Restriction ;
                                      owl:onProperty :hasTextFileValue ;
                                      owl:cardinality "1"^^xsd:nonNegativeInteger
                                    ] ;

                    rdfs:label "a TextRepresentation representing an XSL transformation that can be applied to an XML created from standoff. The transformation's result is ecptected to be HTML."@en ;

                    rdfs:comment "a TextRepresentation representing an XSL transformation that can be applied to an XML created from standoff.  The transformation's result is ecptected to be HTML."@en .


```

Could the problem be that it is not an immediate subclass of `knora-base:Resource` and the current implementation only returns direct children?

tobiasschweizer:
Add V2 responder and messages for standoff, move and refactor V1 code as necessary and call V2 from V1. V2 Standoff is then self contained.

relates to: #803 
closes #818


tobiasschweizer:
@benjamingeer A mapping creation request is a HTTP multipart request (at least in V1). One part contains the mapping as XML, the other the JSON-LD. In `StandoffResponderV1` the parts had to be named "json" and "xml" and would then be merged into one case class.

With the current design, we could just call `fromJsonLD` and then call `copy` on it to get the xml in. This would surely work, but it is not very elegant. Another approach would be to put the XML into the JSON-lD. But this seems ugly too.

```scala
case class CreateMappingRequestV2(xml: String, label: String, projectIri: SmartIri, mappingName: String, userProfile: UserProfileV1, apiRequestID: UUID) extends StandoffResponderRequestV2

object CreateMappingRequestV2 extends KnoraJsonLDRequestReaderV2[CreateMappingRequestV2] {
    override def fromJsonLD(jsonLDDocument: JsonLDDocument,
                            apiRequestID: UUID,
                            userProfile: UserProfileV1): CreateMappingRequestV2 = {

        implicit val stringFormatter: StringFormatter = StringFormatter.getGeneralInstance

        val label: String = jsonLDDocument.requireString(OntologyConstants.Rdfs.Label, stringFormatter.toSparqlEncodedString)

        val projectIri: SmartIri = jsonLDDocument.requireString(OntologyConstants.KnoraApiV2WithValueObjects.AttachedToProject, stringFormatter.toSmartIriWithErr)

        val mappingName: String = jsonLDDocument.requireString(OntologyConstants.KnoraApiV2WithValueObjects.MappingHasName, stringFormatter.toSparqlEncodedString)

        CreateMappingRequestV2(
            xml = "", // TODO: get xml from multipart request
            label = label,
            projectIri = projectIri,
            mappingName = mappingName,
            userProfile = userProfile,
            apiRequestID = apiRequestID
        )
    }
}
```

benjamingeer:
> With the current design, we could just call fromJsonLD and then call copy on it to get the xml in.

Seems OK to me. You have a request that's part JSON-LD and part XML. So you parse the JSON-LD and then add the XML.

No matter how you do it, it will be two steps, because there are two syntaxes.

benjamingeer:
Oh I see, the problem is that you don't want this:

```
xml = ""
```

Then you can make another case class that just has the data from the JSON-LD, and wrap it in something that also has the XML.

tobiasschweizer:
> Oh I see, the problem is that you don't want this:

Exactly :-) 

> Then you can make another case class that just has the data from the JSON-LD, and wrap it in something that also has the XML.

Ok, will do :-)

The code you wrote to turn JSON-LD to case classes looks really good!

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/standoffmessages/StandoffMessagesV2.scala
```diff
@@ -0,0 +1,112 @@
...
+  */
+case class CreateMappingResponseV2(mappingIri: IRI) extends KnoraResponseV2 {
+
+    def toJsonLDDocument(targetSchema: ApiV2Schema, settings: SettingsImpl): JsonLDDocument = {
+
+        implicit val stringFormatter: StringFormatter = StringFormatter.getGeneralInstance
+
+        val body = JsonLDObject(Map(
+            "@id" -> JsonLDString(mappingIri),
+            "@type" -> JsonLDString(OntologyConstants.KnoraBase.XMLToStandoffMapping.toSmartIri.toOntologySchema(targetSchema).toString)
```
tobiasschweizer: @benjamingeer Is this the correct way to use smartIri (following <http://www.knora.org/documentation/manual/rst/03-knora-api-server/design-documentation/api-v2.html?highlight=smartiri#usage>)?

Is calling `toString` on it correct after having converted it to a target schema?

benjamingeer: Yes, `toString` gets you the IRI as a string, which is what you need for `JsonLDString`.

Converting to the target schema and then to JSON-LD in the same method is OK if you only need to support the complex schema, which I guess is the case here. If you need to be able to return JSON-LD in more than one schema, I recommend doing the schema conversion in a separate method.

benjamingeer:
Also, if I were you, I would merge `develop` into this branch before doing any more work on it. I merged a big PR on Friday that removes a lot of V2 dependencies on V1.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/standoffmessages/StandoffMessagesV2.scala
tobiasschweizer: > Converting to the target schema and then to JSON-LD in the same method is OK if you only need to support the complex schema, which I guess is the case here. 

This is the response to a mapping creation request. Write or update operations will always be in v2 complex schema and thus v2 complex will be returned. But still the `targetSchema` is not hard-coded, it is a parameter. So it should theoretically also work for v2 simple, right?

tobiasschweizer:
> Also, if I were you, I would merge develop into this branch before doing any more work on it. I merged a big PR on Friday that removes a lot of V2 dependencies on V1.

Ok, done.

tobiasschweizer:
@benjamingeer Thanks for the hint, I fixed the issue with `UserProfileV1` in https://github.com/dhlab-basel/Knora/pull/811/commits/14415e8ae3ba0fd2d19993cdb707514ba579a7a3

benjamingeer:
> So it should theoretically also work for v2 simple, right?

Yes. Here you have a very simple data structure that just contains one IRI, so it doesn't matter anyway. I was thinking of situations where you have a more complex data structure, so building the JSON-LD is more work. Then doing the schema conversion in a separate method helps clarify simplify the code.

webapi/src/main/scala/org/knora/webapi/responders/v2/StandoffResponderV2.scala
```diff
@@ -79,63 +80,63 @@ class StandoffResponderV2 extends Responder {
...
-            _ = if (!(textLocation.format_name == "XML" && textLocation.origname.endsWith(".xsl"))) {
-                throw BadRequestException(s"$xslTransformationIri does not have a file value referring to a XSL transformation")
+            // check if `xsltFileValue` represents an XSL transformation
+            _ = if (!(xsltFileValue.internalMimeType == "text/xml" && xsltFileValue.originalFilename.endsWith(".xsl"))) {
+                throw BadRequestException(s"$xslTransformationIri does not have a file value referring to an XSL transformation")
             }
-        } yield textLocation
 
-        val recoveredTextLocationFuture = textLocationFuture.recover {
+            xsltUrl: String = s"${settings.internalSipiFileServerGetUrl}/${xsltFileValue.internalFilename}"
```
tobiasschweizer: @subotic Is this Sipi URL correct? 

Using the v2 responder, I could get rid of the URL replacing operation.

webapi/src/main/scala/org/knora/webapi/responders/v2/StandoffResponderV2.scala
```diff
@@ -0,0 +1,785 @@
...
+                }
+                case None => throw BadRequestException(s"Resource $xslTransformationIri is not a ${OntologyConstants.KnoraBase.XSLTransformation}")
+            }
+
+            // check if `xsltFileValue` represents an XSL transformation
+            _ = if (!(xsltFileValue.internalMimeType == "text/xml" && xsltFileValue.originalFilename.endsWith(".xsl"))) {
+                throw BadRequestException(s"$xslTransformationIri does not have a file value referring to an XSL transformation")
+            }
+
+            xsltUrl: String = s"${settings.internalSipiFileServerGetUrl}/${xsltFileValue.internalFilename}"
```
tobiasschweizer: @subotic Is this Sipi URL correct?

Using the v2 responder, I could get rid of the URL replacing operation that we needed for v1.

webapi/src/test/scala/org/knora/webapi/e2e/v2/StandoffRouteV2R2RSpec.scala
```diff
@@ -0,0 +1,151 @@
...
+
+        "create a mapping from a XML" in {
+
+            val xmlFileToSend = new File(RequestParams.pathToLetterMapping)
+
+            val mappingParams =
+                s"""
+                   |{
+                   |    "knora-api:mappingHasName": "LetterMapping",
+                   |    "knora-api:attachedToProject": "$ANYTHING_PROJECT_IRI",
```
tobiasschweizer: @benjamingeer I noticed that you used `knora-api:projectIri` (complex schema) in the ontology route. 

Shall I also use `knora-api:projectIri` (complex schema)?

```
"knora-api:projectIri" : {
        "@id" : "knora-api:projectIri",
        "@type" : "owl:DatatypeProperty",
        "knora-api:objectType" : "http://www.w3.org/2001/XMLSchema#anyURI",
        "rdfs:comment" : "Represents the IRI of a Knora project",
        "rdfs:label" : "project IRI"
      }
``` 

```
"knora-api:attachedToProject" : {
        "@id" : "knora-api:attachedToProject",
        "@type" : "owl:ObjectProperty",
        "knora-api:objectType" : "http://api.knora.org/ontology/knora-api/v2#knoraProject",
        "rdfs:comment" : "Connects a Resource or a ListNode (only the root node) to a project"
      }
```

Neither an ontology nor a mapping are resources. However, I think that `attachedToProject` expresses the relation between an ontology or a mapping and a project better than `projectIri`.

webapi/src/test/scala/org/knora/webapi/e2e/v2/StandoffRouteV2R2RSpec.scala
```diff
@@ -0,0 +1,151 @@
...
+    "The Standoff v2 Endpoint" should {
+
+        "create a mapping from a XML" in {
+
+            val xmlFileToSend = new File(RequestParams.pathToLetterMapping)
+
+            val mappingParams =
+                s"""
+                   |{
+                   |    "knora-api:mappingHasName": "LetterMapping",
```
tobiasschweizer: @benjamingeer I guess I am supposed to add that to `webapi/src/main/scala/org/knora/webapi/messages/v2/responder/ontologymessages/KnoraApiV2WithValueObjects.scala`, right?

webapi/src/test/scala/org/knora/webapi/e2e/v2/StandoffRouteV2R2RSpec.scala
benjamingeer: You're right, I should have used `attachedToProject`. I'll change it in the ontology route.

webapi/src/test/scala/org/knora/webapi/e2e/v2/StandoffRouteV2R2RSpec.scala
benjamingeer: Yes, I think that any properties used in requests should be included in `knora-api` as documentation.

webapi/src/test/scala/org/knora/webapi/e2e/v2/StandoffRouteV2R2RSpec.scala
tobiasschweizer: Then we should change the `rdfs:comment` of `attachedToProject` in `knora-admin.ttl` too. It does not have a `subjectClassConstraint` so it should be fine.

benjamingeer: I'll do that, too. It's already used with ontologies when they're served, so it's definitely fine.

benjamingeer: And it's already attached to ontologies in the triplestore.

tobiasschweizer: Ok, very good. Thank you!

webapi/src/main/scala/org/knora/webapi/util/standoff/StandoffTagUtilV2.scala
```diff
@@ -731,25 +735,25 @@ object StandoffTagUtilV2 {
...
-                        } else if (propPredicates.get(OntologyConstants.KnoraBase.ObjectDatatypeConstraint).isDefined) {
+                        } else if (propPredicates.get(OntologyConstants.KnoraBase.ObjectDatatypeConstraint.toSmartIri).isDefined) {
 
                             // it is a data type property (literal)
-                            val propDataType = propPredicates(OntologyConstants.KnoraBase.ObjectDatatypeConstraint)
+                            val propDataType: PredicateInfoV2 = propPredicates(OntologyConstants.KnoraBase.ObjectDatatypeConstraint.toSmartIri)
 
                             propDataType.objects.headOption match {
-                                case Some(OntologyConstants.Xsd.String) =>
+                                case Some(SmartIriLiteralV2(SmartIri(OntologyConstants.Xsd.String))) =>
```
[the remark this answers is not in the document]
benjamingeer: I don't understand the thing about stable identifiers, but the way you're doing it is correct. I remember I had to write the method `SmartIri.unapply` to make this work.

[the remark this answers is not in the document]
benjamingeer: There's even a test for it:

https://github.com/dhlab-basel/Knora/blob/develop/webapi/src/test/scala/org/knora/webapi/util/StringFormatterSpec.scala#L685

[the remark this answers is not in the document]
tobiasschweizer: I used this myself:

```scala
def f(x: Int, y: Int) = x match {
  case `y` => ...
}
```
(https://www.scala-lang.org/files/archive/spec/2.11/08-pattern-matching.html#stable-identifier-patterns)

`y` holds a value that you want to match for (e.g., 2). But I do not understand how this is related to our case. Maybe because there is the return value of a function (`toSmartIri`) involved that could behave differently although the parameter stays the same?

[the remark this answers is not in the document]
benjamingeer: I have no idea. I was just happy that I could make it possible to use `SmartIri` in match-case expressions.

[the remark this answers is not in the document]
benjamingeer: `toSmartIri` involves some compiler magic, which may interfere with using it in this context:

http://docs.scala-lang.org/overviews/core/value-classes.html

But anyway, I think `case Some(SmartIriLiteralV2(SmartIri(OntologyConstants.Xsd.String)))` looks just fine. :)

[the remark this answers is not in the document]
tobiasschweizer: > But anyway, I think case Some(SmartIriLiteralV2(SmartIri(OntologyConstants.Xsd.String))) looks just fine. :)

Absolutely! I was *just* curious :-) 

I once had a weird issue involving match case where I got a match error at runtime although the compiler could have figured out at compile time that there was not a single possible match. Still it compiled. This is why I want to keep an eye on details ;-) But you are right, if it works, there is no use for further investigations.

tobiasschweizer:
I think about how to refactor the current internal representation of a mapping (#396).

webapi/src/main/scala/org/knora/webapi/responders/v2/StandoffResponderV2.scala
tobiasschweizer: @subotic Will this work on the production server? We had to do some replacement operations for v1.

webapi/src/main/scala/org/knora/webapi/OntologyConstants.scala
```diff
@@ -749,6 +749,9 @@ object OntologyConstants {
 
         val ResourceIcon: IRI = KnoraApiV2PrefixExpansion + "resourceIcon"
 
+        // TODO: Add this to the ontology in the complex schema
```
benjamingeer: Could you do this TODO?

webapi/src/main/scala/org/knora/webapi/responders/v2/StandoffResponderV2.scala
```diff
@@ -0,0 +1,781 @@
...
+            userIri: IRI <- Future {
+                if (userProfile.isAnonymousUser) {
+                    throw ForbiddenException("Anonymous users aren't allowed to create mappings")
+                } else {
+                    userProfile.id
+                }
+            }
+
+            // check if the given project IRI represents an actual project
+            projectInfo: ProjectInfoResponseV1 <- (responderManager ? ProjectInfoByIRIGetRequestV1(
```
benjamingeer: Please make this use messages from `ProjectsMessagesADM.scala` instead of v1 messages.

webapi/src/main/scala/org/knora/webapi/routing/v2/StandoffRouteV2.scala
```diff
@@ -0,0 +1,129 @@
...
+import akka.actor.ActorSystem
+import akka.event.LoggingAdapter
+import akka.http.scaladsl.model.Multipart
+import akka.http.scaladsl.model.Multipart.BodyPart
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.Route
+import akka.stream.ActorMaterializer
+import akka.util.Timeout
+import org.knora.webapi.messages.v2.responder.standoffmessages.{CreateMappingRequestMetadataV2, CreateMappingRequestV2, CreateMappingRequestXMLV2}
+import org.knora.webapi.routing.v1.StandoffRouteV1.getUserADM
```
benjamingeer: This v1 import is unused, please remove it.

webapi/src/main/scala/org/knora/webapi/OntologyConstants.scala
tobiasschweizer: I have already added it to `KnoraApiV2WithValueObjects.scala`, just forgotten to remove the TODO :-).

webapi/src/main/scala/org/knora/webapi/responders/v2/StandoffResponderV2.scala
tobiasschweizer: done in https://github.com/dhlab-basel/Knora/pull/811/commits/2d8e33d09b0c77f5916ef1fb3e6aabe1e9fb61ea

webapi/src/main/scala/org/knora/webapi/routing/v2/StandoffRouteV2.scala
tobiasschweizer: done in https://github.com/dhlab-basel/Knora/pull/811/commits/2d8e33d09b0c77f5916ef1fb3e6aabe1e9fb61ea

webapi/src/main/scala/org/knora/webapi/OntologyConstants.scala
tobiasschweizer: removed TODO in https://github.com/dhlab-basel/Knora/pull/811/commits/2d8e33d09b0c77f5916ef1fb3e6aabe1e9fb61ea

webapi/src/test/scala/org/knora/webapi/e2e/v2/ResponseCheckerR2RV2.scala
```diff
@@ -153,43 +153,69 @@ object ResponseCheckerR2RV2 {
...
+      * @param expectedJSONLD the expected response as JSONLD.
+      * @param receivedJSONLD the received response as JSONLD.
+      */
+    def compareJsonForMappingCreationResponse(expectedJSONLD: String, receivedJSONLD: String): Unit = {
+
+        val expectedResponseAsScala: Map[IRI, Any] = JSONToScala(expectedJSONLD)
+
+        val receivedResponseAsScala: Map[IRI, Any] = JSONToScala(receivedJSONLD)
+
+        assert(expectedResponseAsScala == expectedResponseAsScala, "Mapping creation request response did not match expected response")
```
tobiasschweizer: @benjamingeer This test will always be true :-)

tobiasschweizer: my only excuse is that this was before my holiday ...

PR #812, opened 2018-04-06
feature (webapi): OpenAPI / Swagger API documentation generation

subotic:
### Summary

- add OpenAPI / Swagger API documentation generation

To try it out, run `webapi` and open `http://0.0.0.0:3333/api-docs/swagger.json` in http://petstore.swagger.io

### DIFF
--- a/docs/rst/03-knora-api-server/api_admin/api-general.rst
+++ b/docs/rst/03-knora-api-server/api_admin/api-general.rst
@@ -1,0 +1,74 @@
+.. Copyright © 2015-2018 the contributors (see Contributors.md).
+
+   This file is part of Knora.
+
+   Knora is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Affero General Public License as published
+   by the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   Knora is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public
+   License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
+
+
+Introduction: Using Admin API
+=============================
+
+RESTful API
+-----------
+
+Knora Admin API is a RESTful API that allows for reading and adding of administrative resources from and to Knora and
+changing their values using HTTP requests. The actual data is submitted as JSON (request and response format). The
+diverse HTTP methods are applied according to the widespread practice of RESTful APIs: GET for reading, POST for adding,
+PUT for changing resources and values, and DELETE to delete resources or values (see RESTful_API_).
+
+.. _RESTful_API: http://www.restapitutorial.com/lessons/httpmethods.html
+
+Knora IRIs
+----------
+
+Every resource that is created or hosted by Knora is identified by a unique id, a so called Internationalized Resource
+Identifier (IRI). The IRI is required for every API operation to identify the resource in question. A Knora IRI has
+itself the format of a URL. For some API operations, the IRI has to be URL-encoded (HTTP GET requests).
+
+Admin Path Segment
+------------------
+
+Every request to Admin API includes ``admin`` as a path segment, e.g. ``http://host/admin/users/http%3A%2F%2Frdfh.ch%2Fusers%2Froot``.
+
+Knora API Response Format
+-------------------------
+
+In case an API request could be handled successfully, Knora responds with a 200 HTTP status code. The actual answer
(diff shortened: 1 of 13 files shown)


musicEnfanthen:
The planned use of Swagger etc. is great. Big thumps up!
[Have the issue here, that it does not work with `0.0.0.0:3333` in the browser adress but only with `localhost:3333`, but maybe this is somehow dependent on my settings.]

benjamingeer:
@subotic Looking forward to trying this tomorrow. :)

webapi/src/main/scala/org/knora/webapi/Main.scala
```diff
@@ -52,6 +52,9 @@ object Main extends App with LiveCore with KnoraService {
     // starts zipkin monitoring reporter
     if (arglist.contains("-j")) applicationStateActor ! SetZipkinReporterState(true)
```
benjamingeer: Instead of parsing command-line arguments yourself, you could use a library. I've had good results with scallop:

https://github.com/scallop/scallop

webapi/src/main/scala/org/knora/webapi/routing/SwaggerApiDocsRoute.scala
```diff
@@ -0,0 +1,68 @@
...
+import com.github.swagger.akka.SwaggerHttpService
+import com.github.swagger.akka.model.Info
+import io.swagger.models.auth.BasicAuthDefinition
+import io.swagger.models.{ExternalDocs, Scheme}
+import org.knora.webapi.SettingsImpl
+import org.knora.webapi.routing.admin._
+
+import scala.concurrent.ExecutionContextExecutor
+
+class SwaggerApiDocsRoute(_system: ActorSystem, settings: SettingsImpl, log: LoggingAdapter) extends SwaggerHttpService {
```
benjamingeer: Please add a class comment explaining what this route does.

subotic:
I tried the ReDoc `index.html` yesterday by right-clicking it in IntelliJ and opening it in a browser, and it worked. I don't know if IntelliJ does any magic in the background, or if maybe something changed in the meantime. I will take a look.

benjamingeer:
I don't have that file in IntelliJ, because my IntelliJ just sees `webapi`, so I double-clicked the ReDoc `index.html` in the macOS Finder and it opened a browser window.

subotic:
@musicEnfanthen mentioned that he only got it working with `localhost:3333`. I will try to load it directly in the browser, as soon as I can change branches.

benjamingeer:
Something I don't understand: when I look at, say, the Knora users routes in http://petstore.swagger.io/, there is sample JSON for the requests and responses. Where does this come from? The `@ApiOperation` annotations in `UsersRouteADM` specify the case classes, but how does Swagger know what these will look like in JSON?

Also, is there a way to get IntelliJ to recognise the Swagger annotations so it doesn't think they're syntax errors?

subotic:
> but how does Swagger know what these will look like in JSON?

It knows because of `spray-json` and the definitions in `UsersADMJsonProtocol`

> Also, is there a way to get IntelliJ to recognise the Swagger annotations so it doesn't think they're syntax errors?

 Your probably using IntelliJ 2018.1 where the `scala` plugin is broken, so IntelliJ didn't pick up the Swagger libraries.

subotic:
https://youtrack.jetbrains.com/issue/SCL-13637

And I don't think that it will be fixed anytime soon. I opened an issue last week and downgraded IntelliJ.

benjamingeer:
> It knows because of spray-json and the definitions in UsersADMJsonProtocol

What if I don't use `spray-json` to generate JSON, because I'm generating JSON-LD, and because there isn't a simple one-to-one mapping between the case classes and the generated JSON-LD?

subotic:
> What if I don't use spray-json to generate JSON, because I'm generating JSON-LD, and because there isn't a simple one-to-one mapping between the case classes and the generated JSON-LD?

No idea. We will have to see how we can generate OpenAPI definitions for JSON-LD.

benjamingeer:
It looks like there's no support for JSON-LD in OpenAPI:

https://github.com/OAI/OpenAPI-Specification/issues/1100

I think this means it won't be possible to use Swagger to document Knora API v2 at all.

subotic:
> ReDoc didn't work for me

For me, it only works if it is served by IntelliJ. Opening it directly in Safari throws a CORS error. This is anyway meant to be served on the Knora website, alongside the other documentation.

webapi/src/main/scala/org/knora/webapi/Main.scala
subotic: I know, I looked at Scallop, but it looked complicated and I didn't want to spend to much time on it. Though if I add something else in the future, I will change to scallop.

subotic:
I will keep an eye on how we could also document JSON-LD. Maybe the OpenAPI people will come up with a way in the near future, or we role our own thing. We need to keep us busy after all :-)

PR #814, opened 2018-04-09
feature (webapi): allow loading data into graphdb on windows (internal PR)

musicEnfanthen:
This small PR dupplicates #683 , now as PR from inside the knora repo.

As described in the original PR, it adds the following to achieve some kind of windows compatibility of the webapi:

In order to get data loaded into graphdb on windows there are some small changes necessary to the load files in `webapi/scripts`: 

Because getting `Expect` run as executable under `usr/bin/expect` on windows is kind of hell, instead of `.expect` file I adapted the load script reusing the former `cat xxx.ttl | $CONSOLE` command that was removed in #607  . (Thanks to @subotic for pointing me to this solution.)
In accordance to this it was also necessary to recreate the former `graphdb-knora-test-data.ttl` file, now `win-graphdb-knora-test-data.ttl` . 

Also a small adaption to classpath in `openrdf-console/bin/console.sh` was necessary due to windows use of another file seperator (;) instead of (:).

I named the new files with `win-...`, so none of the existing POSIX files are touched.


webapi/scripts/openrdf-console/bin/win-console.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
subotic: This file is in a folder which is not managed by us. From time to time, we exchange the `openrdf-console` folder with a current version of the distribution, which will remove this file.

musicEnfanthen: Ok, I see. What do you recommend? Moving the win-file to `/scripts` or replacing it every time the `openrdf-console` is exchanged?

subotic: I have downloaded the newest version and now they have now a `console.bat`, so it should work under windows. So shouldn't be a problem when we exchange this folder.

musicEnfanthen: That's nice. Should we use the `console.bat` right now, so no further change is needed? But I can't see this file, wether on this branch nor on `develop`. If not, I can also merge now and change later on.

subotic: I didn't update the directory. I don't have time now to test all scripts.

PR #830, opened 2018-04-17
feature (function calls KnarQL): accept custom function calls in KnarQL

tobiasschweizer:
replaces #705 (the branch still has to be deleted)

requires https://github.com/dhlab-basel/Salsah/pull/235


webapi/src/test/scala/org/knora/webapi/e2e/v2/SearchRouteV2R2RSpec.scala
```diff
@@ -1782,5 +1782,44 @@ class SearchRouteV2R2RSpec extends R2RSpec {
...
+                  |     FILTER(lang(?text) = "en")
+                  |}
+                """.stripMargin
+
+            // TODO: find a better way to submit spaces as %20
+            Get("/v2/searchextended/" + URLEncoder.encode(sparqlSimplified, "UTF-8").replace("+", "%20")) ~> addCredentials(BasicHttpCredentials(anythingUserEmail, password)) ~> searchPath ~> check {
+
+                assert(status == StatusCodes.OK, response.toString)
+
+                // TODO: add some test data that have a language annotation and check for results
```
tobiasschweizer: @SepidehAlassi You would adapt this query to check some of your test data to make sure it really works.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -570,324 +570,373 @@ class SearchResponderV2 extends ResponderWithStandoffV2 {
 
                 case filterCompare: CompareExpression =>
 
-                    // left argument of a CompareExpression is expected to be a QueryVariable
-                    val queryVar: QueryVariable = filterCompare.leftArg match {
+                    // left argument of a CompareExpression is expected to be a QueryVariable or a 'lang' function call
+                    filterCompare.leftArg match {
```
benjamingeer: The method `transformFilterExpression` contains many long, deeply nested `match`-`case` expressions, which makes it very difficult to read. Could you refactor this method so that each `match`-`case` would fit on the screen? You could put each of these into its own method:

* `case filterCompare: CompareExpression =>`
  * `case queryVar: QueryVariable =>`
    * `case propInfo: PropertyTypeInfo =>`
      * `case iriRef: IriRef =>`
    * `case nonPropInfo: NonPropertyTypeInfo =>`
      * `case OntologyConstants.Xsd.Integer =>`
      * `case OntologyConstants.Xsd.Decimal =>`
      * `case OntologyConstants.Xsd.Boolean =>`
      * `case OntologyConstants.Xsd.String =>`
      * `case OntologyConstants.Xsd.Uri =>`
      * `case OntologyConstants.KnoraApiV2Simple.Date =>`
        * `case CompareExpressionOperator.EQUALS =>`
        * `case CompareExpressionOperator.NOT_EQUALS =>`
        * `case CompareExpressionOperator.LESS_THAN =>`
        * `case CompareExpressionOperator.LESS_THAN_OR_EQUAL_TO =>`
        * `case CompareExpressionOperator.GREATER_THAN =>`
        * `case CompareExpressionOperator.GREATER_THAN_OR_EQUAL_TO =>`
  * `case lang: LangFunction =>`
* `case filterOr: OrExpression =>`
* `case filterAnd: AndExpression =>`
* `case regexFunction: RegexFunction =>`
* `case functionCall: FunctionCallExpression =>`

SepidehAlassi:
Travis fails due to new changes in Develop. This branch should be updated after the fix.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
tobiasschweizer: @benjamingeer 

>  Could you refactor this method so that each match-case would fit on the screen?

Couldn't this be solved by buying a bigger screen for you? :-)

Just joking, I will do it ;-)

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -555,16 +555,60 @@ class SearchResponderV2 extends ResponderWithStandoffV2 {
...
+        /**
+          * Handles query variables that represent properties in a [[FilterPattern]].
+          *
+          * @param queryVar the query variable to be handled.
+          * @param comparisonOperator the comparison operator used in the filter expression.
+          * @param iriRef the Iri the property query variable is restricted to.
+          * @param propInfo information about the query variable's type.
+          * @return a [[TransformedFilterPattern]].
+          */
+        private def handlePropertyIriQueryVar(queryVar: QueryVariable, comparisonOperator: CompareExpressionOperator.Value, iriRef: IriRef, propInfo: PropertyTypeInfo): TransformedFilterPattern = {
```
tobiasschweizer: @benjamingeer Do you think it is ok to use `private` for these methods? The method will be only used within the same "template" (https://www.scala-lang.org/files/archive/spec/2.11/05-classes-and-objects.html#private).

also see https://stackoverflow.com/questions/44747514/how-to-inherit-a-protected-private-of-an-abstract-class-in-scala


https://stackoverflow.com/questions/44747514/how-to-inherit-a-protected-private-of-an-abstract-class-in-scala

benjamingeer: Yes, if subclasses don't need to call those methods, it's better for them to be private.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
tobiasschweizer: @benjamingeer I did the refactoring. Could please look at it again?

benjamingeer:
That's much easier to read now, thank you!

PR #826, opened 2018-04-16
Standoff Search

tobiasschweizer:
Search for text that is marked up in a certain way.

This PR is based on #811. Once that PR is merged, this PR will have to be merged back to `develop`.

relates to #813 (allow for KnarQL queries in complex api format)

closes #630

Issue #813, opened 2018-04-09
Allow querying for ListNodes


SepidehAlassi:
Followings should be added:
- route for requesting ListNodes
- query for ListNodes (functionality for Knarql)
Since this functionality is necessary for querying topics in BEOL, I would like that we prioritise resolving this issue.

tobiasschweizer:
The following query gets all resources of type `anything:Thing` that have a `anything:hasListItem`:

```sparql
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/simple/v2#>
PREFIX anything: <http://0.0.0.0:3333/ontology/anything/simple/v2#>

    CONSTRUCT {
        ?thing knora-api:isMainResource true .

        ?thing anything:hasListItem ?listItem .

    } WHERE {
        ?thing a knora-api:Resource .

        ?thing anything:hasListItem ?listItem .

        anything:hasListItem knora-api:objectType xsd:string .

        ?listItem a xsd:string .

    } OFFSET 0
```

If we want to restrict this query to a specific list item, we have to decide whether this should be an `xsd:string` or an Iri. We could modify the search logic for `xsd:string` in a way so that a value for the list node's`rdfs:label` can be used in the Filter:

```sparql
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/simple/v2#>
PREFIX anything: <http://0.0.0.0:3333/ontology/anything/simple/v2#>

    CONSTRUCT {
        ?thing knora-api:isMainResource true .

        ?thing anything:hasListItem ?listItem .

    } WHERE {
        ?thing a knora-api:Resource .

        ?thing anything:hasListItem ?listItem .

        anything:hasListItem knora-api:objectType xsd:string .

        ?listItem a xsd:string .

        FILTER(?listItem = "Baumlistenknoten 02")

    } OFFSET 0
```  

But an `rdfs:label` is not unique (and it could be also available in several languages). I would rather prefer an Iri to identify a list node. But the reason it is an `xsd:string` is the idea to serve the list item's label in simple schema for a list value.

@benjamingeer Could we think of a list item like a resource that is pointed to? Could we handle properties pointing to list items like resource references?

tobiasschweizer:
How is the use case of a list node different from a resource reference? Couldn't they be handled the same way?

[...]

Issue #630, opened 2017-10-11
Implement Standoff Search


tobiasschweizer:
Provide standoff search possibilities: search for a text that is marked up in a certain way.

An example in Sparql:

Search for the word "Mesure" that is marked up as italic and happens to be inside a paragraph. The paragraph does not need to be the immediate parent.
```
PREFIX standoff: <http://www.knora.org/ontology/standoff#>
PREFIX knora-base: <http://www.knora.org/ontology/knora-base#>
PREFIX beol: <http://www.knora.org/ontology/beol#>
select ?textValue ?markedup ?string where { 
    
    BIND("Mesure" AS ?searchVal)
    
    # use index for query optimisation 
    ?string <http://www.ontotext.com/owlim/lucene#fullTextSearchIndex> ?searchVal .
    
    ?textValue a knora-base:TextValue .
    
    ?textValue knora-base:valueHasString ?string .
    
    ?textValue knora-base:valueHasStandoff ?standoffNode .
        
    ?standoffNode a standoff:StandoffItalicTag .
    
    ?standoffNode knora-base:standoffTagHasStart ?start .
    
    ?standoffNode knora-base:standoffTagHasEnd ?end .
     
    # https://www.w3.org/TR/xpath-functions/#func-substring 
    # The first character of a string is located at position 1, not position 0. -> standoff uses a 0 based index
    BIND(SUBSTR(?string, ?start+1, ?end - ?start) AS ?markedup)
    
    FILTER REGEX(?markedup, ?searchVal, "i")
    
    ?standoffNode knora-base:standoffTagHasStartParent* ?standoffParentTag .
    ?standoffParentTag a standoff:StandoffParagraphTag .
} ORDER BY ?textValue ?start
LIMIT 100
```

tobiasschweizer:
Basic idea: use Lucene index to filter out all the text values that do not contain the search term (for optimization). Then select those text values that have an italic standoff tag that contains the search term: first get the whole text markup as italic and then check if it contains the search term using a FILTER with regex. Then check that the italic standoff node has some parent of type paragraph using property path syntax.

Performance: property path syntax is slow in our experience. So I expect queries making use of them to be slow.

Lucene and Regex: Both have their own syntax. We have to think about what possibilities we would like to offer to the user: Boolean Logic, wildcards etc.

tobiasschweizer:
@benjamingeer suggestes

[...]

PR #858, opened 2018-05-16
feature (search v2): limit fulltext search to standoff class

tobiasschweizer:
This PR replaces https://github.com/dhlab-basel/Knora/pull/826 (the branch `wip/standoff_search` still has to be deleted once this PR is merged).

This PR provides the possibility to restrict a fulltext search to a standoff class.

### DIFF
--- a/docs/src/paradox/03-webapi/api-v2/reading-and-searching-resources.md
+++ b/docs/src/paradox/03-webapi/api-v2/reading-and-searching-resources.md
@@ -125,7 +125,8 @@ combined using the Boolean `AND` operator. Please note that the search
 terms have to be URL-encoded.
 
 ```
-HTTP GET to http://host/v2/search/searchValue[limitToResourceClass=resourceClassIRI][limitToProject=projectIRI][offset=Integer]
+HTTP GET to http://host/v2/search/searchValue[limitToResourceClass=resourceClassIRI]
+[limitToStandoffClass=standoffClassIri][limitToProject=projectIRI][offset=Integer]
 ```
 
 Please note that the first parameter has to be preceded by a question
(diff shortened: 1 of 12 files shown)


tobiasschweizer:
@benjamingeer I had to reapply the changes to a new branch because I got a lot of merge conflicts. The previous branch was based on the standoff PR that has been merged. This looked good when I changed the base branch to develop. Bt then I got a lot of merge conflicts in files that I haven't changed in the PR itself, but for the standoff PR.

tobiasschweizer:
Ah, I should add the new param to the docs!

benjamingeer:
> Am I right to assume that it makes no difference whether the user submits Iri params in complex or simple schema since they are converted to the internal one anyway?

I think so, but try it and see. Also, if we do what we talked about in #813, I think we should remove standoff completely from the simple schema. So then we shouldn't allow standoff class IRIs in the simple schema in requests.

tobiasschweizer:
I tried it, it makes no difference at all. The response format has to be chosen either via the header or the "schema" param. 

So I can mere this PR?

benjamingeer:
OK, just merge it then.

tobiasschweizer:
Ok, thx!

mattssp:
Hi @tobiasschweizer @benjamingeer , if I should want to search for a knora-resource that is referenced by a standoffLinkTag that contains a search phrase, would that be possible with KNARL (even if you remove standoff from the simple schema)? Or would it be useful to implement this as a feature in knora?

tobiasschweizer:
@mattssp This is a closed PR, could you please comment here #630 instead?

Maybe #813 helps here: we plan to add support for the complex schema in KnarQL, so one could search for standoff.

PR #883, opened 2018-06-06
Allow a nested block in an OPTIONAL in Gravsearch

### LINKED ISSUE #138: Which things in knora-base can/must be used as is?

benjamingeer:
- [x] Allow nesting in OPTIONAL (but still not in UNION).
- [x] Add tests.

Resolves #138.

@gfoo Thanks for the test case. Does this do what you need?


benjamingeer:
@gfoo Thanks for the review!

PR #892, opened 2018-06-15
doc (webapi): add UBKVP shortcode to the list

subotic:
### Summary

-  add UBKVP shortcode to the list


subotic:
@benjamingeer this is a tough one ;-)

PR #895, opened 2018-06-18
build (graphdb): changes to GraphDB repository initialization

### LINKED ISSUE #893: Load built-in Knora ontologies as read-only imports in GraphDB
I noticed [this interesting thing](http://graphdb.ontotext.com/documentation/standard/reasoning.html#retraction-of-assertions) in the GraphDB docs about reasoning:

> The forward-chaining part of the algorithm terminates as soon as it detects that a statement is read-only, because if it cannot be deleted, there is no need to look for statements derived from it. For this reason, performance can be greatly improved when all schema statements are made read-only by importing ontologies (and OWL/RDFS vocabularies) using the imports repository parameter.

So I guess we should tell it to load all the `.ttl` files in `Knora/knora-ontologies` as read-only imports, like this:

http://graphdb.ontotext.com/documentation/standard/configuring-a-repository.html#configuring-a-repository-configuration-parameters-imports

### LINKED ISSUE #898: Use  owlim:MonitorRepository as repository type
It seems that we initialize our repositories in the wrong way. We have `rep:repositoryType "openrdf:SailRepository" ;` but according to Ontotext we should have `rep:repositoryType "owlim:MonitorRepository"`.  At least, this was the explanation given by them for a ticket that I opened regarding backup restore.

- http://graphdb.ontotext.com/documentation/standard/configuring-a-repository.html#configure-a-repository-programmatically

subotic:
### Summary

- [ ] import system ontologies
- [ ] change repository type to `owlim:MonitorRepository`

### Issues

- resolves #893
- resolves #898


subotic:
work moved to #921

PR #896, opened 2018-06-18
feature (webapi): startup state + health endpoint

### LINKED ISSUE #794: We should probably remove the 'loadDemoData' flag
The `loadDemoData` flag is of limited use, as it can only work if `webapi` is started through SBT. If a user manages to start `webapi` through SBT, then he should be able to run one of the repository initialization scripts that we have.

On the other hand, if we want to have a really easy way for someone to try out the Knora stack, then we should probably think of a "fat" docker image, where everything that is needed (graphdb-free, webapi, sipi, salsah) is included, initialized and started when the user runs this image.

### LINKED ISSUE #891: Knora-Service should wait for GraphDB on startup and provide information about current state
Currently, `KnoraService` requires a running GraphDB instance. In production, this is difficult to guaranty with container orchestration since the GraphDB docker container is started first, but there is no way to know how long GraphDB will take to actually start.

I suppose that also the bigger the GraphDB triplestore gets, the longer it will take.

I propose to start ontology initialization only after GraphDB is started, which needs to be detected somehow. Further, I would like to see that the HTTP server is started immediately so that monitoring and health endpoints could be queried. They can provide information about the state of the `Knora-Service`, e.g., `Waiting for GraphDB`, `Initializing Ontologies`, `Ready`. Only when the startup procedure is finished, the `/admin`, `/v1`, and `/v2` routes should be activated.

The startup workflow:

1. start HTTP server
1. only `/health` and `/monitor` routes should be active, all other routes should return `503 Service Unavailable` with a description of the state, e.g., waiting for GraphDB
1. start periodic checking if GraphDB is up
1. run `LoadOntologiesRequestV2`
1. enable `/admin`, `/v1`, and `/v2`

subotic:
### Summary

- [x] track application state on startup and trigger loading of ontologies when DB is ready
- [x] add `/health` endpoint
- [x] removes `loadDemoData`
- [x] fix E2E tests

### Issues

- resolves #794
- resolves #891


subotic:
@benjamingeer please review whenever you have a spare moment :-)

benjamingeer:
Sorry, missed this, thanks for reminding me. Will review tomorrow morning.

subotic:
@benjamingeer whenever you have time ;-)

benjamingeer:
Sorry, started looking at this this morning, then got stuck in bugs, continuing to look at it now. :)

webapi/src/main/scala/org/knora/webapi/routing/HealthRoute.scala
```diff
@@ -0,0 +1,129 @@
...
+                ).compactPrint
+            )
+        )
+    }
+
+    private def status(s: Boolean) = if (s) "healthy" else "unhealthy"
+
+    private def statusCode(s: Boolean) = if (s) StatusCodes.OK else StatusCodes.ServiceUnavailable
+
+    private def unhealthy(str: String) = {
```
benjamingeer: Poor app, should see a doctor. :)

webapi/src/main/scala/org/knora/webapi/KnoraService.scala
```diff
@@ -241,20 +410,31 @@ trait KnoraService {
...
 
     /**
-      * Stops Knora.
+      * Checks if application state is 'Running' and only then returns.
       */
-    def stopService(): Unit = {
-        system.terminate()
-        CacheUtil.removeAllCaches()
-        //Kamon.shutdown()
+    private def waitForRunningState(): Unit = {
```
benjamingeer: This method is never called, can it be deleted?

subotic: yes, forgot to remove it.

subotic:
@benjamingeer so, how is the review going? ;-)

webapi/src/main/scala/org/knora/webapi/routing/HealthRoute.scala
subotic: :-)

benjamingeer:
> so, how is the review going? ;-)

Done, just waiting for an answer to this question:

> When I do `reStart`, I get `webapi (checkRepository - exception,java.util.concurrent.TimeoutException: Futures timed out after [500 milliseconds])`, even though GraphDB is already running; is that normal?

Otherwise the whole thing looks great.

subotic:
> When I do reStart, I get webapi (checkRepository - exception,java.util.concurrent.TimeoutException: Futures timed out after [500 milliseconds]), even though GraphDB is already running; is that normal?

sorry, didn't see the question., Yes, this is normal. The exception is caught by a try/catch. Maybe I could print out a nicer message. Will do that in a future revision. I thought of a few more cases that need to be handled. But will do it in another PR.

subotic:
thanks :-)

PR #899, opened 2018-06-19
Gravsearch search for list value using complex schema

### LINKED ISSUE #813: Allow querying for ListNodes

benjamingeer:
- [x] Support existing query functionality using the complex schema.
- [x] Check that if the complex schema is used, the simple schema is allowed only in date comparisons in `FILTER` expressions.
- [x] Support searching for list values with particular list nodes using the complex schema.
- [x] Reject properties that don't exist in the schema in which they're submitted.
- [x] Don't allow entities from `rdf`, `rdfs`, or `owl` in a Gravsearch `CONSTRUCT` clause.
- [x] Reject properties that aren't (yet) supported in Gravsearch.
- [x] Type inference improvements:
  - [x] Don't run a rule more than once if it can never return new information after the first iteration.
  - [x] Infer the type of a variable used in a function.
- [x] Add tests.
- [x] Update docs.
- [x] Update release notes.

Resolves #813.


benjamingeer:
TODO: make a second inference rule pipeline, excluding rules that never return new information when run more than once.

benjamingeer:
TODO: if the complex schema is used anywhere in the query, don't allow the simple schema anywhere except in date comparisons.

benjamingeer:
Rules for using the complex schema:

* If the complex schema is used at all in the query, the whole query must be in the complex schema, except for a date literal compared with the return value of the function `knora-api:toSimpleDate`.
* To filter on a literal other than a date, use the predicates `knora-api:tintValueAsInt`, etc.
* Items below the value level may not be mentioned in the `CONSTRUCT` clause or in ORDER BY.
* To order by values, use value object variables in `ORDER BY`, rather than the literals they contain.
* To get the language of a text value, use `knora-api:textValueHasLanguage`.
* The `match` and `regex` functions work on the objects of `knora-api:valueAsString`.

TODO: Do something more sensible if the query contains `rdfs:label`.

benjamingeer:
@tobiasschweizer I disallowed `rdf:type` in Gravsearch `CONSTRUCT` clauses, because it actually serves no purpose there. Will that require a change in SALSAH? Also, could you maybe add a more realistic test with a list value from BEOL?

tobiasschweizer:
@benjamingeer I do not think so, but I will check on Monday!

tobiasschweizer:
>  I disallowed rdf:type in Gravsearch CONSTRUCT clauses, because it actually serves no purpose there. Will that require a change in SALSAH?

I checked the Gravsearch query templates in SALSAH and they work with this branch. 

> Also, could you maybe add a more realistic test with a list value from BEOL?

You mean a v2 search route test for a beol list value? Sure, I will do that now.

benjamingeer:
> You mean a v2 search route test for a beol list value? Sure, I will do that now.

Yes, thanks.

tobiasschweizer:
@benjamingeer and here you go: https://github.com/dhlab-basel/Knora/pull/899/commits/81643cd40a4dbbedf13f88549cc1d3ecb3e8bd55

benjamingeer:
@tobiasschweizer Great, thank you! Would you have time to finish reviewing this today?

docs/src/paradox/03-apis/api-v2/query-language.md
```diff
@@ -182,81 +214,179 @@ is already known, e.g. to request specific information about that resource and
...
-query. To restrict a value, a FILTER must be used. Without a FILTER, all
-the instances of a value are returned.
+- Text values (`xsd:string`)
+- Integer values (`xsd:integer`)
+- Decimal values (`xsd:decimal`)
+- Boolean values (`xsd:boolean`)
+- Date values (`knora-api:Date`)
 
-#### Functions in FILTER Expressions
+A Knora value may not be represented as the literal object of a predicate;
```
tobiasschweizer: I think we should see how hard it is to allow literals in the simple schema. The main reason why we forbade this was because it makes the processing harder in `SearchResponderV2`.

In simple schema, we have to add an extra level by generating a statement for `valueHasString`. However, from the user's perspective it is rather hard to  understand why Gravsearch supports literals in complex schema while it does not in simple schema.

tobiasschweizer:
> Would you have time to finish reviewing this today?

I am doing it now :-)

docs/src/paradox/03-apis/api-v2/query-language.md
benjamingeer: If you used a literal instead of a variable, that would mean you couldn't mention that value in the CONSTRUCT clause, so wouldn't it be excluded from the results? I think this is rarely what the user would want.

benjamingeer: By the way, the rule is currently the same for both schemas: you can't represent a Knora value as the literal object of a predicate. In the complex schema, you can use literals for things below the value level, but not for the value itself.

tobiasschweizer: > If you used a literal instead of a variable, that would mean you couldn't mention that value in the CONSTRUCT clause, so wouldn't it be excluded from the results?

I think in that case you would just use the same statement:

```sparql
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/simple/v2#>

    CONSTRUCT {

        ?letter ?linkingProp1  ?person1 .

        ?person1 beol:hasFamilyName "Meier" .

    } WHERE {
      
        ?letter a beol:letter .

        ?letter ?linkingProp1 ?person1 .
        FILTER(?linkingProp1 = beol:hasAuthor || ?linkingProp1 = beol:hasRecipient )

        ?person1 beol:hasFamilyName "Meier".

    } ORDER BY ?date
```

I think in a first version of Gravsearch, we allowed for literals. But when I refactored it, I did not support them anymore because the processing logic would have been more complex.

In `webapi/src/test/scala/org/knora/webapi/e2e/v2/SearchRouteV2R2RSpec.scala` there are still some old tests involving literals that are ignored.

benjamingeer: OK, let's look at this after standoff search is done.

tobiasschweizer:
> OK, let's look at this after standoff search is done.

Yes, absolutely! I do not think it is urgent since you can do everything with a `Filter` that you can do with a literal.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -2216,22 +2178,25 @@ class SearchResponderV2 extends ResponderWithStandoffV2 {
             typeInspectionResult: GravsearchTypeInspectionResult <- gravsearchTypeInspectionRunner.inspectTypes(inputQuery.whereClause, requestingUser)
             whereClauseWithoutAnnotations: WhereClause = GravsearchTypeInspectionUtil.removeTypeAnnotations(inputQuery.whereClause)
 
-            // Preprocess the query to convert API IRIs to internal IRIs and to set inference per statement.
-
-            preprocessedQuery: ConstructQuery = QueryTraverser.transformConstructToConstruct(
```
tobiasschweizer: Did you get rid of this because in the old way, we just converted everything to internal assuming that we got it in simple schema. Now, we have to check if we get it in simple or complex.

benjamingeer: Yes.

webapi/src/test/resources/test-data/ontologyR2RV2/standoffOntologySimple.rdf
```diff
@@ -12,557 +12,557 @@
...
 	<rdfs:subClassOf rdf:resource="http://api.knora.org/ontology/standoff/simple/v2#StandoffVisualTag"/>
-	<rdfs:subClassOf rdf:nodeID="genid-55a441049d76415ba3999f7d47ee542f-b8"/>
-	<rdfs:subClassOf rdf:nodeID="genid-55a441049d76415ba3999f7d47ee542f-b9"/>
-	<rdfs:subClassOf rdf:nodeID="genid-55a441049d76415ba3999f7d47ee542f-b10"/>
-	<rdfs:subClassOf rdf:nodeID="genid-55a441049d76415ba3999f7d47ee542f-b11"/>
-	<rdfs:subClassOf rdf:nodeID="genid-55a441049d76415ba3999f7d47ee542f-b12"/>
-	<rdfs:subClassOf rdf:nodeID="genid-55a441049d76415ba3999f7d47ee542f-b13"/>
-	<rdfs:subClassOf rdf:nodeID="genid-55a441049d76415ba3999f7d47ee542f-b14"/>
-	<rdfs:subClassOf rdf:nodeID="genid-55a441049d76415ba3999f7d47ee542f-b15"/>
+	<rdfs:subClassOf rdf:nodeID="genid-0e6b49fe8a374047959ae8a349b1a4de-b8"/>
```
tobiasschweizer: Are these blank node Iris?

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -390,81 +295,101 @@ class SearchResponderV2 extends ResponderWithStandoffV2 {
...
 
-                    // linking property: just include the original statement relating the subject to the target of the link
-                    statementPattern +: linkValueStatements
+                // linking property: just include the original statement relating the subject to the target of the link
+                statementPatternToInternalSchema(statementPattern, typeInspectionResult) +: linkValueStatements
 
-                case _ =>
-                    // value property
+            } else {
+                // Is the subject a resource?
```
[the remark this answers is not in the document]
benjamingeer: Yes. Clarified the comments there.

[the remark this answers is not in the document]
tobiasschweizer: Thanks!

webapi/src/test/resources/test-data/ontologyR2RV2/standoffOntologySimple.rdf
benjamingeer: Yes, for cardinalities.

tobiasschweizer: But aren't those different for every request you make?

benjamingeer: Yes. Every time you convert a JSON-LD document containing blank nodes to RDF/XML, you get different randomly generated blank node IDs. Semantically it makes no difference.

tobiasschweizer: I just wondered how the tests work. I assume the blank node Iris are just ignored, right?

benjamingeer: The tests use RDF4J Rio to parse the RDF/XML into an `org.eclipse.rdf4j.model.Model`. If you compare two `Model` objects that contain blank nodes using the `==` operator, the `Model` class understands that the blank node IDs aren't significant. It just compares their contents, not their IDs.

tobiasschweizer: Ok, got it. Thanks!

PR #909, opened 2018-06-25
Exclude list values from fulltext search

### LINKED ISSUE #906: v2 wildcard fulltext search error
I get this error with this request `http://localhost:3333/v2/search/*i*`:
```json
{
error: "org.knora.webapi.InconsistentTriplestoreDataException: Expected http://www.w3.org/2000/01/rdf-schema#label in assertions for a value object of type list value."
}
```
It works with:
- two letters : `http://localhost:3333/v2/search/*ii*`
- **v1** : `http://localhost:3333/v2/search/*i*?searchtype=fulltext`

Update doc ? not found any mentions of:
- the wildcard for full text search
- the 3 characters required for the search value

tobiasschweizer:
List values are excluded from fulltext search because `knora-base:valueHasString` contains the list node's Iri. The only real use case would be a comment on the list value.

resolves #906

### DIFF
--- a/docs/src/paradox/03-apis/api-v2/reading-and-searching-resources.md
+++ b/docs/src/paradox/03-apis/api-v2/reading-and-searching-resources.md
@@ -161,6 +161,20 @@ HTTP GET to http://host/v2/search/searchValue[limitToResourceClass=resourceClass
 Please note that the first parameter has to be preceded by a question
 mark `?`, any following parameter by an ampersand `&`.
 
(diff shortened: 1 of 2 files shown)


tobiasschweizer:
see https://github.com/dhlab-basel/Knora/issues/813#issuecomment-399863278

tobiasschweizer:
> Please update the release notes, too.

done in https://github.com/dhlab-basel/Knora/pull/909/commits/831c524d56b2fdc409ea2040395823e398877854

Issue #889, opened 2018-06-14
Gravsearch response content and type support 


gfoo:
I summarize here all what I would (finally) need in my project to fully use Gravsearch : 
- Resource default response content (like "rdfs:label"):
  - creationDate (+ order by)
  - lastModificationDate (+ order by)
  - attachedToUser
- ~~knora-base:UriValue type support~~ (not relevant, see next answer)

related to #802 #793

Thanks a lot !

n.b. : now, I only use Gravsearch to retrieve and search for data in my project (oh yeah !), works like a charm as we say today 👍

benjamingeer:
Support for `creationDate`, `lastModificationDate`, and `attachedToUser` should be easy to add. But I would like to do this after #813 and #630.

gfoo:
`xsd:anyURI` works for me, many thanks.  (May be update the doc as well: https://docs.knora.org/paradox/03-apis/api-v2/query-language.html)

tobiasschweizer:
@gfoo Don't be afraid, I am just playing around with the emojis :-)

benjamingeer:
@gfoo Do you still need this functionality, or can I close this issue like #793?

gfoo:
You can close, thx.

Issue #763, opened 2018-03-05
refactor splitMainResourcesAndValueRdfData so it uses SparqlExtendedConstructResponse


tobiasschweizer:
- use `SparqlExtendedConstructResponse` so the language tag can be taken into account
- do not return a list node's label for a list value. Instead, the client has to make a request to the v2 list route (requires a change in SALSAH)
- make case class for `Map[IRI, ConstructResponseUtilV2.ResourceWithValueRdfData]`
- make case class for `Map[IRI, Seq[ValueRdfData]]`

tobiasschweizer:
```
 ```
def createValueContentV2FromValueRdfData(valueObject: ValueRdfData, mappings: Map[IRI, MappingAndXSLTransformation]): ValueContentV2 = {
        val stringFormatter = StringFormatter.getGeneralInstance

        // every knora-base:Value (any of its subclasses) has a string representation
        val valueObjectValueHasString: String = valueObject.assertions(OntologyConstants.KnoraBase.ValueHasString).head.toString

        // every knora-base:value (any of its subclasses) may have a comment
        val valueCommentOption: Option[String] = valueObject.assertions.get(OntologyConstants.KnoraBase.ValueHasComment) match {
            case Some(literals: Seq[LiteralV2]) => Some(literals.head.toString)
            case None => None
        }

        valueObject.valueObjectClass match {
            case OntologyConstants.KnoraBase.TextValue =>

                if (valueObject.standoff.nonEmpty) {
                    // standoff nodes given
                    // get the IRI of the mapping
                    val mappingIri: IRI = valueObject.assertions.getOrElse(OntologyConstants.KnoraBase.ValueHasMapping, throw InconsistentTriplestoreDataException(s"no mapping IRI associated with standoff belonging to textValue ${valueObject.valueObjectIri}")).head.toString

                    val mapping: MappingAndXSLTransformation = mappings(mappingIri)

                    val standoffTags: Vector[StandoffTagV1] = StandoffTagUtilV1.createStandoffTagsV1FromSparqlResults(mapping.standoffEntities, valueObject.standoff.map {
                        case (standoffNode: IRI, assertion: Map[IRI, Seq[LiteralV2]]) =>
                            (standoffNode, assertion.map {
                                case (standoffProp, standoffObjs) =>
                                    (standoffProp, standoffObjs.head.toString)
                            })
                    })

[...]

PR #910, opened 2018-06-25
Query standoff markup in Gravsearch

### LINKED ISSUE #630: Implement Standoff Search

benjamingeer:
- [x] Find standoff tags that enclose text containing search terms (optimised using the full-text search index).
- [x] Search for standoff link tags by link target.
- [x] Search for standoff date tags by comparing the date to a `knora-api-simple:Date` in a `FILTER`.
- [x] Support `owl:TransitiveProperty` and `standoffTagHasStartAncestor` as per https://github.com/dhlab-basel/Knora/issues/630#issuecomment-396976505.
- [x] Don't allow project-specific ontologies to use `owl:TransitiveProperty`.
- [x] Type-check all typeable entities used as function arguments.
- [x] Add docs.
- [x] Add tests.
- [x] Update release notes.

To get support for `standoffTagHasStartAncestor`, you have to recreate your repository and restart GraphDB.

Resolves #630.


benjamingeer:
@tobiasschweizer I'm having a problem with the processing of the results of the main query. I want to write a query that finds the letter `http://rdfh.ch/0801/beol/-0tI3HXgSSOeDtkf-SA00w`, which has a standoff link to Claude Jordan (GND `"(VIAF)271899510"`).

This query works:

```sparql
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX knora-api-simple: <http://api.knora.org/ontology/knora-api/simple/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/v2#>

CONSTRUCT {
    ?letter beol:hasText ?text .
} WHERE {
    ?letter a beol:letter .
    FILTER(knora-api:toSimpleDate(?date) = "GREGORIAN:1756-01-03 CE"^^knora-api-simple:Date)
    ?letter beol:letterHasLanguage ?language .
    ?language knora-api:valueAsString "German" .
    ?letter beol:hasText ?text .
    ?letter knora-api:hasStandoffLinkTo ?person .
    ?person a beol:person .
    ?person beol:hasIAFIdentifier ?iafIdentifier .
    ?iafIdentifier knora-api:valueAsString "(VIAF)271899510" .
}
```

But this one returns `ForbiddenResource`:

```sparql
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX knora-api-simple: <http://api.knora.org/ontology/knora-api/simple/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/v2#>

CONSTRUCT {
    ?letter beol:hasText ?text .
} WHERE {
    ?letter a beol:letter .
    FILTER(knora-api:toSimpleDate(?date) = "GREGORIAN:1756-01-03 CE"^^knora-api-simple:Date)
    ?letter beol:letterHasLanguage ?language .
    ?language knora-api:valueAsString "German" .
    ?letter beol:hasText ?text .
    ?text knora-api:textValueHasStandoff ?standoffTag .
    ?standoffTag a knora-api:StandoffLinkTag .
    ?standoffTag knora-api:standoffTagHasLink ?person .
    ?person a beol:person .
    ?person beol:hasIAFIdentifier ?iafIdentifier .
    ?iafIdentifier knora-api:valueAsString "(VIAF)271899510" .
}
```

Could it be that `?person` is being mistaken for a dependent resource, even though it isn't mentioned in the `CONSTRUCT` clause? I'm guessing that it has to do with this code in `SearchResponderV2`, but I'm having trouble understanding how it works:

```scala
// get all the IRIs for variables representing dependent resources per main resource
val dependentResourceIrisPerMainResource: Map[IRI, Set[IRI]] = prequeryResponse.results.bindings.foldLeft(Map.empty[IRI, Set[IRI]]) {
    case (acc: Map[IRI, Set[IRI]], resultRow: VariableResultsRow) =>
        // collect all the values for the current main resource from prequery response

        val mainResIri: String = resultRow.rowMap(mainResourceVar.variableName)

        val dependentResIris: Set[IRI] = dependentResourceVariablesConcat.flatMap {
            dependentResVar: QueryVariable =>

                // check if key exists (the variable could be contained in an OPTIONAL or a UNION)
                val dependentResIriOption: Option[IRI] = resultRow.rowMap.get(dependentResVar.variableName)

                dependentResIriOption match {
                    case Some(depResIri: IRI) =>

                        // IRIs are concatenated, split them
                        depResIri.split(nonTriplestoreSpecificConstructToSelectTransformer.groupConcatSeparator).toSeq

                    case None => Set.empty[IRI] // no value present
                }

        }

        acc + (mainResIri -> dependentResIris)
}
```

tobiasschweizer:
Are you logged in?

tobiasschweizer:
@benjamingeer Do you want me to have look today?

benjamingeer:
> Do you want me to have look today?

Yes please.

tobiasschweizer:
@benjamingeer I think there is a problem in the way the presence is checked for everything specified in the Where clause after permission checking has been done. I would have to look at it more carefully another time.

However, as a temporary solution just bypass it:

https://github.com/dhlab-basel/Knora/blob/af42f6dd709ae75d7a87fbd98967be46c3b5cb0e/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala#L2626

-> `                } yield queryResultsSep //queryResWithFullQueryPathOnlyRequestedValues`

Like this, I get the result which means that the query works :-)

tobiasschweizer:
Ok, this where the resource is filtered out:

https://github.com/dhlab-basel/Knora/blob/af42f6dd709ae75d7a87fbd98967be46c3b5cb0e/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala#L2504-L2532

Maybe it is related to the collection of dependent resource Iris (as you suspected https://github.com/dhlab-basel/Knora/pull/910#issuecomment-400350921). I will have a look at it!

benjamingeer:
> I will have a look at it!

That would be great, this PR can't be merged before this is fixed.

tobiasschweizer:
Note for myself:

```
println("dep res: " + expectedDependentResources .diff(resAndValueObjIris.resourceIris))
println("val objs " + expectedValueObjects.diff(resAndValueObjIris.valueObjectIris))
```

-> 

> dep res: Set(http://rdfh.ch/biblio/up0Q0ZzPSLaULC2tlTs1sA)
> val objs Set(http://rdfh.ch/biblio/up0Q0ZzPSLaULC2tlTs1sA/values/w14VkrZDSeuZv1qwcJGUmA)

It seems that the main query gets information about the linked person, but the link value is not included as a nested resource.

tobiasschweizer:
@benjamingeer I have an idea, but I would still need to verify that:

I think the person is not included as a nested resource because the `hasStandoffLinkToValue` (reification) is not queried. Normally, the necessary statements to query the reification is created from the linking property. But in that case it is not since we ask for the standoff tag's property.

Maybe this is logic that has to be added to the `SearchResponderV2`so that this is done automatically when processing standoff tag properties.

benjamingeer:
Gravsearch re-implementation of full-text search restricted to a standoff tag:

```sparql
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX knora-api-simple: <http://api.knora.org/ontology/knora-api/simple/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/v2#>

CONSTRUCT {
    ?letter beol:hasText ?text .
} WHERE {
    ?letter a beol:letter .
    ?letter beol:hasText ?text .
    ?text knora-api:valueAsString ?textStr .
    ?text knora-api:textValueHasStandoff ?standoffTag .
    ?standoffTag a standoff:StandoffParagraphTag .
    FILTER(knora-api:matchInStandoff(?textStr, ?standoffTag, "Euler"))
}
```

benjamingeer:
Which generates this prequery:

```sparql
SELECT DISTINCT ?letter (GROUP_CONCAT(DISTINCT(?text); SEPARATOR='') AS ?text__Concat)
WHERE {
?letter <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
    ?letter <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
}
?letter <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/0801/beol#letter> .
?letter <http://www.knora.org/ontology/0801/beol#hasText> ?text .
    ?text <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
}
?text <http://www.knora.org/ontology/knora-base#valueHasString> ?textStr .
?text <http://www.knora.org/ontology/knora-base#valueHasStandoff> ?standoffTag .
?standoffTag <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/standoff#StandoffParagraphTag> .
    ?textStr <http://www.ontotext.com/owlim/lucene#fullTextSearchIndex> "Euler"^^<http://www.w3.org/2001/XMLSchema#string> .
}
    ?standoffTag <http://www.knora.org/ontology/knora-base#standoffTagHasStart> ?standoffTag__start .
}
    ?standoffTag <http://www.knora.org/ontology/knora-base#standoffTagHasEnd> ?standoffTag__end .
}
BIND(substr(?textStr, ?standoffTag__start + 1, ?standoffTag__end - ?standoffTag__start) AS ?standoffTag__markedUp)
FILTER(regex(?standoffTag__markedUp, "Euler", "i"))
}
GROUP BY ?letter
ORDER BY ASC(?letter)
LIMIT 25
```

benjamingeer:
After discussion with @tobiasschweizer:

How to deal with the permissions issue: try adding statements to match the `LinkValue` for the property `hasStandoffLinkToValue` whose target resource is `?person`. We know when we have to do this because we know that `?person` is a resource.

tobiasschweizer:
Another thing to keep in mind is the Apache Lucene Syntax: https://lucene.apache.org/core/3_6_2/queryparsersyntax.html and the syntax for the SPARQL Regex function since we pass the same search term to them (wildcards etc.).

benjamingeer:
The problem is how to find out what resource is the source of the link here:

```
 ?standoffTag knora-api:standoffTagHasLink ?person .
```

The search responder could generate:

```
?standoffTag__linkSource knora-base:hasValue ?standoffTag__textValue .
?standoffTag__textValue knora-base:valueHasStandoff ?standoffTag .
```

But this would be redundant if the user has already put statements like that in the query. I don't know if this would make the query slower. I can test it.

The alternative would be to force the user to use a custom function like this:

```
FILTER(knora-api:standoffLink(?letter, ?standoffTag, knora-api:standoffTagHasLink, ?person)
```

Which would generate the same thing.

benjamingeer:
TODO:

* Include function arguments in typeable entities.
* Finish incomplete type-checking of function arguments, including newly added functions.

benjamingeer:
It works!

```sparql
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX knora-api-simple: <http://api.knora.org/ontology/knora-api/simple/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/v2#>

CONSTRUCT {
    ?letter beol:hasText ?text .
} WHERE {
    ?letter a beol:letter .
    FILTER(knora-api:toSimpleDate(?date) = "GREGORIAN:1756-01-03 CE"^^knora-api-simple:Date)
    ?letter beol:letterHasLanguage ?language .
    ?language knora-api:valueAsString "German" .
    ?letter beol:hasText ?text .
    ?text knora-api:textValueHasStandoff ?standoffTag .
    ?standoffTag a knora-api:StandoffLinkTag .
    FILTER knora-api:standoffLink(?letter, ?standoffTag, ?person)
    ?person a beol:person .
    ?person beol:hasIAFIdentifier ?iafIdentifier .
    ?iafIdentifier knora-api:valueAsString "(VIAF)271899510" .
}
```

benjamingeer:
@tobiasschweizer To make a test with a `StandoffDateTag`, I need an XML-to-standoff mapping for it, right? I think there's no `MappingElement` for `StandoffDateTag` in the `StandardMapping`, should I add one?

tobiasschweizer:
The `StandardMapping` complies with the configuration of CKEditor in the Salsah 1 GUI.

For testing you could use `_test_data/test_route/texts/mappingForHTML.xml` and `_test_data/test_route/texts/HTML.xml`. The letter contains markup like this:

```
On the <span class="event" data-description="New year's eve" data-date="GREGORIAN:2016-12-31 CE">31st of December 2016, we will have a party</span>.
```

You could search for `party` that is marked up as a Knora date.

You could work with the existing tests:

https://github.com/dhlab-basel/Knora/blob/e099b18ae8cfc7893c44b6b6ef244c0e291a2cb8/webapi/src/test/scala/org/knora/webapi/e2e/v1/StandoffV1R2RSpec.scala#L846-L875

https://github.com/dhlab-basel/Knora/blob/e099b18ae8cfc7893c44b6b6ef244c0e291a2cb8/webapi/src/test/scala/org/knora/webapi/e2e/v1/StandoffV1R2RSpec.scala#L931-L957

benjamingeer:
@tobiasschweizer I tried that (see the last test in `SearchRouteV2R2RSpec`, but it doesn't work, and I think there's a bug in the code that converts XML to standoff when creating a text value. The text value that's created is supposed to contain an `anything:StandoffEventTag`, but actually it contains a `knora-base:StandoffDateTag`. Could you have a look?

benjamingeer:
Never mind, my mistake. 😜

tobiasschweizer:
but I see a datetag too. Why is that?

benjamingeer:
> but I see a datetag too. Why is that?

You're right, now I'm really confused.

benjamingeer:
Ah, never mind, it's because there is actually a date tag in the input, as well as an event tag.

tobiasschweizer:
Yes, there are two dates:

```
<span class="date" data-date="GREGORIAN:2016-12-24 CE">Christmas 2016</span>

<span class="event" data-description="New year's eve" data-date="GREGORIAN:2016-12-31 CE">31st of December 2016, we will have a party</span>

```

eventtag

datetag

benjamingeer:
I was too stupid to look carefully. :)

tobiasschweizer:
@benjamingeer No problem, I am actually quite excited to see that search work!

Did you check the prequery? Maybe the necessary statements for date comparisons are not correct for the standoff tag.

benjamingeer:
Looks OK to me:

```sparql
SELECT DISTINCT ?thing (GROUP_CONCAT(DISTINCT(?text); SEPARATOR='') AS ?text__Concat)
WHERE {
?thing <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
    ?thing <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
}
?thing <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/0001/anything#Thing> .
?thing <http://www.knora.org/ontology/0001/anything#hasText> ?text .
    ?text <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
}
?text <http://www.knora.org/ontology/knora-base#valueHasStandoff> ?standoffEventTag .
?standoffEventTag <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/0001/anything#StandoffEventTag> .
    ?standoffEventTag <http://www.knora.org/ontology/knora-base#valueHasStartJDN> ?standoffEventTag__valueHasStartJDN .
}
    ?standoffEventTag <http://www.knora.org/ontology/knora-base#valueHasEndJDN> ?standoffEventTag__valueHasEndJDN .
}
FILTER((("2457724"^^<http://www.w3.org/2001/XMLSchema#integer> <= ?standoffEventTag__valueHasEndJDN) && ("2457754"^^<http://www.w3.org/2001/XMLSchema#integer> >= ?standoffEventTag__valueHasStartJDN)))
}
GROUP BY ?thing
ORDER BY ASC(?thing)
LIMIT 25
```

tobiasschweizer:
Do you get the correct result from the prequery?

What does the main query look like and what results do you get from it?

benjamingeer:
> Do you get the correct result from the prequery?

Yes:

screen shot 2018-07-02 at 14 11 01

> What does the main query look like

```sparql
CONSTRUCT {
?mainResourceVar <http://www.knora.org/ontology/knora-base#isMainResource> "true"^^<http://www.w3.org/2001/XMLSchema#boolean> .
?mainAndDependentResource <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
?mainAndDependentResource ?mainAndDependentResourceProp ?mainAndDependentResourceObj .
?mainAndDependentResource <http://www.knora.org/ontology/knora-base#hasValue> ?mainAndDependentResourceValueObject .
?mainAndDependentResource ?mainAndDependentResourceValueProp ?mainAndDependentResourceValueObject .
?mainAndDependentResourceValueObject ?mainAndDependentResourceValueObjectProp ?mainAndDependentResourceValueObjectObj .
?mainAndDependentResourceValueObject <http://www.knora.org/ontology/knora-base#valueHasStandoff> ?standoffNode .
?standoffNode ?standoffProp ?standoffValue .
?mainAndDependentResourceValueObject <http://www.knora.org/ontology/knora-base#valueHasListNode> ?listNode .
?listNode <http://www.w3.org/2000/01/rdf-schema#label> ?listNodeLabel .
} WHERE {
{
VALUES ?mainResourceVar { <http://rdfh.ch/0001/a-thing> }
?mainResourceVar <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
    ?mainResourceVar <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
}
} UNION {
VALUES ?mainAndDependentResource { <http://rdfh.ch/0001/a-thing> }
?mainAndDependentResource <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
    ?mainAndDependentResource <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
}
    ?mainAndDependentResource ?mainAndDependentResourceProp ?mainAndDependentResourceObj .
}
} UNION {
VALUES ?mainAndDependentResourceValueObject { <http://rdfh.ch/0001/a-thing/values/ONYJiqveRVSeEO6KNxCIpg> }
?mainAndDependentResource <http://www.knora.org/ontology/knora-base#hasValue> ?mainAndDependentResourceValueObject .
    ?mainAndDependentResource ?mainAndDependentResourceValueProp ?mainAndDependentResourceValueObject .
}
    ?mainAndDependentResourceValueObject <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
}
    ?mainAndDependentResourceValueObject ?mainAndDependentResourceValueObjectProp ?mainAndDependentResourceValueObjectObj .
}
} UNION {
VALUES ?mainAndDependentResourceValueObject { <http://rdfh.ch/0001/a-thing/values/ONYJiqveRVSeEO6KNxCIpg> }
    ?mainAndDependentResourceValueObject <http://www.knora.org/ontology/knora-base#valueHasStandoff> ?standoffNode .
}
    ?standoffNode ?standoffProp ?standoffValue .
}
} UNION {
VALUES ?mainAndDependentResourceValueObject { <http://rdfh.ch/0001/a-thing/values/ONYJiqveRVSeEO6KNxCIpg> }
    ?mainAndDependentResourceValueObject <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#ListValue> .
}
    ?mainAndDependentResourceValueObject <http://www.knora.org/ontology/knora-base#valueHasListNode> ?listNode .
}
    ?listNode <http://www.w3.org/2000/01/rdf-schema#label> ?listNodeLabel .
}
}
}
```

> and what results do you get from it?

[standoff-date-query-result.ttl.zip](https://github.com/dhlab-basel/Knora/files/2155141/standoff-date-query-result.ttl.zip)

tobiasschweizer:
That looks great!

benjamingeer:
I've just realised that `owl:TransitiveProperty` is not safe in Knora, because it could allow you to add values to a resource that you don't have permission to modify. I'll implement this differently tomorrow.

benjamingeer:
I kept `owl:TransitiveProperty`, because it's convenient, but I allowed it only for built-in Knora ontologies. I added something to the FAQ to explain why.

tobiasschweizer:
I have just started to look at it and I think standoff search offers great possibilities. 

I am unsure about how to query for a resource that has a standoff link to another resources and include the dependent resource in the result set:

I came up with the following query:

```
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/v2#>

CONSTRUCT {
    ?letter beol:hasText ?text .
  
    ?letter knora-api:hasStandoffLinkTo ?letter2 .
} WHERE {
    ?letter a beol:letter .
    ?letter beol:hasText ?text .
    ?text knora-api:textValueHasStandoff ?standoffLinkTag .
    ?standoffLinkTag a knora-api:StandoffLinkTag .
  
    ?text knora-api:valueAsString ?textStr .
    FILTER knora-api:matchInStandoff(?textStr, ?standoffLinkTag, "1707.06.29")
  
    ?letter knora-api:hasStandoffLinkTo ?letter2 .
  
    FILTER knora-api:standoffLink(?letter, ?standoffLinkTag, ?letter2)
    ?letter2 a beol:letter .
    
}
```

However, the statement `?letter knora-api:hasStandoffLinkTo ?letter2 .` seems redundant to me. Could we say that we always return the dependent resource if a standoff link is queries and the text value is contained in the result?

benjamingeer:
> Could we say that we always return the dependent resource if a standoff link is queries and the text value is contained in the result?

Sounds reasonable to me. How do we do that? I don't know how dependent resources are identified.

benjamingeer:
I see that `AbstractSparqlTransformer` has something called `dependentResourceVariables`. But what if the dependent resource is an IRI, not a variable?

benjamingeer:
A related question: should `AbstractSparqlTransformer.handleStandoffLinkFunction` add the target resource to `dependentResourceVariables`?

benjamingeer:
... if it's a variable?

tobiasschweizer:
I think it is already added correctly:

```
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/v2#>

CONSTRUCT {
    ?letter beol:hasText ?text .
} WHERE {
    ?letter a beol:letter .
    ?letter beol:hasText ?text .
    ?text knora-api:textValueHasStandoff ?standoffLinkTag .
    ?standoffLinkTag a knora-api:StandoffLinkTag .
  
    FILTER knora-api:standoffLink(?letter, ?standoffLinkTag, ?letter2)
    ?letter2 a beol:letter .
    
}
```

-> 

```
SELECT DISTINCT ?letter (GROUP_CONCAT(DISTINCT(?letter2); SEPARATOR='') AS ?letter2__Concat) (GROUP_CONCAT(DISTINCT(?text); SEPARATOR='') AS ?text__Concat) (GROUP_CONCAT(DISTINCT(?letter2__LinkValue); SEPARATOR='') AS ?letter2__LinkValue__Concat)
 WHERE {
 ?letter <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     ?letter <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?letter <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/0801/beol#letter> .
 ?letter <http://www.knora.org/ontology/0801/beol#hasText> ?text .
     ?text <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?text <http://www.knora.org/ontology/knora-base#valueHasStandoff> ?standoffLinkTag .
 ?standoffLinkTag <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#StandoffLinkTag> .
 ?letter2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     ?letter2 <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?letter2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/0801/beol#letter> .
     ?letter <http://www.knora.org/ontology/knora-base#hasStandoffLinkTo> ?letter2 .
 }
 ?standoffLinkTag <http://www.knora.org/ontology/knora-base#standoffTagHasLink> ?letter2 .
 ?letter <http://www.knora.org/ontology/knora-base#hasStandoffLinkToValue> ?letter2__LinkValue .
     ?letter2__LinkValue <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#LinkValue> .
 }
     ?letter2__LinkValue <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
     ?letter2__LinkValue <http://www.w3.org/1999/02/22-rdf-syntax-ns#subject> ?letter .
 }
     ?letter2__LinkValue <http://www.w3.org/1999/02/22-rdf-syntax-ns#object> ?letter2 .
 }
 }
 GROUP BY ?letter
 ORDER BY ASC(?letter)
 LIMIT 25
```

`dependentResourceVariablesConcat` -> Set(QueryVariable(letter2__Concat))

tobiasschweizer:
I think the missing part is just a statement in the CONSTRUCT clause: After permission checking, `SearchResponderV2` figures out what the client wants to have included in the answer. Since some statements were added automatically by `AbstractSparqlTransformer.handleStandoffLinkFunction`, these are not present in the CONSTRCU clause and thus not returned. The result is that the dependent resource is not present in the result.

I have the following idea: if the user asks for a text value which links to another resource by a standoff link and the text value is present in the CONSTRUCT clause, the dependent resource should be returned as well.

benjamingeer:
> I think it is already added correctly

What if you remove this line from the query:

```
?letter2 a beol:letter .
```

benjamingeer:
> I think the missing part is just a statement in the CONSTRUCT clause

I guess that would have to be:

```
?letter knora-api:hasStandoffLinkTo ?letter2 .
```

Would that work even if it's not in the WHERE clause? I thought we had a rule that everything in the CONSTRUCT clause also has to be in the WHERE clause.

benjamingeer:
Actually I'm stupid, it is in the WHERE clause, it's generated by `handleStandoffLinkFunction` (line 1363).

tobiasschweizer:
> What if you remove this line from the query:
> ?letter2 a beol:letter .

Then it is absent: 

```
 SELECT DISTINCT ?letter (GROUP_CONCAT(DISTINCT(?text); SEPARATOR='') AS ?text__Concat) (GROUP_CONCAT(DISTINCT(?letter2__LinkValue); SEPARATOR='') AS ?letter2__LinkValue__Concat)
 WHERE {
 ?letter <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     ?letter <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?letter <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/0801/beol#letter> .
 ?letter <http://www.knora.org/ontology/0801/beol#hasText> ?text .
     ?text <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?text <http://www.knora.org/ontology/knora-base#valueHasStandoff> ?standoffLinkTag .
 ?standoffLinkTag <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#StandoffLinkTag> .
     ?letter <http://www.knora.org/ontology/knora-base#hasStandoffLinkTo> ?letter2 .
 }
 ?standoffLinkTag <http://www.knora.org/ontology/knora-base#standoffTagHasLink> ?letter2 .
 ?letter <http://www.knora.org/ontology/knora-base#hasStandoffLinkToValue> ?letter2__LinkValue .
     ?letter2__LinkValue <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#LinkValue> .
 }
     ?letter2__LinkValue <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
     ?letter2__LinkValue <http://www.w3.org/1999/02/22-rdf-syntax-ns#subject> ?letter .
 }
     ?letter2__LinkValue <http://www.w3.org/1999/02/22-rdf-syntax-ns#object> ?letter2 .
 }
 }
 GROUP BY ?letter
 ORDER BY ASC(?letter)
 LIMIT 25
 
```

This is because of this code here:

https://github.com/dhlab-basel/Knora/blob/a810b48d2e198490b1e0a7cdbfbc4c6a670f3ae1/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala#L308-L321

It just gets added if there is a statement `?letter knora-api:hasStandoffLinkTo ?letter2 .`

benjamingeer:
> Then it is absent:

And what if you remove that line *and* use an IRI instead of the variable `?letter2` in the `knora-api:standoffLink` function?

tobiasschweizer:
@benjamingeer I will make an issue for the bug I found https://github.com/dhlab-basel/Knora/pull/910#issuecomment-402661813 

I have a problem with the list node query: the list nodes Iri is also treated as a dependent resource Iri which is wrong. I will have to change the way I get them from type inspector, checking if they represent a resource or not.

tobiasschweizer:
```
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/v2#>

CONSTRUCT {
    ?letter beol:hasText ?text .
} WHERE {
    ?letter a beol:letter .
    ?letter beol:hasText ?text .
    ?text knora-api:textValueHasStandoff ?standoffLinkTag .
    ?standoffLinkTag a knora-api:StandoffLinkTag .
  
    FILTER knora-api:standoffLink(?letter, ?standoffLinkTag, <http://rdfh.ch/biblio/up0Q0ZzPSLaULC2tlTs1sA>)
    #<http://rdfh.ch/biblio/up0Q0ZzPSLaULC2tlTs1sA> a beol:person .
    
}
```

> {
>"error": "org.knora.webapi.GravsearchException: Types could not be determined for one or more >entities: <http://rdfh.ch/biblio/up0Q0ZzPSLaULC2tlTs1sA>"
>}

but it works when you uncomment the last line.

benjamingeer:
> but it works when you uncomment the last line.

Hmm, I think that's a bug, let me see.

tobiasschweizer:
ok, I will be back after lunch :-)

benjamingeer:
OK, I fixed that bug, now that last query returns 14 forbidden resources, but I don't understand why.

tobiasschweizer:
> OK, I fixed that bug, now that last query returns 14 forbidden resources, but I don't understand why.

Maybe it does not find an expected  dependent resource in the results after permission checking

benjamingeer:
> Maybe it does not find an expected dependent resource in the results after permission checking

Could you try it and see?

tobiasschweizer:
> Could you try it and see?
 You mean the query from https://github.com/dhlab-basel/Knora/pull/910#issuecomment-402669099 ?

benjamingeer:
No, the one from https://github.com/dhlab-basel/Knora/pull/910#issuecomment-402676783.

tobiasschweizer:
ah I see, ok!

tobiasschweizer:
It does not find all the value objects it expects: 
https://github.com/dhlab-basel/Knora/blob/a810b48d2e198490b1e0a7cdbfbc4c6a670f3ae1/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala#L2743

```
 expected value objs: Set(http://rdfh.ch/0801/beol/Z6ivoioqTKO5dG4HL4_yyA/values/CgNcMe-kQ-eYyhqUHv4WLQ, http://rdfh.ch/0801/beol/Z6ivoioqTKO5dG4HL4_yyA/values/Mj7fgg-MQrWeYc0Ofpqd-w)
 given value objs: Set(http://rdfh.ch/0801/beol/Z6ivoioqTKO5dG4HL4_yyA/values/CgNcMe-kQ-eYyhqUHv4WLQ)
 expected value objs: Set(http://rdfh.ch/0801/beol/ZJDcAU6yQiWyqIqV58dLdA/values/fTbAZAZnTmSBBaRjwmLSzQ, http://rdfh.ch/0801/beol/ZJDcAU6yQiWyqIqV58dLdA/values/3WR-TzXwQaaCgB1vtk0dfw)
 given value objs: Set(http://rdfh.ch/0801/beol/ZJDcAU6yQiWyqIqV58dLdA/values/fTbAZAZnTmSBBaRjwmLSzQ)
 expected value objs: Set(http://rdfh.ch/0801/beol/dMMrwg4CRtqGKNgUAuQvUA/values/Wwj5HC4LRXaliEBETT3pSQ, http://rdfh.ch/0801/beol/dMMrwg4CRtqGKNgUAuQvUA/values/If_ynceERs6zLCVgXlqpoA)
 given value objs: Set(http://rdfh.ch/0801/beol/dMMrwg4CRtqGKNgUAuQvUA/values/Wwj5HC4LRXaliEBETT3pSQ)
 expected value objs: Set(http://rdfh.ch/0801/beol/yBr2EjBMTkeCeVmGPfjCtw/values/kNEoIj-KS7q1e43tmIQC4w, http://rdfh.ch/0801/beol/yBr2EjBMTkeCeVmGPfjCtw/values/XyJh_NraTvaeU5AAOgouug)
 given value objs: Set(http://rdfh.ch/0801/beol/yBr2EjBMTkeCeVmGPfjCtw/values/kNEoIj-KS7q1e43tmIQC4w)
 expected value objs: Set(http://rdfh.ch/0801/beol/NMGEQzYvTRijL2FWHAc_tw/values/kRKiRn3fRhaaYrmKwO5xhA, http://rdfh.ch/0801/beol/NMGEQzYvTRijL2FWHAc_tw/values/qCfFJ_BKQ4mVaMb6vgVApQ)
 given value objs: Set(http://rdfh.ch/0801/beol/NMGEQzYvTRijL2FWHAc_tw/values/kRKiRn3fRhaaYrmKwO5xhA)
 expected value objs: Set(http://rdfh.ch/0801/beol/DF6WV9VeSSOYltBFIaD1iA/values/eHocSlpGR_alk2rlxjqjfA, http://rdfh.ch/0801/beol/DF6WV9VeSSOYltBFIaD1iA/values/45da6qNPS96WcN2cG_h5vw)
 given value objs: Set(http://rdfh.ch/0801/beol/DF6WV9VeSSOYltBFIaD1iA/values/eHocSlpGR_alk2rlxjqjfA)
 expected value objs: Set(http://rdfh.ch/0801/beol/20dsBNORRzuqx2AolMH_Sg/values/koY7xB0bTmGm2we87j7Z9w, http://rdfh.ch/0801/beol/20dsBNORRzuqx2AolMH_Sg/values/wmQ8HJ_8QMStrj2TnOOQzQ)
 given value objs: Set(http://rdfh.ch/0801/beol/20dsBNORRzuqx2AolMH_Sg/values/koY7xB0bTmGm2we87j7Z9w)
 expected value objs: Set(http://rdfh.ch/0801/beol/4Wfnmh7uQ522AH9OWEbygA/values/Sss4_gFqTvefnkmI027TnQ, http://rdfh.ch/0801/beol/4Wfnmh7uQ522AH9OWEbygA/values/W3Wn_b9uSzqBYX_HnjaUZQ)
 given value objs: Set(http://rdfh.ch/0801/beol/4Wfnmh7uQ522AH9OWEbygA/values/Sss4_gFqTvefnkmI027TnQ)
 expected value objs: Set(http://rdfh.ch/0801/beol/5JsY7N2nQXqv0R4jC1uWGQ/values/CPiYSh_zTceNDOPqjvIRkQ, http://rdfh.ch/0801/beol/5JsY7N2nQXqv0R4jC1uWGQ/values/EghcKq9ATv6-CoDJBIR3ww)
 given value objs: Set(http://rdfh.ch/0801/beol/5JsY7N2nQXqv0R4jC1uWGQ/values/CPiYSh_zTceNDOPqjvIRkQ)
 expected value objs: Set(http://rdfh.ch/0801/beol/f5a92GjJRriC1GTY_NzfvQ/values/Y5C8uumxTy-tCOuf6Vib6A, http://rdfh.ch/0801/beol/f5a92GjJRriC1GTY_NzfvQ/values/5sQSdXGPQRS90EqzfovvCg)
 given value objs: Set(http://rdfh.ch/0801/beol/f5a92GjJRriC1GTY_NzfvQ/values/Y5C8uumxTy-tCOuf6Vib6A)
 expected value objs: Set(http://rdfh.ch/0801/beol/3tct2_3BRLmnSoA45OHBAA/values/VcA57tXxSbKn8muHHbyYCg, http://rdfh.ch/0801/beol/3tct2_3BRLmnSoA45OHBAA/values/X6XE2y9dQyewGqfPRhN7KQ)
 given value objs: Set(http://rdfh.ch/0801/beol/3tct2_3BRLmnSoA45OHBAA/values/VcA57tXxSbKn8muHHbyYCg)
 expected value objs: Set(http://rdfh.ch/0801/beol/YeCW07xVQZGgnxCFbXLoTQ/values/0fPRf-JmRyqmUj20zwTTOg, http://rdfh.ch/0801/beol/YeCW07xVQZGgnxCFbXLoTQ/values/N2epyTluRAOwYO2In2hsIA)
 given value objs: Set(http://rdfh.ch/0801/beol/YeCW07xVQZGgnxCFbXLoTQ/values/0fPRf-JmRyqmUj20zwTTOg)
 expected value objs: Set(http://rdfh.ch/0801/beol/JrlhJFijREOSodX8Ci9OMw/values/5Y09qbITTwibZDNsoMKpeg, http://rdfh.ch/0801/beol/JrlhJFijREOSodX8Ci9OMw/values/UU9p5EUARdqnmPHCDmK03Q)
 given value objs: Set(http://rdfh.ch/0801/beol/JrlhJFijREOSodX8Ci9OMw/values/5Y09qbITTwibZDNsoMKpeg)
 expected value objs: Set(http://rdfh.ch/0801/beol/-0tI3HXgSSOeDtkf-SA00w/values/GJQB9IHYTl2RhG4g4ru0YA, http://rdfh.ch/0801/beol/-0tI3HXgSSOeDtkf-SA00w/values/aMwl67I7RGa5Dvk-lczTdQ)
 given value objs: Set(http://rdfh.ch/0801/beol/-0tI3HXgSSOeDtkf-SA00w/values/GJQB9IHYTl2RhG4g4ru0YA)
```

tobiasschweizer:
I think these are standoff reification nodes.

tobiasschweizer:
@benjamingeer You asked me about this line:

https://github.com/dhlab-basel/Knora/blob/a810b48d2e198490b1e0a7cdbfbc4c6a670f3ae1/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala#L2617-L2618

I think this is necessary because `VariableResultsRow` suppresses empty results. This means there is actually a variable in the SELECT clause sent to the triplestore but it has no results and because of this it is not present as a key in `VariableResultsRow`:

```
/**
  * Represents a row of results in a JSON response to a SPARQL SELECT query.
  *
  * @param rowMap a map of variable names to values in the row. An empty string is not allowed as a variable
  *               name or value.
  */
case class VariableResultsRow(rowMap: ErrorHandlingMap[String, String]) {
    require(rowMap.forall {
        case (key, value) => key.nonEmpty && value.nonEmpty
    }, "An empty string is not allowed as a variable name or value in a VariableResultsRow")
}
```

Does that sound plausible?

benjamingeer:
> Does that sound plausible?

I think I don't really understand. What specifically would the variable represent?

tobiasschweizer:
```

OPTIONAL {
  ?book incunabula:title ?title
}

```

For books that have no title, `?title` would be absent in `VariableResultsRow`, right?

benjamingeer:
> For books that have no title, `?title` would be absent in `VariableResultsRow`, right?

Yes, I think that's right.

tobiasschweizer:
if `?title__Concat` is present in the prequery's SELECT clause but it may return no results since `?title` is in an optional clause.

benjamingeer:
> if `?title__Concat` is present in the prequery's SELECT clause but it may return no results since `?title` is in an optional clause. 

OK, now I understand. I think that just needs a comment that's two or three sentences long to make it clear what's happening.

tobiasschweizer:
> OK, now I understand. I think that just needs a comment that's two or three sentences long to make it clear what's happening.

Ok!

tobiasschweizer:
@benjamingeer Could you have a look at #926 and see if it sufficiently explains the design?

benjamingeer:
Reading it now.

tobiasschweizer:
Please feel free to add remarks and comments and I will improve the text next week (Monday).

tobiasschweizer:
I can dedicate most of upcoming Monday to it and see how far I get.

tobiasschweizer:
@benjamingeer I think I found out why this query does not work as expected:

```
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/v2#>

CONSTRUCT {
    ?letter beol:hasText ?text .
} WHERE {
    ?letter a beol:letter .
    ?letter beol:hasText ?text .
    ?text knora-api:textValueHasStandoff ?standoffLinkTag .
    ?standoffLinkTag a knora-api:StandoffLinkTag .
  
    FILTER knora-api:standoffLink(?letter, ?standoffLinkTag, <http://rdfh.ch/biblio/up0Q0ZzPSLaULC2tlTs1sA>)
    #<http://rdfh.ch/biblio/up0Q0ZzPSLaULC2tlTs1sA> a beol:person .
    
}
```

It is related to the way Iris of dependent resources are collected if they were not given as varibles in the input query.

https://github.com/dhlab-basel/Knora/blob/b220b33ff7d52a21e231b4ea5cc1365819f35938/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala#L2647-L2652

In our case, the set is empty.

Should the Iri `<http://rdfh.ch/biblio/up0Q0ZzPSLaULC2tlTs1sA>` be returned since it is an argument of the function `standoffLink` and the type inspector could infer it is a resource?

benjamingeer:
> Should the Iri `<http://rdfh.ch/biblio/up0Q0ZzPSLaULC2tlTs1sA>` be returned since it is an argument of the function `standoffLink` and the type inspector could infer it is a resource? 

Do you mean, should it be returned in the search results? Not in this query, because it's not mentioned in the `CONSTRUCT` clause of the Gravsearch query. But if it's mentioned in the `CONSTRUCT` clause, it should be returned in the search results.

I believe the type inspector already infers that it's a resource. There's a test for this:

https://github.com/dhlab-basel/Knora/pull/910/files#diff-ff259a561e8217f7b69c74ee01dcd486R145

benjamingeer:
Maybe the problem is actually that the function argument isn't included in `positiveEntities`:

```
     case (iri: TypeableIri, _: NonPropertyTypeInfo) if whereClauseWithoutAnnotations.positiveEntities.contains(IriRef(iri.iri)) => 
```

The Gravsearch parser doesn't add function arguments to `positiveEntities`:

https://github.com/dhlab-basel/Knora/blob/b220b33ff7d52a21e231b4ea5cc1365819f35938/webapi/src/main/scala/org/knora/webapi/util/search/gravsearch/GravsearchParser.scala#L665

benjamingeer:
> The Gravsearch parser doesn't add function arguments to positiveEntities

Actually I think it does, because `makeEntity` does this. Looking more carefully now...

tobiasschweizer:
> Do you mean, should it be returned in the search results? Not in this query, because it's not mentioned in the CONSTRUCT clause of the Gravsearch query. But if it's mentioned in the CONSTRUCT clause, it should be returned in the search results.

No, this is not what I meant. My question was only how to get the Iris of dependent resources from the type inspection results.

benjamingeer:
Fixed: it was a bug in `GravsearchParser`. Now the query works!

tobiasschweizer:
@benjamingeer thx for fixing this. I will add an R2R for the Iri case

benjamingeer:
Once that code in `SearchResponderV2` is refactored, it really needs some design documentation to explain things like this: that it relies on the type inspection result and on `whereClause.positiveEntities` to get the IRIs of dependent resources.

tobiasschweizer:
> Once that code in SearchResponderV2 is refactored, it really needs some design documentation to explain things like this: that it relies on the type inspection result and on whereClause.positiveEntities to get the IRIs of dependent resources.

This is why I started: https://github.com/dhlab-basel/Knora/issues/926

tobiasschweizer:
@benjamingeer Given that the tests pass now, I suggest that I finish the review of this PR today and then do the refactoring in a new PR that I start this week. Would that be ok for you?

I would then also try to figure out how a dependent resource could be returned when the user asks for a text value with a standoff a link and wants the text value to be included in the results.

benjamingeer:
@tobiasschweizer OK, that would be great.

benjamingeer:
> I would then also try to figure out how a dependent resource could be returned when the user asks for a text value with a standoff a link and wants the text value to be included in the results.

Doesn't this work already? If not, can you give an example?

tobiasschweizer:
see for example the test "search for a standoff link using the knora-api:standoffLink function (submitting the complex schema)"

```sparql
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX anything: <http://0.0.0.0:3333/ontology/0001/anything/v2#>

CONSTRUCT {
    ?thing anything:hasText ?text .
} WHERE {
    ?thing a anything:Thing .
    ?thing anything:hasText ?text .
    ?text knora-api:textValueHasStandoff ?standoffTag .
    ?standoffTag a knora-api:StandoffLinkTag .
    FILTER knora-api:standoffLink(?thing, ?standoffTag, ?otherThing)
    ?otherThing a anything:Thing .
}
```

it is expected to return: 

```json
{
  "@graph" : [ {
    "@id" : "http://rdfh.ch/0001/a-thing-with-text-values",
    "@type" : "anything:Thing",
    "anything:hasText" : [ {
      "@id" : "http://rdfh.ch/0001/a-thing-with-text-values/values/1",
      "@type" : "knora-api:TextValue",
      "knora-api:textValueAsXml" : "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<text>Ich liebe die <a href=\"http://rdfh.ch/0001/a-thing\" class=\"salsah-link\">Dinge</a>, sie sind alles für mich.</text>",
      "knora-api:textValueHasMapping" : "http://rdfh.ch/standoff/mappings/StandardMapping"
    }, {
      "@id" : "http://rdfh.ch/0001/a-thing-with-text-values/values/2",
      "@type" : "knora-api:TextValue",
      "knora-api:textValueAsXml" : "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<text>Na ja, die <a href=\"http://rdfh.ch/0001/a-thing\" class=\"salsah-link\">Dinge</a> sind OK.</text>",
      "knora-api:textValueHasMapping" : "http://rdfh.ch/standoff/mappings/StandardMapping"
    } ],
    "rdfs:label" : "Ein Ding für jemanden, dem die Dinge gefallen"
  }, {
    "@id" : "http://rdfh.ch/0001/project-thing-1",
    "@type" : "anything:Thing",
    "anything:hasText" : {
      "@id" : "http://rdfh.ch/0001/project-thing-1/values/2",
      "@type" : "knora-api:TextValue",
      "knora-api:textValueAsXml" : "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<text>This is a <a href=\"http://rdfh.ch/0001/project-thing-2\" class=\"salsah-link\">standoff link</a>.</text>",
      "knora-api:textValueHasMapping" : "http://rdfh.ch/standoff/mappings/StandardMapping"
    },
    "rdfs:label" : "A thing that only project members can see"
  } ],
  "@context" : {
    "rdf" : "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
    "rdfs" : "http://www.w3.org/2000/01/rdf-schema#",
    "knora-api" : "http://api.knora.org/ontology/knora-api/v2#",
    "anything" : "http://0.0.0.0:3333/ontology/0001/anything/v2#"
  }
}
```

I think when the user expects to get back the text `?text` linking to a resource, then the resource referred to should also be returned (its Iri, label and type). 

My rationale is the following: if there are no other statements about that target resource, how could it be included in the CONSTRUCT clause if the user wants to see it?

benjamingeer:
> I think when the user expects to get back the text ?text linking to a resource, then the resource referred to should also be returned (its Iri, label and type).

Why? If the user wants to get that information, why can't they just add one line to the `CONSTRUCT` clause?

```
?otherThing a anything:Thing .
```

Maybe they don't actually want that information. Perhaps the target resource is an IRI, and the user already knows all about it, so they don't need it in the search result.

> if there are no other statements about that target resource

You mean in the `WHERE` clause? But there are statements about it there. There is this one:

```
?otherThing a anything:Thing .
```

And the `FILTER` also generates statements about the target resource. So I don't see why this wouldn't work.

tobiasschweizer:
```sparql
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX anything: <http://0.0.0.0:3333/ontology/0001/anything/v2#>

CONSTRUCT {
    ?thing anything:hasText ?text .
    
    ?otherThing a anything:Thing .
} WHERE {
    ?thing a anything:Thing .
    ?thing anything:hasText ?text .
    ?text knora-api:textValueHasStandoff ?standoffTag .
    ?standoffTag a knora-api:StandoffLinkTag .
    FILTER knora-api:standoffLink(?thing, ?standoffTag, ?otherThing)
    ?otherThing a anything:Thing .
}
```

> {
"error": "org.knora.webapi.GravsearchException: <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> is not allowed in a CONSTRUCT clause"
}

tobiasschweizer:
It's not a problem if you make additional statements like `?otherThing anything:hasBoolean .` etc.

benjamingeer:
Oh yeah, of course, I'm stupid. Hmm. Then I guess there are two options. The user could add this to the `CONSTRUCT` clause:

```
?thing knora-api:hasStandoffLinkTo ?otherThing .
```

It would work because the `FILTER` already adds that statement to the generated `WHERE` clause.

Or we could add that automatically. But I'd rather give the user control over it.

tobiasschweizer:
With the current design, adding `?thing knora-api:hasStandoffLinkTo ?otherThing .` to the CONSTRUCT clause only works if it is also present in the Gravsearch query's WHERE clause. 

Maybe I can have a look at how this logic could be improved so that this would work.

webapi/src/main/scala/org/knora/webapi/responders/v2/OntologyResponderV2.scala
```diff
@@ -3148,12 +3158,18 @@ class OntologyResponderV2 extends Responder {
...
 
-        PropertyInfoContentV2(
+        val propertyDef = PropertyInfoContentV2(
             propertyIri = propertyIri,
             subPropertyOf = subPropertyOf,
             predicates = otherPreds,
             ontologySchema = propertyIri.getOntologySchema.get
         )
+
+        if (!propertyIri.isKnoraBuiltInDefinitionIri && propertyDef.getRdfTypes.contains(OntologyConstants.Owl.TransitiveProperty.toSmartIri)) {
```
tobiasschweizer: Does this make sure that you cannot use `owl:transitiveProperty` in a project ontology because of the reasons outlined in FAQ?

benjamingeer:
> Maybe I can have a look at how this logic could be improved so that this would work.

Or we could just make the user add it to the `WHERE` clause, for consistency. It would be redundant but I think it wouldn't do any harm.

webapi/src/main/scala/org/knora/webapi/responders/v2/OntologyResponderV2.scala
benjamingeer: Yes.

tobiasschweizer:
> Or we could just make the user add it to the WHERE clause, for consistency. It would be redundant but I think it wouldn't do any harm.

I would be ok with that. Could you add that to the docs about the `standoffLink` function?

tobiasschweizer:
@benjamingeer btw, I fixed this https://github.com/dhlab-basel/Knora/issues/889#issuecomment-397270708 in ed03eb7ed79249afa5a734a4dd6f47063ac2985e

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -936,139 +1054,412 @@ class SearchResponderV2 extends ResponderWithStandoffV2 {
...
+            // Generate statements matching the link value that describes the standoff link between the source and target resources.
+            val statementsForLinkValue: Seq[StatementPattern] = generateStatementsForLinkValue(
+                linkSource = linkSourceEntity,
+                linkPred = hasStandoffLinkToIriRef,
+                linkTarget = linkTargetEntity
+            )
+
+            TransformedFilterPattern(
+                None, // FILTER has been replaced with statements
+                linkStatements ++ statementsForLinkValue ++ statementsForTargetResource
```
tobiasschweizer: Do you think we have to care about redundant triples?

```sparql
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
PREFIX standoff: <http://api.knora.org/ontology/standoff/v2#>
PREFIX anything: <http://0.0.0.0:3333/ontology/0001/anything/v2#>

CONSTRUCT {
    ?thing anything:hasText ?text .
    
} WHERE {
    ?thing a anything:Thing .
    ?thing anything:hasText ?text .
    ?text knora-api:textValueHasStandoff ?standoffTag .
    ?standoffTag a knora-api:StandoffLinkTag .
    FILTER knora-api:standoffLink(?thing, ?standoffTag, ?otherThing)
    ?otherThing a anything:Thing .
    
}
```

prequery:

```sparql
 SELECT DISTINCT ?thing (GROUP_CONCAT(DISTINCT(?otherThing); SEPARATOR='') AS ?otherThing__Concat) (GROUP_CONCAT(DISTINCT(?text); SEPARATOR='') AS ?text__Concat) (GROUP_CONCAT(DISTINCT(?otherThing__LinkValue); SEPARATOR='') AS ?otherThing__LinkValue__Concat)
 WHERE {
 ?thing <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     ?thing <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?thing <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/0001/anything#Thing> .
 ?thing <http://www.knora.org/ontology/0001/anything#hasText> ?text .
     ?text <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?text <http://www.knora.org/ontology/knora-base#valueHasStandoff> ?standoffTag .
 ?standoffTag <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#StandoffLinkTag> .
 ?otherThing <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     ?otherThing <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 ?otherThing <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/0001/anything#Thing> .
     ?thing <http://www.knora.org/ontology/knora-base#hasStandoffLinkTo> ?otherThing .
 }
 ?standoffTag <http://www.knora.org/ontology/knora-base#standoffTagHasLink> ?otherThing .
 ?thing <http://www.knora.org/ontology/knora-base#hasStandoffLinkToValue> ?otherThing__LinkValue .
     ?otherThing__LinkValue <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#LinkValue> .
 }
     ?otherThing__LinkValue <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
     ?otherThing__LinkValue <http://www.w3.org/1999/02/22-rdf-syntax-ns#subject> ?thing .
 }
     ?otherThing__LinkValue <http://www.w3.org/1999/02/22-rdf-syntax-ns#object> ?otherThing .
 }
 ?otherThing <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     ?otherThing <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
 }
 GROUP BY ?thing
 ORDER BY ASC(?thing)
 LIMIT 25
```

Note that 

```
 ?otherThing <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.knora.org/ontology/knora-base#Resource> .
     ?otherThing <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 }
``` is redundant. This is because there are two statements involving `?otherThing`.

Could we simply remove duplicates from the sequence of statements before creating the WHERE clause for the prequery?

Comment (tobiasschweizer) [reply webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala]: `createAdditionalStatementsForNonPropertyType ` is called twice: once from within `handleStandoffLinkFunction` for `?otherThing` and once because of ` ?otherThing a anything:Thing .`. We call it from within `handleStandoffLinkFunction` because there could be no statement ` ?otherThing a anything:Thing .` and then the logic would not work correctly.

Comment (tobiasschweizer) [reply webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala]: I guess this is because we look at statements one by one.

Comment (benjamingeer) [reply webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala]: How about if `createAdditionalStatementsForNonPropertyType` maintains a mutable set of statements that has already generated, so it can avoid generating them redundantly?

Comment (tobiasschweizer) [reply webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala]: Could we do that as an optimization in the refactoring we have in mind anyway?

Then we could merge this PR, so it does not block you from what you have to do.

Comment (benjamingeer) [reply webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala]: OK!

File: webapi/src/test/scala/org/knora/webapi/e2e/v2/SearchRouteV2R2RSpec.scala [hunk 3]
```diff
@@ -1869,7 +1912,7 @@ class SearchRouteV2R2RSpec extends R2RSpec {
 
         }
 
-        "search for a anything:Thing with a list value" in {
+        "search for an anything:Thing with a list value" ignore { // ignored because the list node label returned is not deterministic
```
tobiasschweizer: I guess we should get rid of the list node label and query the list v2 route instead.

benjamingeer: Yes, you wanted to do that in another PR, right?

tobiasschweizer: Yes, #763 would be a good occasion to that. Could we work on it together?

benjamingeer: Yes please!

benjamingeer:
Thank you! :)

tobiasschweizer:
I glad this works now with the current design. This proves that we can actually deal with non XML markup and make useful queries :-)

PR #914, opened 2018-06-29
releasing-v1.6.0

subotic:


subotic:
@benjamingeer should be a quick review :-)

subotic:
thanks :-)

subotic:
ah, crap. need to wait for the tests to pass. we need to make them run faster. the unwritten rule is no longer than 15 minutes. just enough time to get a cup of coffee.

benjamingeer:
I think I've found a computer that's fast enough to run our tests in a reasonable amount of time: https://www.energy.gov/articles/summit-supercomputer-ranked-fastest-computer-world

subotic:
I hope so ;-)

PR #920, opened 2018-07-04
guiOrder corrections

SepidehAlassi:
- added the missing guiOrder to properties
- corrected indentations


SepidehAlassi:
@subotic travis build fails due to error installing ruby 2.5.1
screenshot 2018-07-05 09 18 54
Can you please check this?

SepidehAlassi:
@tobiasschweizer just changed cardinalities of couple of properties in beol-onto based on #929. Creating the birthplace resource although a nice idea would mean changing the already existing xml files for nameIndex which I really don't want to do. So I prefer changing the maxCardinality of properties in ontology. a very small change to review!

tobiasschweizer:
Looks ok!

PR #934, opened 2018-07-12
fix (webapi): crash during import test

### LINKED ISSUE #932: Knora crashes during import load test
Running the `load_test` from https://github.com/dhlab-basel/knora-tests crashes Knora. This should never happen. If something goes wrong, then a request can return an error, but Knora should keep on running.

### LINKED ISSUE #935: Travis build fails because GraphDB runs out of heap space
https://travis-ci.org/dhlab-basel/Knora/jobs/403155085

screen shot 2018-07-12 at 17 43 30

This is from https://github.com/dhlab-basel/Knora/pull/927

subotic:
### Summary

- [x] update `akka-actor` and akka-http`
- [x] remove blocking calls in `Authenticator`
- [x] refactor all routes to resolve the `UserADM` future
- [x] downgrade GraphDB to 8.5.0 on Travis to see if heap memory error occur

### Issues

- resolves #932 
- resolves #935


.travis.yml
```diff
@@ -30,7 +30,7 @@ env:
     - CHROME_BIN=google-chrome-stable
     - GDB_HEAP_SIZE="2G"
     - TAG=${TRAVIS_BRANCH//\//-}
-    - GRAPHDB_VERSION="8.6.0-se"
+    - GRAPHDB_VERSION="8.5.0-se"
```
benjamingeer: Could you possibly send Ontotext a message to ask them about the increased memory requirements of 8.6 and whether there's anything we can do about that?

webapi/src/main/scala/org/knora/webapi/routing/v1/CkanRouteV1.scala
```diff
@@ -38,21 +39,23 @@ object CkanRouteV1 extends Authenticator {
...
-                    val limit: Option[Int] = params.get("limit").map(_.toInt)
-                    val info: Boolean = params.getOrElse("info", false) == true
-                    val requestMessage = CkanRequestV1(project, limit, info, userProfile)
 
-                    RouteUtilV1.runJsonRoute(
+                    val requestMessage = for {
+                        userProfile <- getUserADM(requestContext)
+                        params = requestContext.request.uri.query().toMap
+                        project: Option[Seq[String]] = params.get("project").map(_.split(","))
+                        limit: Option[Int] = params.get("limit").map(_.toInt)
```
benjamingeer: I think it's great to make all this code non-blocking. A question: in some cases (like here) you've put all the route processing in a `Future` using a `for` comprehension. But in other cases (like in `GroupsRouteADM`), the request parameters are processed outside the `Future`, and the `Future` is created only at the end of the route. I know that currently, if a route throws an exception, the exception is handled by an Akka exception handler. Does that change now that the code is in a `Future`? For example, on this line, what happens if `toInt` throws an exception? Is the exception handled differently from how it was handled before?

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilV2.scala
```diff
@@ -153,7 +153,7 @@ object RouteUtilV2 {
     /**
       * Sends a message to a responder and completes the HTTP request by returning the response as TEI/XML.
       *
-      * @param requestMessage   a future containing a [[KnoraRequestV2]] message that should be sent to the responder manager.
+      * @param requestMessageF  a future containing a [[KnoraRequestV2]] message that should be sent to the responder manager.
```
benjamingeer: Since `runRdfRoute` is now called only by `runRdfRouteWithFuture`, you could make `runRdfRoute` private.

.travis.yml
subotic: yes, wanted to do that.

webapi/src/main/scala/org/knora/webapi/routing/v1/CkanRouteV1.scala
subotic: good question. just tryied it out:

this call:
```
http://localhost:3333/v1/ckan?limit=one
```

produces this response:

```
{
    "status": 4,
    "error": "java.lang.NumberFormatException: For input string: \"one\""
}
```

So it down's matter if it is thrown here `userProfile <- getUserADM(requestContext)` or here `limit: Option[Int] = params.get("limit").map(_.toInt)`.

subotic: It is not refactored the same everywhere. After the first few routes, I got tired of doing it and looked for the quickest possible way :-)

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilV2.scala
subotic: done

.travis.yml
subotic: I think that we can reproduce this with the same set of GraphDB queries from #937.

subotic:
thanks for the review :-)

PR #948, opened 2018-07-18
Fix infinite recursion when generating XML schemas for bulk import

### LINKED ISSUE #944: timeout or java.lang.OutOfMemoryError when using  /v1/resources/xmlimportschemas/ for some ontologies
Sometimes I'm running into a problem when I'm using /v1/resources/xmlimportschemas/ for the bulk-import. It seems to occur with 'more complex' ontologies only, i.e. that are highly connected with other ontologies. Until now I tried to things:
 - to avoid a timeout error, I increased knora's timeout in 'application.conf'
 - I also increased the memory to avoid a java.lang.OutOfMemoryError ([ERROR] Uncaught error from thread [webapi-akka.actor.default-dispatcher-39]: GC overhead limit exceeded, shutting down JVM since 'akka.jvm-exit-on-fatal-error' is enabled for ActorSystem[webapi])

Like said before, I have no problems with some of our ontologies (e.g. agent, text, text-structure)
but with others (e.g. delille, drcs, kuno-raeber) I run into problems. It might be to do with knora and its memory usage, since I noticed, that the (graphdb-)queries run just fine (for delille this was done in one minute), but afterwards java/knora uses all CPU resources and drops out after some time (for delille after another 20 minutes).

I would be glad, if you can take a look at this problem. The used ontologies are attached to the issue.
Thank you very much in advance. 

[Ontologies.zip](https://github.com/dhlab-basel/Knora/files/2201891/Ontologies.zip)

benjamingeer:
This fixes an infinite recursion that could occur when following dependencies between ontologies. The solution (in `ResourcesRouteV1.getNamedGraphInfos`): don't get the same ontology twice.

Fixes #944.


sacha-alex:
Dear Ben,

  thank you very much for the fix. We have tested it against our ontologies and the bulk import is working, too. It resolves our problems completely.
Thanks again for the great and quick work.

Best,
Sascha

benjamingeer:
@subotic Can I merge this?

benjamingeer:
@subotic Great, thanks!

PR #950, opened 2018-07-18
fixes the small issues with facsimiles representations in BEOL

SepidehAlassi:


tobiasschweizer:
@SepidehAlassi It has just occurred to me that `beol:page` is used for the Meditationes and there is already a lot of data using this class.

We could use the `facsimile` class for LEOO (and BEBB), and still use the `page` for the Meditationes. Would that be ok?

SepidehAlassi:
@tobiasschweizer  I defined a new resource documentImage. Page and facsimiles are subclassof documentImage

Issue #968, opened 2018-08-17
Startup check does not detect running triplestore in production environment with SSL enabled


subotic:
Probably too tight timeouts, as establishing the connection with SSL takes longer.

subotic:
done in #969

PR #969, opened 2018-08-17
fix (webapi): startup check does not detect running triplestore

subotic:
### Summary

- raise timeouts in KnoraService so that the triplestore check has time to complete

### Issues

- #968


subotic:
@benjamingeer small review.

subotic:
thanks for the review :-)

Issue #961, opened 2018-08-07
Return users depending on the requesters permission


subotic:
User is `SystemAdmin`:
- return all users for `/admin/users`
- return all users for `/admin/projects/members`

User is `ProjectAdmin`:
- return FORBIDDEN for `/admin/users`
- return all users for `/admin/projects/members`

User is normal user:
- return FORBIDDEN for `/admin/users`
- return FORBIDDEN for `/admin/projects/members`

PR #975, opened 2018-08-22
feature (webapi): not return or allow changing of built-in users

### LINKED ISSUE #772: Disallow changing or using the system and anonymous user
- Need to check explicitly for `system@localhost` and `anonymous@localhost`, and return an error. 
- Don't allow any changes to those two users
- Those two users should not get returned by the admin API as actual users.

subotic:
### Summary

 - filter out the `System` and `Anonymous` user from any API requests thus not returning it
 - not allow any kind of changes to these users

### Issues

- resolves #772

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
@@ -92,6 +92,8 @@ class UsersResponderADM extends Responder {
 
         //log.debug("usersGetV1")
 
+        // ToDO: need to have certain permissions to allow access
+
         for {
             sparqlQueryString <- Future(queries.sparql.admin.txt.getUsers(
                 triplestore = settings.triplestoreType,
@@ -1126,6 +1128,10 @@ class UsersResponderADM extends Responder {
         // log.debug("updateUserV1 - userUpdatePayload: {}", userUpdatePayload)
 
         /* Remember: some checks on UserUpdatePayloadV1 are implemented in the case class */
+
+        if (userIri.contains(KnoraSystemInstances.Users.SystemUser.id) || userIri.contains(KnoraSystemInstances.Users.AnonymousUser.id)) {
+            throw BadRequestException("Changes to built-in users are not allowed.")
+        }
 
         if (userUpdatePayload.email.nonEmpty) {
             // changing email address, so we need to invalidate the cached profile under this email
--- a/webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
@@ -92,7 +92,7 @@ class UsersResponderADM extends Responder {
 
         //log.debug("usersGetV1")
 
-        // ToDO: need to have certain permissions to allow access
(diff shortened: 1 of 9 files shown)


subotic:
@benjamingeer small review.

webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
```diff
@@ -92,6 +92,8 @@ class UsersResponderADM extends Responder {
 
         //log.debug("usersGetV1")
 
+        // ToDO: need to have certain permissions to allow access
```
benjamingeer: Is there an issue for this?

webapi/src/main/scala/org/knora/webapi/routing/admin/UsersRouteADM.scala
```diff
@@ -156,6 +156,10 @@ class UsersRouteADM(_system: ActorSystem, settings: SettingsImpl, log: LoggingAd
 
                             val userIri = stringFormatter.validateAndEscapeIri(value, throw BadRequestException(s"Invalid user IRI $value"))
 
+                            if (userIri.equals(KnoraSystemInstances.Users.SystemUser.id) || userIri.equals(KnoraSystemInstances.Users.AnonymousUser.id)) {
+                                throw BadRequestException("Changes to built-in users are not allowed.")
```
benjamingeer: Should be `ForbiddenException`?

webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
subotic: yes, should probably add the link?

subotic: #961

webapi/src/main/scala/org/knora/webapi/routing/admin/UsersRouteADM.scala
subotic: Yea, I was debating this with myself. `Forbidden` would imply that the calling user does not have the permission but that somebody else has. In this case, nobody will ever be allowed to perform this operation and that is why I'm treating it like a `BadRequest`.

But you decide ;-)

webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
benjamingeer: I like to add the issue number in the TODO, but I think the important thing is just that there's an issue, because it's easy to forget TODOs.

subotic: issue number added.

webapi/src/main/scala/org/knora/webapi/routing/admin/UsersRouteADM.scala
benjamingeer: I see what you mean. I think either would make sense. So let's leave it like this, it's easier. :)

subotic: :-)

subotic:
thanks for the review :-)

Issue #915, opened 2018-06-29
Knora should recover from temporary failure to access triplestore


subotic:
When the triplestore is not available after the Knora service successfully started, following should happen:

- change state to `unhealthy`
- answer every API request with a `503 Service Unavailable`
- start periodic checking for the  triplestore (much like at startup)
- after triplestore is available again, maybe initiate some deeper checking:
       - reload ontologies update?
       - trigger GraphDB consistency checker?
- change state to `healthy`
- start handling API requests.

subotic:
maybe useful: https://doc.akka.io/docs/akka/current/common/circuitbreaker.html

PR #979, opened 2018-08-23
refactor (webapi): remove unnecessary actor pools

### LINKED ISSUE #916: Knora should be configurable through environment variables
The environment variable can contain either:
 - the value itself
 - the path to a file containing the value

This should allow Knora to support [Docker Secrets](https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images) used for injecting the Docker container at runtime with sensitive information.

### LINKED ISSUE #977: Remove unnecessary actors and actor pools
Following our discussion about actor pools, I would like to remove the unnecessary ones.

The only two actor pools that are going to be kept are the pools of `HttpTriplestoreConnector` and `SipiResponderV1` actors.

This should simplify our architecture, allow us to get rid of a few dispatchers, thus be able to run with fewer threads and use fewer system resources.

### LINKED ISSUE #980: Knora freezes during BEOL import-test
The BEOL import test fails on `develop` (https://github.com/dhlab-basel/Knora/commit/7d379187a74f24eb260c62e6c785f440b93bfcac) after around 700 letters with status `500`:

```
{"status":4,"error":"org.knora.webapi.TriplestoreConnectionException: Connection to triplestore timed out after 10 seconds"}
```

For whatever reason, GraphDB is having a hiccup. Maybe this can be easily solved by raising the timeout.

subotic:
### Summary

- remove unnecessary actor pools
- add developer-centric `docker-compose.yml` for starting the Knora / GraphDB / Sipi
- configure `webapi` and `salsah1` thorough environment variables
- update for Java 10
- comment out the generation of fat jars from `KnoraBuild.sbt` (for now)
- update ehcache
- update sbt
- remove Kamon monitoring (for now) since we don't see anything meaningful there. We probably will have to instrument `Knora` by hand and then use Kamon for access.
- update Dockerfiles for `webapi` and `salsah1`

### Issues

- resolves #977 
- resolves #980 
- resolves #916


subotic:
Thanks for the suggestion. I will try to implement something.

subotic:
I've read the documentation. I'm not sure that this we need to do anything. Maybe #915 can be implemented by overriding the default supervision strategy. But as I understand it, it should only be used to decide what to do in case of a failure, i.e. should the actor be restarted or only the exception propagated. I don't think, that we need to restart an actor. Or do we?

subotic:
ok, the general behavior will not change by removing the pools. The only thing necessary to remove the pools was actually to remove the definition from `application.conf`. The way how actors or pools are used is exactly the same. Creating an actor or a pool is just a matter of configuration. The code for using one actor or the pool is the same.

The propagation of errors doesn't change in this PR.

subotic:
It seems that Travis has some problems today.

subotic:
Thank you for the review :-)

PR #987, opened 2018-09-04
Add support for shared ontologies

### LINKED ISSUE #986: Allow shared ontologies
Allow shared ontologies living under the IRI: http://www.knora.org/ontology/shared/[name]

subotic:
### Summary

- [x] Add built-in `DefaultSharedOntologiesProject` with project code `0000`.
- [x] Change `SmartIri` to support ontology IRIs containing the segment `shared`, using `api.knora.org` as their hostname in API v2:
  - [x] `http://www.knora.org/ontology/shared/example-box` (internal)
  - [x] `http://api.knora.org/ontology/shared/example-box/v2` (external, complex schema)
  - [x] `http://api.knora.org/ontology/shared/example-box/simple/v2` (external, simple schema)
  - [x] The project code of a `shared` ontology IRI defaults to `0000`.
  - [x] `SmartIri` can support additional shared ontology projects (by specifying the project code) if we decide to add them later (see `StringFormatterSpec`). I think this will probably end up being more convenient if we have a lot of shared ontologies.
- [x] Allow shared ontologies to be created via API v2.
- [x] Enforce these rules:
  - [x] An ontology can refer to a Knora ontology in another project only if the other ontology is built-in or shared.
  - [x] Shared ontologies must belong to `DefaultSharedOntologiesProject`, which can contain only shared ontologies.
  - [x] You can't create a resource in one project whose class is defined in a non-shared ontology in another project (so far API v1 only, because we can't yet create resources in v2).
- [x] In API v2 (complex schema), to make things easier for Salsah:
  - [x] Serve shared ontologies with `knora-api:isShared true`.
  - [x] Serve built-in ontologies with `knora-api:isBuiltIn true`.
- [x] Add tests.
- [x] Add a dummy `parole-religieuse` ontology to make `DrawingsGodsV1E2ESpec` work (because it previously created `anything:Thing` resources in the `parole-religieuse` project, which is now forbidden).

### Issues

- resolves #986


benjamingeer:
@subotic I'd like to request a review from you, but GitHub won't let me do it, because you created the PR.

benjamingeer:
@subotic I guess that if you review the PR, I can approve it. :)

subotic:
No problem. I will review it and then you can approve :-)

subotic:
Just a thought, which we can address at a later time. I think that shared ontologies should be hostable at a URL other than `http://api.knora.org/ontology/shared/[name]`. If they can only be hosted at this URL, then anybody wanting to use Knora and add shared ontologies would need to publish them with us. We (as the DaSCH) can only host shared ontologies that have something to do with the DaSCH. We (as the developers of Knora) cannot host anything but the core ontologies.

benjamingeer:
I think you’re right. I could imagine that a group of projects (all hosted on different servers) might create a shared ontology together, which they would all use. The ontology would have to be in the triplestore on each server, but its external IRI would contain the hostname of a single “authoritative” server that would serve the ontology. I guess each participating server would need to have a lookup table in a configuration file, which would enable `SmartIri` to produce the correct hostname for each shared ontology.

benjamingeer:
Or I guess the hostname could be part of the definition of a shared ontology project. Then each shared ontology project would have to be duplicated on each participating server.

subotic:
I was thinking about a default setting of `www.knora.org` and then somewhere a lookup table for remapping IRIs to URLs.

I also think that it shouldn't be such a problem to publish the base ontologies on Github.

benjamingeer:
> I was thinking about a default setting of www.knora.org

In IRIs, the domain `www.knora.org` is reserved for internal IRIs, but we could make `shared.knora.org` or something like that.

> a lookup table for remapping IRIs to URLs

That would be the easiest implementation. It could be in `application.conf`.

> I also think that it shouldn't be such a problem to publish the base ontologies on Github.

Yes, we could do that with Github Pages, but we would need a web server to process the HTTP Accept header and redirect to the file in the requested format (JSON-LD, Turtle, or RDF/XML).

benjamingeer:
Thanks for reviewing!

PR #989, opened 2018-09-05
Automatically add link value prop cardinality with link prop cardinality

### LINKED ISSUE #988: Ontology API error with link values
If adding the cardinality of a resource pointer link value, I get an error that either the link or the link property of the link value oproperty to not have a valid cardinality.

I think the reason is "improved" consistency checking. Using the Ontology API I have to add for links 2 cardinality one after the other: for the link propery and for the linkValue property. If I add the first, the second has not yet the right cardinality (that is none at all) or vice versa. So I can at the moment not add any link property because of this bug...

Suggestion: The linkValue cardinality is automatically added if the link gets a cardinality (in one transaction). I beleive that both cardinalities must always be the same.

Error example (if adding linkValue cardinality:

org.knora.webapi.BadRequestException: Resource class http://0.0.0.0:3333/ontology/1010/postcards/v2#postcard has cardinalities for one or more link value properties without corresponding link properties. The missing (or incorrectly defined) property or properties: http://0.0.0.0:3333/ontology/1010/postcards/v2#externalreference
stdClass Object

benjamingeer:
Fixes #988.

@lrosenth could you try this branch?

### DIFF
--- a/docs/src/paradox/00-release-notes/v1.8.0.md
+++ b/docs/src/paradox/00-release-notes/v1.8.0.md
@@ -13,3 +13,5 @@ See the
 ## Bugfixes:
 
 - trouble with xml-checker and/or consistency-checker during bulk import (@github[#978](#978))
+
+- Ontology API error with link values (@github[#988](#988))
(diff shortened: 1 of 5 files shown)


benjamingeer:
Thanks for the quick review! :)

PR #990, opened 2018-09-05
feature (webapi): list admin api additions part 1

subotic:
### Summary

- [x] Add creating of child nodes
- [x] Allow providing of `name` during the creation of root and child nodes
- [x] Make sure that `name` is unique on the project level
- [x] Update Swagger.json
- [x] Update release notes


subotic:
Thanks for the review.

No, I did it at first, but then after I created the corresponding case classes I saw that I could also use those without the separate OWL classes. The OWL classes would allow for better consistency checking inside the triplestore on one hand, but on the other, would mean a lot of changes to the existing data.

benjamingeer:
OK, makes sense.

subotic:
@lrosenth you should now be able to create child list nodes. This PR is already merged into `develop`.

PR #993, opened 2018-09-10
fix (remove Biblio project)

### LINKED ISSUE #992: beol and biblio ontologies refer to each other
For #987 (Add support for shared ontologies), I'm trying to enforce this rule:

> An ontology can refer to an ontology in another project only if the other ontology is shared.

This is explained [in the Knora FAQ](https://docs.knora.org/paradox/faq.html#can-a-project-use-classes-or-properties-defined-in-another-projects-ontology-):

> Can a project use classes or properties defined in another project’s ontology?
>
> No, and Knora API v2 will forbid this. Each project must be free to change its own ontologies, but this is not possible if they have been used in ontologies or data created by other projects.
>
> Instead, there will be a process for standardising ontologies that can be shared by multiple projects (issue #523).

I've just realised that the `beol` and `biblio` ontologies, which are in separate projects, refer to each other. For example, the definition of `biblio:publicationHasManuscript` refers to `beol:manuscript`.

I think there are two possible solutions:

1. Put the `biblio` ontology in the `beol` project for now, until it's ready to be standardised as a shared ontology.
2. Change the `biblio` ontology so it doesn't refer to the `beol` ontology; then it could become a shared ontology now.

I'm guessing that (1) would be easier. @SepidehAlassi what do you think?

SepidehAlassi:
biblio project is removed, biblio ontology is part of beol project now

Fixes #992.

### DIFF
--- a/webapi/_test_data/all_data/admin-data.ttl
+++ b/webapi/_test_data/all_data/admin-data.ttl
@@ -458,33 +458,6 @@ Die Internetpublikation macht das digitalisierte Korpus dieser Frühdrucke  durc
 
 
 ###################################
-# BIBLIO                          #
-###################################
-
-<http://rdfh.ch/projects/DczxPs-sR6aZN91qV92ZmQ> a knora-base:knoraProject ;
-  knora-base:projectShortname "biblio"^^xsd:string ;
-  knora-base:projectShortcode "0802"^^xsd:string ;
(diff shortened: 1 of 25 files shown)


SepidehAlassi:
issue had blocking label, :/ had to fix it fast :)

benjamingeer:
Much appreciated!

Issue #998, opened 2018-09-20
Create/update file values in API v2


benjamingeer:
### How API v1 does this

* Each resource can have at most one file, and the property IRI depends solely on the file type (in `SipiConstants`), so we need only the resource IRI and the file itself.
* Knora's Lua scripts for Sipi allow anyone to upload a file.
* There are two use cases:
  * The "GUI case", in which SALSAH uploads the file directly to Sipi, then sends a confirmation to Knora.
  * The "non-GUI case", in which the client uploads the file directly to Knora, which sends it to Sipi. 

### Goals for API v2

* Allow creating a resource containing multiple files.
* Provide a way for Sipi to determine if the user has permission to upload a file.

### Design

Akka HTTP can process a request with multiple files:

https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/file-upload-directives/fileUploadAll.html

To authorise an upload in Sipi, I guess we're going to use JWT (#520)? How will this work in SALSAH?

benjamingeer:
Is there actually any reason for the client to submit files to Knora? Couldn't it just submit them to Sipi, then just tell Knora about the metadata?

loicjaouen:
For the user, it is simpler to send a request to knora and then let the knora/sipi magic happen behind the scene. This is very much in the bulk import way of doing (I just don't like the implementation that expect a shared directory between knora and sipi).

For the user again, The current *GUI Case* is more convoluted (sending a file to sipi, keep the returned iri, send it to knora).

benjamingeer:
> For the user, it is simpler to send a request to knora and then let the knora/sipi magic happen behind the scene.

I'm not sure it's really simpler. To make this work in API v2, you would need to send a single `muiltipart/form-data` request to Knora, containing:

* A JSON-LD document with a special magic name (e.g. `request.jsonld`), with content referring to filenames.
* One or more files, whose filenames are referred to in the JSON-LD document.

So in any case this would be different and more complex than a normal POST/PUT request.

If you send the files to Sipi first, you will have to construct exactly same JSON-LD request (referring to the files that you sent to Sipi). And you can make a normal POST/PUT to Knora instead of a special `multipart/form-data`.

In any case, this will all be handled by a Salsah component, so users won't need to care about it.

[...]

PR #999, opened 2018-09-25
Create resources in API v2, part 1

benjamingeer:
- [x] Create a new resource with values (no file values yet).
- [x] Fix the property `knora-api:textValueHasMapping` so its object is a JSON-LD object containing `@id` (also needs to be fixed in Knora-ui)
- [x] Add tests.
  - [x] Change resources route R2R test to an E2E test.
- [x] Add docs.

For future PRs:

* Create resources with file values (depends on #998).
* Update resource metadata.
* Mark resources as deleted.
* Implement API v2 bulk import (the design here tries to allow for this in a less complicated way than in v1).


flavens:
Fixes have been done in Knora-ui - [PR #87](https://github.com/dhlab-basel/Knora-ui/pull/87)
Review requested before merging.

benjamingeer:
@subotic Thank you!

PR #1006, opened 2018-10-10
Rewrite date processing in API v2

### LINKED ISSUE #803: Remove remaining dependencies on API v1 code from API v2 code

### LINKED ISSUE #928: Refactor DateUtilV2
`DateUtilV2` should:

* Not depend on API v1 code
* Be able to convert in both directions between `DateValueContentV2` and Julian day numbers

benjamingeer:
This rewrites date processing in API v2 to use [ICU4J](http://site.icu-project.org/), which supports many different calendars as well as Julian Day Numbers. This PR doesn't include support for calendars other than Gregorian and Julian, but now it should be much easier to add them. It removes the remaining dependencies of API v2 code on API v1 code, and eliminates the need to convert dates to strings when converting them between JSON-LD and Julian Day Numbers.

- [x] Add new date processing classes in `org.knora.webapi.util.date.CalendarDateUtilV2`.
- [x] Change existing v2 code to use the new classes.
- [x] Add tests to check that everything works as before.

Resolves #803.
Resolves #928.


webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
```diff
@@ -0,0 +1,543 @@
...
+
+/**
+  * Represents a date range consisting of two instances of [[CalendarDateV2]]. Both instances must have the same
+  * calendar.
+  *
+  * @param startCalendarDate the start of the range.
+  * @param endCalendarDate the end of the range.
+  */
+case class CalendarDateRangeV2(startCalendarDate: CalendarDateV2, endCalendarDate: CalendarDateV2) {
+    if (startCalendarDate.calendarName != endCalendarDate.calendarName) {
```
tobiasschweizer: You could also check if `startCalendarDate ` is before or equals `endCalendarDate`.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
```diff
@@ -0,0 +1,543 @@
...
+  * Represents the name of the Gregorian calendar.
+  */
+case object CalendarNameGregorian extends CalendarNameGregorianOrJulian {
+    override def toString: String = StringFormatter.CalendarGregorian
+}
+
+/**
+  * Represents the name of the Julian calendar.
+  */
+case object CalendarNameJulian extends CalendarNameGregorianOrJulian {
```
tobiasschweizer: How would that be extended once we support more calendars than Gregorian and Julian?

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
```diff
@@ -0,0 +1,543 @@
...
+        throw AssertionException(s"Invalid date: CalendarDateV2($calendarName, $year, $maybeMonth, $maybeDay, $maybeEra)")
+    }
+
+    calendarName match {
+        case _: CalendarNameGregorianOrJulian =>
+            if (maybeEra.isEmpty) {
+                throw AssertionException(s"Era is required in calendar $calendarName")
+            }
+
+        case _ => ()
```
tobiasschweizer: In this case, we would have a non supported calendar, right? 

Could you explain why there is a `CalendarNameV2` and a trait `CalendarNameGregorianOrJulian`? 

Is your rationale that you would add additional traits for other calendars and include them in the match case statement?

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
```diff
@@ -0,0 +1,543 @@
...
+                    maybeMonth = Some(month),
+                    maybeDay = Some(day),
+                    maybeEra = maybeEra
+                )
+        }
+    }
+
+    /**
+      * Parses a string representing a single date, without the calendar. This method does does not check that the
+      * date is valid in its calendar; to do that, call `toJulianDayRange` on it.
```
tobiasschweizer: The actual validity will only be checked if a `GregorianCalendar` is instantiated and leniency is set to false, right?

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
```diff
@@ -0,0 +1,543 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.knora.webapi.util.date
+
+import java.util.Date
```
tobiasschweizer: Is the Java date class (GregorianCalendar) used anywhere in this code?

webapi/WebapiBuild.sbt
```diff
@@ -318,6 +319,8 @@ lazy val library =
 
         // Java EE modules which are deprecated in Java SE 9, 10 and will be removed in Java SE 11
         val jaxbApi                = "javax.xml.bind"                % "jaxb-api"                 % "2.2.12"
+
+        val icu4j                  = "com.ibm.icu"                   % "icu4j"                    % "62.1"
```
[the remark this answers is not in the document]
tobiasschweizer: Or do you want to leave `DateUtilV1 ` as it is? Could we make call v1 v2 methods?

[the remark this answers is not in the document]
benjamingeer: `DateUtilV1` uses jodd. I could refactor it to use the v2 code, but that would be more work. My idea was to avoid changing v1 if possible, because the priority is to finish v2.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer: Yes, good point.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer:
```scala
case object CalendarNameIslamic extends CalendarNameV2
case object CalendarNameHebrew extends CalendarNameV2
```

And so on.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer: `CalendarNameGregorianOrJulian` is for things like this:

```scala
def getGregorianCalendarChangeDate(calendarName: CalendarNameGregorianOrJulian): Date
```

That method should only accept a Gregorian or Julian calendar, and using the type `CalendarNameGregorianOrJulian` guarantees that this is the case.

benjamingeer: > In this case, we would have a non supported calendar, right?

That's actually not possible with this code. The `case _` is just there for the future, when we have more calendars. Until then, I could remove that line, because it can never be reached.

The design pattern here is called *sealed case objects*, which is an alternative to using Scala enumerations. See:

https://pedrorijo.com/blog/scala-enums/

With sealed case objects, the compiler can guarantee that the pattern matching is exhaustive.

benjamingeer: > Is your rationale that you would add additional traits for other calendars and include them in the match case statement?

Yes.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer: Yes, and this happens in `toJulianDayRange`.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer: No.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer: Done in https://github.com/dhlab-basel/Knora/pull/1006/commits/c1977f8e8caa8bb301921baf988d1eedd6857a2e.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer: We're already using sealed case objects for ontology schemas:

https://github.com/dhlab-basel/Knora/blob/5f7df5d04535463ca9acfbfbd7f36083b3da3981/webapi/src/main/scala/org/knora/webapi/OntologySchema.scala

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer: `com.ibm.icu.util.GregorianCalendar` replaces `java.util.GregorianCalendar`.

benjamingeer: http://icu-project.org/apiref/icu4j/com/ibm/icu/util/GregorianCalendar.html

> ICU's replacement for java.util.GregorianCalendar.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer: The other advantage of sealed case objects is that you can make a hierarchy of subtypes, like `CalendarNameV2` -> `CalendarNameGregorianOrJulian` -> `CalendarNameGregorian`, which you can't do with enumerations.

webapi/WebapiBuild.sbt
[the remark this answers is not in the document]
tobiasschweizer: >  I could refactor it to use the v2 code, but that would be more work. My idea was to avoid changing v1 if possible, because the priority is to finish v2.

I understand. My concern is that both versions of the API have to behave identically when converting from and to JDN and that this  might be more difficult to assure if we use two different libraries for v1 and v2.

But of course, there could be a version update of one of those libraries and then we would have the same issue if the new version behaves differently from the old one.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
tobiasschweizer: great, thx.

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
tobiasschweizer: got it

webapi/WebapiBuild.sbt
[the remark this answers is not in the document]
tobiasschweizer: Ok, could you create a separate issue for that so we can do it later?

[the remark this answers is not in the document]
benjamingeer: #1007

benjamingeer:
OK to merge now?

webapi/src/main/scala/org/knora/webapi/util/date/CalendarDateUtilV2.scala
benjamingeer: I'm using the ICU `GregorianCalendar` class instead of the one in the Java standard library, because the ICU class supports Julian Day Numbers.

benjamingeer:
Thanks! :)

PR #1008, opened 2018-10-15
bug fix simpletext

loicjaouen:
here `$this` is not a good idea as it depends on the context


loicjaouen:
this is a one liner... it is my mistake in the first place

loicjaouen:
Thanks @subotic !

PR #1009, opened 2018-10-15
Implement graph query in API v2

benjamingeer:
This ports the API v1 graph query functionality to v2, with a couple of additional features:

* In v1, both inbound and outbound links are always queried. Here you can query only inbound links, only outbound links, or both.
* In v1, `knora-base:isPartOf` is always excluded. Here you can exclude any property, or no property.

The results are returned in a JSON-LD format that's as compact as possible, to accommodate large graphs. Each resource is represented only by its IRI, class, and label, and direct links are shown instead of link values.

In v1, resource class labels and property labels were returned; these are not included here, because the client can get them from the ontology route.

Results can be returned in either the simple or the complex schema (the only difference being the IRIs).

- [x] Refactor v1 implementation for v2.
- [x] Add tests.
  - [x] Refactor v1 unit tests for v2.
  - [x] Add e2e tests.
- [x] Add documentation.
- [x] Update release notes.


benjamingeer:
Thanks for the review!

PR #1013, opened 2018-10-22
fix (salsah): check if there is a value before accessing it

### LINKED ISSUE #1012: bug: it is not possible to add a boolean
missing-bool

here we see that clicking on the plus to add a boolean property crashes; the input checkbox doesn't appear and nothing more happens.

loicjaouen:
fixes #1012

result after the fix:

missing-bool-fix


loicjaouen:
thanks @subotic

PR #1023, opened 2018-10-31
default view permissions to images

### LINKED ISSUE #1022: Create default permissions for BEOL resources
For BEOL, we should add default permissions, so that image files have view permissions for unknown users. 

Currently unknown users only have restricted view permissions.

SepidehAlassi:
resolves #1022

### DIFF
--- a/webapi/_test_data/all_data/permissions-data.ttl
+++ b/webapi/_test_data/all_data/permissions-data.ttl
@@ -282,7 +282,7 @@
 
                       knora-base:forResourceClass beol:figure ;
 
-                      knora-base:hasPermissions "CR knora-base:Creator|M knora-base:ProjectMember|V knora-base:KnownUser|V knora-base:UnknownUser"^^xsd:string .
+                      knora-base:hasPermissions "CR knora-base:Creator|M knora-base:ProjectMember|V knora-base:KnownUser,knora-base:UnknownUser"^^xsd:string .
 
 
 ### Default Object Access Permissions on beol:documentImage resource class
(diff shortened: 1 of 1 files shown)


webapi/_test_data/all_data/permissions-data.ttl
```diff
@@ -273,6 +273,29 @@
...
+### Default Object Access Permissions on beol:figure resource class
+<http://rdfh.ch/permissions/0801/006-d2>
+
+                      rdf:type knora-base:DefaultObjectAccessPermission ;
+
+                      knora-base:forProject <http://rdfh.ch/projects/yTerZGyxjZVqFMNNKXCDPF> ;
+
+                      knora-base:forResourceClass beol:figure ;
+
+                      knora-base:hasPermissions "CR knora-base:Creator|M knora-base:ProjectMember|V knora-base:KnownUser|V knora-base:UnknownUser"^^xsd:string .
```
tobiasschweizer: Shouldn't this be written as like "`CR knora-base:Creator|M knora-base:ProjectMember|V knora-base:KnownUser,knora-base:UnknownUser"^^xsd:string ."`?

https://docs.knora.org/paradox/02-knora-ontologies/knora-base.html#permissions

webapi/_test_data/all_data/permissions-data.ttl
```diff
@@ -273,6 +273,29 @@
...
+### Default Object Access Permissions on beol:documentImage resource class
+<http://rdfh.ch/permissions/0801/006-d3>
+
+                      rdf:type knora-base:DefaultObjectAccessPermission ;
+
+                      knora-base:forProject <http://rdfh.ch/projects/yTerZGyxjZVqFMNNKXCDPF> ;
+
+                      knora-base:forResourceClass beol:documentImage ;
+
+                      knora-base:hasPermissions "CR knora-base:Creator|M knora-base:ProjectMember|V knora-base:KnownUser|V knora-base:UnknownUser"^^xsd:string .
```
tobiasschweizer: see 
https://github.com/dhlab-basel/Knora/pull/1023/files#r229959057

webapi/_test_data/all_data/permissions-data.ttl
SepidehAlassi: both have the same effect at the end. But you are right according to

> Each permission abbreviation is followed by a space, then a comma-separated list of groups that the permission is granted to.

comma separated list of groups is better to be used.

PR #1048, opened 2018-11-07
feature (webapi): add username field to UserADM

### LINKED ISSUE #1047: Add 'username' field to UserADM
After a discussion together with @kilchenmann and @lrosenth, we would like to add back the `username` field.

The reasoning is that a user can have a page like https://github.com/subotic, and then showing the `email` in the URL would not only be ugly but also reveal the email, which the user maybe doesn't want to make public at all.

subotic:
### Issues

- resolves #1047


benjamingeer:
Will try to do this tomorrow, otherwise on Monday.

subotic:
Thanks :-)

subotic:
@benjamingeer gentle reminder :-)

webapi/src/main/scala/org/knora/webapi/util/StringFormatter.scala
```diff
@@ -699,6 +700,9 @@ class StringFormatter private(val knoraApiHostAndPort: Option[String]) {
     private val SalsahGuiAttributeRegex: Regex =
         """^(\p{L}+)=(.+)$""".r
 
+    // A regex for matching a string containing an email address.
+    private val EmailAddressRegex: Regex = ("""^([0-9a-zA-Z]([-.\w]*[0-9a-zA-Z])*@(([0-9a-zA-Z])+([-\w]*[0-9a-zA-Z])*\.)+[a-zA-Z]{2,9})$""").r
```
benjamingeer: This won't work for email address with non-ASCII characters, such as email addresses at [Internationalized Domain Names](https://www.icann.org/resources/pages/idn-2012-02-25-en).  Some [examples](https://en.wikipedia.org/wiki/Email_address#Internationalization_examples):

* Latin alphabet with diacritics: Pelé@example.com
* Greek alphabet: δοκιμή@παράδειγμα.δοκιμή
* Traditional Chinese characters: 我買@屋企.香港
* Japanese characters: 二ノ宮@黒川.日本
* Cyrillic characters: чебурашка@ящик-с-апельсинами.рф
* Devanagari characters: संपर्क@डाटामेल.भारत

There seems to be a lot of debate on StackOverflow and elsewhere about whether it's really possible to validate an email address with a regex. Other options:

* An Apache Commons class called [EmailValidator](https://commons.apache.org/proper/commons-validator/apidocs/index.html?org/apache/commons/validator/routines/EmailValidator.html). I've just tried it, and it rejects the Greek and Japanese examples above.
* [The javax.mail API](https://kodejava.org/how-do-i-validate-email-address-using-java-mail-api/), which you can [get from Maven](https://mvnrepository.com/artifact/com.sun.mail/javax.mail/1.6.2). But it rejects the Devanagari example above, which is actually the real contact email address of [this hosting service in India](http://www.xn--c2bd4bq1db8d.xn--h2brj9c/).

My suggestion: use a very minimal regex like `^.*@.*$`, and actually send an email confirmation to the user to check that they received the email. Because even if the address is syntactically valid, the user could still have mistyped it. If they can actually receive a message there, we know it's OK.

webapi/_test_data/all_data/admin-data.ttl
```diff
@@ -322,16 +343,16 @@ Die Internetpublikation macht das digitalisierte Korpus dieser Frühdrucke  durc
 
 <http://rdfh.ch/users/multiuser> rdf:type knora-base:User ;
 
-                                        knora-base:userid "multiuser"^^xsd:string ;
```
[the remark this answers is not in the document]
subotic: We had `knora-base:userid`, but removed it some time ago. These are remnants of those time ;-)

webapi/src/main/scala/org/knora/webapi/util/StringFormatter.scala
subotic: Yes, the email confirmation is a good suggestion. I've opened an issue for this #1065.

benjamingeer: I guess we can require at least one character before and after the `@`. :)

subotic:
No problem. Thanks for the review :-)

PR #1074, opened 2018-11-23
Fix unescaping in update check and reading standoff URL

### LINKED ISSUE #1069: Error on update of a richtext property with a mapping
Using the `http://localhost:3333/v2/values` endpoint to update a richtext property with a mapping 
 with for example this body: 
```json
{
    "@id": "http://rdfh.ch/0113/JYIs9DmnR6eUVEf6oZfS0w",
    "@type": "onto:Person",
    "onto:hasBiography": {
        "@id": "http://rdfh.ch/0113/JYIs9DmnR6eUVEf6oZfS0w/values/3AYB5YIgS8ChR2FltUy-Sw",
        "@type": "knora-api:TextValue",
        "knora-api:textValueAsXml": "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n    <text>\n    <p>Test</p>\n    </text>",
        "knora-api:textValueHasMapping": {
            "@id": "http://rdfh.ch/standoff/mappings/StandardMapping"
        }
    },
    "@context": {
        "knora-api": "http://api.knora.org/ontology/knora-api/v2#",
        "onto": "http://0.0.0.0:3333/ontology/0113/lumieres-lausanne/v2#"
    }
}
```

I don't understand the following error while, finally, the property is modified in graphdb as expected:
```json
{"error":"org.knora.webapi.AssertionException: The value saved as http://rdfh.ch/0113/JYIs9DmnR6eUVEf6oZfS0w/values/-8KBL68LSfOEP36hj2ZSOw is not the same as the one that was submitted"}
```

benjamingeer:
This fixes two bugs:

- After a resource was created in API v2, and the values were read back from the triplestore to check that they were saved correctly, the saved values were compared with escaped expected values (e.g. a linefeed would be `\n`). This fixes the comparison so the expected values are unescaped (like the ones read from the triplestore).
- When a `StandoffUriTag` was read from the triplestore, it was given a `StandoffTagIriAttributeV2` instead of a `StandoffTagUriAttributeV2`.

Fixes #1069.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
@@ -59,8 +59,8 @@ class ResourcesResponderV2 extends ResponderWithStandoffV2 {
       *                                       the resource.
       * @param values                         the resource's values for verification.
       */
-    case class ResourceReadyToCreate(sparqlTemplateResourceToCreate: SparqlTemplateResourceToCreate,
-                                     values: Map[SmartIri, Seq[GenerateSparqlForValueInNewResourceV2]])
+    private case class ResourceReadyToCreate(sparqlTemplateResourceToCreate: SparqlTemplateResourceToCreate,
+                                             values: Map[SmartIri, Seq[UnverifiedValueV2]])
 
     override def receive: Receive = {
         case ResourcesGetRequestV2(resIris, requestingUser) => future2Message(sender(), getResources(resIris, requestingUser), log)
@@ -324,7 +324,7 @@ class ResourcesResponderV2 extends ResponderWithStandoffV2 {
                 resourceClassIri = internalCreateResource.resourceClassIri.toString,
                 resourceLabel = internalCreateResource.label
             ),
-            values = valuesWithValidatedPermissions
+            values = sparqlForValuesResponse.unverifiedValues
         )
     }
 
@@ -663,7 +663,7 @@ class ResourcesResponderV2 extends ResponderWithStandoffV2 {
 
             _ = resource.values.foreach {
                 case (propertyIri: SmartIri, savedValues: Seq[ReadValueV2]) =>
-                    val expectedValues: Seq[GenerateSparqlForValueInNewResourceV2] = resourceReadyToCreate.values(propertyIri)
+                    val expectedValues: Seq[UnverifiedValueV2] = resourceReadyToCreate.values(propertyIri)
 
                     if (expectedValues.size != savedValues.size) {
                         throw AssertionException(s"Resource <$resourceIri> was saved, but it has the wrong values")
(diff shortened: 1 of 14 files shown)


webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -324,7 +324,7 @@ class ResourcesResponderV2 extends ResponderWithStandoffV2 {
                 resourceClassIri = internalCreateResource.resourceClassIri.toString,
                 resourceLabel = internalCreateResource.label
             ),
-            values = valuesWithValidatedPermissions
+            values = sparqlForValuesResponse.unverifiedValues
```
benjamingeer: This was doing the update verification with the (escaped) values that were sent to `ValuesResponderV2`, not with the `UnverifiedValueV2` objects returned by `ValuesResponderV2`, which should have been (but weren't) unescaped.

webapi/src/main/scala/org/knora/webapi/responders/v2/ValuesResponderV2.scala
```diff
@@ -554,7 +554,7 @@ class ValuesResponderV2 extends Responder {
             insertSparql = insertSparql,
             unverifiedValue = UnverifiedValueV2(
                 newValueIri = newValueIri,
-                value = valueToCreate.valueContent,
+                valueContent = valueToCreate.valueContent.unescape,
```
benjamingeer: The `UnverifiedValueV2` objects returned by `ValuesResponderV2` needed to be unescaped.

webapi/src/main/scala/org/knora/webapi/util/standoff/StandoffTagUtilV2.scala
```diff
@@ -750,7 +750,8 @@ object StandoffTagUtilV2 {
                                 case Some(SmartIriLiteralV2(SmartIri(OntologyConstants.Xsd.Boolean))) =>
                                     StandoffTagBooleanAttributeV2(standoffPropertyIri = propIri, value = value.toBoolean)
 
-                                case Some(SmartIriLiteralV2(SmartIri(OntologyConstants.Xsd.Uri))) => StandoffTagIriAttributeV2(standoffPropertyIri = propIri, value = value)
+                                case Some(SmartIriLiteralV2(SmartIri(OntologyConstants.Xsd.Uri))) =>
+                                    StandoffTagUriAttributeV2(standoffPropertyIri = propIri, value = value)
```
benjamingeer: The wrong attribute object was constructed here.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -663,15 +663,15 @@ class ResourcesResponderV2 extends ResponderWithStandoffV2 {
...
+                    val expectedValues: Seq[UnverifiedValueV2] = resourceReadyToCreate.values(propertyIri)
 
                     if (expectedValues.size != savedValues.size) {
                         throw AssertionException(s"Resource <$resourceIri> was saved, but it has the wrong values")
                     }
 
                     savedValues.zip(expectedValues).foreach {
                         case (savedValue, expectedValue) =>
-                            if (!(savedValue.valueContent.wouldDuplicateCurrentVersion(expectedValue.valueContent) &&
+                            if (!(expectedValue.valueContent.wouldDuplicateCurrentVersion(savedValue.valueContent) &&
```
tobiasschweizer: see above

webapi/src/main/scala/org/knora/webapi/util/standoff/StandoffTagUtilV2.scala
tobiasschweizer: again: I took pain killers when I wrote this code (dental surgery)

PR #1018, opened 2018-10-29
feat(salsah1): Add case for boolean type in extended search

loicjaouen:
fixes #1017

PR #1077, opened 2018-11-26
Fix broken SALSAH 1 tests

benjamingeer:
@loicjaouen I think this might fix the problem you're having in #1018.


loicjaouen:
super cool! thanks Ben, I am trying that right now

PR #1083, opened 2018-11-29
releasing v3.0.0

subotic:


subotic:
@benjamingeer only a click ;-)

subotic:
@benjamingeer could you maybe do a quick review? I see that we are both having free today :-) Happy Dies Academicus :-)

Issue #520, opened 2017-05-31
Have webapi send JWT to Sipi for auth


benjamingeer:
The Knora API server can use JWT to authorise a user to upload files to Sipi, as well as to view files stored in Sipi.

benjamingeer:
It seems that JWT isn't going to work, because of HTML `img` tags, so it looks like Sipi will need to make a request to the Knora API server for this.

subotic:
relevant: #171

subotic:
As I understand it, this is only a problem in Salsah 1.5, where the `img` tag is used. In Salsah 2.0 this shouldn't be a problem. I will talk to @lrosenth about this.

lrosenth:
I think SIPI is accepting webtokens as cookie or as URL option (don't remember the "keyword" to be used...). May be this can be used for Salsah 1.5 bay adding the token to the URL in the img-ag.

Sent from my iPad

On 9 Aug 2017, at 11:00, Ivan Subotic <notifications@github.com<mailto:notifications@github.com>> wrote:


As I understand it, this is only a problem in Salsah 1.5, where the img tag is used. In Salsah 2.0 this shouldn't be a problem. I will talk to @lrosenth<https://github.com/lrosenth> about this.

subotic:
> May be this can be used for Salsah 1.5 bay adding the token to the URL in the `img-tag`.

Yes, this is what I also had in mind. We could generate so called `access tokens` and add them to every image URL. Those access tokens would only be valid for one particular image and only for a short amount of time. The pre-flight script in SIPI could then check the validity of the token and give access, all without the need to contact `webapi`.

benjamingeer:
Just talked to Lukas about this again. Here's the problem:

Putting a JWT in an image URL is fine for SALSAH 1.5, but it won't work for a IIIF viewer. The IIIF auth standard expects to get a JWT in a cookie or in an HTTP auth header. There's no way to set an auth header for an `<img>` tag, so the cookie is the only option. But we can't make one giant cookie containing the user's permissions on all images. We can make a cookie that says who the user is, but then Sipi still has to make a request to Knora to ask for the user's permissions on a particular image. And then we're back where we started.

So my question is this: If the purpose of using JWT is to avoid the Sipi-to-Knora request, but we're actually going to have to make that request anyway, why go to the trouble of using JWT? Why not just keep using cookies as we do now?

subotic:
[...]

PR #1084, opened 2018-11-29
fix (webapi): image permission

### LINKED ISSUE #1003: After an image has been uploaded, the user is unable to see it (only RV)
Using Salsah 1.5, after a user has uploaded an image (i.e. `anything.admin@example.org` creating a `anything:Dingbild`), he is unable to view it properly afterwards. 

The image is only available to him in restricted view. 

It looks like there is a bug in permissions that didn't exist before.

### LINKED ISSUE #1086: Make preview image URLs returned by make_thumbnail configurable

### LINKED ISSUE #1091: Sipi should use Knora's session cookie
Sipi should use Knora's session cookie for authentication, which is sent to Sipi anyway for each request and not open a separate session.

subotic:
### Issues

- resolves #1003 
- resolves #1086 
- resolves #1091


subotic:
`webapi` is not the problem. Added tests to check `v1/files` route and it seems to work as it should.

subotic:
There is a problem with `Sipi` where the threeway authentication (Salsah -> Sipi -> webapi) over the cookie is not working as it should.

gfoo:
is this pull request also related to #520 ?

subotic:
No, if I get the cookies again to work with Sipi. If not, then I will enable JWT but only for V1 of the API, so that Salsah1 can work correctly. Work will commence on #520 in January and will incorporate V1 and V2 of the API. But there we still have to find a solution that is compatible with generic IIIF viewers (Mirador, etc.).

mrivoal:
Yes, thanks @subotic !

subotic:
> Could you also update the design docs

Yes, will do that.

PR #1085, opened 2018-11-30
Update technical user docs

### LINKED ISSUE #1045: Explanation of "How does handle Knora (XML-)Files"
possible cases according to @tobiasschweizer : 

- TextValue  (--> StandOff)
- FileValue (--> SIPI, XSL Transformation)

It is really hard to find information about how xml files are stored with Knora in the documentation. Something seems to be here: https://docs.knora.org/paradox/03-apis/api-v1/adding-resources.html#generate-xml (but not clear which "data" is meant at all)

For new users, maybe something like "If you have an image, this happens. If you have an XML file that, and if you only have text, this will happen" could be somewhere explained without much text and the technical details (which are very detailed in the documentation), only concentrating what and especially "why" you can do this or that?

### LINKED ISSUE #1046: Digital vs. non-digital representation
When it comes to the (quite often-used) distinction of non-digital vs. digital representations In the documentation, it is not quite clear what is meant at all. Finally I found some explanation about Representations (without attribute) in

"/Knora Onotologies/The Knora Base Ontology/Resources/Representations" 

what can only be found by incident or full text search.

### LINKED ISSUE #1061: Get XML Schemas in Bulk Import API v1
Hi
I am trying the bulk import v1 as documented in  https://docs.knora.org/paradox/03-apis/api-v1/adding-resources. I try to get the XML Schemas and interpret the 

HTTP GET to http://host/v1/resources/xmlimportschemas/ontologyIRI as follows:

http://localhost:3333/v1/resources/xmlimportschemas/0802/biblio

With this I get a 404 Error. What is wrong with the link? Thanks for your help.

benjamingeer:
- [x] Rewrite **What Is Knora**? (The new sections there can then be linked from the knora.org home page.)
- [x] Add an introduction to standoff.
- [x] Clarify `Representation`.
- [x] Clarify bulk import docs.
- [x] Add more docs about shared ontologies.

Resolves #1045.
Resolves #1046.
Resolves #1061.

### DIFF
--- a/docs/src/paradox/01-introduction/standoff-rdf.md
+++ b/docs/src/paradox/01-introduction/standoff-rdf.md
@@ -17,15 +17,8 @@ You should have received a copy of the GNU Affero General Public
 License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
-# Introduction
+# Standoff/RDF Text Markup
 
-@@toc { depth=2 }
+@@toc
 
-@@@ index
-
-* [What Is Knora?](what-is-knora.md)
-* [Data Formats in Knora](data-formats.md)
-* [Standoff/RDF Text Markup](standoff-rdf.md)
-* [An Example Project](example-project.md)
-
-@@@
+TODO
--- a/docs/src/paradox/01-introduction/standoff-rdf.md
+++ b/docs/src/paradox/01-introduction/standoff-rdf.md
@@ -45,25 +45,24 @@ such as XML:
 In Knora's Standoff/RDF, a tag is an RDF entity that is linked to a @ref:[text value](../02-knora-ontologies/knora-base.md#textvalue). Each tag points to a substring
 of the text, and has semantic properties of its own. You can define your own tag classes
 in your ontology by making subclasses of `knora-base:StandoffTag`, and attach your own
-properties to them.
+properties to them. You can then search for those properties using Knora's search language, @ref:[Gravsearch](../03-apis/api-v2/query-language.md).
 
 The built-in @ref:[knora-base](../02-knora-ontologies/knora-base.md) and `standoff` ontologies
 provide some basic tags that can be reused or extended. These include tags that represent
 Knora data types. For example, `knora-base:StandoffDateTag` represents a date in exactly the
 same way as a Knora @ref:[date value](../02-knora-ontologies/knora-base.md#datevalue), i.e. as a
 calendar-independent astronomical date. You can use this tag as-is, or extend it by making
-a subclass, to represent dates in texts. Knora's search language, @ref:[Gravsearch](../03-apis/api-v2/query-language.md), provides built-in functionality for
-searching for these data type tags. Thus you can search for text containing a date that
+a subclass, to represent dates in texts. Gravsearch includes built-in functionality for
+searching for these data type tags. For example, you can search for text containing a date that
 falls within a certain @ref:[date range](../03-apis/api-v2/query-language.md#matching-standoff-dates).
 
 Knora's APIs support automatic conversion between XML and Standoff/RDF. To make this work,
 Standoff/RDF stores the order of tags and their hierarchical relationships. You must define an @ref:[XML-to-Standoff Mapping](../03-apis/api-v2/xml-to-standoff-mapping.md) for your standoff tag classes and properties.
 Then you can import an XML document into Knora, which will store it as Standoff/RDF. The text and markup
-can then be searched using @ref:[Gravsearch](../03-apis/api-v2/query-language.md). When you retrieve
-the document, Knora converts it back to the original XML.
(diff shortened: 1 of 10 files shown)


docs/src/paradox/01-introduction/what-is-knora.md
```diff
@@ -19,51 +19,101 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
...
+a content management system for the long-term preservation and reuse of
+humanities data. It is designed to accommodate data with a complex internal
+structure, including data that could be stored in relational databases.
+
+A key problem in the long-term preservation of data is the multitude of
+data formats in use, many of which are proprietary and quickly become obsolete.
+It is not practical to maintain all the software that was used to create and
+read old data files.
+
+Instead of preserving a multitude of different data formats, Knora supports
```
tobiasschweizer: I think you should stress the difference between a classical digital archive and Knora: In a digital archive, contents are preserved by using standard archival formats like PDF/A but they are not directly accessible (https://en.wikipedia.org/wiki/Open_Archival_Information_System). The contents in a digital archive are only searchable via metadata. Once you have found something, it is handed over to you.

When it comes to databases, these are stored in a non functional state, e.g., a relational database is serialized to XML.

Knora on the other hand preserves data by keeping it alive. It is a keep-alive archive. You can import a relational database into Knora and still query it. Of course, we do not emulate the original SQL queries.

tobiasschweizer: I tried to describe that here: http://dhlab.unibas.ch/documents/schweizer//archiving_2014.pdf

It is outdated because it was written before Knora came into existence. But I think the basic thought is still valid.

docs/src/paradox/01-introduction/what-is-knora.md
```diff
@@ -19,51 +19,101 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
...
+the project-specific data models, and meets Knora's minimum requirements
+for interoperability and reusability of data.
+
+## Linked Open Data
+
+Knora supports publishing data online as as [Linked Open Data](http://linkeddata.org/),
+using open standards to allow interoperability between different repositories
+on the web.
+
+## Build Your Own Application
```
tobiasschweizer: This is outdated. You can now build your own VREs using modules provided by KnoraUI (https://github.com/dhlab-basel/Knora-ui). SALSAH is just one possible VRE that make use of Knora UI.

docs/src/paradox/faq.md
```diff
@@ -21,6 +21,26 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
...
+### Does Knora store XML files?
+
+XML files do not lend themselves to searching and linking. Knora's RDF storage
+is better suited to its goal of facilitating data reuse.
+
+If your XML files represent text with markup (e.g. [TEI/XML](http://www.tei-c.org/)),
+the recommended approach is to allow Knora to store it as
+@ref:[Standoff/RDF](01-introduction/standoff-rdf.md). This will allow both text and
+markup to be searched using @ref:[Gravsearch](03-apis/api-v2/query-language.md). Knora
+can also regenerate the original XML document at any time.
```
tobiasschweizer: `original XML document` -> does this refer to its parsed state? I think there might be differences if you just diff the XML files (order of attributes), but these differences do not matter.

docs/src/paradox/01-introduction/standoff-rdf.md
```diff
@@ -0,0 +1,72 @@
...
+separately as RDF data. This approach has some advantages over commonly used markup systems
+such as XML:
+
+- XML and other hierarchical markup systems assume that a document is a hierarchy, and
+  have difficulty representing
+  [non-hierarchical structures](http://www.tei-c.org/release/doc/tei-p5-doc/en/html/NH.html)
+  or multiple overlapping hierarchies. Standoff markup can easily represent these structures.
+- Markup languages are typically designed to be used in text files. But there is no
+  standard system for searching and linking together many different text files containing
+  markup. By storing markup as RDF, Knora can search for markup structures
```
tobiasschweizer: There is http://exist-db.org/exist/apps/homepage/index.html 

However, I do not know how they handle links to entities defined in other places, like a link to a person defined in a global index.

docs/src/paradox/faq.md
benjamingeer: “do not matter”: that’s why I didn’t think it was important to split hairs here by saying “a document that is semantically equivalent to the original document” or something like that.

docs/src/paradox/01-introduction/standoff-rdf.md
benjamingeer: I said “there is no standard system”. ExistDB is not a standard.

benjamingeer:
TODO: add more docs about shared ontologies.

PR #1097, opened 2018-12-05
build (docs): update ruby dependencies

subotic:
- update jekyll version
- Should resolve the Github Alert


benjamingeer:
With current `develop`, I get this error when doing `docs/makeSite` in sbt:

```
[error] /Library/Ruby/Gems/2.3.0/gems/bundler-1.17.1/lib/bundler/runtime.rb:319:in `check_for_activated_spec!': You have already activated public_suffix 3.0.3, but your Gemfile requires public_suffix 3.0.2. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```

subotic:
You probably need to update your gem files. You can try the command from section 6: https://github.com/dhlab-basel/Knora/blob/develop/docs/Readme.md

Unfortunately publishing the doc requires that ruby and all necessary gems are correctly installed on the local machine. I hate it, since it is always broken and requires some poking around. I should dockerize it.

PR #1118, opened 2018-12-07
Bulk import: check for correct handling of entities

### LINKED ISSUE #1105: Bulk import: transform entities before saving them as a simple string
If a simple string without markup is submitted using the bulk import, XML standard entities should not be saved directly in the triplestore. 

`&amp;` should be saved as `&`, the same goes for `<`, `>` etc. https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Predefined_entities_in_XML

tobiasschweizer:
This PR adds a test to check that entities in XML are handled correctly. 
I think the bulk import works correctly.

closes #1105


tobiasschweizer:
> Let's merge this, it can't hurt to add another test.

Thx. I will merge this once the tests have passed.

PR #1131, opened 2018-12-18
Update resource metadata in API v2

### LINKED ISSUE #1015: Update permission of an existing resource
In recent v2 create resource API, I've seen that we can set the permission on the fly 👍 , but can we hope to update it later ?

This could be **particularly** useful for us in the following (common) use case in our app:
```
Bibliography -- hasNote --> Note
```
Any project member can create and add a Note to a Bibliography, then during the lifetime of the Note, the user can modify the access of this Note:
- visible/editable only by project members
- visible/editable only by users of a group list
- visible to any users

From permissions Knora point of view, it is just playing with the permission property of a Resource, isn't it ?

thanks.

### LINKED ISSUE #1088: Update internal modification date of a resource
If it is possible to change the modification date of a resource, I'm interested to have an API entry point to manage a common use case in our app due to what we call a _composite resource_, for example :
```
Person -- hasNote -- Note
```
Each time a User modify a Note I would like to update the modification date of the related `Person`(s). I have of course many other ways to implement this function, the fatest way being obviously to manage myself a lastModification date property in `Person`.

Thanks.

benjamingeer:
This adds an API v2 route for updating the following resource metadata:

- [x] `rdfs:label`
- [x] permissions
- [x] last modification date

Also:

- [x] Add unit tests.
- [x] Add E2E tests.
- [x] Update docs.
- [x] Update release notes.

Resolves #1088.
Resolves #1015.

### DIFF
--- a/docs/src/paradox/03-apis/api-v2/editing-resources.md
+++ b/docs/src/paradox/03-apis/api-v2/editing-resources.md
@@ -201,7 +201,8 @@ The submitted JSON-LD object must also contain one or more of the following pred
 
 - `rdfs:label`: a string
 - `knora-api:hasPermissions`, in the format described in @ref:[Permissions](../../02-knora-ontologies/knora-base.md#permissions)
-- `knora-api:newModificationDate`: an `xsd:dateTimeStamp`
+- `knora-api:newModificationDate`: an [xsd:dateTimeStamp](https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp).
+  This is used to make sure that the resource has not been modified by someone else since you last read it.
 
 Here is an example:
 
--- a/docs/src/paradox/03-apis/api-v2/editing-resources.md
+++ b/docs/src/paradox/03-apis/api-v2/editing-resources.md
@@ -173,6 +173,66 @@ The response is a JSON-LD document containing a
 @ref:[preview](reading-and-searching-resources.md#get-the-preview-of-a-resource-by-its-iri)
 of the resource.
 
-## Modifying a Resource
+## Modifying a Resource's Values
 
-To modify a resource, you can modify its values; see @ref:[Editing Values](editing-values.md).
+See @ref:[Editing Values](editing-values.md).
+
+## Modifying a Resource's Metadata
+
+You can modify the following metadata attached to a resource:
+
+- label
+- permissions
+- last modification date
+
+To do this, use this route:
+
+```
+HTTP PUT to http://host/v2/resources
+```
+
+The request body is a JSON-LD object containing the following information about the resource:
+
+- `@id`: the resource's IRI
+- `@type`: the resource's class IRI
+- `knora-api:lastModificationDate`: an `xsd:dateTimeStamp` representing the last modification date that is currently attached to the resource, if any
+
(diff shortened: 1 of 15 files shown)


benjamingeer:
@gfoo Could you try this branch and see if it does what you want?

benjamingeer:
@gfoo I was hoping you would say that. 🙂

gfoo:
Just took a look at the doc:
- `knora-api:hasPermissions`: looks fine for me, I will use it from my code next week
- `knora-api:newModificationDate`: fine too, I will use it in the next few weeks, maybe add a link about the accepted string format of `xsd:dateTimeStamp`? (https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp)
- `knora-api:lastModificationDate`: what is the purpose of this date, I don't understand why I have to provide it?

benjamingeer:
> knora-api:lastModificationDate: what is the purpose of this date, I don't understand why I have to provide it?

You have to provide it to make sure that the resource has not been modified by someone else since you last read it. Otherwise, you would be able to overwrite someone else's changes without knowing that you are doing so.

benjamingeer:
Will clarify docs:

* Date format
* Reason for last mod date

gfoo:
In the json resource response, you provide a _very high precision_ date value: `2019-01-07T13:53:13.255630Z`

And when I try to use it in javascript I lost this precision:
```
new Date('2019-01-07T13:53:13.255630Z').toISOString()
"2019-01-07T13:53:13.255Z"
```
Following my previous link of the `xsd:dateTimeStamp` , I think that the date format should follow the [D ISO 8601 Date and Time Formats](https://www.w3.org/TR/xmlschema-2/#formatdetails), that is:

> s -- represents a digit used in the time element "second". The two digits in a ss format can have values from 0 to 60. In the formats described in this specification the whole number of seconds ·may· be followed by decimal seconds to an arbitrary level of precision. This is represented in the picture by "ss.sss". A value of 60 or more is allowed only in the case of leap seconds.

As explained, you can provide the level of precision you want, but is it really necessary? it maybe implies some probs for frontenders :)

n.b.:  I can of course keep in my objects the string instead of a js Date, no prob for me.

Nice to know (little bit old): https://nbsoftsolutions.com/blog/iso-8601-and-nanosecond-precision-across-languages

docs/src/paradox/03-apis/api-v2/editing-resources.md
```diff
@@ -173,6 +173,67 @@ The response is a JSON-LD document containing a
...
+You can modify the following metadata attached to a resource:
+
+- label
+- permissions
+- last modification date
+
+To do this, use this route:
+
+```
+HTTP PUT to http://host/v2/resources
```
subotic: Shouldn't this be `http://host/v2/resources/[ID]`? I thought our convention was to have the ID in the URL for something that we wanted to change?

benjamingeer: With JSON-LD, the IRI of the thing you want to change is in the body of the request, so it would be redundant to also put it in the URL. For example, this is how you update a value:

https://docs.knora.org/paradox/03-apis/api-v2/editing-values.html#updating-a-value

PR #1133, opened 2018-12-20
build(knora): fix docker build

subotic:
resolves #1133


subotic:
All done on my iPhone. I love Github 😀

PR #1146, opened 2019-01-08
refactor (webapi): non actor responders

### LINKED ISSUE #1139: Make responders not be actors
Since there's just one instance of each responder, and they do all their work in futures, they don't need to be actors.

A theoretically easy way to do this:

* Make `Responder` not extend `Actor`.
* Make `ResponderManager` a singleton, not an actor, containing one instance of each responder.
* Give `ResponderManager` a `?` method that takes a Knora request message as argument, and has the same return type as Akka's `?`.
* Have that `?` method call `receive` directly on the relevant responder instance and return the result.
* Write an alternative to `future2Message` that takes care of error-handling in this new context, and use it in responders instead of the existing `future2Message`.

subotic:
### Summary

- [x] change responders to scala classes
- [x] update documentation

### Issues
- resolves #1139


benjamingeer:
Cool, will look at this tomorrow.

subotic:
Thanks :-)

subotic:
Thanks for the review:-) I’m currently refactoring them in another branch.

PR #1159, opened 2019-01-14
feature (webapi): extend v2 auth

### LINKED ISSUE #1165: Add session creation to V2 authentication API
As a first step in implementing the IIIF Authentication API, the `/v2/authentication` should beside the token also create a cookie.

Part of #520

subotic:
### Summary

- [x] session cookie creation during V2 authentication
- [x] update docs
- [x] simple HTML login form http://localhost:3333/v2/login (needed for IIIF Authentication API support)
- [x] update release notes

### Issues

- resolves #1165


webapi/src/main/scala/org/knora/webapi/routing/Authenticator.scala
```diff
@@ -125,6 +126,49 @@ trait Authenticator {
 
     } yield httpResponse
 
+    def presentLoginFormV2(requestContext: RequestContext)(implicit system: ActorSystem, executionContext: ExecutionContext): Future[HttpResponse] = {
```
benjamingeer: Does the HTML login form need to be generated by the server, or could it be generated by Knora-ui in the browser?

PR #1161, opened 2019-01-14
Generate ARK URLs for resources

### LINKED ISSUE #844: Define ARK IDs
# ARK existing server

here is an ARK ID:

[`http://`NMAH`/`]`ark:/`NAAN`/`Name[Qualifier]

where:

- NMAH, Name Mapping Authority Host, for us is: `ark.dasch.swiss`
- NAAN, Name Assigning Authority Number, for us is: `72163`

resulting in this ark-id for a resource in salsah v1:

`http://ark.dasch.swiss/ark:/72163/0801-31ec6eab334-c.201852`

which is translated into:

`http://data.dasch.swiss/resources/3491?citdate=201852`

In the ark-id, the string `0801-31ec6eab334-c.201852` means:
- `0801` is the project id, and that it is mapped to `http://data.dasch.swiss/resources/`
- the resource code `31ec6eab334` is mapped to `3491`
- and append the transformed `c.201852` to `?citdate=201852`

# Knora and ARK

So an ark-id could be:
`http://ark.dasch.swiss/ark:/72163/0105--AU-w879ReyDWm6wpwZnAw-c.201852`

Where:
- project `0105` would be mapped in the ark mapping server to `http://knora-demo.unil.ch/v1/resources/http://rdfh.ch/0105/drawings-gods/`
- `-AU-w879ReyDWm6wpwZnAw` is the resource IRI and would remain the same

So it would end up as the knora call:
- `http://knora-demo.unil.ch/v1/resources/http://rdfh.ch/0105/drawings-gods/-AU-w879ReyDWm6wpwZnAw`

# Project specific web app

Registering a project specific web app, we could set the project mapping to `http://ddd.unil.ch/permalink/` and the same `http://knora-demo.unil.ch/v1/resources/http://rdfh.ch/0105/drawings-gods/-AU-w879ReyDWm6wpwZnAw` would redirects to `http://ddd.unil.ch/permalink/-AU-w879ReyDWm6wpwZnAw`

benjamingeer:
- [x] In API v2, return an [ARK URL](https://tools.ietf.org/html/draft-kunze-ark-18) as part of the metadata of each resource (in `ReadResourceV2.toJsonLD`).
- [x] Add documentation:
  - [x] `docs/src/paradox/03-apis/api-v2/resource-permalinks.md`
  - [x] `docs/src/paradox/05-internals/design/api-v2/ark.md`
- [x] Change all resources in test data to use [Knora's standard resource IRI format](https://docs.knora.org/paradox/03-apis/api-v2/knora-iris.html#iris-for-data), so we can generate ARK URLs for them.

Resolves #844.


tobiasschweizer:
@benjamingeer I have data that still has  http://data.knora.org instead of http://rdfh.ch

loicjaouen:
we don't (at Lausanne)

benjamingeer:
@tobiasschweizer Then this will be a breaking change. You'll have to do a global search-and-replace in your data. You can dump the data from GraphDB to a Turtle file, do the search-and-replace in the file, then load it back into GraphDB.

benjamingeer:
@tobiasschweizer Do your resource IRIs have project shortcodes in them?

tobiasschweizer:
@benjamingeer The Iris would have to be updated in the XMLs, too. 

> Do your resource IRIs have project shortcodes in them?

I will have to check for that and get back to you.

tobiasschweizer:
@benjamingeer I have old style Iris that look like this http://data.knora.org/xyz and Iris that look like this http://rdfh.ch/beol/xyz

Could we check the search and replace thing on Monday (especially for the value Iris)?

benjamingeer:
@tobiasschweizer We can fix all your IRIs together next week, it's not difficult, I've just done this for the Incunabula test data.

tobiasschweizer:
@benjamingeer I also think that it won't be hard. I actually used emacs regex search and replace to adapt the permissions.

 I just have to do the same thing in the XMLs where the IRIs are used.

@subotic Maybe the internal beol server should be re-installed from scratch. This time, I would like to use docker. Could you give me a hand?

subotic:
> @subotic Maybe the internal beol server should be re-installed from scratch. This time, I would like to use docker. Could you give me a hand?

Sure, we can take the deployment configuration from production and change it for the internal beol server.

webapi/src/test/scala/org/knora/webapi/e2e/v2/SearchRouteV2R2RSpec.scala
```diff
@@ -70,7 +70,7 @@ class SearchRouteV2R2RSpec extends R2RSpec {
     private val password = "test"
 
     // If true, writes all API responses to test data files. If false, compares the API responses to the existing test data files.
-    private val writeTestDataFiles = false
+    private val writeTestDataFiles = true
```
subotic: Maybe this should be set back to `false`?

subotic: Maybe this should be set back to `false`?

benjamingeer:
> Can we please move ark.py to its own repository? 

Yes, good idea.

subotic:
I've taken the liberty and moved it into its own repository :-)

https://github.com/dhlab-basel/ark-resolver/

benjamingeer:
Great, thank you!

subotic:
@benjamingeer Any particular reason why the resolver is written in python and not in scala? You know that you are now the sole responsible person for maintaining the code  ;-)

benjamingeer:
> Any particular reason why the resolver is written in python and not in scala?

Two reasons:

1. It's such a simple program, I think it doesn't need a JVM consuming gigabytes of memory.
2. I think some people might want to write their own ARK resolvers for use with Knora, they'll want to look at this one to see how it works, and it might be easier for them to read code in Python than in Scala.

> you are now the sole responsible person for maintaining the code ;-)

That's OK, I can live with maintaining this, because it hardly does anything. :)

ezbinden:
Hi Ben, Ivan
Could this be a breaking change for some of our partners? If yes they should be notified via the DaSCH team list
Cheers, E.

PR #1170, opened 2019-01-17
fix (webapi): add domain to cookie

### LINKED ISSUE #1169: Add 'domain' to HTTP cookie
`Domain=example.com` will be sent to `*.example.com`

> Specifies those hosts to which the cookie will be sent. If not specified, defaults to the host portion of the current document location (but not including subdomains). Contrary to earlier specifications, leading dots in domain names are ignored. If a domain is specified, subdomains are always included.

subotic:
### Summary
- [x] add domain to cookie
- [x] update docs
- [x] update release notes
- [x] update test data from `localhost` to `0.0.0.0`

### Issues
- resolves #1169


subotic:
The config needs to be set to `0.0.0.0` or there is an error with involving project ontologies (some tests fail).

For local testing, you should always use `localhost`.

subotic:
Ok, it basically boils down to this.

When running locally, If you want to use the authentication cookie with `sipi`, which happens when using `Salsah1` and probably in the future also with `Salsah2` (as `v2/authentication` now also uses cookies), then both `webapi` and `sipi` need to be accessed over `0.0.0.0`.

If no cookies are involved, or only `webapi` is accessed without `sipi`, then both `0.0.0.0` and `localhost` should work.

benjamingeer:
@subotic OK could you add something about this to the documentation? The choice between `0.0.0.0` and `localhost` seems to be a FAQ, so it would be good to write something down about it (maybe in `faq.md`).

loicjaouen:
isn't it outside of knora's scope?

- `localhost` (a.k.a `127.0.0.1`) is the logical host's network: services running on your host will see each others, nobody else will
- `0.0.0.0` is the hardware's network of the host: services outside of the host but connected to your host through can talk to you

So if you have different network interfaces attached to you host, instead of `0.0.0.0` you might want to specify one or another, like a `86.46.x.x` for internet access and `10.0.x.x` for access within the organization (to cut access from a public DoS attack but still allow *inside* access).

It might also depends if you use docker or not, and if you use docker on linux or macos (with a virtual machine)?

subotic:
@loicjaouen  `localhost` or `0.0.0.0` in this context is only relevant when running the stack locally. Any other deployment should use resolvable domain names. It is only relevant if you want authorization cookies to be shared between `webapi` and `sipi` and you are using the default configuration. In this case, `0.0.0.0` must be used, or otherwise, the cookie will not be sent to `sipi`.

I will add a note to `faq.md`

PR #1178, opened 2019-01-20
Gravsearch query with an optional sort criterion

### LINKED ISSUE #1010: Error when combining optional property and orderby
Hi,
found an error with gravsearch when I try to order the result by using an optional property.
I think we could expect to get all the resources even all that do not have a date (at the end of the last page of course and not sorted) ? 

```sparql
PREFIX onto: <http://0.0.0.0:3333/ontology/0001/anything/simple/v2#>

CONSTRUCT {
?thing onto:hasDate ?date .
} WHERE {
  
  ?thing a knora-api:Resource .
  ?thing a onto:Thing .
  
OPTIONAL {

?thing onto:hasDate ?date .
onto:hasDate knora-api:objectType knora-api:Date .
?date a knora-api:Date .

}

}
ORDER BY DESC(?date)
```
Error response:
`{
    "error": "org.knora.webapi.InconsistentTriplestoreDataException: entity http://rdfh.ch/0113/CQinAnP5TxmAY_2tEVacCA has no creator"
}`

tobiasschweizer:
This PR fixes a bug when using an optional value as a sort criterion.

closes #1010


tobiasschweizer:
@gfoo This PR should fix the problem you reported in #1010. Could you check out this branch and try? Any more test scenarios needed?

gfoo:
looks good to me.

tobiasschweizer:
Ok, I will add some additional tests and then this PR can be reviewed

benjamingeer:
> I noticed that there are two variables created for the same literal: one for the sorting and one for the filter. How could this be unified?

We talked about having a mutable set of all statements that have been added, so we don't add the same statement twice.

Can you look at why we aren't using the same variable name in both cases? Perhaps the same function could be used.

tobiasschweizer:
`addGeneratedVariableForValueLiteral` returns a Boolean that indicates whether the variable had to be added or already was part of the collection. A statement is only added if the variable was **not** part of the collection. This works fine for the simple schema. In the additional tests that I created, I checked that the same statement was not created several times for order by and filters. 

In the complex schema, however, a statement has to be created for the sort criterion:

> If the query is written in the complex schema, items below the level of Knora values may not be used in ORDER BY.

In the example above, the user submitted a statement `?decimal knora-api:decimalValueAsDecimal ?decimalVal ` and `?decimalVal` is used in filter. 

So maybe for the case of the complex schema when handling the sort criterion, there should be a check whether such a statement already exists. Maybe this can be done when handling the statements of the where clause. If yes, we should use that value literal var in order by instead of adding a new one. 

I admit that I was not aware of the way the complex schema is handled since I worked on the simple schema. For dates, everything stays the same because the special date function is used.

tobiasschweizer:
I suggest that we solve the problem in a separate issue (#1195)

tobiasschweizer:
@benjamingeer I will add a few tests for the count query and then this PR can be reviewed.

tobiasschweizer:
@benjamingeer Thanks, but actually I am not quite done yet :-)

PR #1191, opened 2019-01-28
Use project shortcode in IIIF URLs

### LINKED ISSUE #874: Use project shortcode instead of knora as the Sipi IIIF prefix
Currently, `knora` is used as the prefix, so that under the `images` folder a `knora` folder holds all images. The goal is to have a folder per project, and for this, we can use the project shortcode.

I suppose the SipiResponder will need to be changed and the lua scripts.

The corresponding issue in Sipi: https://github.com/dhlab-basel/Sipi/issues/233

benjamingeer:
- [x] Change API v1 and v2 to produce IIIF URLs containing the project shortcode instead of `knora`.
- [x] Change API v1 and v2 to send the project shortcode to Sipi when requesting file storage.
- [x] Change Knora's Sipi scripts to use the prefix as the name of a storage directory.
- [x] Add functionality to `SmartIri` to get project shortcodes from resource and value IRIs.
- [x] In routes where we're creating a resource, we have to get the project shortcode from the projects responder. Here it would help to have #1188.
- [x] Change old `/reps/` to `/values/` in value IRIs in test data, so we can parse value IRIs to get their project shortcodes.
- [x] Have the Sipi-to-Knora route include the IIIF prefix, which is the project shortcode (not used yet).
- [x] Move the Sipi responder to admin.

If you still have file value IRIs like `http://rdfh.ch/PROJECT_SHORTCODE/RESOURCE_UUID/reps/VALUE_UUID` (which was the format before #649), you will have to change `/reps/` to `/values/`.

Resolves #874.


subotic:
Hmm, no idea, sorry. But maybe if you would change this function, to pass the error message back to `webapi`, then you could maybe have a better picture of what is going wrong here:

https://github.com/dhlab-basel/Knora/blob/020a95ae64e0750eb323079038ddffb6c9dcea5b/sipi/scripts/send_response.lua#L37-L78

subotic:
The error is probably happening in `upload.lua` but the error message is never sent.

PR #1208, opened 2019-02-07
feature (webapi): secure routes returning user information

### LINKED ISSUE #961: Return users depending on the requesters permission

### LINKED ISSUE #1209: Reorganize user and project routes

subotic:
### Summary
- [x] reorganize users route
- [x] reorganize projects route
- [x] secure routes returning user information
- [x] add input validation and escaping for project IRI, shortname and shortcode
- [x] add input validation and escaping for user IRI, username and email

### Issues
- resolves #961 
- resolves #1209


webapi/src/main/scala/org/knora/webapi/messages/admin/responder/usersmessages/UsersMessagesADM.scala
```diff
@@ -555,6 +555,42 @@ case class UserADM(id: IRI,
         }
     }
 
+    /**
+      * Given an identifier, returns true if it is the same user, and false if not.
+      */
+    def isSelf(identifier: UserIdentifierADM): Boolean = {
+
+        val iriEquals = if (identifier.toIriOption.isDefined) {
```
benjamingeer: I was going to say that there's a shorter way to write this method:

```scala
val iriEquals = identifier.toIriOption.contains(id)
val emailEquals = identifier.toEmailOption.contains(email)
val usernameEquals = identifier.toUsernameOption.contains(id)

iriEquals || emailEquals || usernameEquals
```

But after thinking about it some more, I think there's a small design problem. Suppose you have two instances of `UserIdentifierADM` that represent the same user, but one instance contains the email address and the other instance contains the username. Then `isSelf` will return `false`, though it should return `true`. So I think the `isSelf` method belongs somewhere else, not in this class, so it can determine whether two `UserIdentifierADM` instances really represent the same user.

Also, I think it's not ideal that `value` is just a string that has to be parsed every time you use it (in `hasType`, `toIri`, and `toIriOption`). I think it would be simpler, safer, and more efficient to use types.

```scala
sealed trait UserIdentifierADM
case class UserIriADM(iri: IRI) extends UserIdentifierADM
case class UserEmailADM(email: String) extends UserIdentifierADM
case class UsernameADM(username: String) extends UserIdentifierADM
```

Looking at the places where a `UserIdentifierADM` is constructed, it looks like you almost always know what type of identifier you have, so you don't need `UserIdentifierADM` to figure it out for you. The exceptions that I see are in `UsersRouteADM` and `AuthenticationRouteV2`, where you get an identifier from the client, but you don't know what type it is. Here I would really suggest changing the route so that the client has to specify the type of the identifier. I think it doesn't seem very safe to rely on guessing the type of the identifier from its format.

subotic: The idea behind `UserIdentifierADM` was exactly to allow the user to call a route and supply either an IRI, username or email, without the need to provide further information, as each uniquely identifies the user.

subotic: > But after thinking about it some more, I think there's a small design problem. Suppose you have two instances of `UserIdentifierADM` that represent the same user, but one instance contains the email address and the other instance contains the username. Then `isSelf` will return `false`, though it should return `true`.

Why would it return `false`? If in the first case the `email` matches, then the result will be `true`. If in the second case the `username` matches, then also the result will be `true`.

benjamingeer: > The idea behind `UserIdentifierADM` was exactly to allow the user to call a route and supply either an IRI, username or email, without the need to provide further information, as each uniquely identifies the user.

What if my username is `foo@example.org`?

And there could be more types of identifiers in the future, creating more ambiguity.

> Why would it return `false`?

Sorry, I got confused, I thought the `isSelf` method was in `UserIdentifierADM`. Now I see it's in `UserADM`. But, it does seem to have a bug. I added a test to show this.

I think that if you used different classes for different types of identifiers, this kind of bug would be less likely.

subotic: > But, it does seem to have a bug. I added a test to show this.

I've added already one in the commit before ;-)

benjamingeer: > I've added already one in the commit before ;-)

But my test still fails.

subotic: > And there could be more types of identifiers in the future, creating more ambiguity.

hmm, then I should change the routes to something like this:

- `/admin/users/iri/<userIri>`
- `/admin/users/email/<email>`
- `/admin/users/username/<username>`
- `/admin/users/isni/<isni>`

and also the same structure for projects. There we have the same problem.

This will certainly make my life easier but will make it a bit harder for the consumers of the API.

benjamingeer: > hmm, then I should change the routes to something like this:

That seems OK to me. You could also do it with one route:

- `/admin/users?email=user%40example.org`
- `/admin/users?username=foo`

For logging in, you get a `POST` request, so the client can specify `email` , `username`, etc., instead of `identifier`.

benjamingeer: I think it's usually not a bad thing to make the client say explicitly what they want. It forces them to think about it more, which could prevent client bugs.

subotic:
@benjamingeer If you could take another look. I've made the routes in `/admin/users` and `/admin/projects` explicit in regard to the `identifier`.

benjamingeer:
Great, will look tomorrow morning.

webapi/src/main/scala/org/knora/webapi/messages/v2/routing/authenticationmessages/AuthenticationMessagesV2.scala
```diff
@@ -33,11 +33,25 @@ import spray.json._
   * @param identifier   the user's IRI, username, or email.
   * @param password     the user's password.
   */
-case class LoginApiRequestPayloadV2(identifier: String,
-                                    password: String) {
+case class LoginApiRequestPayloadV2(iri: Option[IRI] = None,
```
benjamingeer: All these inputs (IRI, email, and username) are inserted into SPARQL queries, right? If so, they should be escaped using `StringFormatter.toSparqlEncodedString()` to prevent SPARQL injection attacks.

webapi/src/main/scala/org/knora/webapi/routing/admin/UsersRouteADM.scala
```diff
@@ -121,108 +129,71 @@ class UsersRouteADM(routeData: KnoraRouteData) extends KnoraRoute(routeData) wit
...
+
+    /**
+      *  return a single user identified by username
+      */
+    private def getUserByUsername: Route = path("admin" / "users" / "username" / Segment) { value =>
+        get {
+            requestContext =>
+                val requestMessage: Future[UserGetRequestADM] = for {
+                    requestingUser <- getUserADM(requestContext)
+                } yield UserGetRequestADM(UserIdentifierADM(username = Some(value)), UserInformationTypeADM.RESTRICTED, requestingUser)
```
benjamingeer: The username needs to be escaped with `StringFormatter.toSparqlEncodedString()`.

webapi/src/main/scala/org/knora/webapi/routing/Authenticator.scala
```diff
@@ -495,13 +496,29 @@ object Authenticator {
...
         val params: Map[String, Seq[String]] = requestContext.request.uri.query().toMultiMap
 
-        // check for email (old name) or identifier (new name) parameters
-        val identifierList: Iterable[String] = params.get("email").map(_.head) ++ params.get("identifier").map(_.head)
-        val maybeIdentifier: Option[String] = identifierList.headOption
-        val maybePassword: Option[String] = params get "password" map (_.head)
+        log.debug("extractCredentialsFromParametersV2 - params: {}", params)
+
+        // check for iri, email, or username parameters
+        val maybeIriIdentifier: Option[String] = params.get("iri").map(_.head)
```
benjamingeer: Validate/escape IRI, email, and username to prevent SPARQL injection

webapi/src/main/scala/org/knora/webapi/messages/v2/routing/authenticationmessages/AuthenticationMessagesV2.scala
subotic: ok, so the general workflow would be:

1. validate
2. StringFormatter.toSparqlEncodedString()

Correct?

subotic:
Thanks, for the review. Yes, I agree that it would be easier to centralize all the checking. It was something that I wanted to "someday" but maybe better do do it now.

webapi/src/main/scala/org/knora/webapi/messages/v2/routing/authenticationmessages/AuthenticationMessagesV2.scala
benjamingeer: * For IRIs: `StringFormatter.validateAndEscapeIRI()`
* For email addresses: `StringFormatter.validateEmail()`, then `StringFormatter.toSparqlEncodedString()` (you could make a `validateAndEscapeEmail` function to do both)
* For usernames: Are there any restrictions on what characters can be in a username? Maybe there should be. Whatever restrictions you choose, validate, then `StringFormatter.toSparqlEncodedString()` (you could make a `validateAndEscapeUsername` function to do both)

benjamingeer: Also maybe `StringFormatter` should have a maximum allowed length for IRIs, email addresses, and usernames.

subotic:
What happens if I run `StringFormatter.toSparqlEncodedString()` twice?

benjamingeer:
> What happens if I run `StringFormatter.toSparqlEncodedString()` twice?

Then the escapes get escaped, so don't do that. :)

benjamingeer:
```
sbt:knora> webapi / console
[info] Starting scala interpreter...
Welcome to Scala 2.12.4 (OpenJDK 64-Bit Server VM, Java 10.0.2-adoptopenjdk).
Type in expressions for evaluation. Or try :help.

scala> import org.knora.webapi.util._
import org.knora.webapi.util._

scala> StringFormatter.initForTest()

scala> val stringFormatter = StringFormatter.getGeneralInstance
stringFormatter: org.knora.webapi.util.StringFormatter = org.knora.webapi.util.StringFormatter@5226c11e

scala> val r1 = stringFormatter.toSparqlEncodedString("""\n""", throw new Exception())
r1: String = \\n

scala> val r2 = stringFormatter.toSparqlEncodedString(r1, throw new Exception())
r2: String = \\\\n
```

subotic:
>> What happens if I run `StringFormatter.toSparqlEncodedString()` twice?

> Then the escapes get escaped, so don't do that. :)

Ok, thanks. I thought as much.

subotic:
I think that now every input except passwords is validated and escaped. I've opened an issue #1213 for passwords.

docs/src/paradox/03-apis/api-admin/users.md
```diff
@@ -19,153 +19,187 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
 
 # Users Endpoint
 
-  - **Get users**
+## Endpoint Overview
+
+**User Operations:**
+- `GET: /admin/users` : return all users
+- `GET: /admin/users/[iri | email | username]/<userIri>` : return single user identified by [IRI | email | username]
```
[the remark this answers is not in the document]
subotic: changed

docs/src/paradox/03-apis/api-admin/users.md
```diff
@@ -19,153 +19,187 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
...
+
+### Get users
+
+  - Required permission: SystemAdmin
+  - GET: `/admin/users`
+
+### Get user
+
+  - Required permission: SystemAdmin / self
+  - GET:`/admin/users/[iri | email | username ]/<userIri>`
```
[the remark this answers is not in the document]
subotic: changed

webapi/src/test/scala/org/knora/webapi/responders/admin/UsersResponderADMSpec.scala
```diff
@@ -28,6 +28,7 @@ import java.util.UUID
 import akka.actor.{ActorRef, ActorSystem}
 import akka.actor.Status.Failure
 import akka.testkit.ImplicitSender
+import arq.iri
```
benjamingeer: All these imports of `arq.iri` seem to be unused, did you add them by mistake?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/usersmessages/UsersMessagesADM.scala
```diff
@@ -697,39 +698,55 @@ object UserIdentifierType extends Enumeration {
...
+        new UserIdentifierADM(
+            maybeIri = sf.validateAndEscapeOptionalUserIri(maybeIri, throw BadRequestException(s"Invalid user IRI $maybeIri")),
+            maybeEmail = sf.validateAndEscapeOptionalEmail(maybeEmail, throw BadRequestException(s"Invalid email $maybeEmail")),
+            maybeUsername = sf.validateAndEscapeOptionalUsername(maybeUsername, throw BadRequestException(s"Invalid username $maybeUsername")))
+    }
+}
+
+/**
+  * Represents the user's identifier. It can be an IRI, email, or username.
+  * @param value the user's identifier.
```
benjamingeer: Could you update the parameters in the Scaladoc comment?

webapi/src/main/scala/org/knora/webapi/util/StringFormatter.scala
```diff
@@ -1544,7 +1562,7 @@ class StringFormatter private(val maybeSettings: Option[SettingsImpl], initForTe
       * @param iri the IRI to be checked.
       */
     def isKnoraDataIriStr(iri: IRI): Boolean = {
-        DataIriStarts.exists(startStr => iri.startsWith(startStr))
+        isIri(iri) && DataIriStarts.exists(startStr => iri.startsWith(startStr))
```
benjamingeer: The method `isKnoraDataIriStr` is called in the constructor of `SmartIriImpl`, where `validateAndEscapeIri` has already been called, so we already know it's a valid IRI. Could you refactor this to avoid the redundant validation?

webapi/src/test/scala/org/knora/webapi/responders/admin/UsersResponderADMSpec.scala
subotic: it's IntelliJ. sometimes when I set `IRI` as the type, IntelliJ adds this import. I will remove them.

subotic: removed

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/usersmessages/UsersMessagesADM.scala
subotic: done

webapi/src/main/scala/org/knora/webapi/util/StringFormatter.scala
subotic: done

subotic:
Thanks for the review :-)

Issue #1115, opened 2018-12-07
Implement version history in API v2


benjamingeer:
Wish list:

- Given a resource IRI and a timestamp, retrieve the resource as it was at that time (optionally, specifying a property IRI and getting the resource's values for just that property).
- Given a resource IRI, a start date, and an end date, retrieve a list of all the modifications that were done in a resource between those dates (no need to return the changes themselves, just when they were done and by whom).

benjamingeer:
Ontotext replied:

> Unfortunately, SPARQL supports only dateTime comparisons and GraphDB does not offer an extension on it.

So before we can implement querying version history, we'll have to change all `xsd:dateTimeStamp` to `xsd:dateTime`.

benjamingeer:
This is how you get the values that were current at a given time in the past:

```sparql
?resource knora-base:hasValue ?currentValue .                
?currentValue knora-base:previousValue* ?valueObject .
?valueObject knora-base:valueCreationDate ?valueObjectCreationDate .
FILTER(?valueObjectCreationDate < "2019-02-12T08:05:10Z"^^xsd:dateTime)
                
FILTER NOT EXISTS {
  ?currentValue knora-base:previousValue* ?othervalueObject .
  ?othervalueObject knora-base:valueCreationDate ?othervalueObjectCreationDate .
  FILTER((?othervalueObjectCreationDate < "2019-02-12T08:05:10Z"^^xsd:dateTime) && (?othervalueObjectCreationDate > ?valueObjectCreationDate))
}
```

PR #1211, opened 2019-02-11
Change all xsd:dateTimeStamp to xsd:dateTime in the triplestore

benjamingeer:
To implement version history (#1115), we need to be able to compare timestamps in the triplestore. So far we have used [xsd:dateTimeStamp](https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp), but it turns out that SPARQL supports comparisons with `xsd:dateTime` but not with `xsd:dateTimeStamp`. So this PR changes `xsd:dateTimeStamp` to `xsd:dateTime` in:

- `knora-base`
- Test data
- Twirl templates that generate SPARQL

The only difference between the two types is that in `xsd:dateTimeStamp`, the timezone is required. All the timestamps generated by Knora specify the UTC timezone (`Z`) in any case.

We can still use `xsd:dateTimeStamp` in the API.

Users will have to change their own data, but this is very easy to do (just dump to Turtle and do a search-and-replace).


tobiasschweizer:
@benjamingeer In my data, I just need to change `xsd:dateTimeStamp` to `xsd:dateTime`?

Example from my data: `"2017-11-29T12:01:47.467Z"^^xsd:dateTimeStamp .`

I assume that the format of the timestamp (YYYY-MM-DDT...Z) itself was never changed since mid 2016.

benjamingeer:
@tobiasschweizer That's right.

benjamingeer:
@subotic Whenever you get a chance. :)

subotic:
sorry, didn’t see the review request. will do it tomorrow.

PR #1214, opened 2019-02-13
Query version history

### LINKED ISSUE #1079: Concurrent updates in API v2
Because `ValuesResponderV2` uses Gravsearch to verify updates, it doesn't handle this very unlikely situation described in the Knora design docs under **Triplestore Updates**:

> Users A and B are looking at version 1. User A submits an update and
> it succeeds, creating version 2. Before User A has time to do a
> SELECT, user B reads the new value and updates it again. Both users
> then do a SELECT, and find that both their new value IRIs are
> present in the value's version history.

What will actually happen in API v2 is that user A will be told that their update failed.

### LINKED ISSUE #1115: Implement version history in API v2

### LINKED ISSUE #1215: Record who marked a resource or value as deleted
Currently, if a resource or value is marked as deleted, we don't have a record of who did it. This should be stored along with `deleteComment` and `deleteDate`.

benjamingeer:
- [x] `getResourcePropertiesAndValuesGraphDB.scala.txt`: Given a resource IRI and a timestamp, retrieve the resource as it was at that time.
- [x] `ResourcesRouteV2`: in the route `/v2/resources/IRI`, accept the query parameter `version` (an `xsd:dateTimeStamp`, or a Knora ARK URL timestamp).
- [x] `getResourcePropertiesAndValuesGraphDB.scala.txt`: optionally specify a property IRI, and get the resource's values for just that property at the specified time. Needed only internally, for #1079.
- [x] `ValuesResponderV2.verifyValue()`: Use version history to verify value updates (#1079).
- [x] `ReadResourceV2.toJsonLD()`: When returning a resource, include a `knora-api:versionArkUrl` containing a timestamp, referring to the version that's being returned.
- [x] `StringFormatter`: Reimplement the formatting and parsing of ARK URL timestamps to make them as short as possible, by removing trailing zeros in fractional seconds (depends on https://github.com/dhlab-basel/ark-resolver/commit/b004034b5915e19a91973c6f158a789e575d44a6).
- [x] Record the IRI of the user who marked a resource or value as deleted (#1215), so we can include that information when querying the version history of a resource.
- [x] `getResourceVersionHistoryGraphDB.scala.txt`: Given a resource IRI, a start date, and an end date, retrieve a list of all the changes that were made in the resource's values between those dates (no need to return the changes themselves, just when they were made and by whom). Include the creation of the resource, even if it was created with no values.
- [x] `ResourcesRouteV2`: add the route `/v2/resources/history/IRI`, to get the version history of the resource's values.
- [x] When returning resource values, include their `knora-api:valueCreationDate`.
- [x] `header.xsl`: Adapt the XSLT transformation for generating TEI/XML headers from BEOL letters to handle the fact that, in RDF/XML, target resources may or may not be nested in link values.

A lot of the changes in this PR are in test data, to include `knora-api:versionArkUrl` and `knora-api:valueCreationDate`.

Resolves #1115.
Resolves #1215.
Fixes #1079.


benjamingeer:
TODO in getting a past version:

- Should we be able to get a past version of a resource that has been marked as deleted?
- What if a past version of a link points to a resource that was later deleted? Should we return the (now deleted) target resource nested in the link value?

Need to think about use cases for dealing with deleted items.

TODO in `/v2/resources/history`:

- Get the resource preview to check its permissions and get its creator and creation date.
- Include the creation of the resource in the history.

benjamingeer:
@kilchenmann This PR adds `knora-api:versionArkUrl` to each resource, and @tobiasschweizer says this means that Knora-ui code and test data will need to be adapted. Tobias says he can help if needed.

benjamingeer:
@subotic Tobias and I need a little more time to fix a broken test here, no need to review yet.

benjamingeer:
@tobiasschweizer I implemented a function in `header.xsl` like you suggested, and it works, thanks! I also added a test for it in `XMLUtilSpec`.

benjamingeer:
@subotic I think this is finally ready for review now.

benjamingeer:
@subotic Could you possibly find time to review this PR this week?

benjamingeer:
No worries, thanks!

subotic:
Yes, sorry. Will do it today.

benjamingeer:
@subotic The Travis build is failing because it can't find the command `typedoc` to build the documentation:

```
Makefile:14: *** The 'typedoc' command was not found. Make sure you have Typedoc installed..  Stop.
```

kilchenmann:
We will update the knora-ui modules to use it with this version of knora.
[PR #185](https://github.com/dhlab-basel/Knora-ui/pull/185)

webapi/src/main/scala/org/knora/webapi/routing/v2/ResourcesRouteV2.scala
```diff
@@ -176,16 +177,43 @@ class ResourcesRouteV2(routeData: KnoraRouteData) extends KnoraRoute(routeData)
...
+                            requestContext = requestContext,
+                            settings = settings,
+                            responderManager = responderManager,
+                            log = log,
+                            responseSchema = ApiV2WithValueObjects
                         )
                     }
                 }
             }
+        } ~ path("v2" / "resources" / "history" / Segment) { resourceIriStr: IRI =>
```
subotic: I would personally prefer the following:

```
/v2/resources/[resourceIRI]/history
```

webapi/src/main/scala/org/knora/webapi/routing/v2/ResourcesRouteV2.scala
```diff
@@ -176,16 +177,43 @@ class ResourcesRouteV2(routeData: KnoraRouteData) extends KnoraRoute(routeData)
...
+                        requestMessageF = requestMessageFuture,
+                        requestContext = requestContext,
+                        settings = settings,
+                        responderManager = responderManager,
+                        log = log,
+                        responseSchema = ApiV2WithValueObjects
+                    )
+                }
+            }
         } ~ path("v2" / "resources" / Segments) { resIris: Seq[String] =>
```
subotic: Here I would prefer:

```
/v2/resources/[resourceIRI]/version/[timestamp]
```

subotic: and for multiple IRIs:

```
/v2/resources/version/[timestamp]?Id=iri1,iri2,iri3
```

subotic: I think that the more information is in the URL path the better. I don't like URL parameters very much.

webapi/src/main/scala/org/knora/webapi/routing/v2/ResourcesRouteV2.scala
benjamingeer: This would conflict with the existing route:

```
/v2/resources/[resourceIRI]/[resourceIRI]
```

webapi/src/main/scala/org/knora/webapi/routing/v2/ResourcesRouteV2.scala
benjamingeer: I don't know whether changing this would break Knora-ui, BEOL, etc.

subotic:
we are not the only ones: https://github.com/TypeStrong/typedoc/issues/976

webapi/src/main/scala/org/knora/webapi/routing/v2/ResourcesRouteV2.scala
subotic: Definitely. The comments are only meant as "food for thought". We can make an issue and do it sometime when @kilchenmann has more time.

subotic:
we could comment out the documentation building test from travis for the time being.

benjamingeer:
> we are not the only ones

Oh, great 🤦🏻‍♂️

> we could comment out the documentation building test from travis for the time being.

OK.

PR #1224, opened 2019-02-21
Ignore order of attributes when comparing standoff

### LINKED ISSUE #1221: richtext property with a mapping update AssertionException
I created a standoff mapping to manage this particular html code in a richtext property:
```html
<text>
<span title="titre" data-corr="correction" class="sic">test</span>
</text>
``` 
I successfully imported those data with a bulk import but I get this error when I try to update it:
```json
{"error":"org.knora.webapi.AssertionException: The value saved as http://rdfh.ch/0113/H_rE2o3hSLm8qxevXlcDQA/values/kASk1JdIRrqrpBxg713MOA is not the same as the one that was submitted"}
```
I `PUT http://0.0.0.0:3333/v2/values` this json:
```json
{
    "@id": "http://rdfh.ch/0113/H_rE2o3hSLm8qxevXlcDQA",
    "@type": "onto:Transcription",
    "onto:transcriptionHasText": {
        "@id": "http://rdfh.ch/0113/H_rE2o3hSLm8qxevXlcDQA/values/G4d2-EKNSgOOTpBgOwToRw",
        "@type": "knora-api:TextValue",
        "knora-api:textValueAsXml": "<text><span title=\"titre\" data-corr=\"correction\" class=\"sic\">new value</span></text>",
        "knora-api:textValueHasMapping": {
            "@id": "http://rdfh.ch/projects/0113/mappings/manuscript-mapping"
        }
    },
    "@context": {
        "knora-api": "http://api.knora.org/ontology/knora-api/v2#",
        "onto": "http://0.0.0.0:3333/ontology/0113/lumieres-lausanne/v2#"
    }
}
```

Sounds like #1069

benjamingeer:
When verifying whether a value was saved correctly, we compare the saved value with the one that was supposed to be saved. Before we can compare standoff, we call `StandoffTagUtilV2.makeComparableStandoffCollection`, which normalises insignificant differences to make the standoff comparable. The bug in #1221 is that one of these insignificant differences is not taken into account: if the order of standoff tag attributes is different, the standoff incorrectly appears to be different.

Fixes #1221.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/util/standoff/StandoffTagUtilV2.scala
+++ b/webapi/src/main/scala/org/knora/webapi/util/standoff/StandoffTagUtilV2.scala
@@ -1006,7 +1006,18 @@ object StandoffTagUtilV2 {
         // then make a sequence of sets of tags sorted by index.
 
         standoff.groupBy(_.startIndex).map {
-            case (index: Int, standoffForIndex: Seq[StandoffTagV2]) => index -> standoffForIndex.map(_.copy(uuid = "")).toSet
+            case (index: Int, standoffForIndex: Seq[StandoffTagV2]) =>
+                // Set each tag's UUID to the empty string (because these should not affect the comparison),
+                // and sort its attributes by standoff property IRI.
+                val comparableTags = standoffForIndex.map {
+                    tag =>
+                        tag.copy(
+                            uuid = "",
+                            attributes = tag.attributes.sortBy(_.standoffPropertyIri)
+                        )
+                }
+
+                index -> comparableTags.toSet
         }.toVector.sortBy {
             case (index: Int, standoffForIndex: Set[StandoffTagV2]) => index
         }.map {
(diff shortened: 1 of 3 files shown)


benjamingeer:
@tobiasschweizer Have you got a minute to review this? It should be very quick.

tobiasschweizer:
Ok, I am looking at it now

webapi/src/main/scala/org/knora/webapi/util/standoff/StandoffTagUtilV2.scala
```diff
@@ -1006,7 +1006,18 @@ object StandoffTagUtilV2 {
...
         standoff.groupBy(_.startIndex).map {
-            case (index: Int, standoffForIndex: Seq[StandoffTagV2]) => index -> standoffForIndex.map(_.copy(uuid = "")).toSet
+            case (index: Int, standoffForIndex: Seq[StandoffTagV2]) =>
+                // Set each tag's UUID to the empty string (because these should not affect the comparison),
+                // and sort its attributes by standoff property IRI.
+                val comparableTags = standoffForIndex.map {
+                    tag =>
+                        tag.copy(
+                            uuid = "",
+                            attributes = tag.attributes.sortBy(_.standoffPropertyIri)
```
tobiasschweizer: This looks familiar to me: We also have to it in tests where the order is not determined.

benjamingeer:
Thanks for the review!

tobiasschweizer:
you are welcome

PR #1225, opened 2019-02-21
fix (bulk import): knora-base:Region as object class constraint

tobiasschweizer:
Adds `knora-base:Region` to the XML schema for bulk import validation (as in #1171).


tobiasschweizer:
@benjamingeer The tests actually failed before I did 
https://github.com/dhlab-basel/Knora/pull/1225/commits/a3339c5c98d3c70a28db55312c410d71446c10ef

so let's see if they pass now with he adaption in the schema

tobiasschweizer:
ok, let's wait for travis to finish the job ;-)

PR #1229, opened 2019-02-25
Don't allow conflicting cardinalities

### LINKED ISSUE #1228: Forbid conflicting cardinalities
This shouldn't be allowed:

- property hasName
- property hasTitle subPropertyOf hasName
- property hasSubtitle subPropertyOf hasName
- class NamedThing
  - hasName maxCardinality 1
- class Text subClassOf NamedThing
  - hasTitle maxCardinality 1
  - hasSubtitle maxCardinality 1

It would result in an error if you tried to create a Text with a title and a subtitle, because that would violate hasName maxCardinality 1.

benjamingeer:
Fixes #1228.


benjamingeer:
Thanks for the review!

PR #1239, opened 2019-03-04
feat (api-admin): Add data loading route

### LINKED ISSUE #1180: Add route for loading data/ontology '.ttl' files
Currently, when loading data with the scripts in `webapi/scripts`, it is required to restart `webapi` so that the ontology cache is refreshed. Also, these scripts require for the data to be accessible by `webapi`, which makes things difficult in production.

Goal: It should be possible to upload the data and ontology `.ttl` files to a route, which will forward them to GraphDB and finally trigger an ontology cache refresh. This operation will only be allowed for a `SystemAdmin`. It will be a destructive operation, as it will first remove any existing data.

subotic:
### Issues

- Resolves #1180


subotic:
abandoning

PR #1242, opened 2019-03-04
Unescape standoff string attributes when verifying text value update

### LINKED ISSUE #1241: property with xml value AssertionException
I get this error: 
```json
{"error":"org.knora.webapi.AssertionException: Resource <http://rdfh.ch/0001/1QK-3k3HQfe91BbPguKpMA> was saved, but one or more of its values are not correct"}
```
when I try to create/update an xml value with `encoded` tags inside an attribute `data-corr="correction with <span class="overline">paragrah</span>"`:
```xml
<text xmlns="">
 <span data-corr="correction with &lt;span class=&quot;overline&quot;&gt;paragrah&lt;/span&gt;" title="titre" class="sic">test</span>
</text>
```
(seems to be related to the ` class="overline"`)
To reproduce: [TestLL.zip](https://github.com/dhlab-basel/Knora/files/2926405/TestLL.zip)
```shell
# LL onto + mapping
./load-onto.expect  http://localhost:7200
# bulk import ok
curl -s -u root@example.com:test -X POST -d @create.xml http://localhost:3333/v1/resources/xmlimport/http%3A%2F%2Frdfh.ch%2Fprojects%2F0001
# failed with v2 API
curl -s -u root@example.com:test -X POST -d @create.json http://localhost:3333/v2/resources
```

benjamingeer:
Fixes #1241.

### DIFF
--- a/docs/src/paradox/00-release-notes/next.md
+++ b/docs/src/paradox/00-release-notes/next.md
@@ -9,7 +9,9 @@ Also, please change the **HINT** to the appropriate level:
 
 - FEATURE: the changes introduced warrant a minor number increase
 
-- FIX: the changes introduced warrant a bugfix number increase
+- FIX: the changes introduced warrant a bug fix number increase
 
 
 ## HINT => MAJOR CHANGE
+
+- FIX: Unescape standoff string attributes when verifying text value update (@github[#1242](#1242))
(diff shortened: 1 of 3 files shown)


benjamingeer:
@gfoo Could you please try this branch?

gfoo:
works for me, thanks!

PR #1251, opened 2019-03-11
feat(api-v2): Get resources in a particular class from a project

### LINKED ISSUE #1235: Add global project filter
I would like to be able to set a global project filter, e.g., in the HTTP request header, and then the API should behave as if this project is the only one on the server:

- `GET /admin/users` - should only return members of the project
- `GET /v2/ontologies/metadata` - should only return the system and project ontologies
- `GET /v2/resources` - should return all the resources in the project (currently GET is not allowed on this endpoint)

These are just some queries from the top of my mind. In the case of `/v2/resources` I don't think that it should be necessary to use Gravsearch just to get some project relevant data out of Knora. It should be possible to add `GET /v2/resources` which under the hood would then run the appropriate Gravsearch query.

Use Case: project-specific GUI.

benjamingeer:
- [x] Implement `X-Knora-Accept-Project` header as per #1235.
- [x] Support `X-Knora-Accept-Project` when requesting ontology metadata.
- [x] Add a `/v2/resources` route that returns resources in a particular class from a project, using `X-Knora-Accept-Project`.
- [x] Add docs.

I first tried to implement this using Gravsearch, but it was a bit slow, I think because the resulting SPARQL SELECT prequery needed to have an `OPTIONAL` block for every property. So I rewrote it to use a custom SELECT prequery template, `getResourcesInProjectPrequery.scala.txt`, which just returns resource IRIs, which are then passed to the full resource query `CONSTRUCT` template, `getResourcePropertiesAndValues.scala.txt`. It's somewhat faster this way.

Resolves #1235.


benjamingeer:
@subotic Could you have a look at this and see if it does what you need?

benjamingeer:
@tobiasschweizer Could you please have a look at `SearchResponderV2.searchResourcesByProjectAndClassV2()` and see if it makes sense in terms of how permissions are handled?

tobiasschweizer:
@benjamingeer I will have a look at it. I can do it by tomorrow, ok?

tobiasschweizer:
>  Could you please have a look at SearchResponderV2.searchResourcesByProjectAndClassV2() and see if it makes sense in terms of how permissions are handled?

If I understand correctly, the `ForbiddenResource` is returned whenever a resource as a whole is not visible. But if the user does not have the permission to see one ore more of a resource's values, the resource is still returned with its visible values.

I think this makes sense for the use case at hand. It is like a full resource request (although there we return a not found exception if a resource is not visible, but the user explicitly submits the resources' Iris).

benjamingeer:
@subotic Thanks, glad to hear it!

PR #1257, opened 2019-03-18
feat(api-v2): Return user's permission on resources and values

### LINKED ISSUE #1204: Return more helpful permission information
With each resource or value, we should return, instead of our internal permission string, just a representation of what the requesting user has permission to do (view, modify, etc.).

If the client wants to change the permissions on the object, they should make another request, which returns the whole set of current permissions, but also in a more user-friendly format.

benjamingeer:
This PR adds `knora-api:userHasPermission` to every resource and value returned in API v2 in the complex schema. This indicates the requesting user's maximum permission on the entity. The values are (in ascending order):

- `RV`: restricted view permission
- `V`: view permission
- `M` modify permission
- `D`: delete permission
- `CR`: change rights permission

Resolves #1204.


benjamingeer:
I didn't remove the existing `knora-api:hasPermissions` from the resource response, because it's just a short string and it doesn't seem worth requiring the user to make a separate request to get it.

tobiasschweizer:
please keep in mind that this requires a change in Knora-ui

Issue #1256, opened 2019-03-18
Generate client code from ontologies


benjamingeer:
The basic idea is to generate client code in different programming languages (at minimum, TypeScript and Python) that makes it easier to work with the Knora APIs. This code could be generated from the Knora ontologies and from project-specific ontologies, so it could be automatically updated when those ontologies change.

### Use case 1: admin API client code

Given the class `knora-base:User`, generate code in Python, TypeScript, etc., that can get a user from the admin API and return a `User` object in the target language. Would have methods like:

- `getUsername: String` (because of `owl:cardinality 1` on property `username`)
- `getIsInGroup: Set[IRI]` (because of `owl:minCardinality 0` on property `isInGroup`)

There would be some kind of factory method that would get `User` objects given their IDs. This method would know which API route to use and what to submit to it. This would be defined in some machine-readable file that the code generator would read.

### Use case 2: client code for a project-specific ontology

Similar to (1), but for a project-specific ontology. For example, given `incunabula:book`, there would be a class called `Book` with methods like:

- `getTitle: Set[TextValue]` (because of `owl:minCardinality 1` on property `title`)
- `getPubdate: Option[DateValue]` (because of `owl:maxCardinality 1` on property `pubdate`)

### Use case 3: generic client code for project-specific data

Here there would be a `Resource` class with methods like:

- `getResourceProperties: Set[IRI]`: returns the set of property IRIs that are subproperties of `knora-api:resourceProperty` (and therefore have Knora values), by querying the relevant ontology
- `getLinkProperties: Set[IRI]`
- `getPropertyValues(propertyIri): Set[Value]`

## Updating data

These classes could also have `set` methods or constructors (whatever makes sense in the target language) for creating/updating data in Knora.

This is just a sketch of an idea, what do you all think?

@subotic @andreas-aeschlimann @kilchenmann @flavens @tobiasschweizer @lrosenth @gfoo @loicjaouen @mrivoal

gfoo:
An another way is to have a [ORM](https://fr.wikipedia.org/wiki/Mapping_objet-relationnel)-like library, a proof of concept with Knora [here](https://github.com/LaDHUL/oknoram).
It is only based on gravsearch BUT it does not update data.

benjamingeer:
@gfoo That looks very cool, thanks.

As a first step, I'm going to try to generate this:

https://github.com/dhlab-basel/knora-api-js-lib/blob/develop/andreas/src/classes/admin/user.ts

[...]

PR #1259, opened 2019-03-19
feat: Generate client API code, part 1

benjamingeer:
This PR is a first step towards #1256.

- [x] Add `ClientApi`, a programming-language-independent representation of a Knora client API. Each `KnoraRoute` can implement `ClientEndpoint` to define client API functions that should be generated.
- [x] Add a route, `/clientapi/TARGET`,  that generates client API code for the specified programming language target, returning a Zip file.
  - [x] Implement `GeneratorFrontEnd`, which converts ontology class definitions into instances of `ClientClassDefinition`, which are suitable for use in code generation.
- [x] Add a Scala DSL for defining client functions (see `docs/src/paradox/05-internals/design/client-api/index.md`).
- [x] Each target is implemented in a class that extends `GeneratorBackEnd`.
  - [x] Implement the target `typescript` in `TypeScriptBackEnd`, which generates source code for use with [knora-api-js-lib](https://github.com/dhlab-basel/knora-api-js-lib), using templates in `twirl/clientapi/typescript`.
- [x] Add `AdminClientApi`, defining the admin client API. Implemented so far:
  - [x] `UsersRouteADM`
  - [x] `PermissionsRouteADM`
  - [x] `GroupsRouteADM`
  - [x] `ProjectsRouteADM`
- [x] Add documentation.

To test locally, start Knora and visit http://0.0.0.0:3333/clientapi/typescript.

To do in subsequent PRs: 

* Finish defining the admin client API.
* Generate code for Knora API v2, using [knora-api-js-lib](https://github.com/dhlab-basel/knora-api-js-lib) and/or [knora-jsonld-simplify](https://github.com/dasch-swiss/knora-jsonld-simplify).


subotic:
This pull request has been mentioned on **Discuss DaSCH**. There might be relevant details there:

https://discuss.dasch.swiss/t/generating-client-libraries/25/1

benjamingeer:
To be generated for the admin API:

- `api/admin/admin-endpoint.ts`
- `api/admin/users/users-endpoint.ts`

benjamingeer:
TODO:

- Adapt the admin ontology external schema so it matches the JSON actually produced.
- Can we make the client API DSL easier to read?

benjamingeer:
TODO:

- Add a description to `ClientClassDefinition` and `ClientPropertyDefinition`, populate them from `rdfs:comment`, and use them to generate code comments.
- Generate [knora-api-connection.ts](https://github.com/dhlab-basel/knora-api-js-lib/blob/develop/andreas/src/knora-api-connection.ts)

benjamingeer:
@andreas-aeschlimann Here's some source code generated using this PR. Could you have a look and see if it's correct? I tried to make it as similar as possible to the handwritten code.

[src.zip](https://github.com/dhlab-basel/Knora/files/3320161/src.zip)

One thing I'm not sure about is the interfaces: I generated them, but they're not used for anything. Are there specific places where they need to be used? I think you mentioned that this could be necessary to prevent import cycles.

benjamingeer:
Thanks, glad you like it!

docs/src/paradox/05-internals/development/generating-client-apis.md
```diff
@@ -0,0 +1,31 @@
...
+License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
+-->
+
+# Generating Client API Code
+
+The following route returns a Zip file containing generated client API
+code for the specified target:
+
+```
+HTTP GET to http://host/v2/clientapi/TARGET
```
lrosenth: `v2` should be removed from the path

benjamingeer: Thanks, will fix in a new PR when I get back from holiday.

docs/src/paradox/05-internals/development/generating-client-apis.md
```diff
@@ -0,0 +1,31 @@
...
+You should have received a copy of the GNU Affero General Public
+License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
+-->
+
+# Generating Client API Code
+
+The following route returns a Zip file containing generated client API
+code for the specified target:
+
+```
```
tobiasschweizer: v2 is not part of the path

andreas-aeschlimann:
I have looked into the TypeScript source and found the following issues:

**Classes**
- Class names: It would be good if we kept the same structure. For example, `admin/groups/groups-endpoint.ts` instead of `groups-api-endpoint.ts`.
- Import path: Some import paths are wrong, they start with `models/...` instead of the relative path.
- Import name: Import strings must not end with `.ts`, there should not be any file extension given.
- Code style: I suggest to drop the `public` keyword everywhere. It is the default and thus redundant. 
- Code style 2: There should be no new line before the first import line, and only one new line after the class end. There should not be more than 1 empty line in a row.
- Documentation style: There is a space missing in the documentation code when you start a documentation block with `/**`. 

**Types**
- In some cases you used references like `project.id` where `id` is used in the API path. However, you have set the type of `id` to `undefined | string` because it is an optional property. This leads to a compile error when you use `project.id` as a string. We need a type check at every use (for example: `if (project.id !== undefined) ...`), or we do not make those properties optional.

As a general remark, if we really auto-generate everything, then we do not need to create the interfaces.

I think it looks good so far. I could not compile the library yet due to these issues, but I suppose it would be possible to fix them.

andreas-aeschlimann:
As answer to your question: We do not need the interfaces/base classes until we bump into circular references. Indeed, I have found circular imports already which will lead to compile errors. For example, the class `Project` imports `User` and `User` imports `Project`.

benjamingeer:
@andreas-aeschlimann Thanks for checking the generated code.

> Class names: It would be good if we kept the same structure. For example, `admin/groups/groups-endpoint.ts` instead of `groups-api-endpoint.ts`.

Are you talking about the directory structure or the filenames? I tried to keep the same directory structure, but some of the filenames are different. I assumed that the filenames were not important, but if you like the old filenames better, we can change them.

>  Import path: Some import paths are wrong, they start with `models/...` instead of the relative path.

Can you be more specific about which ones are incorrect and what they should be?

> Code style 2: There should be no new line before the first import line, and only one new line after the class end. There should not be more than 1 empty line in a row.

It can be difficult to control whitespace in the output while still keeping the templates readable. I will try.

> - Documentation style: There is a space missing in the documentation code when you start a documentation block with `/**`. 

I don't understand. Where should the space be? Can you give an example?

> We need a type check at every use (for example: `if (project.id !== undefined) ...`)

That's easy to add.

> As a general remark, if we really auto-generate everything, then we do not need to create the interfaces.

OK, that makes things simpler. :)

Since this PR has already been merged, please open a new issue for what needs to be changed. I can look at it when I get back from holiday on 12 August. Or you can try changing the TypeScript backend code and templates yourself:

https://github.com/dhlab-basel/Knora/blob/develop/webapi/src/main/scala/org/knora/webapi/util/clientapi/TypeScriptBackEnd.scala

https://github.com/dhlab-basel/Knora/tree/develop/webapi/src/main/twirl/clientapi/typescript

benjamingeer:
> We do not need the interfaces/base classes until we bump into circular references. Indeed, I have found circular imports already which will lead to compile errors. For example, the class `Project` imports `User` and `User` imports `Project`.

Can you make a list of which classes need interfaces?

andreas-aeschlimann:
1) Name convention is: `file-name.ts` should be class `FileName`. `file-name.service.ts` should be injectable class `FileNameService`, for example.

2) Some imports were `models/bla/bla/file.ts`. The path always must be relative, `models` is not the root folder.

3) The first line of your documentation block `/**` has three spaces before the slash instead of 4. Our style checker whines about it, because it is usually 4 spaces.

I will be back mid August as well, so I will gladly respond you in detail then. Have a great time!

Issue #1141, opened 2019-01-07
Custom Object Class Constraints for Standoff Links


tobiasschweizer:
Currently, standoff links have the object class constraint of `knora-base:Resource`.

Would it make sense to add the possibility to make a subclass of `StandoffLink` whose linking property is more specific? E.g., one could make a `PersonStandoffLink` linking to a `beol:Person`.

The advantage would be that Knora not only makes sure that the target exists but also that it has the correct type.

loicjaouen:
I think that it would make sense.

Standoff is so flexible and accessible that it defeats an otherwise well structured data model with constraints on links.

And it could end up making some searches impossible, like give me resources that have a link to a resource class Person.

On the other hand, if you have a given person (a `Person` instance), you can ask for linked resources and find them.

As another example: if you link `Notes` to `Person` with standoff links, you might want to distinguish between different kind of links like character, author, etc.

benjamingeer:
I agree, and I think this goes in the same direction as #438.

Issue #913, opened 2018-06-28
Processing fails when a search returns 25 text values with standoff


benjamingeer:
Try this on the `develop` branch with the BEOL data loaded:

http://localhost:3333/v2/search/Hochwohlgebohrner?limitToStandoffClass=http%3A%2F%2Fapi.knora.org%2Fontology%2Fstandoff%2Fsimple%2Fv2%23StandoffParagraphTag

The response:

```
{
"error": "java.util.NoSuchElementException: key not found: http://www.knora.org/ontology/knora-base#standoffTagHasEnd"
}
```

The stack trace on the console:

```
java.util.NoSuchElementException: key not found: http://www.knora.org/ontology/knora-base#standoffTagHasEnd
	at scala.collection.immutable.Map$Map2.apply(Map.scala:129)
	at org.knora.webapi.util.standoff.StandoffTagUtilV2$.$anonfun$createStandoffTagsV2FromSparqlResults$1(StandoffTagUtilV2.scala:769)
	at org.knora.webapi.util.standoff.StandoffTagUtilV2$$$Lambda$1584/1506996791.apply(Unknown Source)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
	at scala.collection.TraversableLike$$Lambda$12/1676605578.apply(Unknown Source)
	at scala.collection.immutable.HashMap$HashMap1.foreach(HashMap.scala:231)
	at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:462)
	at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:462)
	at scala.collection.TraversableLike.map(TraversableLike.scala:234)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:227)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at org.knora.webapi.util.standoff.StandoffTagUtilV2$.createStandoffTagsV2FromSparqlResults(StandoffTagUtilV2.scala:706)
	at org.knora.webapi.util.ConstructResponseUtilV2$.createValueContentV2FromValueRdfData(ConstructResponseUtilV2.scala:463)
	at org.knora.webapi.util.ConstructResponseUtilV2$.$anonfun$constructReadResourceV2$3(ConstructResponseUtilV2.scala:686)
	at org.knora.webapi.util.ConstructResponseUtilV2$$$Lambda$1560/390556254.apply(Unknown Source)
[...]

PR #1260, opened 2019-03-20
fix (tei beol): fix problems in XSLT

### LINKED ISSUE #1258: BEOL: TEI export of letters
`<ref>` must not occur outside of a `<p>`.

```
<!--

        http://www.knora.org/ontology/knora-base#StandoffLinkTag

        http://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-ref.html

    -->
    <xsl:template match="entity|entity2">
        <ref>
            <xsl:attribute name="target"><xsl:value-of select="@ref"/></xsl:attribute>
            <xsl:apply-templates/>
        </ref>
    </xsl:template>
```

Check if there is a parent of type `<p>`. If not, wrap the element.

tobiasschweizer:
closes #1258


tobiasschweizer:
@subotic after updating this branch with develop, the build works again on travis :-)

tobiasschweizer:
So far, the TEI route obtains all the information making a Gravsearch query. I think that for the text value (and the standoff links) it would make sense to make a full resource request.

So there would be two request for the TEI export:
1. Gravsearch query to get the metadata (TEI header) without the text. In the use case of BEOL, we need to get the author and recipient as nested resources including their IAF (GND), and therefor we need to do a Gravsearch query.
2. Full resource query to get the text incl. all the standoff and the standoff links

The drawback is that the full resource request returns information that is thrown away. I have noticed that a property can be submitted for restricting the request (optional). Maybe we could allow for several properties. But maybe this would be complicated to do in the Sparql template `webapi/src/main/twirl/queries/sparql/v2/getResourcePropertiesAndValuesGraphDB.scala.txt` (I guess a UNION would be needed to achieve that).

In this use case, I need the text property and the standoff links.

Given the standoff links, the references in the TEI/XML could include the type of the target and perhaps also the label.

benjamingeer:
>  I think that for the text value (and the standoff links) it would make sense to make a full resource request.

Why?

> But maybe this would be complicated to do in the Sparql template 

I think you could do it with a `FILTER`.

tobiasschweizer:
I thought about it again and maybe the full resource request is not necessary:

```sparql
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/simple/v2#>

    CONSTRUCT {

        ?letter beol:hasText ?text .
        
        ?letter knora-api:hasStandoffLinkTo ?standoffLinks .

        ?letter beol:hasAuthor ?person1 .

        ?person1 beol:hasFamilyName ?name1 .

        ?person1 beol:hasGivenName ?givenName1 .

        ?person1 beol:hasIAFIdentifier ?iaf1 .

        ?letter beol:hasRecipient ?person2 .

        ?person2 beol:hasFamilyName ?name2 .

        ?person2 beol:hasGivenName ?givenName2 .

        ?person2 beol:hasIAFIdentifier ?iaf2 .


    } WHERE {
        BIND(<http://rdfh.ch/0801/m7WpUEviTHOXStg8tsnosQ> as ?letter)

        ?letter a beol:letter .

        ?letter beol:hasText ?text .
        
        OPTIONAL {
        	?letter knora-api:hasStandoffLinkTo ?standoffLinks .
        }

        OPTIONAL {

          ?letter beol:hasAuthor ?person1 .

          ?person1 beol:hasFamilyName ?name1 .

          ?person1 beol:hasGivenName ?givenName1 .

          ?person1 beol:hasIAFIdentifier ?iaf1 .

        }

        OPTIONAL {

          ?letter beol:hasRecipient ?person2 .

          ?person2 beol:hasFamilyName ?name2 .

          ?person2 beol:hasGivenName ?givenName2 .

          ?person2 beol:hasIAFIdentifier ?iaf2 .

        }

    }
```

Do you think we could always return all standoff information if the Gravsearch query contains a BIND (applies to one resource only)? The Gravsearch query above is kind of a full resource request.

benjamingeer:
> Do you think we could always return all standoff information if the Gravsearch query contains a BIND 

You can do a BIND and still get multiple results, if you're binding something other than the main resource IRI.

Since this feature is needed only internally (not in the API), why not add a boolean parameter, `returnCompleteTextValues`, to `GravsearchRequestV2`?

tobiasschweizer:
> Since this feature is needed only internally (not in the API), why not add a boolean parameter, returnCompleteTextValues, to GravsearchRequestV2?

The Gravsearch query is generated from a template and is only used internally, yes. Yes, I think the parameter would make sense. Shall I add this to the issue https://github.com/dhlab-basel/Knora/issues/913 where we discussed not retuning all standoff so we do not forget about is?

benjamingeer:
> Shall I add this to the issue #913

Yes please.

tobiasschweizer:
see 
https://github.com/dhlab-basel/Knora/issues/913#issuecomment-475564123

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -1274,6 +1274,24 @@ class ResourcesResponderV2(responderData: ResponderData) extends ResponderWithSt
                 }
             }
 
+            // get standoff link values
+            standoffLinkValuesOption: Option[Seq[ReadValueV2]] = resource.values.get(OntologyConstants.KnoraBase.HasStandoffLinkToValue.toSmartIri)
+
+            standoffLinkValues: Seq[ReadLinkValueV2] = standoffLinkValuesOption match {
+
+                case Some(valueSeq: Seq[ReadValueV2]) =>
```
tobiasschweizer: @benjamingeer In a first approach, I tried to match directly for `Seq[ReadLinkValueV2]` but I got a compiler warning: "non-variable type argument is unchecked since it is eliminated by erasure".

I found a [post](https://stackoverflow.com/questions/25222989/erasure-elimination-in-scala-non-variable-type-argument-is-unchecked-since-it) which discusses that issue and use `collect` now which works fine. 

Do you think that is a good solution?

tobiasschweizer: I am reading http://squidarth.com/scala/types/2019/01/11/type-erasure-scala.html now ...

benjamingeer: > got a compiler warning

Yes, I think this because you’re matching on a subtype of the element type in a collection. If `collect` does what you need, that’s great. Otherwise you could iterate over the elements and `match` each element.

tobiasschweizer: `collect` does the job :-)

tobiasschweizer:
@benjamingeer I added the property `knora-base:standoffLinkTagTargetHasType` to knora-base. It is only used for the TEI/XML export: whenever there is a standoff link tag, the type of the referred resource gets added automatically to the tag's attributes:

```xml
Der Brief Hermanns an Varignon, in dem er einige Exemplare der
<hi rend="italic">Mém. Paris</hi>
für Scheuchzer bestellt hatte und auf dessen Beantwortung er wartete, ist anscheinend nicht erhalten. Die Antwort Varignons traf bei Hermann im April 1706 ein. Hermann informierte Johannes Scheuchzer darüber in seinem Brief von
<ref target="http://rdfh.ch/0801/pVo6TJS1QgWp_SoujgQaZQ" type="http://0.0.0.0:3333/ontology/0801/beol/simple/v2#letter">1706.04.07</ref>
und zitiert darin die für Scheuchzer relevante Passage.
```

Like this, the client knows whether a linked resource can be obtained as TEI/XML as well.

Do you think this is a good way to handle it? This way, the standoff to xml conversion works as is.

benjamingeer:
@tobiasschweizer why not use `objectClassConstraint`, which we already have?

You could make a project-specific subproperty of `standoffTagHasLink`. Its `objectClassConstraint` would be the specific resource type that it points to. Your mapping could include that type in the XML.

tobiasschweizer:
In TEI/XML, the actual type of the referred resource should be displayed although the object class constraint can be generic, i.e `knora-base:Resource`.

I would like to return the type of each link's target not the one defined in the model.

benjamingeer:
Sorry, I don't understand. What is the object of `standoffLinkTagTargetHasType`? I see it has `objectDatatypeConstraint xsd:string`. What is this string? Is it a type IRI? If so, it would be redundant with `objectClassConstraint`.

>  the actual type of the referred resource should be displayed although the object class constraint can be generic, i.e knora-base:Resource.

The object class constraint doesn't have to be generic if you make a subproperty of `standoffTagHasLink` as I suggested. Your subproperty would have a more specific `objectClassConstraint`.

tobiasschweizer:
> The object class constraint doesn't have to be generic if you make a subproperty of standoffTagHasLink as I suggested. Your subproperty would have a more specific objectClassConstraint.

If I understand correctly, your approach is model-based (ontology). So for each target type (link type) there would be a separate subproperty of `standoffTagHasLink` with a custom object class constraint. I believe this corresponds to https://github.com/dhlab-basel/Knora/issues/1141, right? And then at creation time, there would be checks.
 
What I tried to do in this PR, is simply putting the information contained in `knora-base:hasStandoffLinkToValue` into the TEI/XML. This will always give me the actual type of the instance also if the object class constraint is more general. I added some code in scala that automatically creates a standoff attribute with the target's type (and therefore I had to add `standoffLinkTagTargetHasType `).

`standoffLinkTagTargetHasType` is not necessarily the same as `objectClassConstraint` because it could be more specific (the type could be a subclass of the constraint). 

I would prefer a way without having `standoffLinkTagTargetHasType` in the ontology but I think this would require me to change the standoff to XML conversion logic.

tobiasschweizer:
@benjamingeer I am also available on Facetime today.

tobiasschweizer:
after discussion with @benjamingeer: I will undo the changes I made to add the target's type in the XML and create a separate issue so we can continue on that once we have a concrete use case for the TEI/XML export.

tobiasschweizer:
@SepidehAlassi Once you are back from holiday, could we look at the table issue in the Goldbach letters?

webapi/_test_data/test_route/texts/beol/standoffToTEI.xsl
```diff
@@ -441,10 +441,22 @@
...
+            <xsl:when test="ancestor::p">
+                <ref>
+                    <xsl:attribute name="target"><xsl:value-of select="@src"/></xsl:attribute>
+                    <xsl:apply-templates/>
+                </ref>
+            </xsl:when>
+            <xsl:otherwise>
+                <p>
+                    <ref>
+                        <xsl:attribute name="target"><xsl:value-of select="@src"/></xsl:attribute>
```
tobiasschweizer: @SepidehAlassi Once you have fixed the original XML, this is not necessary anymore.

SepidehAlassi: @tobiasschweizer In Goldbach letters, there were no errors about facsimiles links that did not occur inside a paragraph. There were cases of endnote references which were not in the paragraph, that I have fixed.

tobiasschweizer: @SepidehAlassi Great, so I will undo this hack :-)

tobiasschweizer: undone in https://github.com/dhlab-basel/Knora/pull/1260/commits/d0f85365ba67a98ec489a43cfead4229385760ba

SepidehAlassi:
@tobiasschweizer how can I test this?

SepidehAlassi:
@tobiasschweizer how can I test this?

tobiasschweizer:
@SepidehAlassi run `webapi/_test_data/test_route/texts/beol/beolTEI.py`. The script will return the following parameters:

- gravsearchTemplateIri
- mappingIRI
- teiHeaderXSLTIri
- textProperty

You can use https://github.com/dhlab-basel/beol/pull/135 and add those parameters to the config. Then the letter template will display a link to obtain a TEI/XML file.

I have also made a script that gets the TEI/XML file for all the letters. It is in switchdrive/tei and requires a file with all the letter Iris (letterIris.csv).

@benjamingeer merged a PR (https://github.com/dhlab-basel/Knora/pull/1263) yesterday that requires the data to be converted. Shall I do it for beol-data.ttl and put the file back on switchdrive?

tobiasschweizer:
@SepidehAlassi I added `beol-data_converted.ttl` to switchdrive.

tobiasschweizer:
@SepidehAlassi I could fix the table problem (empty rows) for the BEBB letters in XSLT. Thanks a lot for pointing this out!

webapi/_test_data/test_route/texts/beol/standoffToTEI.xsl
SepidehAlassi: @tobiasschweizer I just checked this. I exported the TEI of the L0001 which has a facsimile. Validation of it fails because of the reference of facsimile outside of the paragraph. This is weird because this letter was not among the problematic xmls you had asked me to check before!
So sorry I have not noticed that there are actually references of facsimiles which should be wrapped in <p>
Now, either you have to leave the hack to be, or I have to change my xslt again to fix these. Then I would need to give you the new data again.

SepidehAlassi:
@tobiasschweizer I just checked this. I exported the TEI of the L001-O which has a facsimile. Validation of it fails because of the reference of facsimile outside of the paragraph. This is weird because this letter was not among the problematic xmls you had asked me to check before!
So sorry I have not noticed that there are actually references of facsimiles which should be wrapped in <p>
Now, either you have to leave the hack to be, or I have to change my xslt again to fix these. Then I would need to give you the new data again.

tobiasschweizer:
@SepidehAlassi This is because before I just wrapped the facsimile with a paragraph when transforming it to TEI/XML ;-) Sorry for the confusion.

Would it be possible for you to wrap the facsimile links with paragraphs before importing them? If so, please use the current version of `develop` so the data does not have to be converted again (due to #1263 ). New data is fine for me.

SepidehAlassi:
@tobiasschweizer ok, I'll do it first thing tomorrow.

tobiasschweizer:
Awesome! Thanks for the review!

PR #1278, opened 2019-03-26
a handfull of changes in documentation and error messages

### LINKED ISSUE #1277: amend documentation
there are a couple of little changes to do on the documentation, as I go through it again, I make changes here.

loicjaouen:
fixes #1277


loicjaouen:
@subotic , while building the documentation, I had the error:

```bash
.rvm/gems/ruby-2.3.4/gems/bundler-2.0.1/lib/bundler/spec_set.rb:87:in `block in materialize': Could not find concurrent-ruby-1.1.3 in any of the sources (Bundler::GemNotFound)
```

which I fixed locally with:

```bash
cd docs/src/jekyll
bundle install
```

should it be mentioned somewhere in the documentation or is it a local oddity?

subotic:
It is mentioned at the end of https://github.com/dhlab-basel/Knora/blob/develop/docs/Readme.md but maybe it needs to be extended.

subotic:
I would like to dockerize building of the documentation. I always have problems, too.

loicjaouen:
@subotic : thanks :)

loicjaouen:
@subotic and @benjamingeer , my next problem is to build and run the `webapi` in intelliJ, the current description fails on a missing `scala-xml` jar (not included within scala since 2.1). I guess it is because we import the project from the `Knora/webapi` folder as now the sbt build config is on the root folder `Knora`, shouldn't we import the project from the root folder?

I tried so... but I get into other problems of multiple declaration:
```
/Users/ljaouen/Soft/Code/github/KnoraB/webapi/src/test/scala/org/knora/webapi/util/MutableTestIri.scala
    Error:(27, 7) MutableTestIri is already defined as class MutableTestIri
class MutableTestIri {
    Error:(60, 12) TestIriException is already defined as case class TestIriException
case class TestIriException(message: String) extends Exception(message)
```

How do you set-up a knora development environment?

benjamingeer:
@loicjaouen Just import the root folder `Knora` to create a project in IntelliJ. Select the SBT option when prompted. This needs to be updated in the Knora docs.

loicjaouen:
@benjamingeer thanks, yes:
- I was about to update the doc
- I did import the root folder as sbt, but then got this error about `already defined` classes, I'll re-try

benjamingeer:
>  got this error about already defined classes, I'll re-try


Try this in the root folder:

```
$ sbt
> clean
```

benjamingeer:
Also try deleting any existing `.idea` directories you might have (in the root folder or in `webapi`).

loicjaouen:
I did all of that and still get the same build exceptions :(

loicjaouen:
Here are the screenshots I took to amend the docs, @benjamingeer could you have a look if something catches your eye?

starting point, import project:
Capture d’écran 2019-03-27 à 17 10 46

as `sbt`:
Capture d’écran 2019-03-27 à 17 11 36

I'm not touching the prefs:
Capture d’écran 2019-03-27 à 17 12 02

Result of the import:
Capture d’écran 2019-03-27 à 17 14 08

Adding a configuration to run/debug the webapi:
Capture d’écran 2019-03-27 à 17 14 41

giving it a name, selecting a class and a module:
Capture d’écran 2019-03-27 à 17 15 55

forcing the way through, right clicking to get to the option and exclude the two files from the build... then it works, I can then build and debug `webapi`. But that might not be the way we want to document it :)

benjamingeer:
I'll try this tomorrow and see if I can reproduce the problem you're having.

benjamingeer:
OK, I deleted my `Knora/.idea` folder and did a fresh import into IntelliJ, with no errors.

Start IntelliJ:

Screen Shot 2019-03-29 at 10 02 35

Select "Import Project":

Screen Shot 2019-03-29 at 10 02 44

SBT is selected automatically. Click "Next":

Screen Shot 2019-03-29 at 10 03 07

Select a JDK:

Screen Shot 2019-03-29 at 10 03 18

IntelliJ imports the project without errors:

Screen Shot 2019-03-29 at 10 05 27

@loicjaouen One more idea for you: try deleting the whole `Knora` folder and cloning it from GitHub again.

benjamingeer:
Ah, I think I finally understand what you mean now. You're getting that error not when you import Knora into IntelliJ, but when you try to use IntelliJ as a build tool and debugger. I've never tried to do that. I always build using SBT, and I don't like debuggers. I believe that @SepidehAlassi and @tobiasschweizer have figured out how to make this work, though.

benjamingeer:
I wonder if this option in the "Import Project" dialog might do what you want:

Screen Shot 2019-03-29 at 10 24 45

loicjaouen:
@benjamingeer thanks for taking the time to look at it

Let's say that at that point the person going through the set-up should be skilled enough to either deal with the possible errors (like i did by excluding those files) or ask the community.

Do you mind reviewing the PR?

loicjaouen:
Thanks @benjamingeer and @subotic  :)

PR #1285, opened 2019-03-29
https://github.com/dhlab-basel/Knora/blob/fix/change-password/webapi/…

### LINKED ISSUE #1284: admin api: change password is broken
from the docs:
https://github.com/dhlab-basel/Knora/blob/develop/docs/src/paradox/03-apis/api-admin/users.md#update-users-password

```
PUT: /admin/users/iri/<userIri>/Password
{
  "oldPassword": "test",
  "newPassword": "test1234"
}
```

should change the password.

What we get is a `400 Bad Request` response:
```
The request content was malformed:
Too few parameters sent for password change.
```

loicjaouen:
fixes  #1284

### DIFF
--- a/docs/src/paradox/03-apis/api-admin/users.md
+++ b/docs/src/paradox/03-apis/api-admin/users.md
@@ -109,7 +109,7 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
   - BODY:
     ```
     {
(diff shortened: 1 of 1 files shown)


loicjaouen:
apparently, it was on my local set-up only that I received the exception described in the bug, it works fine on prod, let's only fix the doc for now.

PR #1306, opened 2019-04-12
feat(api-v2): Allow a resource to be created as a specified user

### LINKED ISSUE #1280: Set attachedToUser on resource creation
It will be very helpful for us to set the attached user IRI on resource creation (like creation/modif dates etc.).

The only solution I see is to login with the concerned user, create resource, logout. In a database migration scenario it is no really efficient.

Related to #1217

benjamingeer:
Resolves #1280.


benjamingeer:
@gfoo Please try this and see if it does what you need.

gfoo:
Won't have time right now, on holydays for the next week, back on April 24th. I will test at that time.

benjamingeer:
@gfoo OK, bonnes vacances!

benjamingeer:
Thank you!

PR #1291, opened 2019-04-03
feat(admin): Give the admin ontology an external schema

benjamingeer:
Resolves #1289.

PR #1307, opened 2019-04-15
feat(api-v2): Return standoff markup separately from text values

### LINKED ISSUE #913: Processing fails when a search returns 25 text values with standoff

### LINKED ISSUE #1293: Errors on creation of resources with huge Richtext value
Don't really know if it is errors in my data/standoff mapping or bugs?
Could you give some help by trying to reproduce these errors:

```json
{"error":"org.knora.webapi.InconsistentTriplestoreDataException: Expected http://www.w3.org/2000/01/rdf-schema#label in assertions for a value object of type list value."}
{"error":"java.util.NoSuchElementException: key not found: http://www.knora.org/ontology/knora-base#standoffTagHasEnd"}
{"error":"java.util.NoSuchElementException: key not found: http://www.knora.org/ontology/knora-base#standoffTagHasStart"}
```
All tests are described in the README of [FailedImportTrans.zip](https://github.com/dhlab-basel/Knora/files/3047508/FailedImportTrans.zip)

benjamingeer:
This PR enables text markup to be returned as standoff/RDF, separately from text values. Standoff is returned in pages of a maximum size.

It also queries standoff in pages internally when returning markup as XML, to reduce the size of SPARQL query results.

- [x] Change `getResourcePropertiesAndValues.scala.txt` so it can return:
  - [x] Text values without their standoff.
  - [x] Text values with the first page of standoff from each one.
  - [x] A page of standoff from a text value (given the value IRI), without `valueHasString`.
- [x] Add `knora-base:valueHasMaxStandoffStartIndex` to each text value with standoff, so that when Knora queries a page of standoff, it knows whether it has reached the last page.
- [x] To optimise querying a page of standoff, change the format of standoff IRIs so that the last component is the start index (instead of a UUID). This allows us to get the start indexes of the parent tags from their IRIs, without having to query them.
- [x] Add a script to update existing repositories to change standoff IRIs and add `knora-base:valueHasMaxStandoffStartIndex`.
- [x] Change `ReadTextValueV2` and `TextValueContentV2` so they can represent either:
  - [x] A complete text value with its markup.
  - [x] A text value without its markup.
  - [x] A page of standoff tags, without `valueHasString`.
- [x] Handle those three cases in `ConstructResponseUtilV2.makeTextValueContentV2`.
- [x] Support “schema options” (see `OntologySchema.scala`) to enable to client to choose (via an HTTP header or a URL parameter) whether markup should be returned as XML or as standoff in the complex schema. (The default is XML.)
- [x] In the simple schema, don’t query standoff at all when getting resources from the triplestore.
- [x] If the schema options say that markup will be returned as standoff:
  - [x] Don’t query standoff when getting resources from the triplestore.
  - [x] When returning a text value, include a boolean `knora-api:textValueHasMarkup` to indicate whether it has any markup, and if so, include `knora-api:textValueHasMaxStandoffStartIndex`, to give the client an idea of how much data to expect.
- [x] Implement the standoff query in `StandoffResponderV2`, using `getResourcePropertiesAndValues.scala.txt` (so it can use `ConstructResponseUtilV2`).
- [x] Add a simple task queue mechanism to `ActorUtil`, in which each task returns a result and an optional next task.
- [x] When querying all the standoff in a text value (either to verify an update or to return the text value as XML), query it internally one page at a time (in `StandoffResponderV2`, using the task queue mechanism), to avoid getting huge responses from the triplestore. The first page is queried with the rest of the text value, and each subsequent page is queried separately. This process starts:
  - [x] In `ResourcesResponderV2`, using `getResourcePropertiesAndValues.scala.txt` to get the first page of standoff with each text value.
  - [x] in `SearchResponderV2`, using `GravsearchMainQueryGenerator` and `FullTextMainQueryGenerator` to get the first page of standoff with each text value.
- [x] Add a route for requesting a page of standoff from a text value, given an offset.
  - [x] Generate JSON-LD from standoff case classes.
  - [x] With each page of standoff, return the next available offset (if any) as `knora-api:nextStandoffStartIndex`.
- [x] In `OntologyConstants`, rename `KnoraApiV2WithValueObjects` -> `KnoraApiV2Complex` (which I forgot to do in #1291).
- [x] Update existing test data.
- [x] Add tests.
- [x] Add documentation.

Resolves #913. (Open a new issue for related functionality to be developed later.)
Fixes #1293.


tobiasschweizer:
@benjamingeer I would be happy to do the review once this PR is ready. I could also test the data transformation scripts.

I will be back from holiday on Tuesday, 30 April.

benjamingeer:
@tobiasschweizer I have a problem: I'm using `ConstructUtilV2` to process each page of standoff (to check permissions, etc.). But `ConstructUtilV2` requires a mapping for each text value with standoff. I don't want to query a mapping every time I get a page of standoff. The problem is that `ConstructUtilV2.makeTextValueContentV2` calls `StandoffTagUtilV2.createStandoffTagsV2FromSparqlResults`, which requires a `StandoffEntityInfoGetResponseV2`, which comes from `StandoffResponderV2.getStandoffEntitiesFromMappingV2`. Can you explain why?

It seems to me that if we are returning standoff rather than XML, the text value shouldn't need to have a mapping at all. Currently every text value with standoff is required to have a mapping, but I don't understand why.

Can you help me refactor this code so that text values with standoff don't need mappings?

tobiasschweizer:
@benjamingeer We can look at it together on Monday.

gfoo:
To plan my futur devs, in this PR: 
- are you going to remove the xml representation of a `Resource` json response?
- If so, will you add a new API enpoint allowing to retrieve this xml data?

gfoo:
> To plan my future devs, in this PR:
> 
>     * are you going to remove the xml representation of a `Resource` json response?

the xml representation value of `Richtext` properties of the resource of course!
> 
>     * If so, will you add a new API enpoint allowing to retrieve this xml data?

benjamingeer:
> are you going to remove the xml representation of a Resource json response?

Not in this PR and not anytime soon. Eventually I would like to try to convert standoff markup to XML in the client, but this will take some time to develop. In the meantime, the current functionality will still be supported. This PR fixes the issues associated with huge SPARQL results caused by querying standoff, so there is no urgent reason to remove the existing XML generation functionality.

tobiasschweizer:
@benjamingeer I will review this PR on Monday. Is that ok?

benjamingeer:
Monday is perfect, thanks!

docs/src/paradox/03-apis/api-v2/reading-and-searching-resources.md
```diff
@@ -61,6 +61,124 @@ See the interfaces `Resource` and `ResourcesSequence` in module
...
+The object `knora-api:valueAsString` contains the text without markup. The predicate
+`knora-api:textValueHasMarkup` indicates that the text value has markup,
+and the value of the predicate `knora-api:textValueHasMaxStandoffStartIndex` gives the start
+index of the last standoff tag; this gives the client some idea of how much markup there is.
+
+You can then request the text value's standoff/RDF, which is returned in pages of a limited
+size. To get each page:
+
+```
+HTTP GET to http://host/v2/standoff/RESOURCE_IRI/TEXT_VALUE_IRI/OFFSET
```
tobiasschweizer: I had to provide `?markup=standoff` as a param to make it work in the browser (did not modify any headers). In this case (`v2/standoff/resIri/textValIri`), shouldn't standoff be the default? What else would the client want to get when using this route?

benjamingeer:
With GraphDB 8.5 and a large repository, we get consistency checking errors when running the SPARQL-based upgrade script, and upgrade is very slow anyway. This reminds me that last time I wrote an upgrade script, I also initially wrote it to use SPARQL, and it worked with the latest GraphDB, but with GraphDB 8.5 we got consistency checking errors. I wonder if there is a bug in GraphDB 8.5 consistency checking, which perhaps appears only when there is a lot of data. For now I'm going to solve this problem the way I did last time: rewrite the upgrade script to download the data to files and update it in Python rather than in SPARQL.

@subotic When you get a chance, it would be great it if we could find a way to use the latest version of GraphDB.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -959,23 +992,30 @@ class ResourcesResponderV2(responderData: ResponderData) extends ResponderWithSt
...
                                propertyIri: Option[SmartIri] = None,
                                versionDate: Option[Instant] = None,
+                               targetSchema: ApiV2Schema,
+                               schemaOptions: Set[SchemaOption],
                                requestingUser: UserADM): Future[ReadResourcesSequenceV2] = {
 
         // eliminate duplicate Iris
         val resourceIrisDistinct: Seq[IRI] = resourceIris.distinct
 
+        val queryStandoff = SchemaOptions.queryStandoffWithTextValues(targetSchema = targetSchema, schemaOptions = schemaOptions)
```
tobiasschweizer: Please add a comment that this finds out whether standoff has to be queried and that this passed to `makeTextValueContentV2`.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -916,24 +934,39 @@ class ResourcesResponderV2(responderData: ResponderData) extends ResponderWithSt
...
+                                            targetSchema: ApiV2Schema,
+                                            schemaOptions: Set[SchemaOption],
                                             requestingUser: UserADM): Future[Map[IRI, ResourceWithValueRdfData]] = {
 
         // eliminate duplicate Iris
         val resourceIrisDistinct: Seq[IRI] = resourceIris.distinct
 
+        // If we're supposed to query standoff, get the indexes delimiting the first page of standoff. (Subsequent
+        // pages, if any, will be queried separately.)
+        val (maybeStandoffMinStartIndex: Option[Int], maybeStandoffMaxStartIndex: Option[Int]) = StandoffTagUtilV2.getStandoffMinAndMaxStartIndexesForTextValueQuery(
```
tobiasschweizer: Is this necessary? Isn't this already determined in `getResourcesV2 `:

`val queryStandoff = SchemaOptions.queryStandoffWithTextValues(targetSchema = targetSchema, schemaOptions = schemaOptions)`

webapi/src/main/scala/org/knora/webapi/responders/v2/search/gravsearch/mainquery/GravsearchMainQueryGenerator.scala
```diff
@@ -214,18 +219,44 @@ object GravsearchMainQueryGenerator {
...
+            val queryStandoff: Boolean = SchemaOptions.queryStandoffWithTextValues(targetSchema, schemaOptions)
+
+            // WHERE patterns for querying the first page of standoff in each text value
+            val wherePatternsForStandoff: Seq[QueryPattern] = if (queryStandoff) {
+                Seq(
+                    mainAndDependentResourcesValueObjectsValuePattern,
+                    StatementPattern.makeExplicit(subj = mainAndDependentResourceValueObject, pred = IriRef(OntologyConstants.KnoraBase.ValueHasStandoff.toSmartIri), obj = standoffNodeVar),
+                    StatementPattern.makeExplicit(subj = standoffNodeVar, pred = standoffPropVar, obj = standoffValueVar),
+                    StatementPattern.makeExplicit(subj = standoffNodeVar, pred = IriRef(OntologyConstants.KnoraBase.StandoffTagHasStartIndex.toSmartIri), obj = standoffStartIndexVar),
+                    FilterPattern(
```
tobiasschweizer: This FILTER makes sure that only the first page of standoff is queried, right? Any subsequent pages would be queried in `ConstructResponseUtilV2.makeTextValueContentV2`, right?

docs/Makefile
```diff
@@ -38,7 +38,7 @@ jsonformat:
 
 jsonformattest:
 	$(TYPEDOC) --out $(FORMATDOCDIRV1) --readme none --name "Knora API V1 Format Documentation" --module commonjs $(FORMATSRCDIRV1)
-	$(TYPEDOC) --out $(FORMATDOCDIRV2) --readme none --name "Knora API V2 JSON-LD Format Documentation" --module commonjs $(FORMATSRCDIRV2)
+	# $(TYPEDOC) --out $(FORMATDOCDIRV2) --readme none --name "Knora API V2 JSON-LD Format Documentation" --module commonjs $(FORMATSRCDIRV2)
```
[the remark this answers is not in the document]
benjamingeer: Yes.

knora-ontologies/knora-base.ttl
```diff
@@ -2591,6 +2599,10 @@
                              owl:onProperty :valueHasStandoff ;
                              owl:minCardinality "0"^^xsd:nonNegativeInteger
                            ],
+                           [ rdf:type owl:Restriction ;
+                             owl:onProperty :valueHasMaxStandoffStartIndex ;
+                             owl:maxCardinality "1"^^xsd:nonNegativeInteger
```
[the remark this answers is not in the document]
benjamingeer: That's right.

docs/src/paradox/03-apis/api-v2/reading-and-searching-resources.md
benjamingeer: You're right, I'll fix that.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
benjamingeer: OK.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
benjamingeer: You're right, I'll fix that.

webapi/src/main/scala/org/knora/webapi/responders/v2/search/gravsearch/mainquery/GravsearchMainQueryGenerator.scala
benjamingeer: That's right. This the same FILTER used in `getResourcePropertiesAndValuesGraphDB.scala.txt`.

docs/src/paradox/03-apis/api-v2/reading-and-searching-resources.md
benjamingeer: Done in fe74ba6.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
benjamingeer: Done in fe74ba6.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
benjamingeer: Done in fe74ba6.

benjamingeer:
@tobiasschweizer I rewrote the upgrade script and ran it with the BEOL data, and it took 20 minutes and seemed to work. Let me know how it works for you.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -959,23 +989,31 @@ class ResourcesResponderV2(responderData: ResponderData) extends ResponderWithSt
...
         for {
 
             queryResultsSeparated: Map[IRI, ResourceWithValueRdfData] <- getResourcesFromTriplestore(
                 resourceIris = resourceIris,
                 preview = false,
                 propertyIri = propertyIri,
                 versionDate = versionDate,
+                queryStandoff = queryStandoff,
                 requestingUser = requestingUser
             )
```
tobiasschweizer: Shouldn't getting the mappings depend on `queryStandoff`? If no standoff is to be queried, no mapping will be needed.

tobiasschweizer: or does `queryResultsSeparated` just does not contain any standoff in that case and nothing is done?

tobiasschweizer:
@benjamingeer I could convert the Meditationes data. I could not diff the data because the order seems changed. I think this is because the Iris are different.

Do you want to inform @SepidehAlassi and @subotic about the need of data conversion since both of them will be affected by this PR?

benjamingeer:
> Do you want to inform @SepidehAlassi and @subotic about the need of data conversion since both of them will be affected by this PR?

I think you just did. :) Also, I sent a message about it to the knora-user mailing list a few days ago.

benjamingeer:
Is there anything else that needs to be done in this PR before it's merged?

webapi/scripts/graphdb-se-knora-test-repository-config.ttl.tmpl
```diff
@@ -17,6 +17,6 @@
        owlim:repository-type "file-repository" ;
        owlim:query-timeout "30" ;
        owlim:throw-QueryEvaluationException-on-timeout "true";
-       owlim:query-limit-results "50000";
+       # owlim:query-limit-results "50000";
```
tobiasschweizer: Do you think there is no limit we could define to protect the triplestore? Now with standoff queried in iterations, the previous limit should not be  a problem. I agree that such a limit will always be somehow arbitrary.

benjamingeer:
> Do you think there is no limit we could define to protect the triplestore? 

Opened #1320 for this.

tobiasschweizer:
@benjamingeer I think this is ready to merge :-) I will then work on #1317

subotic:
This pull request has been mentioned on **Discuss DaSCH**. There might be relevant details there:

https://discuss.dasch.swiss/t/keeping-response-sizes-small/26/1

PR #1311, opened 2019-04-30
fix(api-v2): Cache base class IRIs correctly when creating/updating class

### LINKED ISSUE #1310: Base class IRIs not cached correctly when creating/updating class
`ReadClassInfoV2.allBaseClasses` ends up empty in the ontology cache.

Thanks @lrosenth for spotting this.

benjamingeer:
Fixes #1310.


benjamingeer:
Thanks for the quick review!

PR #1314, opened 2019-05-03
Support list nodes in Gravsearch

### LINKED ISSUE #1312: Search for list nodes v2
- complex schema: find given list node or any of its subnodes:
   generated SPAQRL:
```sparql
 ?propVal0 <http://www.knora.org/ontology/knora-base#valueHasListNode> ?listNode .
           <http://rdfh.ch/lists/0801/algebra> knora-base:hasSubListNode* ?listNode .
```
- simple schema: do string match for list node's name, e.g. "summer"

tobiasschweizer:
This PR supports list nodes in Gravsearch:
- [x] complex schema: match given list node or any of its subnodes
- [x] simple schema: perform an exact match for the list node's label (without subnodes)

In a future PR I will do some refactoring so the list node label (simple schema) is returned in a deterministic way (user's preferred language), using the list responder (see #1317 ).

closes #1312

### DIFF
--- a/docs/src/paradox/00-release-notes/next.md
+++ b/docs/src/paradox/00-release-notes/next.md
@@ -24,3 +24,5 @@ Also, please change the **HINT** to the appropriate level:
 
 - MAJOR: Separate the `knora-admin` ontology from `knora-api` (@github[#1263](#1263)).
   Existing repositories must be updated; see `upgrade/1263-knora-admin` for instructions.
+  
+- FEATURE: Add support for searching for specific list values in Gravsearch for both the simple and complex schema (@github[#1314](#1314)).  
--- a/docs/src/paradox/03-apis/api-v2/query-language.md
+++ b/docs/src/paradox/03-apis/api-v2/query-language.md
@@ -90,8 +90,8 @@ The response to a count query request is an object with one predicate,
 A Gravsearch query can be written in either of the two
 @ref:[Knora API v2 schemas](introduction.md#api-schema). The simple schema
 is easier to work with, and is sufficient if you don't need to query
-anything below the level of a Knora value. If your query needs to refer to
-list nodes or standoff markup, you must use the complex schema. Each query must use a single
+anything below the level of a Knora value. If your query needs to refer to 
+standoff markup, you must use the complex schema. Each query must use a single
 schema, with one exception (see @ref:[Date Comparisons](#date-comparisons)).
 
 Gravsearch query results can be requested in the simple or complex schema;
@@ -277,6 +277,12 @@ expressions in the simple schema:
 - Decimal values (`xsd:decimal`)
 - Boolean values (`xsd:boolean`)
 - Date values (`knora-api:Date`)
+- List values (`knora-api:ListNode`)
+
+List values can only be searched for using the equal operator (`=`), 
+performing an exact match on a list node's label. Labels can be given in different languages for a specific list node.
+If one of the given list node labels matches, it is considered a match.
+Note that in the simple schema, uniqueness is not guaranteed (as opposed to the complex schema). 
 
 A Knora value may not be represented as the literal object of a predicate;
 for example, this is not allowed:
@@ -785,7 +791,9 @@ PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
 
 ### Searching for a List Value Referring to a Particular List Node
 
-This is possible only in the complex schema:
+Since list nodes are represented by their Iri in the complex schema, 
+uniqueness is guranteed (as opposed to the simple schema).
+Also all the subnodes of the given list node are considered a match.
 
 ```
 PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>
--- a/webapi/src/main/scala/org/knora/webapi/OntologyConstants.scala
+++ b/webapi/src/main/scala/org/knora/webapi/OntologyConstants.scala
@@ -952,6 +952,7 @@ object OntologyConstants {
         val Color: IRI = KnoraApiV2PrefixExpansion + "Color"
         val Interval: IRI = KnoraApiV2PrefixExpansion + "Interval"
         val Geoname: IRI = KnoraApiV2PrefixExpansion + "Geoname"
+        val ListNode: IRI = KnoraApiV2PrefixExpansion + "ListNode"
 
         val Resource: IRI = KnoraApiV2PrefixExpansion + "Resource"
         val ForbiddenResource: IRI = KnoraApiV2PrefixExpansion + "ForbiddenResource"
@@ -1016,7 +1017,7 @@ object OntologyConstants {
             KnoraBase.DateValue -> KnoraApiV2Simple.Date,
             KnoraBase.ColorValue -> KnoraApiV2Simple.Color,
             KnoraBase.GeomValue -> KnoraApiV2Simple.Geom,
-            KnoraBase.ListValue -> Xsd.String,
+            KnoraBase.ListValue -> KnoraApiV2Simple.ListNode,
             KnoraBase.IntervalValue -> KnoraApiV2Simple.Interval,
             KnoraBase.GeonameValue -> KnoraApiV2Simple.Geoname,
             KnoraBase.FileValue -> KnoraApiV2Simple.File,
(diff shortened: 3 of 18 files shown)


tobiasschweizer:
@benjamingeer In order to return a list node's label in a deterministic way for the simple schema, we could use the list responder and then choose the language based on the user's settings. 

Should we also return the label in the complex schema? 

If we do return the label in the simple but not in the complex schema, where would it be queried? To my understanding, schema conversions transform given data but do not query additional data.

benjamingeer:
> Should we also return the label in the complex schema?

I don't think so, it's like an ontology, the client can ask for it separately and cache it.

tobiasschweizer:
Ok, but where would it be queried? If we only need it for the simple schema, would the information be thrown away in case of the complex schema?

benjamingeer:
> Ok, but where would it be queried? If we only need it for the simple schema, would the information be thrown away in case of the complex schema?

I'm currently doing something like this for #1307: in the simple schema, there's no need to query standoff in the triplestore, because we never return it. I think this is a good optimisation!

Your case would be the opposite: getResourcePropertiesAndValues.sparql.txt could get the list node's labels, but only if the result is going to be returned in the simple schema. The SPARQL template could take a boolean parameter to tell it whether to query the labels. Then in `ConstructResponseUtilV2`, you could choose the label in the appropriate language.

tobiasschweizer:
I think we should use the list responder to get the node's label because it handles different languages correctly. Then, the existing logic that does not work in a deterministic way could be removed from the templates.

tobiasschweizer:
see #763

tobiasschweizer:
the client will e able to query and cache lists ( dhlab-basel/Knora-ui#170)

benjamingeer:
> I think we should use the list responder to get the node's label because it handles different languages correctly. Then, the existing logic that does not work in a deterministic way could be removed from the templates.

In that case, you can send a message to the list responder from `ConstructResponseUtilV2.createValueContentV2FromValueRdfData`. I'm doing something like this now to query standoff separately from text values.

tobiasschweizer:
Ok, but I only do that if the user requests information in the simple schema, right?

benjamingeer:
> Ok, but I only do that if the user requests information in the simple schema, right?

Yes. You could pass a boolean `queryListNodeLabels` to `ConstructResponseUtilV2.constructReadResourceV2`, which would pass it down to `createValueContentV2FromValueRdfData`. The responder would decide how to set this boolean parameter depending on the requested schema.

tobiasschweizer:
Ok, thx! 

Btw: will you be in Basle next week?

benjamingeer:
> Btw: will you be in Basle next week?

Yes, looking forward to it!

tobiasschweizer:
After discussion with @benjamingeer: to distinguish list values from text values in the simple schema, a custom data type for list values is added to the simple schema. So far, both text and list values had the datatype `xsd:string`.

This will simplify the handling of list values in Gravsearch in the simple schema. For the complex schema, no change is necessary.

tobiasschweizer:
@benjamingeer I added a custom data type in https://github.com/dhlab-basel/Knora/pull/1314/commits/aada49bb9507cd43211cbf82559ca624649bca9d Does it look good to you?

tobiasschweizer:
@benjamingeer 

In `org/knora/webapi/e2e/v2/ResourcesRouteV2E2ESpec.scala` there is a failing test:
> should create a resource with values *** FAILED *** (2 seconds, 88 milliseconds)
[info]   org.knora.webapi.AssertionException: Property http://0.0.0.0:3333/ontology/0001/anything/simple/v2#hasListItem requires an IRI referring to an instance of http://api.knora.org/ontology/knora-api/simple/v2#ListNode, but object content was received instead

I think it happens here:

https://github.com/dhlab-basel/Knora/blob/f0560788b13a9afaad0fb2cf4767a2637da98786/webapi/src/test/scala/org/knora/webapi/e2e/v2/ResourcesRouteV2E2ESpec.scala#L582-L586

I do not understand the role of the function pointer `doGetRequest`. Is there some test data I have to adapt?

benjamingeer:
> Does it look good to you?

Is the regular expression just there to forbid an empty string?

I would add:

```scala 
subClassOf = Some(OntologyConstants.Xsd.String),
```

> there is a failing test:

Fixed in 050dd94. `InstanceChecker` needs to know which types are custom datatypes. Sorry, I should have thought of this and told you before.

> I do not understand the role of the function pointer doGetRequest.

It's an abstraction to allow `InstanceChecker` to make requests to Knora in an E2E or R2R test, so it can use the ontology route to get entity definitions.

benjamingeer:
> I would add:
>
> ```scala 
> subClassOf = Some(OntologyConstants.Xsd.String),
> ```

Sorry, forget I said that, it's correct the way it is.

tobiasschweizer:
@benjamingeer Thanks a lot!

tobiasschweizer:
@benjamingeer All tests passed, thanks! I will now add some more tests for search e2e and improve the code a bit.

tobiasschweizer:
In the simple schema, only exact matches (equality) are allowed. 

"not equals" would not work correctly because the label could be available in several languages. For example, you could submit a query that gets all the nodes that do not match "tree list 02" but still get "Baumlistenknoten 02" which is not what you want. 

I think conceptually list nodes cannot be fully represented in the simple schema because we cannot enforce uniqueness using labels (Iris would be required for that).

benjamingeer:
> For example, you could submit a query that gets all the nodes that do not match "tree list 02" but still get "Baumlistenknoten 02" which is not what you want.

Would it be possible to say, I want all the nodes with labels in English that do not match "tree list 02"?

benjamingeer:
> I think conceptually list nodes cannot be fully represented in the simple schema because we cannot enforce uniqueness using labels (Iris would be required for that).

I guess you're right. Understanding list nodes means understanding Knora, which means using the complex schema.

tobiasschweizer:
> Would it be possible to say, I want all the nodes with labels in English that do not match "tree list 02"?

I think we could enforce the use of the lang function. But I am not sure if it is worth the effort.

tobiasschweizer:
Also, you could match an English string and still get the German label back if you preferred language is German (once  #1317 is done).

tobiasschweizer:
@benjamingeer I suggest that I now go on and open a PR that handles getting list node labels using the list responder (for the simple schema only).

benjamingeer:
Maybe it doesn't really make sense to search for list node labels in Gravsearch in the simple schema?

tobiasschweizer:
> Maybe it doesn't really make sense to search for list node labels in Gravsearch in the simple schema?

I think it's ok that you can search for labels now. I am also saying that because I spent several hours yesterday afternoon to make that possible ;-)

benjamingeer:
> I think it's ok that you can search for labels now. I am also saying that because I spent several hours yesterday afternoon to make that possible ;-)

OK! :) Please just make it clear in the documentation what the limitations of this feature are.

tobiasschweizer:
> Please just make it clear in the documentation what the limitations of this feature are.

doing that right away ...

tobiasschweizer:
done in https://github.com/dhlab-basel/Knora/pull/1314/commits/25a7ba03a283378c8f321194f5ce6057038b9295

benjamingeer:
Just one question: If I use the `lang` function in the simple schema with a list node label, will it work? Could you add a test for this?

tobiasschweizer:
I don't think it would work with the current design. I will look into it next week.

tobiasschweizer:
@benjamingeer 

A possible way to do this would be to support both the `str` and the `lang` function for type `ListNode` in the simple schema:

```sparql
PREFIX anything: <http://0.0.0.0:3333/ontology/0001/anything/simple/v2#>

CONSTRUCT {

?mainRes knora-api:isMainResource true .

?mainRes anything:hasListItem ?propVal0 .

} WHERE {

  ?mainRes anything:hasListItem ?propVal0 .
 
  FILTER(str(?propVal0) = "Tree list node 02"^^knora-api:ListNode && lang(?propVal0) = "en") 
}
```

However, this would require a change since the `lang` function requires an argument of type string. Also we already handle the `lang` function in a special way for text values (`handleLangFunctionCall`).

I think I would need your help to do this.

tobiasschweizer:
or maybe it would be possible like this without the `str`-function:

```
FILTER(?propVal0 = "Tree list node 02"^^knora-api:ListNode && lang(?propVal0) = "en") 
```

benjamingeer:
> this would require a change since the `lang` function requires an argument of type string

So basically, `handleLangFunctionCall` assumes its argument is a text value. It would also need to handle the case where its argument is a list value.

I don't know whether this is the right approach, I'm just thinking about the use case where you want to publish open linked data in the simple schema. What would be the most useful way to handle multilingual list nodes? We have the concept of the user's preferred language, but what if the user is `UnknownUser`? Would it make more sense to handle this by handling the `Accept-Language` HTTP header, instead of specifying a language in the Gravsearch query?

tobiasschweizer:
> Would it make more sense to handle this by handling the Accept-Language HTTP header, instead of specifying a language in the Gravsearch query?

I think the language you restrict the node label to should be the language you get the response in. It should be consistent. 

How should we handle this? It kind of blocks me since I am supposed to work on #1317 and your PR.

tobiasschweizer:
@benjamingeer Let's create an issue right away. Shall I do it?

benjamingeer:
Yes please!

tobiasschweizer:
ok. will do

benjamingeer:
Could you also send a brief message to the knora-user mailing list, summarising what users need to know about the changes in this PR, and also pointing them to the new issue about further work to be done?

tobiasschweizer:
issue created #1318

PR #1321, opened 2019-05-17
Handle List Nodes in Response

### LINKED ISSUE #1317: Return list node label using the list responder
Use the list responder to query list node labels (only for the simple schema), do not return the list node labels in the complex schema (requires a change in Knora-ui).

https://github.com/dhlab-basel/Knora-ui/pull/233 changes Knora-ui so it queries Knora's list route and caches a list.

tobiasschweizer:
This PR changes the handling of list nodes:

- [x] remove list node label from list value response in the complex schema 
- [x] remove getting list node label from `webapi/src/main/twirl/queries/sparql/v2/getResourcePropertiesAndValuesGraphDB.scala.txt`
- [x] remove getting list node label from main query in Gravsearch
- [x] adapt logic in `org/knora/webapi/util/ConstructResponseUtilV2.scala`: make the node label optional and request it from the list responder if information is requested in the simple schema
- [x] adapt tests
- [x] remove default values for `targetSchema` from messages
- [x] add tests for list value responses in the simple schema
- [x] adapt Knora-ui core so it fetches the information from the list route: dhlab-basel/Knora-ui#233 (ready to be reviewed)
- [x] adapt ontology and response test data in Knora-ui
 
closes #1317

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
@@ -67,7 +67,7 @@ sealed trait ResourcesResponderRequestV2 extends KnoraRequestV2 {
 case class ResourcesGetRequestV2(resourceIris: Seq[IRI],
                                  propertyIri: Option[SmartIri] = None,
                                  versionDate: Option[Instant] = None,
-                                 targetSchema: ApiV2Schema = ApiV2Complex,
+                                 targetSchema: ApiV2Schema,
                                  schemaOptions: Set[SchemaOption] = Set.empty,
                                  requestingUser: UserADM) extends ResourcesResponderRequestV2
 
@@ -78,7 +78,7 @@ case class ResourcesGetRequestV2(resourceIris: Seq[IRI],
   * @param targetSchema   the schema of the response.
   * @param requestingUser the user making the request.
   */
-case class ResourcesPreviewGetRequestV2(resourceIris: Seq[IRI], targetSchema: ApiV2Schema = ApiV2Complex, requestingUser: UserADM) extends ResourcesResponderRequestV2
+case class ResourcesPreviewGetRequestV2(resourceIris: Seq[IRI], targetSchema: ApiV2Schema, requestingUser: UserADM) extends ResourcesResponderRequestV2
 
 /**
   * Requests the version history of the values of a resource.
--- a/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
@@ -75,9 +75,10 @@ case class ResourcesGetRequestV2(resourceIris: Seq[IRI],
   * Requests a preview of one or more resources. A successful response will be a [[ReadResourcesSequenceV2]].
   *
   * @param resourceIris   the IRIs of the resources to obtain a preview for.
+  * @param targetSchema   the schema of the response.
   * @param requestingUser the user making the request.
   */
-case class ResourcesPreviewGetRequestV2(resourceIris: Seq[IRI], requestingUser: UserADM) extends ResourcesResponderRequestV2
+case class ResourcesPreviewGetRequestV2(resourceIris: Seq[IRI], targetSchema: ApiV2Schema = ApiV2Complex, requestingUser: UserADM) extends ResourcesResponderRequestV2
 
 /**
   * Requests the version history of the values of a resource.
--- a/docs/src/paradox/00-release-notes/next.md
+++ b/docs/src/paradox/00-release-notes/next.md
@@ -26,3 +26,5 @@ Also, please change the **HINT** to the appropriate level:
   Existing repositories must be updated; see `upgrade/1263-knora-admin` for instructions.
   
 - FEATURE: Add support for searching for specific list values in Gravsearch for both the simple and complex schema (@github[#1314](#1314)).  
+
+- REFACTOR: List value responses contain the list node's label in the simple schema only (@github[#1321](#1321))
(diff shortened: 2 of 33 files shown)


tobiasschweizer:
@benjamingeer It would be great if you could help me regenerating the tests data.

benjamingeer:
Which test data specifically? How does it need to change?

tobiasschweizer:
> Which test data specifically? How does it need to change?

I removed the node label from the complex response in https://github.com/dhlab-basel/Knora/pull/1321/commits/39e35549af73248c98caf108cdb7c8ca573cb1c9.

I think it is the complex ontology tests data and the responses in the complex schema involving list values. I expect the test to fail after this first commit. Can I come back to you once I know which tests have to be fixed?

benjamingeer:
To regenerate ontology response test data:

1. In `OntologyV2R2RSpec`, change `writeGetTestResponses` from `false` to `true`.
2. Run `OntologyV2R2RSpec`.
3. Change `writeGetTestResponses` back to `false`.

To regenerate resource response test data:

1. In `ResourcesRouteV2E2ESpec` and `SearchRouteV2R2RSpec`, change `writeTestDataFiles` from `false` to `true`.
2. Run `ResourcesRouteV2E2ESpec` and `SearchRouteV2R2RSpec`.
3. Change `writeTestDataFiles` back to `false`.

tobiasschweizer:
@benjamingeer Great, thanks! And I have to do the same for the Knora-ui test data.

benjamingeer:
You will notice that when you regenerate the test data in `OntologyV2R2RSpec`, it regenerates all the `.rdf` (RDF/XML) files, even for ontologies that didn't actually change. To see which ontologies actually changed, you can look at the diffs of the `.ttl`  and `.jsonld` files. Then you can just commit those files and the corresponding `.rdf` files, and discard the changes in the other `.rdf` files.

tobiasschweizer:
@benjamingeer Do you have a minute on FaceTime? I am looking for a convenient way to find out about the requested target schema in `ConstructResponseUtilV2 `.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
```diff
@@ -75,9 +75,10 @@ case class ResourcesGetRequestV2(resourceIris: Seq[IRI],
   * Requests a preview of one or more resources. A successful response will be a [[ReadResourcesSequenceV2]].
   *
   * @param resourceIris   the IRIs of the resources to obtain a preview for.
+  * @param targetSchema   the schema of the response.
   * @param requestingUser the user making the request.
   */
-case class ResourcesPreviewGetRequestV2(resourceIris: Seq[IRI], requestingUser: UserADM) extends ResourcesResponderRequestV2
+case class ResourcesPreviewGetRequestV2(resourceIris: Seq[IRI], targetSchema: ApiV2Schema = ApiV2Complex, requestingUser: UserADM) extends ResourcesResponderRequestV2
```
tobiasschweizer: @benjamingeer I was wondering if we should remove the default value for the target schema. 

With `ApiV2Complex` as the default schema, list node labels are not queried which could lead to problem when trying to serialize the information in the simple schema.

benjamingeer: Yes, now I try to avoid default values, they're often a source of bugs.

tobiasschweizer: >  list node labels are not queried 

if no schema passed by accident

tobiasschweizer: @benjamingeer I propose we could look a this next week.

benjamingeer: OK, but is there something difficult about it?

tobiasschweizer: The only difficulty I see is internal use (checking updates etc.).

benjamingeer: All you have to do is remove the default value, and set the parameter explicitly to `ApiV2Complex` wherever it was previously set by default. Then everything will continue to work as before.

benjamingeer: Now that I think about it, `ResourcesPreviewGetRequestV2` doesn't query values at all, so actually I think this problem can never occur, right?

tobiasschweizer: I think the notion of what a schema is is changed with this and your previous standoff PR because at query time you now have to get more or less information.

Before, I just thought of schemas as something that only the Route has to know about.

tobiasschweizer: > ResourcesPreviewGetRequestV2 doesn't query values

no values, no problems ;-)

benjamingeer: > I think the notion of what a schema is is changed with this and your previous standoff PR because at query time you now have to get more or less information.

OK, but I don't see how that matters. If you remove a default value from a method, and change all the code that calls the method so that it passes the same value explicitly, you will get the same result. It doesn't matter what the method does.

> no values, no problems ;-)

Yes.

benjamingeer:
Please send a message to the `knora-user` mailing list to alert users to breaking changes in this PR.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
tobiasschweizer: > All you have to do is remove the default value, and set the parameter explicitly to ApiV2Complex wherever it was previously set by default. Then everything will continue to work as before.

done in https://github.com/dhlab-basel/Knora/pull/1321/commits/586762554438343d79cbd69a854dfb87e001dab4

tobiasschweizer:
> Please send a message to the knora-user mailing list to alert users to breaking changes in this PR.

done just now

PR #1322, opened 2019-05-17
feat(api-v2): Make values citable

benjamingeer:
This PR aims to make it possible to cite Knora values, as per [this comment](https://github.com/dhlab-basel/Knora/issues/941#issuecomment-482460728). See the discussion in [Fine-grained citations](https://discuss.dasch.swiss/t/fine-grained-citations/24).

- [x] Add `knora-base:valueHasUUID` to `knora-base:Value`. This is stored only in the current version of each value.
- [x] Create and update value UUIDs when creating and updating values.
- [x] Read value UUIDs when reading values from the triplestore.
- [x] (refactor) move the functionality of `KnoraIdUtil` into `StringFormatter`.
- [x] Add a route for getting a resource along with a value specified by its UUID.
- [x] Give values ARK URLs in API responses.
- [x] Update tests.
  - [x] Update `.ttl` files.
  - [x] Update existing test code.
  - [x] Update expected responses.
  - [x] Add tests for citing values.
- [x] Add an upgrade script to add `knora-base:valueHasUUID` to existing repositories.
- [x] Update docs.

Existing repositories must be updated; see `upgrade/1322-cite-values/README.md`.

Resolves #941.

PR #1326, opened 2019-05-21
Leibniz ontology 

SepidehAlassi:
- adds ontology for BEOL-Leibniz connection


SepidehAlassi:
@tobiasschweizer for the next import this PR is needed.

tobiasschweizer:
@SepidehAlassi So this PR just adds the new ontology for Leibniz?

SepidehAlassi:
@tobiasschweizer yes

SepidehAlassi:
@tobiasschweizer I wanted to do the refactoring of the ontologies in this branch, but realised that would need some changes also in GUI which would take long. So for now, it is better to first import Leibniz data also to beol server, and take care of refactoring later.

tobiasschweizer:
got it, thanks for the clarification

SepidehAlassi:
@tobiasschweizer yes, you need this for the data on the switchdrive
Thanks for checking this PR, As soon as travis is done with the checks I will merge this.

tobiasschweizer:
Thanks. I will also convert the data according to https://github.com/dhlab-basel/Knora/pull/1322 and put the converted data back on switchdrive as a separate file.

tobiasschweizer:
@SepidehAlassi I think your are good to merge!

Issue #1246, opened 2019-03-05
Avoiding broken links when a cited resource is (marked as) deleted


benjamingeer:
Currently if you try to access a resource that’s been marked as deleted, Knora pretends it’s never heard of that resource. This would be annoying if a resource was cited (e.g. using an ARK URL in a publication), and then someone casually marked it as deleted. We also need to think about what happens when there’s a link within a Knora repository to a resource that’s been marked as deleted.

Things that might help:

- The user who marks the resource as deleted could optionally specify an alternative resource that should be returned instead (a kind of internal redirect). Knora could then update all links within the repository, to make them point to the alternative resource.
- If there’s no alternative resource, Knora could notify the project admins of the resources that have links to the deleted resource (e.g. by email), to let them know that something needs to be done about those links.

musicEnfanthen:
Another thing that might help:
- before deletion, specify a list of resources that point to the object to be deleted and present it to the user, so he can be sure that it is safe to delete (or not) and adjust the dependent resources

benjamingeer:
> before deletion, specify a list of resources that point to the object to be deleted

We could also do this, but it wouldn't show citations made with ARK URLs, since those could be anywhere on the Internet. Also, the user might not have permission to adjust the dependent resources, especially if they're in another project.

musicEnfanthen:
> marked as deleted, Knora pretends it’s never heard of that resource.

>it wouldn't show citations made with ARK URLs, since those could be anywhere on the Internet.

Yes, of course. So there possibly should be kind of a "This resource has been deleted" message (with date? with contact info?) as a very last fallback (after redirecting and informing) ?

> Also, the user might not have permission to adjust the dependent resources, especially if they're in another project.

Would it be feasable then to send an (automated?) notification to this other project (e.g. the admin) to inform about the deletion? But that is, indeed, your second point at the top.

benjamingeer:
> So there possibly should be kind of a "This resource has been deleted" message (with date? with contact info?) as a very last fallback (after redirecting and informing) ?

I guess there might be cases where it's really necessary to pretend that the resource never existed. But we could let the user decide this when they delete the resource. We already store the ID of the user who deleted the resource, the date when it was deleted, and an optional comment about why it was deleted. We just don't yet provide any way to access this information. We could let the user decide whether they want to make this information public or not.

Issue #1218, opened 2019-02-18
Query and copy deleted items


benjamingeer:
## Query deleted items

Get a list of deleted resources, using various search criteria:

- Deletion date
- Project
- User who deleted the resource
- User who created the resource

For each deleted resource, we only need to show metadata (not values):

  - Deletion date
  - Resource class
  - Label

Also, get a list of deleted values in a resource.

We don't want to make it possible to cite deleted resources or values.

## Copy deleted items

Make a copy of a deleted resource, including all its version history, with a new resource IRI and new value IRIs.

PR #1329, opened 2019-05-24
fix(sipi): Don't try to read a file value in a deleted resource

### LINKED ISSUE #1328: Incorrect error handling when a value in a deleted resource is requested
I am not sure if it is very dangerous but our log fills up with:
```
api_1_e9a49bba8d18 | org.knora.webapi.InconsistentTriplestoreDataException: entity http://rdfh.ch/0107/gjs0RTRkTZeYDT_37nbcCg/values/Dy-pWIujTIqdqKqrswwqfA has no project
```

I am not sure of the meaning of "has no project" but it is a `Value` and values have no `knora-base:attachedToProject`.

Is it something to worry about?

If not, is it hard to remove from the logs?

benjamingeer:
- [x] If Sipi requests information from Knora about a file in a deleted resource, return 404 Not Found.
- [x] Remove some obsolete things that should have been removed before.

Fixes #1328.

### DIFF
--- a/webapi/_test_data/all_data/anything-data.ttl
+++ b/webapi/_test_data/all_data/anything-data.ttl
@@ -1328,6 +1328,33 @@
   knora-base:hasPermissions "V knora-admin:UnknownUser|M knora-admin:ProjectMember";
   knora-base:attachedToUser <http://rdfh.ch/users/9XBCrDV3SRa7kS1WwynB4Q> .
 
+<http://rdfh.ch/0001/gjs0RTRkTZeYDT_37nbcCg/values/85et-o-STOmn2JcVqrGTCQ> a anything:ThingPicture ;
+  rdfs:label "A_100-1-A4_1_406 p. 405" ;
+  knora-base:isDeleted "true"^^xsd:boolean ;
+  knora-base:attachedToProject <http://rdfh.ch/projects/0001>;
+  knora-base:hasStillImageFileValue <http://rdfh.ch/0001/gjs0RTRkTZeYDT_37nbcCg/values/Dy-pWIujTIqdqKqrswwqfA> ;
+  knora-base:creationDate "2019-04-26T16:44:17.485Z"^^xsd:dateTime ;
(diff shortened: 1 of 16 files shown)


subotic:
Hmm, so when the file value is deleted, what happens to the image in sipi? We cannot delete the image in sipi, because if someone wants a past version of the resource where the file value was not deleted, then the image should be servable, right? Shouldn't then the `admin/files` route return `resource deleted` or something of sorts? There is no status code for this, unfortunately.

Ahh, a deleted resource. If we request the resource from a time before deletion, then the same question arises, right?

benjamingeer:
> We cannot delete the image in sipi, because if someone wants a past version of the resource where the file value was not deleted, then the image should be servable, right?

Yes. Deleting a value is considered a change in the resource (like deleting a word in a text file, or the contents of a cell in a spreadsheet), so you can request the resource as it was before that change.

> Shouldn't then the `admin/files` route return `resource deleted` or something of sorts? There is no status code for this, unfortunately.

Knora deals with deleted items by pretending that they never existed, and returning 404 Not Found, with a message saying "X was not found (it may have been deleted)". I believe we originally decided that this would be safest from the point of view of privacy and possible legal issues. There's a related discussion in #1246.

> Ahh, a deleted resource. If we request the resource from a time before deletion, then the same question arises, right?

Currently there's no way to request the contents of a deleted resource (not even a past version from the time before deletion), and I actually think it shouldn't be allowed. In macOS, if you want to view a file that's in the trash, you have to move it out of the trash first.

Screen Shot 2019-05-24 at 16 07 18

Otherwise, "deleted" wouldn't mean much.

Years ago I asked @lrosenth whether he thought there should be a way to undelete things, and he said no, but you should be able to copy the contents of a deleted resource into a new resource. There's an issue for this: #1218.

subotic:
thanks for the clarifications.

benjamingeer:
Thanks for the review!

benjamingeer:
Now I realise I need to make another PR for the case where you request an image from a past version of a resource...

benjamingeer:
#1331

PR #1333, opened 2019-05-28
fix(scripts): add admin ontology data to correct graph

subotic:


subotic:
@benjamingeer sorry, this is a hard one ;-)

subotic:
Thanks for the review :-)

PR #1337, opened 2019-06-03
feat(api-v2): No custom permissions higher than defaults

### LINKED ISSUE #1334: Weird use case on resource creation with permissions
Let say we have a group `Test` with `AdministrativePermission - ProjectResourceCreateRestrictedPermission` on resource `Person`.

Let say we have `DefaultObjectAccessPermission` `forGroup` `Test` = `D Creator`.

By this way we want to specify that users of group `Test` can *create* and *delete* a `Person` but cannot change the permissions.

The funny fact is when a user can *create* a resource with the v2 API, he can set the permissions he wants, so bypass the default `D Creator` rules!

Is that normal?

benjamingeer:
This PR ensures that when a user creates a resource/value, they can't provide custom permissions to give themselves a higher permission on the resource/value than they would get from its default permissions, unless they're a system admin or an admin of the containing project.

Resolves #1334.


benjamingeer:
TODO: This restriction should not apply to a `ProjectAdmin` or `SystemAdmin`.

benjamingeer:
@subotic Do you think this is OK now?

benjamingeer:
@subotic Would you have a minute to review this?

PR #1339, opened 2019-06-04
fix(sipi): Allow Sipi to get permissions from Knora for a past file version

### LINKED ISSUE #1331: Allow Sipi to get permissions from Knora for a past file version
If you request a past version of a file value from Knora, and then request the file from Sipi, Sipi asks Knora for the file permissions. Currently, Knora won't find the file value, because it's not the current version.

benjamingeer:
Fixes #1331.


benjamingeer:
Thanks for the review!

PR #1341, opened 2019-06-05
chore: Upgrade to new GraphDB version

### LINKED ISSUE #1098: Update GraphDB to 8.9

subotic:
## Failed tests:
- org.knora.webapi.responders.v2.OntologyResponderV2Spec
- org.knora.webapi.responders.v2.ResourcesResponderV2Spec
- org.knora.webapi.responders.v1.SearchResponderV1Spec
- org.knora.webapi.other.v1.DrawingsGodsV1E2ESpec
- org.knora.webapi.e2e.v2.SearchRouteV2R2RSpec
- org.knora.webapi.e2e.v2.OntologyV2R2RSpec

resolves #1098


benjamingeer:
On my machine, all the tests pass. Any idea why they fail on Travis?

subotic:
Tests pass locally (more or less) when GraphDB has 3GB of heap memory available (GDB_HEAP_SIZE). On Travis we had 2GB configured before and 3GB now, which is to much for the VM to run everything.

Also, locally just now, SearchRouteV2R2RSpec failed.

benjamingeer:
Locally I don't set `GDB_HEAP_SIZE`. I only set `GDB_JAVA_OPTS="-Xmx4G -Xss6M"`, and with GraphDB 8.10.0, all the tests pass for me.

subotic:
> Locally I don't set GDB_HEAP_SIZE

You really should do it. Knora should be able to run locally with minimal resources. Try to set `GDB_HEAP_SIZE` to 2GB and then run the tests.

benjamingeer:
> Knora should be able to run locally with minimal resources.

Why? Giving it that much memory locally hasn’t caused any problems for me as far as I can remember. Anyway, if there’s nothing we can do about GraphDB’s memory consumption, don’t we just have to accept it?

subotic:
> Why? Giving it that much memory locally hasn’t caused any problems for me as far as I can remember. 

I'm not concerned that you are going to have any problems but rather that it will run on your machine with a lot of memory and then it will break under load in production.

> don’t we just have to accept it?

Accepting is not a problem. Travis is. There the memory of the VM is limited to 7.5GB, which we cannot change.

benjamingeer:
OK, I understand. But 7.5GB of RAM seems very low to me. Even a MacBook has 8GB, and I wouldn't expect to run the DaASCH on a MacBook. I think we need to find a way to run our tests on a server with more memory. What about SWITCHengines, how much memory do we get there?

subotic:
It is planned to run all our tests on the machines we already have in the DHLab. This will take some time to setup.

Until then, we are stuck with 8.5.

benjamingeer:
OK. At least I can continue to test locally with the latest GraphDB. I'm just worried that if something changes in GraphDB that makes Knora stop working, we should find out about it sooner rather than later.

subotic:
GraphDB 9.0.0 is out. I will try it and see if it will maybe work on Travis.

subotic:
done in https://github.com/dasch-swiss/knora-api/pull/1435

PR #1348, opened 2019-06-12
feature(webapi): Add returning of fixed public user information

### LINKED ISSUE #1220: Minimal user's informations
Due to PR #1208, it seems to me that we don't have any more informations about any user.
Is there a way to get the username/givenName/familyName from a user's IRI (`knora-api:attachedToUser`) to simply print out in a UI who is the  Resource creator for example?

Maybe related to #1214 ? (probably same use case, nice to have the history but without minimal informations about users, doesn't make sense for UI)

n.b. : we can still display user's name in Salsah by using the v1 API on value which returns `valuecreatorname` and `valuecreator`.

subotic:
resolves #1220 for now


subotic:
Very simple implementation. The user itself and the SystemAdmin see everything. Everybody else sees only the public information.

subotic:
Docs are updated

subotic:
Thanks for the review :-)

gfoo:
works for me, thanks!

PR #1357, opened 2019-06-20
chore(sipi): Set subdir levels config to 0

subotic:


loicjaouen:
Will sipi handle the change automatically?

subotic:
> Will sipi handle the change automatically?

Yes :-)

subotic:
Thanks for the review :-)

PR #1359, opened 2019-06-22
chore(webapi): Fix logging

subotic:
Using Akka logging in non-actors does not work (for me at least). This PR adds plain Scala Logging to plain Scala objects and classes.


subotic:
sure. done.

subotic:
Thanks for the review :-)

PR #1361, opened 2019-06-27
fix(upgrade): Upload whole repository as Trig file during update

### LINKED ISSUE #1353: Handle dependencies between named graphs during repository updates
The simple solution: upload the whole repository as a single Trig file, so all named graphs are loaded in a single transaction.

benjamingeer:
Fixes #1353.


benjamingeer:
Great, thank you!

PR #1362, opened 2019-07-01
feat(triplestore): Update Lucene index on startup

### LINKED ISSUE #925: Manual incremental updating of the Lucene Full-Text-Index
As we don't use the GraphDB Lucene Connector, we need to manually update the Lucene Full-Text-Index.

We do this by running the following SPARQL query after each update:

```
PREFIX luc: <http://www.ontotext.com/owlim/lucene#>
INSERT DATA { luc:fullTextSearchIndex luc:updateIndex _:b1 . }
```

According to the [docs](http://graphdb.ontotext.com/documentation/standard/full-text-search.html#incremental-update) we should instead/also use the following SPARQL query:

```
PREFIX luc: <http://www.ontotext.com/owlim/lucene#>
INSERT DATA {  luc:fullTextSearchIndex luc:addToIndex <resourceIRI> . }
```

The first query adds all **new** resources to the index. If an update only **changes** an existing resource, then the index for this resource will not be updated. At least this is how I understand the documentation.

benjamingeer:
- [x] Rebuild the Lucene index on startup.
- [x] When a resource's label changes, update the index for that resource. Other data updates create new RDF subjects (because of versioning), so this should only be needed when resource labels change.

Resolves #925.


benjamingeer:
Thanks for the review!

PR #1367, opened 2019-07-02
refactor(api-v2): Use SparqlExtendedConstructResponse in ConstructResponseUtilV2

### LINKED ISSUE #763: refactor splitMainResourcesAndValueRdfData so it uses SparqlExtendedConstructResponse

benjamingeer:
This PR improves type safety in `ConstructResponseUtilV2` by using `SparqlExtendedConstructResponse`, which provides the types returned by the triplestore.

- [x] Change `SparqlExtendedConstructResponse` to use `SmartIri` for predicates.
- [x] Refactor `ConstructResponseUtilV2` and related classes.
- [x] Fix datatype typo in `v2/generateInsertStatementsForValueContent.scala.txt` (was `xsd:valueHasDecimal`, should be `xsd:decimal`).
- [x] Add upgrade script in case anyone was affected by that typo.
- [x] Add test.

Resolves #763.


benjamingeer:
@tobiasschweizer Issue #763 says to:

* Take the language tag into account. But we don't use language tags in data in the triplestore, do we?
* Not return a list node's label for a list value. This was done in #1321, right?

tobiasschweizer:
> Take the language tag into account. But we don't use language tags in data in the triplestore, do we?

No, I don't think so. We do have a language property for text values (ISO language code) but it's not the same that is used for labels etc. in the ontology.

> Not return a list node's label for a list value. This was done in #1321, right?

This was already done, yes.

tobiasschweizer:
@benjamingeer I think I can review this PR next week. Would that be ok?

benjamingeer:
> I think I can review this PR next week. Would that be ok?

Sure, no problem. Would you have time to look at #1259 with @andreas-aeschlimann this week?

tobiasschweizer:
> Would you have time to look at #1259 with @andreas-aeschlimann this week?

He said he would look at it. I will remind him :-)

benjamingeer:
> I will remind him :-)

Thanks!

tobiasschweizer:
The update script is needed in case someone created decimal values using v2, correct?

benjamingeer:
> The update script is needed in case someone created decimal values using v2, correct?

Yes.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -824,20 +824,23 @@ class SearchResponderV2(responderData: ResponderData) extends ResponderWithStand
...
             // collect the IRIs of main resources returned
             mainResourceIris: Set[IRI] = searchResourceByLabelResponse.statements.foldLeft(Set.empty[IRI]) {
-                case (acc: Set[IRI], (subjIri: IRI, assertions: Seq[(IRI, String)])) =>
-                    //statement.pred == OntologyConstants.KnoraBase.IsMainResource && statement.obj.toBoolean
-
+                case (acc: Set[IRI], (subject: SubjectV2, assertions: Map[SmartIri, Seq[LiteralV2]])) =>
                     // check if the assertions represent a main resource and include its IRI if so
-                    val subjectIsMainResource: Boolean = assertions.contains((OntologyConstants.Rdf.Type, OntologyConstants.KnoraBase.Resource)) && assertions.exists {
-                        case (pred, obj) =>
-                            pred == OntologyConstants.KnoraBase.IsMainResource && obj.toBoolean
```
tobiasschweizer: So with the type provided by the triplestore, no additional type conversion has to be performed, right?

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -824,20 +824,23 @@ class SearchResponderV2(responderData: ResponderData) extends ResponderWithStand
...
+                    val subjectIsMainResource: Boolean = assertions.get(OntologyConstants.KnoraBase.IsMainResource.toSmartIri) match {
+                        case Some(objs) => objs.contains(BooleanLiteralV2(true))
```
tobiasschweizer: Couldn't you just check for the first element with headOption? There should be only one statement `IsMainResource`.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
benjamingeer: I’m on the train but if you have IntelliJ add a type annotation for `obj`, it‘s a `LiteralV2`, right? Look at the implementation of its `toBoolean` method. If the `LiteralV2` is a `BooleanLiteralV2`, the conversion has already been done in `HttpTriplestoreConnector` when the Turtle was parsed.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
benjamingeer: Yes, you‘re right.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
benjamingeer: Sorry, got confused there, ignore my reply above. But yes, the conversions are done in `HttpTriplestoreConnector`, using the types provided by the triplestore.

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -735,55 +940,55 @@ object ConstructResponseUtilV2 {
             case OntologyConstants.KnoraBase.IntValue =>
                 FastFuture.successful(IntegerValueContentV2(
                     ontologySchema = InternalSchema,
-                    valueHasInteger = valueObject.assertions(OntologyConstants.KnoraBase.ValueHasInteger).toInt,
+                    valueHasInteger = valueObject.requireIntObject(OntologyConstants.KnoraBase.ValueHasInteger.toSmartIri),
```
tobiasschweizer: This is really a good improvement! I never liked doing these casts.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
tobiasschweizer: ok, no problem

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
tobiasschweizer: ok

benjamingeer:
> I think it would make sense to add a basic test case for ConstructResponseUtilV2. Do you think that is feasible?

I could refactor `HttpTriplestoreConnector` to make it possible to parse Turtle and construct a `SparqlExtendedConstructResponse` without actually connecting to the triplestore. Then I could add a test that took some hard-coded Turtle, parsed it, and ran it through `ConstructResponseUtilV2`. Would that be what you have in mind?

tobiasschweizer:
> Would that be what you have in mind?

Yes, no actual connection the triplestore is needed. Also we can get an idea how much time complex results need for processing (nested resources).

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
benjamingeer: Done in 3d26cf3.

benjamingeer:
> Yes, no actual connection the triplestore is needed.

Done in 3d26cf3.

> Also we can get an idea how much time complex results need for processing (nested resources).

We can't really see this now, because `ConstructResponseUtilV2` does additional SPARQL queries, e.g. to get project info.

PR #1368, opened 2019-07-02
feature(webapi):Add Redis cache

### LINKED ISSUE #1352: Deleted group still related and enable in user's reponse
I use `DELETE /admin/groups/<groupIri>` route to delete a temporary group related to all my users (for temporary import permissions).

Using the `GET /admin/groups/<groupIri>` route, the `status` is correctly set to `false`.

But if I ask for user's data (`GET admin/users/email/`), the group is still returned in the response (ok, could be discussed) but its `status` is `true`.

### LINKED ISSUE #1364: Add Redis Cache

subotic:
### Summary
- [x] add Redis manager with support for writing/reading `UserADM` and `ProjectADM`
- [x] add future instrumentation with Kamon allowing to time a whole future chain
- [x] add separate metrics logger, which can be turned on or off in `logback.xml`
- [x] change `UsersResponderADM` and `ProjectsResponderADM` to use Redis for caching
- [x] add checks to see if a user wants to change its username or email, that the new ones are not already taken
- [x] fix group member route to only allow access if project or system admin
- [x] fix delete project so that all members are removed from the deleted project.
- [x] Add group status change route (`/admin/groups/<groupIri>/status` and message. All members should be removed from the deleted group.
- [x] Update group documentation
- [x] Add Redis to documentation

### Work for another PR
- Split all UserADM, ProjectADM, and GroupADM in `simple` and `extended` variants. The simple variant will have only IRIs pointing to other admin objects, while the extended variant will return the dereferenced objects. Only simple variant will be cached.
- Add other admin objects to Redis cache (groups, permissions)
- opened #1378

### Issues
- resolves #1364
- resolves #1352

### BREAKING CHANGES
- [x] Moved `/admin/groups/members/<groupIri>` to `/admin/groups/<groupIri>/members`


subotic:
@benjamingeer Redis is f'ing fast:-) It takes 1-2ms to get a ProjectADM :-)

benjamingeer:
Great, let's cache everything in it. :)

subotic:
@kilchenmann @flavens FYI there is a small breaking change in the groups API.

subotic:
@benjamingeer I think this is enough for this PR. One thing led to another ...

benjamingeer:
I guess we could eliminate `CacheUtil` and cache everything in Redis?

benjamingeer:
It it possible to find out how fast Redis is compared to the in-memory cache used by `CacheUtil`?

subotic:
> It it possible to find out how fast Redis is compared to the in-memory cache used by `CacheUtil`?

The `CacheUtil` is faster. It is in the sub ms range. Writing and reading from Redis takes at least 5-10 ms.

benjamingeer:
> The CacheUtil is faster. It is in the sub ms range. Writing and reading from Redis takes at least 5-10 ms.

OK, then I guess we can keep it just for things that need to be very fast, like ontologies, and use Redis for everything else.

subotic:
> If our Scala case classes change, it won't be possible to deserialise any instances already stored in Redis.

At least for `UserADM` and `ProjectADM` this does not matter. The deserialization error will not be propagated and the `UsersResponderADM` and `ProjectsRespnderADM` will get a response back that the object was not found and write a fresh copy to the cache, thus overwriting the bad one. But yes, currently there is no versioning of the stored objects. I still need to create a cachable simplified version of the admin objects. Then by using Protobuff there could be a way to version cached objects. 

> Also, if the sysadmin stops Knora, removes something from the triplestore, then restarts Knora, there could be information still cached in Redis that should no longer be there. So I think that `ApplicationStateActor` should empty the Redis cache when Knora starts.

In those cases, the administrator could go and remove everything from the cache by hand. Redis has a CLI tool. Removing everything on each startup is, in my opinion, a bit too coarse. How often will this happen? In the future, I would like to expose a few admin commands over the API, so then the administrator could go into the Admin Interface and clear the cache when necessary.

subotic:
> Could you rename `RedisManager` to something like `CacheServiceManager`, and rename all the `Redis*` messages and `settings.useRedisCache` accordingly? Maybe someday we'll switch from Redis to some other caching service, but the rest of the application doesn't need to know about that.

yes, good idea.

benjamingeer:
> The deserialization error will not be propagated and the `UsersResponderADM` and `ProjectsRespnderADM` will get a response back that the object was not found and write a fresh copy to the cache, thus overwriting the bad one.

OK.

> But yes, currently there is no versioning of the stored objects.

I think it's not necessary, because the data can always be reloaded from the triplestore.

> In those cases, the administrator could go and remove everything from the cache by hand. Redis has a CLI tool.

OK.

> How often will this happen?

Only when upgrading Knora.

> In the future, I would like to expose a few admin commands over the API, so then the administrator could go into the Admin Interface and clear the cache when necessary.

OK.

subotic:
opened: #1378

subotic:
Thanks for the review :-)

PR #1369, opened 2019-07-04
feat(admin): Add admin route to dump project data

### LINKED ISSUE #1358: Add admin route to dump project data
Admin route, which returns all data (all ontology graphs, the data graph, project, user, permissions) for a certain project as a trig file.

benjamingeer:
- [x] Add a route, `/admin/projects/iri/<identifier>/AllData`, that returns a [TriG](https://www.w3.org/TR/trig/) file containing a project's ontologies, resource data, admin data, and permissions.
- [x] Add docs.
- [x] Add tests.

Resolves #1358.


benjamingeer:
It must be Akka setting the file extension, I didn't set it. I'll see if there's a way to specify one.

subotic:
Ah, ok. Thanks :-)

benjamingeer:
It's actually a Safari bug:

https://forums.macrumors.com/threads/safari-erroneously-adding-dms-extension-to-downloads.2080108/

I fixed it by adding a `Content-Disposition` header so the download filename is always `project-data.trig`.

subotic:
Ah cool. But this is an old bug/feature. I know that 15 years ago (give or take) I worked with Lukas on something in PHP and I remember that we had to add something like this (a memorable name for a header). It is time for Apple to fix it.

benjamingeer:
My pleasure, thanks for the review!

PR #1372, opened 2019-07-09
feat(api-v2): Change value permissions

### LINKED ISSUE #1231: Change value permissions
API v2 needs a route for changing the permissions on a value without changing its content.

benjamingeer:
Following the discussion in [Changing the permissions of a value](https://discuss.dasch.swiss/t/changing-the-permissions-of-a-value/45/15):

- [x] When retrieving a past version of a value, use the permissions from the current version.
- [x] Only store value permissions on the current version.
- [x] Add an upgrade plugin to update existing repositories accordingly.
- [x] Allow the value update route to accept requests to change only the permissions of a value.
- [x] Add tests.
- [x] Update docs.

Resolves #1231.


tobiasschweizer:
I will have to update some data (a dump in gitlab for the regions) but I cannot do it before next week. In case I need assistance, can I come back to you?

benjamingeer:
@tobiasschweizer Yes, I'll be available next week.

tobiasschweizer:
@benjamingeer Is it possible just to run the update for this PR (instead of `update-respository.py`)?

benjamingeer:
> Is it possible just to run the update for this PR (instead of `update-respository.py`)? 

Why do you want to do that? `update-respository.py` only applies the updates you need.

tobiasschweizer:
I cannot read the diff between the two dumps. In earlier version, I could do that.

benjamingeer:
Whether you run just one update or several, I think you'll have the same problem. The input files are formatted by GraphDB, and the output files are formatted by `rdflib`, so I wouldn't expect the formatting and order of triples to be exactly the same. You could write a little Python script using `rdflib` to reformat them.

tobiasschweizer:
I always dump the data from graphdb once they were loaded back. In my experience, graphdb dumps the data in a certain order and diffing was possible (you just have to use the same version of graphdb for both dumps).

benjamingeer:
> I cannot read the diff between the two dumps.

If you think there's a bug in the update script, could you please let me know specifically what you think the bug is? "I cannot read the diff between the two dumps" isn't a bug report. It would help if you gave me a (small) test case, showing what you expect and how it's different from what you got.

tobiasschweizer:
I am not saying that the update script does not work, I simply say that I cannot verify its results. 

Earlier, I made a diff to verify.

benjamingeer:
I see three possibilities:

1. GraphDB's output is not as predictable as you think
2. The diff algorithm is simply unable to make readable output in some cases
3. There is a bug in the update script

I think that using diff is not going to help us figure out which of these is the case.

Do you actually have any values with version history?

Could you do a CONSTRUCT query to look at some of your resources and see what changed?

tobiasschweizer:
give me a second, I am trying FileMerge to get a visual output. It didn't crash yet ;-)

tobiasschweizer:
Ok, after looking at it in FileMerge I can say:

- permissions were removed on some values. I checked one and it is pointed to by `knora-base:previousValue`, so this is correct. The permission was removed from a value that appears in the version history.
- `knora-base:valueHasUUID` appear in different places in the two dumps, so perhaps this is because of what you say in 1.

I was only able to see that with a graphical tool like FileMerge.

benjamingeer:
Great, thank you! Can you see whether anything was removed besides value permissions?

tobiasschweizer:
> Can you see whether anything was removed besides value permissions?

I think my brain.

benjamingeer:
Thank you!

PR #1392, opened 2019-08-13
feat(clientapi): Generate client code, part 2

### LINKED ISSUE #1390: Correct generated TypeScript client code
@andreas-aeschlimann Please list the necessary changes in this issue.

benjamingeer:
- [x] Refactor the calculation of relative file paths for importing a file into another file.
- [x] Fix problems in generated TypeScript code for the admin API.
- [x] For each data class (`User`, `Group`, etc.), generate:
    - [x] A base class without read-only properties (like related existing data objects) and without an ID.
    - [x] A `Stored*` subclass (`StoredUser`, etc.) with just the ID.
    - [x] A `Read*` subclass of `Stored*` (`ReadUser`, etc.) with the read-only properties. To avoid circular references, `Read*` classes point to `Stored*` classes, not to other `Read*` classes.
- [x] When `?mock=true` is added to the code generation route URL, include mock TypeScript library dependencies, so the generated code can be tested for compile errors without integrating it into `knora-api-js-lib` (thanks @tobiasschweizer!)
- [x] Add missing cardinality for `knora-admin:hasSelfJoinEnabled` to `knora-admin:UserGroup`.
- [x] Generate test requests and responses.
  - [x] Refactor existing E2E tests to share test requests with the code generation route.
  - [x] Get test responses from Knora when generating client code.
- [x] Add E2E test for the client code route.
- [x] Add more design docs.

Fixes #1390.


benjamingeer:
@andreas-aeschlimann I've tried to fix the problems you mentioned. Here is the result:

[typescript-client-api.zip](https://github.com/dhlab-basel/Knora/files/3501186/typescript-client-api.zip)

You said that the interfaces are needed to prevent circular references, but there are also circular references in the interfaces. For example, `IUser` imports `IProject`, and vice versa. Is this OK?

@tobiasschweizer Are you working on the TypeScript client library these days?

tobiasschweizer:
> Are you working on the TypeScript client library these days?

Yes, in this very moment.

benjamingeer:
@tobiasschweizer Would you be able to check this generated code? I'm working on TypeScript code generation in #1392.

tobiasschweizer:
I am working on v2, not on the admin API. But I could surely look at it together with @andreas-aeschlimann

benjamingeer:
That would be great, thank you. We also need some way to test it.

tobiasschweizer:
@benjamingeer Are you in Basel next week? I would like to discuss what could/needs to be done for client lib v2 endpoints  in Knora. 

I think the following points are important for v2:

1. Create interfaces/classes that describe the structure (structure of the ontology etc.) to convert serialized JSONLD into an internal structure in TypeScript, e.g., 
https://github.com/dhlab-basel/knora-api-js-lib/blob/924e0dd56e728b5d6440c0237aba1f9205b6cc1d/src/models/v2/ontologies/class-definition.ts#L118-L171 I created `class-definition.ts` manually, but I think this could be automized. I think this corresponds to what you have already done for the admin API in `models/admin`.

2. Create JSON-LD test data, so it can be imported into the client lib as a dependency and does not have to be maintained manually.

3. Description of the v2 endpoints: segments of the URL, parameters for GET / classes/interfaces of JSON-LD in body for POST, PUT, and the resulting response type. 

The following code is from typescript-client-api.zip:

```ts
export class GroupsEndpoint extends Endpoint {
    
    /**
     * Returns a list of all groups.
     */
    getGroups(): Observable<ApiResponseData<GroupsResponse> | ApiResponseError> {
    
        return this.httpGet("").pipe(
            map(ajaxResponse => ApiResponseData.fromAjaxResponse(ajaxResponse, GroupsResponse, this.jsonConvert)),
            catchError(error => this.handleError(error))
        );
    
    }

}
```


If I understand correctly, Knora combines the available endpoints to the expected response types. So the client library "knows" that it can ask for groups and gets a `GroupsResponse` back. 

Could this result be achieved by a formal description? The code snipped above contains code that is specific to the implementation with `rxjs` which could change when upgrading the `rxjs` version, but nothing wold have changed in Knora.

benjamingeer:
> Are you in Basel next week?

Yes.

> I would like to discuss what could/needs to be done for client lib v2 endpoints in Knora.

Let's open a new issue for this. This PR is just for correcting the existing problems with the generated admin API code.

tobiasschweizer:
> Let's open a new issue for this. This PR is just for correcting the existing problems with the generated admin API code.

Ok, let's do that next week then.

benjamingeer:
> If I understand correctly, Knora combines the available endpoints to the expected response types.

Knora's admin routes define the client endpoints, and include function definitions written in a target-language-independent Scala DSL. The current client code generation collects these definitions along with class definitions from the admin ontology, and passes them to a target-language-specific backend (currently only the TypeScript backend is implemented). The backend uses Twirl templates to generate TypeScript classes, interfaces, endpoint definitions containing those functions, etc.

The TypeScript functions are generated in a Twirl template, using the function definitions in the admin routes. For example, the `getGroups()` function is defined like this in the Scala DSL:

```scala
private val GroupsResponse = classRef(OntologyConstants.KnoraAdminV2.GroupsResponse.toSmartIri)

private val getGroupsFunction: ClientFunction =
    "getGroups" description "Returns a list of all groups." params() doThis {
        httpGet(BasePath)
    } returns GroupsResponse
```

The `GroupsResponse` class is thus taken from the admin ontology.

How you actually do an HTTP GET in the client code is up to you. If it needs to change, you can just change the Twirl template.

benjamingeer:
I think it doesn't make sense to have an abstract `AdminClass` in TypeScript. The admin classes and their properties are built into Knora and defined in the admin ontology. It's more convenient to have TypeScript classes that match the ones in the admin ontology, e.g. `User`, with members like `username`, rather than an abstract class with a `propertiesList`.

benjamingeer:
You can see how the TypeScript functions are generated in this template:

https://github.com/dhlab-basel/Knora/blob/6a2e7712d7099c7c3bcc201a7facfc931943b96e/webapi/src/main/twirl/clientapi/typescript/generateTypeScriptEndpoint.scala.txt

tobiasschweizer:
I could imagine a thing like this:

```ts
    convert<T>(httpResponse: Observable<AjaxResponse>,
               dataType?: { new(): T }): Observable<ApiResponseData<T> | ApiResponseError> {

        return httpResponse.pipe(
                map(ajaxResponse => ApiResponseData.fromAjaxResponse(ajaxResponse, dataType, this.jsonConvert)),
                catchError(error => this.handleError(error))
        );

    }

    /**
     * Gets a user by a property.
     */
    getUser(property: "iri" | "email" | "username",
            value: string): Observable<ApiResponseData<UserResponse> | ApiResponseError> {

        let userId = value;

        if (property === "iri") userId = encodeURIComponent(userId);

        return this.convert(this.httpGet("/" + property + "/" + userId), UserResponse);

    }

```

`convert` would have to be implemented in the client lib. Knora could create an abstract class `AbstractUsersEndpoint` with `abstract convert(...)`.

benjamingeer:
That’s totally up to you. But I don’t understand would you need an abstract base class with `abstract convert`. Couldn’t `convert` just be in `Endpoint` along with `httpGet`?

tobiasschweizer:
>  Couldn’t convert just be in Endpoint along with httpGet?

Yes, maybe you are right. I will try it out next week.

My point is just that Knora would create all the necessary information for interfacing (how to call a specific endpoint and the type of its response), but leave the implementation to the library. So when there is a change in the http lib that would only affect the client lib, and would not require a change in Knora. 

Probably that would just require a change in the template, and logic in Knora is already sufficient :-)

benjamingeer:
> Probably that would just require a change in the template, and logic in Knora is already sufficient

I think so. In addition to the templates, the TypeScript backend includes a class that runs the templates, which you can also customise.

tobiasschweizer:
> I think it doesn't make sense to have an abstract AdminClass in TypeScript. The admin classes and their properties are built into Knora and defined in the admin ontology. It's more convenient to have TypeScript classes that match the ones in the admin ontology, e.g. User, with members like username, rather than an abstract class with a propertiesList.

The design I made is still to be discussed. If it does not make sense, AdminClass can be removed. Let's discuss this next week.

benjamingeer:
TODO:

- Generate separate derived classes with IDs.
- Add `, true` to `@JsonProperty` if property is optional.
- For circular dependencies, make base class without the property that causes the circular reference, make derived class with only that property, and import the base class in the other class.

benjamingeer:
To generate the `Read*` classes, have each `ClientApi` provide a collection of class with cardinalities that are read-only, e.g. `id` and `members` in `Project`. For each class in this collection, the generator front end can then:

* Generate the `Read*` subclass, in which all properties have other `Read*` classes (if available) as object types.
* Remove the read-only properties from the base classes.

`ClientClassDefinition` needs to provide a set of direct base classes, and the logic that recursively gets class definitions needs to look at these, in case there is a `Read*` class whose base class isn’t used anywhere else.

The API function definitions that do GET operations need to be changed to have the `Read*` classes as return value types.

The TypeScript backend can then transform the class definitions to deal with circular imports. In practice, this should affect only the `Read*` classes.

The TypeScript class template needs to refer to (and include) the base classes specified in the `ClientClassDefinition`.

benjamingeer:
To avoid circular dependencies, add a class called `KnownProject` between, e.g. `Project` and `ReadProject`. A `KnownProject` has an ID, but it doesn't have any of the other read-only properties in `ReadProject`. Then `ReadUser` can refer to `KnownProject`.

benjamingeer:
@tobiasschweizer To sum up:

- The generated TypeScript code compiles, and has no circular references.
- The test data includes:
  - For `GET` requests: real Knora responses to admin API requests.
  - For `PUT` and `POST` requests: the same test requests that are used in the admin E2E tests.

So once this is merged, I think the next steps are:

- Add TypeScript tests that use the test data.
- Combine the generated code with `knora-api-js-lib` and test it with Knora.

tobiasschweizer:
@benjamingeer Thanks. I will look at it tomorrow.

tobiasschweizer:
I carried out the following steps:

- generated the client code with param `mock` set to true
- ran `npm i`
- ran `./node_modules/typescript/bin/tsc` without any errors
- from generated client code source dir `cp -r api/admin/ .../knora-api-js-lib/src/api/admin/` and `cp -r models/admin/ .../knora-api-js-lib/src/models/admin/`

Then I compiled knora-api-js-lib and got 11 errors.

After some time wondering wtf is different I thought of the compiler options defined in `tsconfig.ts`. I set them to what we have in

https://github.com/dhlab-basel/knora-api-js-lib/blob/c0f18432fe031fd00d90da20470061f05d83929e/tsconfig.json#L2-L18

and could reproduce the errors (using `"strictNullChecks": true`), e.g.

```
api/admin/users/users-endpoint.ts:144:49 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.

144         return this.httpPut("/iri/" + encodeURI(user.id) + "/BasicUserInformation", user).pipe(
                                                    ~~~~~~~

```

The problem is that id is optional and this was not detected because of the compiler options settings in the generated code's `tsconfig`. Once this is identical to what we use in the actual lib, it should behave the same.

benjamingeer:
OK, I see why this is happening: in the `Stored*` classes, the `id` property shouldn't be optional. Will fix.

tobiasschweizer:
that was quite some brain teaser :-) 

Once you adapt the compiler options, the E2E test should detect the problem which is great.

benjamingeer:
That fixed some of the compile errors, but there are a few that I don't know how to fix. For example, the TypeScript compiler doesn't like this:

```typescript
@JsonObject("GroupResponse")
export class GroupResponse {

    /**
     * A single group.
     */
    @JsonProperty("group", ReadGroup)
    group: ReadGroup = null;

}
```

It says:

```
models/admin/group-response.ts:15:5 - error TS2322: Type 'null' is not assignable to type 'ReadGroup'.

15     group: ReadGroup = null;
       ~~~~~
```

What should I do about that? The `group` property has a cardinality of 1, but it seems that `json2typescript` requires a default value. This seems like it might be a design flaw in `json2typescript`.

tobiasschweizer:
I think you would have to define the type as `group: ReadGroup | null = null;`, see https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types

benjamingeer:
In that case, should it be optional? Should it be like this?

```typescript
    @JsonProperty("group", ReadGroup, true)
    group?: ReadGroup | null = null;
```

benjamingeer:
Why can't the `GroupResponse` class have a constructor that guarantees that `group` will never be null?

tobiasschweizer:
No, I think it should not be optional, it is either a `ReadGroup` or `null`.

tobiasschweizer:
> Why can't the GroupResponse class have a constructor that guarantees that group will never be null?

As far as I have understood, `json2typescript` does not use a class's constructor, this is why there need to be the decorators. @andreas-aeschlimann Is this correct?

benjamingeer:
OK, please try it again now.

tobiasschweizer:
> The group property has a cardinality of 1, but it seems that json2typescript requires a default value. This seems like it might be a design flaw in json2typescript.

Sorry, maybe I misunderstood your question: is there an actual `null` in the response? 

If you want to make it optional, then the default value would be `undefined` and the question mark would have to be used. Also `true` has to be used as a param.

```
    @JsonProperty(Constants.Label, String, true)
    label?: string = undefined;
```

benjamingeer:
> Sorry, maybe I misunderstood your question: is there an actual `null` in the response? 

The response will always contain a `ReadGroup` object, never a `null`.

> If you want to make it optional

My question is how to deal with a property that's **not** optional. I think I don't understand the difference between `null` and `undefined` in TypeScript. To me, `null` means that something is not there. But that can only happen if it's optional. I don't want it to be optional, so it should not be possible for it to be null. It must always have a value. In strongly typed languages, there is usually a way for the compiler to enforce this.

tobiasschweizer:
I think then it all comes to setting a valid default value:

> Properties need to have a default value (or undefined), otherwise the mapper will not work

https://github.com/dhlab-basel/json2typescript#mapping-example

So then this would mean:

```ts
@JsonProperty("group", ReadGroup)
    group: ReadGroup  = new ReadGroup;
```

I think `undefined` does not work because of the strict null checks.

benjamingeer:
> I think then it all comes to setting a valid default value

In this case, the property isn't optional and there's no sensible default. That's why I think it would be better to require the user to provide a value (e.g. using a constructor). But I don't know how the mapper works.

tobiasschweizer:
@andreas-aeschlimann Is this correct? We need a default value for which is non optional with the strict null checking option.

```ts
@JsonProperty("group", ReadGroup)
    group: ReadGroup  = new ReadGroup();
```

tobiasschweizer:
I have spotted another problem in `ReadUser`:

```

Fatal error in JsonConvert. Failed to map the JSON object to the class "StoredProject" because of a type error.

	Class property: 
		description

	Expected type: 
		[string]

	JSON property: 
		description

	JSON type: 
		[object]

	JSON value: 
		[{"value":"Anything Project"}]

	Reason: Given object does not match the expected primitive type.
```

tobiasschweizer:
Could you change the method `getUser` of the users endpoint:

```ts
getUser(property: "iri" | "email" | "username", value: string): Observable<ApiResponseData<UserResponse> | ApiResponseError> {

        let userId = value;

        if (property === "iri") userId = encodeURIComponent(userId);

        return this.httpGet("/" + encodeURIComponent(property) + "/" + userId).pipe(
            map(ajaxResponse => ApiResponseData.fromAjaxResponse(ajaxResponse, UserResponse, this.jsonConvert)),
            catchError(error => {console.log(error); return this.handleError(error)})
        );
    
    }
```

I think for URL segments, `encodeURIComponent` should be used.

benjamingeer:
> Could you change the method getUser of the users endpoint:

The DSL currently doesn't support `if` statements, and I don't see any way to add something like this generically:

```
        if (property === "iri") userId = encodeURIComponent(userId);
```

We can't put things like `encodeURIComponent` in the function DSL, because how you encode URIs is dependent on your HTTP library. It has to go in the TypeScript template, which means it has to be generated in a consistent way for all URI components.

The call to `encodeURI` is currently generated for any URL segment that isn't a string literal. Should I use `encodeURIComponent` instead?

tobiasschweizer:
https://github.com/dhlab-basel/Knora-ui/tree/master/projects/knora/core/src/lib/declarations/api/admin

tobiasschweizer:
> Fatal error in JsonConvert. Failed to map the JSON object to the class "StoredGroup" because of a type error.

	Class property: 
		description

	Expected type: 
		StringLiteral

	JSON property: 
		description

	JSON type: 
		string

	JSON value: 
		"A group for thing searchers."


in src/models/admin/group.ts

I had to change:

```ts
    @JsonProperty("description", String, true)
    description?: string = undefined;
```

tobiasschweizer:
I could integrate the generated code here https://github.com/dhlab-basel/knora-api-js-lib/pull/16

I just had to manually adapt group.ts, see https://github.com/dhlab-basel/Knora/pull/1392#issuecomment-525825177

benjamingeer:
So `description` is a string in `Group` but an object in `Project`? @subotic is there a reason why they're not consistent?

benjamingeer:
@tobiasschweizer I think I've fixed that, please try it again.

tobiasschweizer:
ok,  will do it today. I added a script in the client-lib that automizes the copying of generated files.

tobiasschweizer:
@benjamingeer I can confirm that it works now with existing specs for the user endpoint.

Does this PR change anything in the admin API (braking existing client libs)?

benjamingeer:
> I can confirm that it works now with existing specs for the user endpoint.

Great, thank you.

> Does this PR change anything in the admin API (braking existing client libs)?

No, nothing.

Are the docs OK?

https://github.com/dhlab-basel/Knora/blob/e99d13bed7e0bb32fcd25e1222c2169e9b7ca13d/docs/src/paradox/05-internals/design/client-api/index.md

https://github.com/dhlab-basel/Knora/blob/e99d13bed7e0bb32fcd25e1222c2169e9b7ca13d/docs/src/paradox/05-internals/development/generating-client-apis.md

tobiasschweizer:
I generated and read the docs yesterday. I think they are fine.

I will handle the integration logic in the client lib itself, see https://github.com/dhlab-basel/knora-api-js-lib/pull/16/files#diff-8e21c98b609b59b6b605875e8ed27329

I will approve this PR now. Next week, we can add more specs to the client lib for the other endpoints. Any required changes will be done in a follow-up PR so this PR is not blocked from being merged.

benjamingeer:
Thanks for reviewing! Looking forward to working on v2, too.

tobiasschweizer:
I am working on parsing resources. I can show you the existing code next week, so we can think about how to do this for v2 and to what extent.

PR #1402, opened 2019-08-27
build: enable testing of docs generation

### LINKED ISSUE #1401: Enable testing docs generation in Travis

subotic:
resolves #1401


subotic:
@benjamingeer could you please make a quick review?

subotic:
> It's cumbersome to have to type (or copy and paste) a 200-character command starting with `docker run` to compile something (whether it's code or documentation). Could we instead always have a Makefile that does this for us, so we could just type `make docs`?

This is only temporary. I don't want to waste to much time on this. In the future, I would like to move from paradox/jekyll/sbt to MkDocs (python based). This is how the docs for `docs.dasch.swiss` are generated. Then it will be just a short command again.

benjamingeer:
> This is only temporary.

Temporary solutions often last a long time.

> I don't want to waste to much time on this.

I don't want to waste time typing/pasting a 200-character command 500 times a day to build things. I have the same problem with building Docker in Sipi.

> In the future, I would like to move from paradox/jekyll/sbt to MkDocs (python based).

That sounds like it will take several months.

subotic:
Yes, I know that the current situation is suboptimal (docs and sipi). I will try to make the transition of the docs as fast as possible and find a fix for sipi soon. If I see that it takes longer, then I will add a simplification for running the docker commands for the docs and sipi.

Could you please nevertheless approve this PR?

subotic:
Thanks!

PR #1415, opened 2019-08-30
 chore: change gren config

subotic:


subotic:
thanks for the review :-)

PR #1417, opened 2019-09-02
ontology refactors

SepidehAlassi:
- [x]  add citation to Leibniz letters

- [x] add topics to med entries


SepidehAlassi:
@tobiasschweizer yes

Issue #1078, opened 2018-11-27
Create a file value using a file that is already in Sipi's permanent storage


benjamingeer:
Use case: bulk import has been done, and needs to be redone because data has changed, but images have not changed.

PR #1429, opened 2019-09-04
feat(api-v2): Erase a resource

### LINKED ISSUE #1421: Physically deleting a resource
Sometimes it will necessary that a project administrator is able to _really_ delete a resource completely (physically). May be a flag can be added to the delete route.
If the resource is references by another resource, the delete should fail.

benjamingeer:
This PR introduces a route `/v2/resources/erase`, for erasing a resource from the triplestore.

- [x] Add route.
- [x] Add implementation in `ResourcesResponderV2`.
- [x] Add tests in `ResourcesResponderV2Spec`.
- [x] Add test in `ResourcesRouteV2E2ESpec`.
- [x] Add docs.

Resolves #1421.


tobiasschweizer:
Have you included https://en.wikipedia.org/wiki/Eraser_(film) in the docs? :-)

benjamingeer:
@tobiasschweizer Sorry, I only include references to films I've seen. :)

tobiasschweizer:
What happens if the resource (e..g, a `StillImageRepresentation`) has an attached file that is stored in Sipi?

benjamingeer:
> What happens if the resource (e..g, a `StillImageRepresentation`) has an attached file that is stored in Sipi? 

Currently Knora never deletes files from Sipi, and I think this is still the safest policy, because of use cases like #1078.

benjamingeer:
(from Sipi's permanent storage I mean)

webapi/src/test/scala/org/knora/webapi/e2e/v2/ResourcesRouteV2E2ESpec.scala
```diff
@@ -956,6 +956,38 @@ class ResourcesRouteV2E2ESpec extends E2ESpec(ResourcesRouteV2E2ESpec.config) {
                 )
            }
         }
+
+        "erase a resource" in {
```
tobiasschweizer: Could you add a test attempting to erase a resource that is referred to to make sure the request is rejected?

webapi/src/main/twirl/queries/sparql/v2/isEntityUsed.scala.txt
```diff
@@ -20,17 +20,19 @@
...
- * Checks whether an ontology entity is used (i.e. is the object
- * of any statements).
+ * Checks whether an entity is used (i.e. is the object of any statements).
  *
  * @param triplestore the name of the triplestore being used.
- * @param entityIri the IRI of the ontology.
+ * @param entityIri the IRI of the entity.
  * @param ignoreKnoraConstraints if true, knora-base:subjectClassConstraint and knora-base:objectClassConstraint will be ignored.
  * This is necessary when modifying the cardinalities of a class.
+ * @param ignoreRdfSubjectAndObject if true, rdf:subject and rdf:object will be ignored. This is necessary when checking
```
tobiasschweizer: So this ignores reifications (link values) referring to the IRI of the resource to be deleted if set to true? 

Wouldn't this mean that  a resource is referenced and hence cannot be deleted?

tobiasschweizer:
Since this is a killer feature in the proper sense, I would like to discuss this together with @benjamingeer and @lrosenth.

webapi/src/main/twirl/queries/sparql/v2/isEntityUsed.scala.txt
tobiasschweizer: I would expect that the IRI can be used as a rdf:subject (outgoing link) but not as an rdf:object.

webapi/src/test/scala/org/knora/webapi/e2e/v2/ResourcesRouteV2E2ESpec.scala
benjamingeer: There's a test for that in `ResourcesResponderV2Spec`:

https://github.com/dhlab-basel/Knora/pull/1429/files#diff-b1ef5075da04ef708a48de2a6b007436R2062

subotic:
We could always create placeholder resources for deleted resources with incoming links.

webapi/src/main/twirl/queries/sparql/v2/isEntityUsed.scala.txt
benjamingeer: Suppose we want to erase resource `A`, and resource `B` has a link value that points to `A`, and the link value is marked as deleted. The link value will contain an `rdf:object` pointing to `A`. It's OK to ignore it, because `B` no longer refers to `A`.

benjamingeer:
> We could always create placeholder resources for deleted resources with incoming links.

The use case that @lrosenth had in mind is that he's creating a new project by doing an import, and he wants to remove some incorrect resources and import them again.

subotic:
I know, but what if project A creates a resource and project B links to it. So now project A wants to delete the resource. Are we not going to allow this? The data belongs to project A and they should be able to do what they want IMHO.

benjamingeer:
> I know, but what if project A creates a resource and project B links to it. So now project A wants to delete the resource. Are we not going to allow this? The data belongs to project A and they should be able to do what they want IMHO.

Currently project A can mark the resource as deleted, but they can't erase it. I guess in this case, if they want to erase it, it would make sense to replace it with a dummy resource. @lrosenth ?

benjamingeer:
Although the problem would be that the placeholder resource would have to be of the correct class, which could have required cardinalities for links to other resources. So we could end up having to create a whole graph of dummy resources just to erase one resource. This is the same problem as the bulk import problem.

webapi/src/test/scala/org/knora/webapi/e2e/v2/ResourcesRouteV2E2ESpec.scala
tobiasschweizer: Great. Sorry, I missed this one.

webapi/src/main/twirl/queries/sparql/v2/isEntityUsed.scala.txt
tobiasschweizer: But the link value would have an `rdf:object` with the IRI of `A` that does not occur anymore as a subject in the triplestore, right?

benjamingeer:
Since erasing resources in production is supposed to be something very rare, and can only be done by a sysadmin or project admin, maybe it would be best to leave it up to the admin to decide how to deal with this scenario. If they want to create a graph of dummy resources and change existing links first before erasing, they could do that via the API.

lrosenth:
OK – I would suggest the following rules:

1. if a resource is **not** referenced by another resource: **erase it completely** (including associated SIPI-Files!)

2. if the resource is referenced by another resource: You cannot erase it at all! The user has first to remove the links to it (that is, the links from other resources to the one to be deleted have to be marked as deleted).

3. if a resource has only "incoming" links marked as deleted to it, these (already marked as deleted) links have to be removed/erased also completely. If it's simpler to "downgrad" the resource to a "dummy resource" that sais "I?ve been deleted"  that would be a solution too – whatever is easier to implement!

Case (3) is not as urgent as (1) and (2)

Erasing a resource is always a "last resort action" that needs admin privileges

benjamingeer:
@lrosenth 

> 1. if a resource is not referenced by another resource: erase it completely (including associated SIPI-Files!)

We have a use case, #1078, where the same Sipi files could be reused in different resources: "Use case: bulk import has been done, and needs to be redone because data has changed, but images have not changed."

That seems similar to your use case here. Then you would want to keep the Sipi files, right?

Also, when you mark a resource or value as deleted, Knora doesn't remove its Sipi files. Should it?

> 3. if a resource has only "incoming" links marked as deleted to it, these (already marked as deleted) links have to be removed/erased also completely.

That seems feasible.

webapi/src/main/twirl/queries/sparql/v2/isEntityUsed.scala.txt
benjamingeer: @tobiasschweizer 

> But the link value would have an `rdf:object` with the IRI of `A` that does not occur anymore as a subject in the triplestore, right?

That's correct. This wouldn't be a problem for Knora or for the GraphDB consistency checking. @lrosenth are you saying that these (marked-as-deleted) references would need to be erased for legal reasons?

tobiasschweizer:
> Currently Knora never deletes files from Sipi, and I think this is still the safest policy, because of use cases like #1078.

Once a resource with an attached file value gets erased/deleted, will the file in Sipi then be unaccessible because no permissions will be granted for an erased/deleted resource?

In other words, the file would still be in Sipi, but it would never be served.

benjamingeer:
> In other words, the file would still be in Sipi, but it would never be served.

It would never be served, but it could be used again to create another resource or value, which is the use case in #1078:

> Use case: bulk import has been done, and needs to be redone because data has changed, but images have not changed.

tobiasschweizer:
> It would never be served, but it could be used again to create another resource or value, which is the use case in #1078:

Which permissions would be required to use an existing file? If the original resource was erased, the cannot be determined. I am thinking of a copyright protected information that required the erasure of a resource.

benjamingeer:
> Which permissions would be required to use an existing file? If the original resource was erased, the original permissions cannot be determined. I am thinking of a copyright protected information that required the erasure of a resource.

To reuse an existing file, you'd have to move it to Sipi's `tmp` directory first. Then you can create a new resource as if you had just uploaded the file to Sipi. No permissions will be checked.

If you don't want to reuse the file, you can delete it from Sipi's storage using `rm`.

tobiasschweizer:
> To reuse an existing file, you'd have to move it to Sipi's tmp directory first. Then you can create a new resource as if you had just uploaded the file to Sipi. No permissions will be checked.

> If you don't want to reuse the file, you can delete it from Sipi's storage using rm.

So this would require write access to Sipi's file system. I think that's fine since you would need admin privileges to use the functionality added with this PR.

tobiasschweizer:
This PR looks fine to me. I leave the final approval to @lrosenth

benjamingeer:
> This PR looks fine to me. I leave the final approval to @lrosenth 

Thanks. I'm currently writing the code to erase marked-as-deleted link values.

tobiasschweizer:
You are the Eraser ...

benjamingeer:
@lrosenth Is this OK now? It erases deleted link values pointing to the resource that's being erased. For the reasons explained above, it doesn't touch files in Sipi. Does that seem reasonable to you?

lrosenth:
Seems very reasonable to me! Maybe at some point we need a "Purge" algorithm whoiich detect unreferenced SIPI files and allows the to be deleted... But that's not urgent!

benjamingeer:
Thanks for the review!

PR #1433, opened 2019-09-10
test(api-v2): Test an ontology with only one entity

### LINKED ISSUE #1428: Test JSON-LD rendering of an ontology containing only one entity
In response to a question from @tobiasschweizer .

benjamingeer:
- [x] Add `minimal-onto.ttl` and test how it's served in JSON-LD (always with `@graph`).
- [x] When updating test response files in `OntologyV2R2RSpec`, don't update RDF/XML files if only their blank node IDs have changed (makes it easier to check the really changed files into git).
- [x] Remove a generated client code file from git (this was causing confusing random errors in the client code test).

Resolves #1428.


tobiasschweizer:
Thanks! I will try that JSOn-LD in knora-api-lib-js.

tobiasschweizer:
I have added this in https://github.com/dhlab-basel/knora-api-js-lib/pull/28

benjamingeer:
> So  "@graph" is also an array if it contains just one element?

Yes, Knora always inserts `@graph` in an ontology response, and the JSON-LD processor will keep the `@graph` if there is something else at the top level (you can see this behaviour in the JSON-LD Playground). The JSON-LD processor moves the contents of `@graph` to the top level only if the graph contains one node and document contains nothing else.

benjamingeer:
Thanks!

tobiasschweizer:
Thank you, you are a true minimalist!

benjamingeer:
https://www.youtube.com/watch?v=doJk4yPwJDk

tobiasschweizer:
I know Steve Reich :-)

PR #1447, opened 2019-09-24
feat (admin): Add edit list item route

### LINKED ISSUE #1387: Edit list item
At the moment, only the information about the list root node (which represents the whole list) can be changed. Updating child nodes must be implemented.

### LINKED ISSUE #1520: Updating the name of a list
`CreateListApiRequestADM` includes `name`, but `ChangeListInfoApiRequestADM` does not, is this correct?

### LINKED ISSUE #1528: Redundant list routes?
I think these two routes are the same:

* `PUT /admin/lists/IRI` https://github.com/dasch-swiss/knora-api/blob/develop/webapi/src/main/scala/org/knora/webapi/routing/admin/ListsRouteADM.scala#L141
* `PUT /admin/lists/infos/IRI` https://github.com/dasch-swiss/knora-api/blob/develop/webapi/src/main/scala/org/knora/webapi/routing/admin/ListsRouteADM.scala#L235

Is that correct? If so, the client API only needs to implement one of them.

subotic:
Resolves #1387 
Resolves #1528
Resolves #1520


benjamingeer:
@LukasStoeckli With the current client code generation framework, each endpoint has to have a single base URL path for all its routes. For the lists endpoint, this is currently `admin/lists`. In this PR you have some list routes that use `admin/lists` and others that use `admin/nodes`. Can you change `admin/nodes` to `admin/lists/nodes`, so that all the list routes start with `admin/lists`?

LukasStoeckli:
sure

tobiasschweizer:
if you need my help integrating this PR into knora-api-s-lib, please let me know. Ideally, this would be automated, see #1554

benjamingeer:
> if you need my help integrating this PR into knora-api-s-lib, please let me know.

@tobiasschweizer Yes please. The semantics of the list update API have changed, so some of the existing client API functions and test data had to be replaced.

tobiasschweizer:
What has changed exactly and how could this affect existing code?

benjamingeer:
> What has changed exactly and how could this affect existing code?

@tobiasschweizer The client function `updateListInfo` has been removed, and the following client functions have been added:

- `updateListName`
- `updateListLabels`
- `updateListComments`
- `updateListNodeName`
- `updateListNodeLabels`
- `updateListNodeComments`

Each of the new functions has new test data.

tobiasschweizer:
So this means that the test for `updateListInfo` has to be removed and tests for the new methods have to be added using the new test data?

This may also affect @kilchenmann and @flavens since the admin method are already in use.

benjamingeer:
> So this means that the test for updateListInfo has to be removed and tests for the new methods have to be added using the new test data?

That's right.

> This may also affect @kilchenmann and @flavens since the admin method are already in use.

Yes. I had understood that the existing functionality wouldn't be changed, but it looks like that wasn't feasible.

LukasStoeckli:
> Yes. I had understood that the existing functionality wouldn't be changed, but it looks like that wasn't feasible.

What functionality exactly?

flavens:
@benjamingeer @LukasStoeckli @tobiasschweizer we do not need to edit anything in the ui modules or lib. All the methods come from the knora-api-js-lib. If Tobias does change the name of the methods, it is all good for us on this point!

flavens:
@benjamingeer @LukasStoeckli will this PR be part of the next release?

benjamingeer:
@flavens We can certainly wait if you need more time.

flavens:
> @flavens We can certainly wait if you need more time.

That would be great! We will do it in February to be part of R2020-02.

SepidehAlassi:
The issues addressed in this PR are implemented in #1753.

PR #1459, opened 2019-10-02
chore: Upgrade sipi to 2.0.1

### LINKED ISSUE #1457: Upgrade sipi to 2.0.1

subotic:
resolves #1457

### DIFF
--- a/KnoraBuild.sbt
+++ b/KnoraBuild.sbt
@@ -283,7 +283,7 @@ lazy val knoraSipi = knoraModule("knora-sipi")
       Docker / dockerExposedPorts ++= Seq(1024),
       Docker / dockerCommands := Seq(
           // FIXME: Someday find out how to reference here Dependencies.Versions.sipiImage
-          Cmd("FROM", "dhlabbasel/sipi:v1.4.3"),
+          Cmd("FROM", "dhlabbasel/sipi:v2.0.1"),
           Cmd("LABEL", s"""MAINTAINER="${maintainer.value}""""),
           Cmd("ADD", "opt/docker/scripts", "/sipi/scripts"),
       )
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -49,7 +49,7 @@ object Dependencies {
         akkaHttpVersion := "10.1.7",
         jenaVersion := "3.4.0",
         metricsVersion := "4.0.1",
-        sipiImage := "dhlabbasel/sipi:v1.4.3",
+        sipiImage := "dhlabbasel/sipi:v2.0.1",
         gdbSEImage := "ontotext/graphdb:8.5.0-se",
         gdbFreeImage := "dhlabbasel/graphdb:8.10.0-free"
     )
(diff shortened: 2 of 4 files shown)


subotic:
@benjamingeer This PR is now ready for review. I had to disable a test because I couldn't make it dynamically configurable. The URL of the IIIF images is hardcoded in a JSON file and doesn't correspond to the setup, i.e., it should be `http://sipi:1024/...` instead of `http://0.0.0.0:1024/...`.

Makefile
```diff
@@ -0,0 +1,353 @@
...
+
+include vars.mk
+
+#################################
+# Documentation targets
+#################################
+
+.PHONY: publish-docs
+docs-publish: ## build and publish docs
+	docker run --rm -it -v $(PWD):/knora -v $(HOME)/.ivy2:/root/.ivy2 -v $HOME/.ssh:/root/.ssh sbt-paradox /bin/sh -c "cd /knora && git config --global user.email $(GIT_EMAIL) && sbt docs/ghpagesPushSite"
```
benjamingeer: You have `$(HOME)` and `$HOME`, is that correct?

subotic:
> > The URL of the IIIF images is hardcoded in a JSON file
> 
> Where is that file?

in `ResourcesResponderV1SpecContextData` line 43: `src/test/resources/test-data/v1/expectedBookContextResponse.json`

Makefile
subotic: I don't think that it matters, but I've added parentheses. Now both should be resolved by `make`.

benjamingeer:
> in ResourcesResponderV1SpecContextData line 43: src/test/resources/test-data/v1/expectedBookContextResponse.json

OK, I understand now. But that's an important test, we shouldn't disable it. You could turn the JSON file into a template, e.g. replace all `0.0.0.0:1024` with `SIPI_HOST`. Then in the test, when the file is loaded, replace `SIPI_HOST` with the configured value.

subotic:
Would be great :-)

benjamingeer:
I mean, just use `String.replaceAll()`, it will just be one line of code.

subotic:
> I mean, just use `String.replaceAll()`, it will just be one line of code.

done. added the test back.

subotic:
@benjamingeer could you take a quick look?

benjamingeer:
Looks OK but something isn't working in the build, and I keep getting emails from Codacy saying that there are issues.

subotic:
yes, I've renamed some makefile targets and forgot to update `.travis.yml`. Doing it now.

subotic:
should be fixed now.

Regarding Codacy, I don't have access to the 'dasch-swiss' organization.

benjamingeer:
> Regarding Codacy, I don't have access to the 'dasch-swiss' organization.

I don't understand much about Codacy, but for me it displays this PR under `dhlab-basel`:

Screenshot 2019-10-17 at 14 01 40

subotic:
ah, makes sense. I will change it.

subotic:
your welcome and thanks for the review :-)

PR #1478, opened 2019-10-17
fix(clientapi): Fix generated TypeScript admin client code to correspond to test data

### LINKED ISSUE #1476: Admin client code generation
A `StoredUser` has the following definition (https://github.com/dasch-swiss/knora-api-js-lib/blob/master/src/models/admin/stored-user.ts):

```
/**
 * Represents a Knora user.
 */
@JsonObject("StoredUser")
export class StoredUser extends User {

    /**
     * The ID of the enclosing object.
     */
    @JsonProperty("id", String)
    id: string = "";

}
```

The expected payload for `updateUserBasicInformation` looks like this (https://github.com/dasch-swiss/knora-api-js-lib/blob/master/test/data/api/admin/users/update-user-request.json):

```
{
    "username": "donald.big.duck",
    "email": "donald.big.duck@example.org",
    "givenName": "Big Donald",
    "familyName": "Duckmann",
    "lang": "de"
}
```

Because `StoredUser` is a subclass of `User`, it has more properties than present in the expected payload.

benjamingeer:
Fixes #1476.


tobiasschweizer:
I got into some issues because properties are now declared as optional:

```
@JsonObject("UsersResponse")
export class UsersResponse {

    /**
     * The users returned in a UsersResponse.
     */
    @JsonProperty("users", [ReadUser])
    users?: ReadUser[] = [];

}
```

```
@JsonObject("GroupsResponse")
export class GroupsResponse {

    /**
     * A collection of groups.
     */
    @JsonProperty("groups", [ReadGroup])
    groups?: ReadGroup[] = [];

}
```

Is this intended?

benjamingeer:
I changed the TypeScript class generation template so that if a property has cardinality `0-n`, it is optional. I did this in response to your problem with the `members` property in `ReadProject`. Some API operations return a project with its members, and other API operations return a project without its members. But the cardinality of `members` is `0-n`, so the value of `members` could always be an empty array, and I think that in that case, it would be reasonable for the client library to also accept that the property may be absent. The easiest way to do this in TypeScript seems to be to make the property optional, and the easiest way to do that is to make all `0-n` properties optional. What sort of problem does this cause? Do you see a better solution?

tobiasschweizer:
It's fine if the users  array is empty, but it shouldn't be optional.

```ts
knoraApiConnection.admin.usersEndpoint.getUsers().subscribe(
                (response: ApiResponseData<UsersResponse>) => {
                    expect(response.body.users.length).toEqual(18);
                    expect(response.body.users[0].familyName).toEqual("Admin-alt");

                    done();
                });
```

It's ok if the users array is empty (length zero), but this code breaks if users is `undefined`. If users is optional, this would require a check if it is undefined each time before accessing it. This is why an empty array is preferable.

benjamingeer:
OK then I guess I have to make a `ProjectBasicInfo` class without `members`, with a `ProjectMembershipsResponse` class to wrap it in.

tobiasschweizer:
> OK then I guess I have to make a ProjectBasicInfo class without members, with a ProjectMembershipsResponse class to wrap it in.

If that is possible, this would be great!

I am just writing a short guideline how to integrate and try the generated code, will be ready soon.

subotic:
I thought that the admin API will always return an empty array if the sequence is empty. Are there any routes that don't behave like this?

benjamingeer:
@subotic The route `/admin/users/iri/USER_IRI/project-memberships` returns a response like this:

```json
{
   "projects":[
      {
         "description":[
            {
               "value":"Anything Project"
            }
         ],
         "id":"http://rdfh.ch/projects/0001",
         "keywords":[

         ],
         "logo":null,
         "longname":"Anything Project",
         "ontologies":[
            "http://www.knora.org/ontology/0001/anything",
            "http://www.knora.org/ontology/0001/something"
         ],
         "selfjoin":false,
         "shortcode":"0001",
         "shortname":"anything",
         "status":true
      }
   ]
}
```

In other admin API responses, a project has a `members` property (with cardinality `0-n`), but in this response, it doesn't have that property.

subotic:
The `ProjectADM` case class doesn't have a `members` property at all, and this is what is returned here.

Which routes return the `members` property together with a project? Strange, I wouldn't know how that could be possible.

benjamingeer:
@subotic You're right, this must be my mistake. Somehow I must have got confused and thought that the project class needed a `members` property in some cases. That makes it easier to fix this bug. :)

tobiasschweizer:
@benjamingeer I suggest to combine the review of this PR with the review of https://github.com/dasch-swiss/knora-api-js-lib/pull/72. https://github.com/dasch-swiss/knora-api-js-lib/pull/72 should be ready soon.

benjamingeer:
@tobiasschweizer Thanks for your help and patience with this!

tobiasschweizer:
@benjamingeer please coordinate with @subotic for merging, he is preparing a new release.

tobiasschweizer:
> Thanks for your help and patience with this!

Thank you :-)

benjamingeer:
@subotic Should I merge this today, or wait till next month? It's not a breaking change and only affects client code generation.

PR #1480, opened 2019-10-20
build: add bazel (DSP-437)

subotic:
Link to 

- [x] docker
  - [x] build knora-api image
  - [x] build knora-salsah1 image
  - [x] build knora-jena-fuseki image
  - [x] build knora-sipi image
- [x] docs (building with SBT as before)
- [x] knora-ontologies
- [x] salsah1
  - [x] build salsah1
  - [x] run salsah1
  - [x] run headless browser tests (ignoring not working tests)
- [x] sipi
  - [x] expose lua scripts
- [x] test_data
- [x] webapi
  - [x] compile twirl templates (yeah baby!!!)
  - [x] generate `BuildInfo.scala`
  - [x] build webapi
  - [x] run webapi
  - [x] run test
  - [x] run it
  - [x] allow tests to generate test data
  - [x] generate Contributors.md
- [x] Github CI
- [x] Get everything up and running after bunch of merge conflicts

YT: https://dasch.myjetbrains.com/youtrack/issue/DSP-437

Steps to take to try it out:
 1. install [Bazel](https://bazel.build/). I used `brew install bazel`.
 2. run `make test`

You probably will need to override some `.bazelrc` settings in your own copy created at `~/.bazelrc`:
```
build --action_env=PATH="/usr/local/bin:/opt/local/bin:/usr/bin:/bin"
build --strategy=Scalac=worker
build --worker_sandboxing
query --package_path %workspace%:/usr/local/bin/bazel/base_workspace
startup --host_jvm_args=-Djavax.net.ssl.trustStore=/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home/lib/security/cacerts \
        --host_jvm_args=-Djavax.net.ssl.trustStorePassword=changeit
```

Use Cases:
- Ben: you're developing in knora-api, and you need to run it and try some requests with Postman, then run some unit tests, then go back to trying things with Postman, etc.


subotic:
https://docs.bazel.build/versions/master/user-manual.html#workspace_status

benjamingeer:
On my machines, the build fails with:

Screenshot 2019-12-10 at 10 35 00

I wonder if it is because of this Bazel issue:

https://github.com/bazelbuild/bazel/issues/5741

According to that issue, Bazel embeds its own JDK, which may not have the SSL certificates needed for some downloads over HTTPS. There's a workaround, but I don't know if it makes sense for builds on macOS and/or Docker:

https://github.com/bazelbuild/bazel/issues/5741#issuecomment-418071387

benjamingeer:
The workaround seems to work on my local machine. I made a `.bazelrc` file in my home directory, containing:

```
startup --host_jvm_args=-Djavax.net.ssl.trustStore=/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home/lib/security/cacerts \
        --host_jvm_args=-Djavax.net.ssl.trustStorePassword=changeit
```

Now I get the same error that GitHub CI reports:

Screenshot 2019-12-10 at 10 46 49

subotic:
great, thanks for the fix. I've added it to the checked-in `.bazelrc`. I've fixed the remaining errors. Now the build should run through.

benjamingeer:
>  I've added it to the checked-in .bazelrc.

But that will only work if you have `/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk` installed. Will that work in a Docker container?

subotic:
> > I've added it to the checked-in .bazelrc.
> 
> But that will only work if you have `/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk` installed. Will that work in a Docker container?

crap, no :-(

subotic:
I've followed the example from the issue you linked. Let's see if it will work on Github CI. The checked-in `.bazelrc` contains the fixed from the linked issue and my private `.bazelrc`contains your variant.

subotic:
Great, the build is running through on Github CI

subotic:
I've finally managed to get to an "agreement" with the whole logging configuration. Now running `bazel run //webapi:api` seems to actually start the api :-)

subotic:
@benjamingeer running the tests is kinda working. There are still a few issues:

- There are still some issues with EHCache. It seems that everything is run inside the same VM so the caches are there when they are tried to be created
- I've moved everything under `webapi/_test_data` and `webapi/src/test/resources/test-data` to a top-level directory called `test-data`. Bazel runs everything in isolation inside a sandbox, so I needed to tell it which directory to provide to the tests. Now it is safe to use something like `FileUtil.readTextFile("test-data/...)` because Bazel will make sure that this directory is available.

- There is still a funky error when `ClientApiRouteE2ESpec.scala` is trying to run. I tries to write inside my home directory ( ~/.npm), which is not allowed.

benjamingeer:
> There is still a funky error when ClientApiRouteE2ESpec.scala is trying to run. I tries to write inside my home directory ( ~/.npm), which is not allowed.

I guess that's because it runs `npm install`.

It looks like Bazel has its own way of building TypeScript:

https://docs.bazel.build/versions/master/build-javascript.html

https://www.npmjs.com/package/@bazel/typescript

subotic:
@SepidehAlassi I should probably give you a tour around this PR. A lot of stuff happening here. It took also only 9 months to get this far :-)

SepidehAlassi:
> @SepidehAlassi I should probably give you a tour around this PR. A lot of stuff happening here. It took also only 9 months to get this far :-)

:D I see, let's schedule a meeting for this then.

docker/knora-api/BUILD.bazel
```diff
@@ -0,0 +1,38 @@
+package(default_visibility = ["//visibility:public"])
+
+load("@io_bazel_rules_docker//container:container.bzl", "container_image", "container_push")
+
+# builds the docker image
+container_image(
+    name = "image",
+    base = "@openjdk11//image",
```
SepidehAlassi, line 8: I see that you have added the build of docker images into Bazel and have removed the corresponding docker-compose of every single image? I feel like it makes the workflow harder to understand. What do you see as an advantage here?

subotic: One of the problems we now have is the disconnect between `docker build` (or any of the make targets) and `sbt compile`. You only want to build a new image if the source code has changed or any of the inputs to the docker image. For this, we need to build Docker images with Bazel. With Bazel, we can generate a nice graph where all the dependencies are depicted. This should help to understand how things are connected.

The current plumbing with Docker and SBT was more of a hack than a nice implementation. SBT has its own way of defining Docker images inside SBT, but I had great problems understanding how to do it. In general, everything we now have with Bazel should/needed to be also be done in a similar way completely with SBT. The only difference is that it took me way less time to do these things in Bazel. I tried many times to do the same in SBT over the past 5 years, but I simply don't understand SBT. To complicated for me ;-)

Makefile
```diff
@@ -93,153 +90,163 @@ else
...
+.PHONY: stack-without-api-and-salsah1
+stack-without-api-and-salsah1: stack-up ## starts the knora-stack without knora-api and salsah1: graphdb, redis, sipi.
+	@docker-compose -f docker-compose.yml stop api
+	@docker-compose -f docker-compose.yml stop salsah1
 
-.PHONY: test-unit
-test-unit: build-all-images ## runs the unit tests (equivalent to 'sbt webapi/testOnly -- -l org.knora.webapi.testing.tags.E2ETest').
-	@echo $@  # print target name
-	sbt 'webapi/testOnly -- -l org.knora.webapi.testing.tags.E2ETest'
+.PHONY: stack-db-only
```
SepidehAlassi, line 192: I have difficulty understanding why, in order to just start the `db` container, we need to start all then stop all except `db`?!

Makefile
```diff
@@ -93,153 +90,163 @@ else
...
 
-## stack without api and sipi
 .PHONY: stack-without-api-and-sipi
 stack-without-api-and-sipi: stack-up ## starts the knora-stack without knora-api and sipi: graphdb, redis, salsah1.
-	docker-compose -f docker/knora.docker-compose.yml stop api
-	docker-compose -f docker/knora.docker-compose.yml stop sipi
+	@docker-compose -f docker-compose.yml stop api
+	@docker-compose -f docker-compose.yml stop sipi
 
-.PHONY: test-only
```
SepidehAlassi, line 194: Here you have removed the `test-only`, what is supposed to be used instead?

SepidehAlassi:
> This PR is very hard to review because 3000 files are changed, and GitHub can't display them all. Would it be possible to split it into several smaller PRs?

@subotic yes, from 3564 files, I can only see 3000 changed files. Github does not show me the rest to review. I see that most of these files are caused by changing the path of `.png` files. Maybe you can reverse that commit.

subotic:
> This PR is very hard to review because 3000 files are changed, and GitHub can't display them all. Would it be possible to split it into several smaller PRs?
> 
Sorry, splitting this into multiple PRs is possible, but will require major refactoring. This would basically mean starting from the beginning. This PR introduces a large number of changes because now all the dependencies that we have, have been formalized. This formalization is necessary because Bazel provides hermetic builds that run in a sandbox.

> Can I ask how the file `dependencies.txt` is generated? Is it hand-written?
> 
> Can you give an overview of what we need to do to specify dependencies using Bazel? I'm concerned that it might be a lot more work than what we're doing now (even with all the problems we have now).

No, this file is generated and not used by the build system. It is only for humans to find out what dependencies a library we pull in has. I will remove it.

All the libraries that we want to use in our workspace are defined here: https://github.com/dasch-swiss/knora-api/blob/wip/1479-add-bazel/third_party/dependencies.bzl

this file corresponds roughly to this file: https://github.com/dasch-swiss/knora-api/blob/develop/project/Dependencies.scala

Makefile
subotic: you can run the following:

`bazel test //webapi/src/test/scala/org/knora/webapi/e2e/v1:SearchV1R2RSpec`

Makefile
subotic: I guess, at that time it was easier to implement like that. Sure, this can be simplified now.

Makefile
subotic: You can call single spec files. You can also create filegroups, that group certain spec files in a meaningful way. The final goal is that you should be able to always call `bazel test //...` which will only run the tests to which the input has changed.

benjamingeer: Is there a way to run a single test within a spec? Currently I do this in SBT like this:

```
webapi/testOnly *OntologyV2R2RSpec -- -z select1
```

Where `select1` is something unique in the string description of the test. This is described here:

https://www.scalatest.org/user_guide/using_the_runner#selectingSuitesAndTests

Makefile
benjamingeer: That would be really helpful, because starting all the containers is slow.

benjamingeer: (especially on my slow iMac)

subotic:
@benjamingeer @SepidehAlassi I would suggest that we set up a meeting, where I could demonstrate everything. Then you can ask everything that is on your mind and we can see what is missing. What do you think?

benjamingeer:
> I would suggest that we set up a meeting, where I could demonstrate everything. Then you can ask everything that is on your mind and we can see what is missing. What do you think?

Good idea!

Makefile
subotic: Ah, cool. I was thinking about this the order day. Bazel provides a way of passing stuff to the underlying test runner. I will try it out and report my findings.

subotic: ok, found it:

`bazel test //webapi/src/test/scala/org/knora/webapi/e2e/v1:SearchV1R2RSpec --test_arg=-z --test_arg="gaga"`

where `gaga` is the universal unique string :-) This will only run the tests inside the spec which description contains `gaga`.

benjamingeer: Cool, can you please document it so I can copy and paste that whenever I need it? :)

benjamingeer:
Wth SBT, I can type `webapi/console` and get a Scala console with the whole application and all its dependencies already in the classpath, so I can import them directly in Scala on the command line. Is there something like that with Bazel?

subotic:
> Wth SBT, I can type `webapi/console` and get a Scala console with the whole application and all its dependencies already in the classpath, so I can import them directly in Scala on the command line. Is there something like that with Bazel?

Yes, there is a rule (`scala_repl`) for this. I will add it.

Makefile
subotic: We should get you a fast notebook if it means less work for me ;-)

benjamingeer: And for Sepideh and Kalthoum, they have the same problem. I think your machine is much faster than everyone else's...

subotic: I've optimized it a bit. It is difficult to do it in a generalized way. If you have a specific workflow that you call all the time, then we can try and add a target that does exactly what and how you need it.

Here are the new steps:

1. `make docker-build` # run at least once so that you have all Docker images
2. `make init-db-test` # stops webapi, deletes and runs only the DB container, creates the repository and loads the data
3. `make stack-up-fast` # starts the whole stack and only rebuilds the `knora-api` Docker container if necessary

After you made some changes to webapi:

1. `make init-db-test` # stops, webapi, deletes and runs only the DB container, creates the repository and loads the data
2. `make stack-up-fast` # starts the whole stack and only rebuilds the `knora-api` Docker container if necessary

subotic:
> > Wth SBT, I can type `webapi/console` and get a Scala console with the whole application and all its dependencies already in the classpath, so I can import them directly in Scala on the command line. Is there something like that with Bazel?
> 
> Yes, there is a rule (`scala_repl`) for this. I will add it.

REPL added. To start: `bazel run //webapi:main_library_repl`

benjamingeer:
> REPL added. To start: bazel run //webapi:main_library_repl

Thanks! Please add all these things to the docs, otherwise I'll never find them again.

subotic:
> > REPL added. To start: bazel run //webapi:main_library_repl
> 
> Thanks! Please add all these things to the docs, otherwise I'll never find them again.

I've added everything to `docs/bazel_notes.md`. As soon as the MkDocs branch is merged, I will move it to the docs.

Makefile
SepidehAlassi: @subotic have you already implemented these new steps? I had made a YouTrack issue about this with suggestions
[DSP-463](https://dasch.myjetbrains.com/youtrack/issue/DSP-463)

subotic:
Thanks for the review :-)

Makefile
subotic: I did optimize it a bit. Your suggested way of doing it cannot be implemented at the moment. I've written a comment explaining it.

PR #1486, opened 2019-10-22
feature (salsah1): Add displaying of ARK URLs

loicjaouen:
add arks to salsah1.5

Salsah1.5 is supposed to call Knora API v1 which does not provide ark.
Here we make a call to Knora API v2.

And it looks like this:
salsah1-5_ark


subotic:
Wouldn't it be easier to return the ARK ID in the V1 response? This is how it is done on Salsah.org.

@benjamingeer how difficult would it be to return the ARK ID in `V1`?

benjamingeer:
> how difficult would it be to return the ARK ID in V1?

Would be easy, will look next week.

subotic:
> > how difficult would it be to return the ARK ID in V1?
> 
> Would be easy, will look next week.

Great, thanks.

@loicjaouen would this be OK?

loicjaouen:
Adding ark in the api v1 response also has its pros, the scale is fairly balanced to me, do as you see fit.

Here are the reasons why I did it this way:  
Changing api v1 code is not that obvious and requires more code change than in salsah1.5 (to my knowledge).  
API v1 should be frozen and possibly deprecated with little to no work done there (as well as salsah1.5).

benjamingeer:
And adding something to v1 API responses might also confuse v1.5 JavaScript code, requiring changes there to make it work.

PR #1490, opened 2019-10-24
build(make): add graphdb home directory setting

### LINKED ISSUE #1489: Allow setting GraphDB data dir for make
Currently, when using ˋmake stack-upˋ a Docker volume is used (for performance reasons) for storing the GraphDB data.

There is a requirement to have multiple GraphDB data folders which should be configurable via an environment variable.

subotic:
Resolves #1489

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
+os: linux
 dist: bionic
-
-sudo: required
 
 addons:
   chrome: stable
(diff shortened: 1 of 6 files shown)


subotic:
@lrosenth I‘ve added the GraphDB data directory environment variable. There is even a small documentation in the readme.

PR #1505, opened 2019-11-05
feat(clientapi): Support collection classes

### LINKED ISSUE #1504: feat(clientapi): Represent collections
We need to be able to represent data structures like this:

```json
{
  "permissions": {
    "administrativePermissionsPerProject": {
      "http://rdfh.ch/projects/0001": [
        {
          "additionalInformation": null,
          "name": "ProjectResourceCreateAllPermission",
          "permissionCode": null
        }
      ]
    },
    "groupsPerProject": {
      "http://rdfh.ch/projects/0001": [
        "http://rdfh.ch/groups/0001/thing-searcher",
        "http://www.knora.org/ontology/knora-admin#ProjectMember"
      ]
    }
  }
}
```

This needs datatypes for `array of T` and `map of string -> T` in `ClientApi`, and a way to define them in RDF.

This is needed for https://github.com/dasch-swiss/knora-api-js-lib/issues/92.

The TypeScript implementation will probably need custom converter decorators: https://github.com/AppVision-GmbH/json2typescript#custom-converter-decorators.

benjamingeer:
This PR adds support for the collection types `Map` and `Array`, by allowing a collection type signature to be defined in an IRI. Collection types defined in this way can then be represented as collection types in the target language.

Resolves #1504.
Needs https://github.com/dasch-swiss/knora-api-js-lib/pull/99.

### DIFF
--- a/docs/src/paradox/05-internals/design/client-api/index.md
+++ b/docs/src/paradox/05-internals/design/client-api/index.md
@@ -57,7 +57,7 @@ generates each file using a Twirl template, and arranges the files in the
 correct directory structure.
 
 Currently one back end, `TypeScriptBackEnd`, is implemented; it generates code
-for use with [knora-api-js-lib](https://github.com/dhlab-basel/knora-api-js-lib).
+for use with [knora-api-js-lib](https://github.com/dasch-swiss/knora-api-js-lib).
 
 ## Client Function DSL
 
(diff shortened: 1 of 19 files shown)


benjamingeer:
TODO:

`json2typescript` doesn't like collection type annotations in `@JsonProperty`. Try defining a class for each one and using the class name instead.

benjamingeer:
Thanks very much for teaching me more TypeScript!

PR #1514, opened 2019-11-11
feat(clientapi): Return v2 test data

### LINKED ISSUE #1513: Generate V2 Values Route Test Data
In https://github.com/dasch-swiss/knora-api-js-lib/pull/86 (support of v2 values route), I would like to use test data created by Knora (client code generation).

Basically, I need JSON test data (requests and responses) for each value type and operation (edit value, create value, delete value) from `org/knora/webapi/e2e/v2/ValuesRouteV2E2ESpec.scala` and  `org/knora/webapi/e2e/v2/ValuesV2R2RSpec.scala`.

benjamingeer:
This PR adds test requests and responses for the `/v2/values` route to the generated client API code.

I refactored the values E2E tests to move the generation of JSON-LD into functions and constants into `SharedTestDataADM`, where they can be used both by E2E tests and by client code generation.

I also refactored the TypeScript backend a bit to make the endpoint template easier to read.

Resolves #1513.


tobiasschweizer:
That's great, thank you. Will try to integrate the generated v2 values endpoint test data in https://github.com/dasch-swiss/knora-api-js-lib/pull/86 and then review this PR.

benjamingeer:
> I cannot find test data for a create still image file value request.

That’s because there is no such request. A file value can only be created when creating a resource.

benjamingeer:
https://docs.knora.org/paradox/03-apis/api-v2/editing-values.html#submit-a-file-value-to-knora

tobiasschweizer:
Ok, I removed the unit test that would not have worked with Knora.

tobiasschweizer:
I will think of a way to prevent clients from performing a create request for file values in the values endpoint.

tobiasschweizer:
`get-still-image-file-value-response.json` contains

{"knora-api:error":"org.knora.webapi.NotFoundException: Value with UUID goZ7JFRNSeqF-dNxsqAS7Q not found (maybe you do not have permission to see it, or it is marked as deleted)","@context":{"knora-api":"http://api.knora.org/ontology/knora-api/v2#"}}

benjamingeer:
Well, it’s a valid response, isn’t it? 😀

tobiasschweizer:
> Well, it’s a valid response, isn’t it? 😀

I like it!

I am reloading my test data and regenerate the code to rule out a local problem.

webapi/src/main/scala/org/knora/webapi/routing/v2/ValuesRouteV2.scala
```diff
@@ -22,174 +22,612 @@ package org.knora.webapi.routing.v2
...
-                    }
+                    } yield requestMessage
+
+                    RouteUtilV2.runRdfRouteWithFuture(
+                        requestMessageF = requestMessageFuture,
+                        requestContext = requestContext,
+                        settings = settings,
+                        responderManager = responderManager,
+                        log = log,
+                        targetSchema = ApiV2Complex,
```
tobiasschweizer: Would the simple schema be rejected?

benjamingeer:
> Would the simple schema be rejected?

Yes. Added tests for this.

PR #1515, opened 2019-11-13
fix/597-duplicates

loicjaouen:
fix for #597 : when clicking fast, the click event is called multiple times which results in duplicate entries.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,32 @@
+name: CI
+
+on: [push]
+
+jobs:
+  compile:
+    name: Compile
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+      - name: Run compile and stage
+        run: sbt clean stage
+  # unit-tests:
+  # e2e-tests:
+  # integration-tests:
+  # upgrade-tests:
+
+  publish:
+    name: Publish
+    needs: compile
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/develop' || github.ref == 'refs/tags/*'
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          fetch-depth: 50
+      - name: Build and publish all images to Dockerhub
+        run: |
+          echo ${{ secrets.DOCKER_HUB_TOKEN }} | docker login -u ${{ secrets.DOCKER_USER }} --password-stdin
+          make publish-all-images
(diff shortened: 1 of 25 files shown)


loicjaouen:
this initial attempt to fix by calling the freeze screen as early as possible failed as well as this recommended way [here](https://stackoverflow.com/questions/16715075/preventing-multiple-clicks-on-button)

loicjaouen:
as well as this [event tricks](https://stackoverflow.com/questions/11556571/prevent-click-event-from-firing-multiple-times-issue-with-fading)

loicjaouen:
back to the root [JS concurrency model](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop)

so a click on the button puts in the message queue the onclick event, when processing the onclick event, a function to process this click will created through a frame added on the stack, the stack processing will execute the function all the way until it returns.

Then the next message (from the queue) will be processed and there is the next onclick.

So there is no way to prevent a second click to generate the message because processing the first one might not be fast enough to visually block the button on the web page.

A software lock in the function processing the click will have no effect as the said lock will be released at the end of the function and won't block the next lock at the next function execution.

loicjaouen:
so what should be done is to add an event listener with in capture mode (is the first to get the event notification) and unique mode (suppress the event listener after the first execution).

it would look like:
```js
options = {
    // events of this type will be dispatched to the registered listener before being dispatched to any EventTarget beneath it in the DOM tree.
    'capture': true, 
    // the listener should be invoked at most once after being added
    'once': true
}
element.addEventListener('click', function, options);
```

But we are in jQuery and the `jQueryElement.click()` does not have such options...

loicjaouen:
with the jQuery way, that makes far less code changes
testing...

loicjaouen:
it works on firefox, safari, chrome

loicjaouen:
hmmm, sometimes we want the code to run more than once :(

if the first submit attempt fails on validation check, we want the user to fix his data and re-submit

loicjaouen:
there was only one case for validation

loicjaouen:
This PR is quite verbose because we started to discuss it with @gfoo and so it would have been easier to catch-up if our discussion was to continue.  

But finally there is not so much to review.

Would one of you guys be kind enough to review this PR?

subotic:
Is this working for you, or are there still problems?

loicjaouen:
it's working now :)

loicjaouen:
thanks @subotic 

codacy checks indirectly revealed a case that I overlooked, it is fixed too now

Issue #1238, opened 2019-03-03
Migrate Knora off of Travis CI


subotic:
Saw this issue https://github.com/ansible/molecule/issues/1770 referencing these tweets:

- https://twitter.com/ReinH/status/1098663375985229825
- https://www.reddit.com/r/devops/comments/at3oyq/it_looks_like_ibera_is_gutting_travis_ci_just_a/
- https://twitter.com/hynek/status/1098908925091610624

It seems that we should slowly think about alternatives. There are a few out there: https://github.com/ligurio/awesome-ci

---

needs dasch-swiss/dsp-meta#7

subotic:
We are going to slowly migrate off of Travis

PR #1516, opened 2019-11-14
build: Fix publish to Dockerhub

subotic:
related:  #1238


subotic:
@benjamingeer It was faster to find out how to use Github Actions and test it this way than to wait for Travis :-) As soon as Travis passes, this should be good to go.

subotic:
Thanks for the review :-)

PR #1517, opened 2019-11-15
build: Fix publish to Dockerhub

subotic:


subotic:
@benjamingeer sorry, copy-paste is sometimes really difficult to get right :-)

subotic:
There is always going to be a need for orchestration. And when changes happen on different services, manual interventions are going to be necessary. Ideally, you wouldn't even notice, because the operations team would take care of it. But because the operations team is me and myself, I need someone to approve :-)

benjamingeer:
I'm not complaining, I'm just thinking this would be a lot less work for you if GitHub and Docker merged into one service. :)

PR #1526, opened 2019-11-20
fix(build): Fix E2E tests

### LINKED ISSUE #1525: Makefile not running the right tests
I’ve just realised that `ResourcesRouteV2E2ESpec` fails if I run it locally with the develop branch. It fails (correctly) because of changes I made in this PR, which was merged two weeks ago:

https://github.com/dasch-swiss/knora-api/pull/1505

However, we never got any failures for it in Travis or GitHub CI, because it was never run. Looking at the Travis test runs that were done for that PR when it was merged, as well as the latest GitHub CI run for the develop branch, you can see that most of the E2E tests are skipped during the "E2E tests" job:

https://github.com/dasch-swiss/knora-api/runs/300674005

https://github.com/dasch-swiss/knora-api/runs/311653127

I think the problem is in the Makefile:

https://github.com/dasch-swiss/knora-api/blob/develop/Makefile#L260

This line seems to be excluding the E2E tests when it should be running the E2E tests. I think this means that tagged E2E tests have not been run on Travis/GitHub since this PR was merged, which introduced that Makefile:

https://github.com/dasch-swiss/knora-api/pull/1459

(I also now see that I forgot to tag E2E tests, so they've still been run.)

benjamingeer:
- [x] Fix `e2e-tests-with-coverage` target in `Makefile`.
- [x] Don't check collection types in `InstanceChecker`.
- [x] Give `InstanceChecker` access to the non-standard names of certain JSON properties.
- [x] Update test data for:
  - [x] `ResourcesRouteV2E2ESpec`
  - [x] `InstanceCheckerSpec`
- [x] Fix `ClientApiRouteE2ESpec` to load `anything-data.ttl`, now needed for client code generation.
- [x] Add E2E tag to E2E specs that didn't have it.
- [x] Use `IIIF_BASE_URL` placeholder in E2E test files, because the Sipi URL is different when tests are run in Docker.
- [x] Ignore tests that create large texts, because they use too much memory for GitHub CI.

Fixes #1525.


webapi/src/it/scala/org/knora/webapi/ITKnoraLiveSpec.scala
```diff
@@ -127,9 +127,7 @@ class ITKnoraLiveSpec(_system: ActorSystem) extends Core with StartupUtils with
         getResponseString(request).parseJson.asJsObject
     }
 
-
-
-    protected def singleAwaitingRequest(request: HttpRequest, duration: Duration = 5999.milliseconds): HttpResponse = {
+    protected def singleAwaitingRequest(request: HttpRequest, duration: Duration = 7999.milliseconds): HttpResponse = {
```
subotic: Just in case you wonder why this odd number. It is easier to find the correct setting that needs to be tweaked. The timeout output is often not clear where it happened.

webapi/src/test/resources/test-data/resourcesR2RV2/Testding.jsonld
```diff
@@ -105,6 +105,52 @@
       "@value" : "http://0.0.0.0:3336/ark:/72163/1/0001/H6gBWUuJSuuO=CilHV8kQwk/bXMwnrHvQH2DMjOFrGmNzgz.20180528T155203897Z"
     }
   },
+  "anything:hasGeometry" : {
```
tobiasschweizer: I assume these values were introduced in #1514 but since the e2e tests were not run the difference in the test data was not detected.

benjamingeer:
> I assume these values were introduced in #1514 but since the e2e tests were not run the difference in the test data was not detected.

Actually, what happened is that at some point we added ARK URLs and UUIDs to values. After that, some hard-coded data that was supposed to fail validation (e.g. it had a boolean property with a string object, to check whether `InstanceChecker` would notice) started to fail validation because the hard-coded values in the test didn’t contain ARK URLs or UUIDs. So I had to add ARK UURLs and UUIDs to those values, to make them fail validation for the right reasons.

benjamingeer:
@subotic I don't understand what's happening. It does:

```
Setting up npm (3.5.2-0ubuntu4) ...
```

But then it says:

```
java.io.IOException: Cannot run program "npm" (in directory "/tmp/clientapi18252897938348417182/src"): error=2, No such file or directory
```

There is also this strange error, as if GraphDB sent a header saying that its response would be in GZIP format but the client couldn't unpack the file:

```
java.util.zip.ZipException: Not in GZIP format
2679
	at java.base/java.util.zip.GZIPInputStream.readHeader(GZIPInputStream.java:166)
2680
	at java.base/java.util.zip.GZIPInputStream.<init>(GZIPInputStream.java:80)
2681
	at java.base/java.util.zip.GZIPInputStream.<init>(GZIPInputStream.java:92)
2682
	at org.apache.http.client.entity.GZIPInputStreamFactory.create(GZIPInputStreamFactory.java:61)
2683
	at org.apache.http.client.entity.LazyDecompressingInputStream.initWrapper(LazyDecompressingInputStream.java:51)
2684
	at org.apache.http.client.entity.LazyDecompressingInputStream.read(LazyDecompressingInputStream.java:69)
2685
	at java.base/sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:284)
2686
	at java.base/sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326)
2687
	at java.base/sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
2688
	at java.base/java.io.InputStreamReader.read(InputStreamReader.java:185)
2689
	at java.base/java.io.Reader.read(Reader.java:144)
2690
	at org.apache.http.util.EntityUtils.toString(EntityUtils.java:227)
2691
	at org.apache.http.util.EntityUtils.toString(EntityUtils.java:308)
2692
	at org.knora.webapi.store.triplestore.http.HttpTriplestoreConnector.$anonfun$getSparqlHttpResponse$1(HttpTriplestoreConnector.scala:662)
2693
	at scala.util.Try$.apply(Try.scala:213)
2694
	at org.knora.webapi.store.triplestore.http.HttpTriplestoreConnector.getSparqlHttpResponse(HttpTriplestoreConnector.scala:652)
```

benjamingeer:
When I run all the tests outside Docker, they all pass.

benjamingeer:
When I run `make e2e-tests` on the command line, it does some stuff and then stops with this error message:

```
make: *** No rule to make target `init-knora-test-unit', needed by `e2e-tests'.  Stop.
```

benjamingeer:
When I run the tests using `make e2e`, is there a way to stop them? I tried Control-C, but it didn't have any effect, the tests kept running. I looked in the Docker Desktop menu but didn't find anything for stopping containers. I tried installing Kitematic, but it doesn't show any local containers either. I ended up finding the sbt process ID and using `kill -9`.

subotic:
> When I run all the tests outside Docker, they all pass.

Then `npm` is missing inside the Docker container. I will add it to the image.

> When I run the tests using make e2e, is there a way to stop them?

No, the only way would be to kill the created `api` container from terminal:

```
$ docker container ls
$ docker container rm <container_id> // you could try 'api' as the container_id
```

subotic:
I've updated the Docker image and restarted the tests. Let's see if this all that is needed.

benjamingeer:
> I've updated the Docker image

So do we still need this in `main.yml`?

```
      - name: install requirements
        run: sudo apt-get install nodejs npm
```

subotic:
> > I've updated the Docker image
> 
> So do we still need this in `main.yml`?
> 
> ```
>       - name: install requirements
>         run: sudo apt-get install nodejs npm
> ```

I don't think so.

benjamingeer:
OMG the tests passed!

benjamingeer:
Now it's failing because of this:

```
Triplestore responded with HTTP code 500:
Query evaluation error:
Insufficient free Heap Memory 240Mb for group by and distinct, threshold:250Mb, reached 0Mb (HTTP status 500)
```

subotic:
> Now it's failing because of this:
> 
> ```
> Triplestore responded with HTTP code 500:
> Query evaluation error:
> Insufficient free Heap Memory 240Mb for group by and distinct, threshold:250Mb, reached 0Mb (HTTP status 500)
> ```

Which test changed after the merge? GraphDB doesn't have enough memory.

benjamingeer:
> Which test changed after the merge? GraphDB doesn't have enough memory.

No tests changed, it was the merge of #1527. Can we give GraphDB more memory?

benjamingeer:
@subotic Does this mean we can't use GitHub CI?

subotic:
> > Which test changed after the merge? GraphDB doesn't have enough memory.
> 
> No tests changed, it was the merge of #1527. Can we give GraphDB more memory?

I don‘t think so.

> Does this mean we can't use GitHub CI?

This means that we need to find a way how it can work with Github CI. Unfortunately, there is no alternative. After doing some calculations, running it on our own hardware is not feasible cost wise.

We could try and isolate the high memory tests with a HighMemory test tag and run them separately?

benjamingeer:
OK the tests that use a lot of memory are the ones that create large texts with lots of standoff. I think we don't really need to run these in CI anyway, so I set them to `ignore`. Otherwise everything seems to work. Can we merge this?

PR #1529, opened 2019-11-27
build: Fix publish

subotic:


subotic:
@benjamingeer should be a quick review :-)

subotic:
> I have no idea what this means but it looks fine. :)

As if I know what it does. We will see :-)

PR #1537, opened 2019-11-28
build: Add env var to set triplestore actor pool

loicjaouen:
The `application.conf` sets many parameters.

Some are hard coded to be changed once in a blue moon, others are more accessible through the setting of environment variables.

This PR is to allow such environment variable to set the size of the thread pool dedicated to querying the tripple store.


loicjaouen:
@subotic , testing this takes longer than expected I tested the effect on a local run, but it also needs to run in the images as we set-up stacks for deployment. So could you already have a look at the changes to check if something obviously wrong meets your eyes?

PR #1560, opened 2019-12-17
fix(clientapi): Fix HTTP method in createChildNode function

### LINKED ISSUE #1559: Admin lists: createChildNode should send a httpPost request, not httpPut
In the method `createChildNode()` in the generated code of knora-api-js-lib, the API request is wrong: it is not `httpPut()` but `httpPost()`.

method to update:
```typescript
createChildNode(node: CreateChildNodeRequest): Observable<ApiResponseData<ListInfoResponse> | ApiResponseError> {
        return this.httpPut(.... ..... ....);
}
```

reference to #1509

benjamingeer:
Fixes #1559.


flavens:
@benjamingeer we should not get back `ListInfoResponse` but `ListNodeInfoResponse` because we create a child node here (see `return`)..

benjamingeer:
@flavens Thanks again!

PR #1561, opened 2019-12-17
fix(webapi): unique username/email check on change user

LukasStoeckli:


webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
```diff
@@ -261,28 +261,19 @@ class UsersResponderADM(responderData: ResponderData) extends Responder(responde
...
-                    // check if current email differs from the one in the change request
-                    if (!user.email.equals(changeUserRequest.email.get)) {
-                        // check if new email is free
-                        userByEmailExists(changeUserRequest.email.get).map(result =>
-                            if (result) throw DuplicateValueException("A user with this email exists already. Cannot change")
-                        )
-                    }
+                val email = currentUserInformation.get.email
+                if (!email.equals(changeUserRequest.email.get)) {
+                    val result = userByEmailExists(changeUserRequest.email.get)
```
subotic: Ups, using Await is unfortunately a „no go“. Await is blocking and thus evil.

tobiasschweizer: https://doc.akka.io/docs/akka/2.5/futures.html#use-with-actors

tobiasschweizer: > Warning

> Await.result and Await.ready are provided for exceptional situations where you must block, a good rule of thumb is to only use them if you know why you must block. For all other cases, use asynchronous composition as described below.

LukasStoeckli: I see, but why would we want concurrent execution of the query, as long as we can't guarantee the uniqueness of the values? Otherwise we would have to rollback, as soon as the results are in...

benjamingeer: The idea is to free up CPU threads so they can do something else (e.g. process a request from another user) while waiting for a result from the triplestore.

subotic:
@LukasStoeckli what is the status of this PR?

LukasStoeckli:
works, but:
* 'check if current email/username differs from the one in the change request' not in place, as I could not figure out how this nested conditions and exceptions stuff works
* stringFormatter.validate... from develop needs to be merged

benjamingeer:
I haven't looked closely at this, but if you need to lock something to guarantee uniqueness of the values, you can use the Knora utility class `IriLocker` instead of Akka's `Await`.

LukasStoeckli:
there is a global IRI lock. The point would be not to perform an update if the value hasn't changed. The issue is the scope of the conditions respectively the exceptions thrown in it. As in the code on develop, the exception is thrown but doesn't lead to not executing the rest of the function. As in this PR, everything works as intended, but there's no longer a check whether the new value is actually new.

benjamingeer:
> the exception is thrown but doesn't lead to not executing the rest of the function

I don't see how that's possible. Are you sure? is there a test for that?

webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
```diff
@@ -260,30 +260,15 @@ class UsersResponderADM(responderData: ResponderData) extends Responder(responde
             )
 
             // check if we want to change the email
-            _ = if (changeUserRequest.email.isDefined) {
```
benjamingeer, line 263: This line throws away the `Future` returned by `userByEmailExists`. The failure containing the exception is inside that `Future`, so if you throw away the `Future`, the exception can't be propagated. You have to "unpack" the `Future`, using `<-` as you do in this PR, to propagate the exception.

LukasStoeckli: ok thanks. I still have no idea how you would do that, so I found another solution.

benjamingeer: The way you did it looks fine. Once this PR is done I'll be glad to review it.

LukasStoeckli: great. thanks for your help. It should be good now.
Is the failing ResourcesRouteV2E2E a common issue in the CI?

benjamingeer:
> Is the failing ResourcesRouteV2E2E a common issue in the CI?

Yes, GraphDB often runs out of memory in GitHub CI, but we don't know why.

webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
```diff
@@ -1420,34 +1403,50 @@ class UsersResponderADM(responderData: ResponderData) extends Responder(responde
...
-    private def userByUsernameExists(username: String): Future[Boolean] = {
-        for {
-            askString <- Future(queries.sparql.admin.txt.checkUserExistsByUsername(username = username).toString)
-            // _ = log.debug("userExists - query: {}", askString)
-
-            checkUserExistsResponse <- (storeManager ? SparqlAskRequest(askString)).mapTo[SparqlAskResponse]
-            result = checkUserExistsResponse.result
-
-        } yield result
+    private def userByUsernameExists(username: String, current: String = ""): Future[Boolean] = {
```
benjamingeer, line 1409: Please don't use `""` to represent a value that does not exist. That's why Scala has the `Option` class. Use `None` instead:

```scala
private def userByUsernameExists(username: String, current: Option[String] = None)
```

webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
```diff
@@ -1420,34 +1403,50 @@ class UsersResponderADM(responderData: ResponderData) extends Responder(responde
...
-    private def userByEmailExists(email: String): Future[Boolean] = {
-        for {
-            askString <- Future(queries.sparql.admin.txt.checkUserExistsByEmail(email = email).toString)
-            // _ = log.debug("userExists - query: {}", askString)
-
-            checkUserExistsResponse <- (storeManager ? SparqlAskRequest(askString)).mapTo[SparqlAskResponse]
-            result = checkUserExistsResponse.result
-
-        } yield result
+    private def userByEmailExists(email: String, current: String = ""): Future[Boolean] = {
```
benjamingeer, line 1434: Please use `Option` here too instead of the empty string:

```scala
private def userByEmailExists(email: String, current: Option[String] = None)
```

webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
```diff
@@ -1420,34 +1403,50 @@ class UsersResponderADM(responderData: ResponderData) extends Responder(responde
...
       *
       * @param username the username of the user.
+      * @param current the current username of the user.
       * @return a [[Boolean]].
       */
-    private def userByUsernameExists(username: String): Future[Boolean] = {
-        for {
-            askString <- Future(queries.sparql.admin.txt.checkUserExistsByUsername(username = username).toString)
-            // _ = log.debug("userExists - query: {}", askString)
+    private def userByUsernameExists(username: Option[String], current: Option[String] = None): Future[Boolean] = {
```
benjamingeer, line 1409: Why use an `Option` for `username`? Is there a reason to call this function if you don't have a new username?

webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
```diff
@@ -1420,34 +1403,50 @@ class UsersResponderADM(responderData: ResponderData) extends Responder(responde
...
       *
       * @param email the email of the user.
+      * @param current the current email of the user.
       * @return a [[Boolean]].
       */
-    private def userByEmailExists(email: String): Future[Boolean] = {
-        for {
-            askString <- Future(queries.sparql.admin.txt.checkUserExistsByEmail(email = email).toString)
-            // _ = log.debug("userExists - query: {}", askString)
+    private def userByEmailExists(email: Option[String], current: Option[String] = None): Future[Boolean] = {
```
benjamingeer, line 1434: Why use an `Option` for `email`? Is there a reason to call this function if you don't have a new email address?

benjamingeer:
ping

webapi/src/main/scala/org/knora/webapi/responders/admin/UsersResponderADM.scala
LukasStoeckli: Because the username might or might not be defined in the request. And if I check this before calling the function, the exception is ignored.

benjamingeer: OK, now I understand, thanks.

benjamingeer:
@subotic Should we merge this PR?

subotic:
@benjamingeer I guess so. The tests are passing.

PR #1571, opened 2020-01-06
fixed date string format

### LINKED ISSUE #1569: broken ark
on PR #1486 I copied/pasted broken code from salsah-v1 where the ark timestamp is computed and suffers from a bad formatting, 2020-01-06 is rendered as `202016` instead if `20200106`:
```diff
- 202016
+ 20200106
```

loicjaouen:
fixes #1569

### DIFF
--- a/salsah1/src/public/js/imagebase.js
+++ b/salsah1/src/public/js/imagebase.js
@@ -48,10 +48,9 @@ $(function() {
 			propedit.append(arkdiv);
 			SALSAH.ApiGet('resourcesv2', resource.resdata.res_id, {resinfo: true, reqtype: 'context'}, function(datav2) {
 				if (datav2['knora-api:arkUrl'] && datav2['knora-api:arkUrl']['@value']) {
-					var now = new Date();
 					// toISOString() returns a string like: 2020-01-06T14:40:16.187Z
 					// removing seperators: -:.
-					var nowstr = now.toISOString().replace(/-|:|\./gi,'');
+					var nowstr = (new Date()).toISOString().replace(/-|:|\./gi,'');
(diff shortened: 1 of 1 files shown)


loicjaouen:
hello @lrosenth , I originally stole this code from you (I guess) and you might be interested.

Now, is it the best way to do it in the salsah1.5 case?

Right now we take the ARK and we append the current date:

```json
    "knora-api:arkUrl" : {
      "@type" : "xsd:anyURI",
      "@value" : "http://0.0.0.0:3336/ark:/72163/1/0001/55UrkgTKR2SEQgnsLWI9mgR/c8zmKe=eRjWMOGIOw=5GyAh"
    },
```

An alternative would be to get the versioned  ARK straight from the API and get the last changed version, not an ever renewed ARK without concern of the resource being the same for years:

```json
    "knora-api:versionArkUrl" : {
      "@type" : "xsd:anyURI",
      "@value" : "http://0.0.0.0:3336/ark:/72163/1/0001/55UrkgTKR2SEQgnsLWI9mgR/c8zmKe=eRjWMOGIOw=5GyAh.20191129T100000673298Z"
    }
```

wouldn't it be better?

loicjaouen:
right now, the ARK dereferences to knora-app which displays the resource at the requested version. The version is a time-stamp which is not forwarded in the exploration of the data graph (by clicking on a linked resource). 

So if we ask for a resource with the last change date or the now timestamp, we get the same resource. And we also get the same result when clicking on a link of that resource.

Conclusion: let's grab the last resource.

loicjaouen:
Thanks @lrosenth :)

PR #1582, opened 2020-01-24
fix(api-v2): Change link value comment

### LINKED ISSUE #1574: Error updating LinkValue with change of commentary only
When updating a LinkValue where only the commentary is changed, but the referenced resource remains the same, an error message occurs:

"org.knora.webapi.SparqlGenerationException: linkUpdateForNewLink.directLinkExists must be false in this SPARQL template"

**Query:**
PUT /v2/values HTTP/1.1
Accept: */*
Origin: http://localhost:4200
Content-Type: application/json; charset=utf-8
Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJLbm9yYSIsInN1YiI6Imh0dHA6Ly9yZGZoLmNoL3VzZXJzL3Jvb3QiLCJhdWQiOlsiS25vcmEiLCJTaXBpIl0sImV4cCI6MTU4MTQ2Njk5MiwiaWF0IjoxNTc4ODc0OTkyLCJqdGkiOiJiWWxsYzIyVFJmQ1UzWVJXeTJoWGN3In0.d5OBil4BXvgVQC7FWqNqj4_HNCzqJS8rVs2CHn8Qtys
Referer: http://localhost:4200/
Content-Length: 517
Host: 0.0.0.0:3333
Accept-Language: de-ch
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_2) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.4 Safari/605.1.15
Accept-Encoding: gzip, deflate
Connection: keep-alive

**Body:**
```json
{
  "@type":"http://0.0.0.0:3333/ontology/0807/mls/v2#Lemma",
  "@id":"http://rdfh.ch/0807/BK8mvpzgREuYnuH5mA3BXg",
  "http://0.0.0.0:3333/ontology/0807/mls/v2#hasSubLemmaValue":{
    "@id":"http://rdfh.ch/0807/BK8mvpzgREuYnuH5mA3BXg/values/CLb4DBxvRQ6osNmnNFgF2g",
  "@type":"http://api.knora.org/ontology/knora-api/v2#LinkValue",
    "http://api.knora.org/ontology/knora-api/v2#valueHasComment":"Dies ist ein Test!",
    "http://api.knora.org/ontology/knora-api/v2#linkValueHasTargetIri":{
      "@id":"http://rdfh.ch/0807/0ZsvWKf6Ri-qQegLG_lN_g"
    }
  }
}
```

**Answer:**
HTTP/1.1 500 Internal Server Error
Access-Control-Allow-Credentials: true
Content-Type: application/json
Access-Control-Expose-Headers: Server
Date: Mon, 13 Jan 2020 21:17:48 GMT
Content-Length: 212
Access-Control-Allow-Origin: http://localhost:4200
Server: webapi/10.0.0-SNAPSHOT akka-http/10.1.7

benjamingeer:
This PR fixes the case where the client changes only the comment on a `LinkValue`.

Fixes #1574.


tobiasschweizer:
@benjamingeer I suggest that @lrosenth first try this PR. I am happy to have a look at the implementation details in knora-api after his approval.

benjamingeer:
@tobiasschweizer It looks like @lrosenth doesn't have time to review this. Should I ask someone else?

tobiasschweizer:
I can do it, is Thursday ok?

benjamingeer:
OK thanks!

webapi/src/test/scala/org/knora/webapi/e2e/v2/ValuesRouteV2E2ESpec.scala
```diff
@@ -2701,6 +2701,44 @@ class ValuesRouteV2E2ESpec extends E2ESpec {
             savedTargetIri should ===(linkTargetIri)
         }
 
+        "update a link between two resources, changing only the comment" in {
```
tobiasschweizer: The link value that gets updated here was created in

https://github.com/dasch-swiss/knora-api/blob/b823a3f2d38552654c82236c636b7340b3a26325/webapi/src/test/scala/org/knora/webapi/e2e/v2/ValuesRouteV2E2ESpec.scala#L1697

without a comment. Then a comment gets added in "update a link between two resources, changing only the comment". Shouldn't we ideally have test cases for

- create a value with an initial comment that then gets changed
- create a value without a comment that is added

benjamingeer: Good idea, I'll do that.

tobiasschweizer: Ideally, you could also do that for `https://github.com/dasch-swiss/knora-api/pull/1582/files#diff-0cbc5ab548dbaeaf4be4a8807abb05b7`.

Maybe it would not hurt to do it for a non link value too. I guess these have a common template so testing one non linking value would be ok to begin with, right?

benjamingeer: Yes, it's only link values that are handled by separate templates.

tobiasschweizer: Great! 

Is this understanding correct?

> To create a new version of a value either the value itself has to be different (regardless of its comment being different from the previous version) or just its comment (this includes submitting a comment for a value that did not have any comment before).

benjamingeer: It should be correct. I'll add the additional tests you asked for to make sure. :)

tobiasschweizer: What about submitting the same value without a comment for a value that had a comment before?

tobiasschweizer: I am about to create validators for the GUI elements and they should disallow the submission of invalid combinations.

benjamingeer: That should remove the comment from the value.

tobiasschweizer: > That should remove the comment from the value.

Is there already a test for this? ;-)

benjamingeer: Will add a test for that.

tobiasschweizer: great!

benjamingeer: I added a lot of tests, see if this looks OK to you.

benjamingeer:
> That looks paranoid enough to me ;-)

That's high praise. :)

PR #1584, opened 2020-01-24
fix(api-v2): Fix ontology deletion

### LINKED ISSUE #974: Deleting ontologies using API v2
Using the API I can not delete an ontology.
To test, run the attached PHP script 2 times. Second time it failed with an error

[ontoldeltest.php.zip](https://github.com/dhlab-basel/Knora/files/2310967/ontoldeltest.php.zip)

benjamingeer:
This PR fixes ontology deletion so that the ontology is removed not only from the triplestore, but also from the ontology cache.

- [x] Fix incorrect test.
- [x] Fix ontology deletion in `OntologyResponderV2`.

Fixes #974.


benjamingeer:
@kilchenmann Thanks for the review, and sorry for the bug! :)

PR #1587, opened 2020-01-28
fix (webapi): Add enforcing of restrictions for username and email

### LINKED ISSUE #1585: Check username for allowed characters when new user is created
Creating a user with `"username": "donald.duck@example.org"` should not be possible.

subotic:
resolves #1585


subotic:
@benjamingeer whenever you have a spare moment, a review would be great. Thanks!

subotic:
Didn't you add the regex? It was there already. I have no idea what it exactly does.

benjamingeer:
You added it in this commit as part of PR #1208:

https://github.com/dasch-swiss/knora-api/commit/3c4d9652dd1b5d8bad862b7c2776277195d0dbf0#diff-659c65d367a0897435d7612e0d1b8a4bR788

I agree that it was there already and so it's not a problem with this PR, but it would be nice to have it explained a bit. If you've already forgotten how it works, that's a good example of why it could use a comment. :)

subotic:
@benjamingeer Thanks for the review.

PR #1594, opened 2020-02-07
chore: Add Flaky test tag and run these tests separately

subotic:


benjamingeer:
Can we discuss this first?

Would you consider `AuthenticationV2E2ESpec` flaky?

https://github.com/dasch-swiss/knora-api/issues/1536#issuecomment-566443466

Or all the tests that get default permissions?

https://github.com/dasch-swiss/knora-api/issues/1536#issuecomment-570198947

I think these these failures are random: any SPARQL query can fail randomly when it's run on GitHub CI.

Can you please explain what makes you think that some particular tests are "flaky"?

subotic:
We need to find a solution for the intermittent test failures and for this, we need to find the root cause. Until then, we can maybe mitigate the problem by isolating some tests that fail. Sometimes these tests also fail locally. This is not a Github CI problem. I suspect this is a GraphDB problem. It all started when we moved GraphDB from 8.5 to 9.0.

Unfortunately, we cannot run the tests with more resources. Github CI is what we have, so we need to fix the situation somehow.

You're on vacation. Don't worry about this. I don't think that this PR will solve the problem. For this, I will need your help, but it can wait until next week. When you are in Basel, we can talk about it.

benjamingeer:
> Sometimes these tests also fail locally.

I haven’t been able to make any of these failures locally. Maybe it’s because I don’t run GraphDB in Docker locally.

> You're on vacation. Don't worry about this. I don't think that this PR will solve the problem. For this, I will need your help, but it can wait until next week. When you are in Basel, we can talk about it.

OK, thanks. :)

subotic:
this PR is deprecated

PR #1601, opened 2020-02-19
fix(api-v2): Fix custom datatypes in knora-api simple ontology

### LINKED ISSUE #1593: virtual "knora-api" problems
I encountered same problems with the virtual `knora-api`- ontology which does not conform to the OWL standard.

- rdfs:label my not be subclass of a restriction class to assign cardinality. See [here](http://protege-project.136.n4.nabble.com/Why-it-is-not-possible-to-have-cardinality-restriction-for-rdfs-label-or-rdfs-comment-td4658506.html). A solution might be - as mentioned - by adding a declaration of rdfs:label as a data property
- Datatype restriction should probably not use "subClassOf" but "rdfs:range":
```
knora-api:Date a rdfs:Datatype;
  rdfs:comment "Represents a date as a period with different possible precisions.";
  rdfs:label "Date literal";
  rdfs:range [ a rdfs:Datatype;
      owl:onDatatype xsd:string;
      owl:withRestrictions [
          xsd:pattern "(GREGORIAN|JULIAN):\\d{1,4}(-\\d{1,2}(-\\d{1,2})?)?( BC| AD| BCE| CE)?(:\\d{1,4}(-\\d{1,2}(-\\d{1,2})?)?( BC| AD| BCE| CE)?)?"
        ]
    ] .
```

Eventually we should follow OWL-standards in order to use tools to visualize ontologies and SPARQL queries using knora-api:

benjamingeer:
This PR:

- Fixes the definitions of custom datatypes in the `knora-api` ontology in the simple schema.
- Adds `rdfs:label a owl:DatatypeProperty` so it can be used in cardinalities.

@tobiasschweizer Do you think this will affect `knora-api-js-lib`?

Fixes #1593.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/ontologymessages/KnoraBaseToApiV2ComplexTransformationRules.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/ontologymessages/KnoraBaseToApiV2ComplexTransformationRules.scala
@@ -32,10 +32,15 @@ import org.knora.webapi.util.{SmartIri, StringFormatter}
 object KnoraBaseToApiV2ComplexTransformationRules extends OntologyTransformationRules {
     private implicit val stringFormatter: StringFormatter = StringFormatter.getInstanceForConstantOntologies
 
-    override val ontologyMetadata = OntologyMetadataV2(
+    override val ontologyMetadata: OntologyMetadataV2 = OntologyMetadataV2(
         ontologyIri = OntologyConstants.KnoraApiV2Complex.KnoraApiOntologyIri.toSmartIri,
         projectIri = Some(OntologyConstants.KnoraAdmin.SystemProject.toSmartIri),
         label = Some("The knora-api ontology in the complex schema")
+    )
+
+    private val Label: ReadPropertyInfoV2 = makeProperty(
+        propertyIri = OntologyConstants.Rdfs.Label,
+        propertyType = OntologyConstants.Owl.DatatypeProperty
     )
 
     private val Result: ReadPropertyInfoV2 = makeProperty(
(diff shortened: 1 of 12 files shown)


tobiasschweizer:
> @tobiasschweizer Do you think this will affect knora-api-js-lib?

knora-api-js-lib uses the complex schema. But it is better to be sure. Could you regenerate and integrate the client code and test data and see if there is any relevant change or if they unit tests break?

see https://github.com/dasch-swiss/knora-api-js-lib/blob/master/Makefile for detailed instructions.

benjamingeer:
The generated client code and test data are all in the complex schema. Also, I'm about to do a PR to remove the client code generation...

tobiasschweizer:
btw: @subotic is working on automating this in #1577

tobiasschweizer:
> Also, I'm about to do a PR to remove the client code generation...

I know. I need some instructions how to generate the test data for admin and v2. Or will the route stay the same?

benjamingeer:
> Or will the route stay the same?

I'll try to change it as little as possible.

tobiasschweizer:
> The generated client code and test data are all in the complex schema.

If this PR has no effects on data in the complex schema, that it won't affect knora-api-lib-js. But as I said, it is better to try it out ;-)

benjamingeer:
> If this PR has no effects on data in the complex schema

It doesn't affect data at all, only the `knora-api` simple ontology.

tobiasschweizer:
> I'll try to change it as little as possible.

great, so this should be easy

tobiasschweizer:
> It doesn't affect data at all, only the knora-api simple ontology.

Do you want me to try it out just to be sure?

benjamingeer:
> Do you want me to try it out just to be sure?

Yes please, if you have time.

tobiasschweizer:
ok, will do

tobiasschweizer:
looks fine, there is just a new element in `knora-api-ontology.json`:

```
{
    "@id" : "rdfs:label",
    "@type" : "owl:DatatypeProperty"
  }
```

benjamingeer:
OK to merge this then?

benjamingeer:
@tobiasschweizer Thanks very much for your help with this.

PR #1602, opened 2020-02-21
feat(api-v2): Return value UUID on value creation and update

### LINKED ISSUE #1595: Obtaining UUID on Value Creation
Is there a way to obtain a new value's UUID in the response to the value creation? 

I would like to read a new value back from Knora right after its creation . But in order to do that, I need its UUID. The only other possibility I can think of is making a full resource request but this is not ideal.

benjamingeer:
This PR makes API v2 return the UUID of a value when it is created or updated.

Resolves #1595.


benjamingeer:
@tobiasschweizer This is a bit more work than I thought it would be, because currently all the UUID creation is done in Twirl templates.

tobiasschweizer:
> This is a bit more work than I thought it would be, because currently all the UUID creation is done in Twirl templates.

It's not a problem if you need more time to work on this. So far, we still have a lot of work to do that does not depend on this PR.

benjamingeer:
> It's not a problem if you need more time to work on this.

It's done now.

tobiasschweizer:
Sorry, I kind of missed this. Thank you very much, I will do the review asap!

tobiasschweizer:
I will also do a PR on knora-api-js-lib and change the message classes.

tobiasschweizer:
Is there JSON test data for a successful response to a value creation / update?

tobiasschweizer:
I integrated this PR here: https://github.com/dasch-swiss/knora-api-js-lib/pull/151

tobiasschweizer:
@benjamingeer Do you think you could create test data for the responses too (see https://github.com/dasch-swiss/knora-api/pull/1602#issuecomment-590222947)?

benjamingeer:
@tobiasschweizer There's no need to ask me three times. :)

benjamingeer:
So far, the generation of test data doesn't actually update the repository. This is why there is no test response for creating a value. So I guess there are two options:

1. Actually create a value and return the response. Then you have to be aware that your repository has been updated and other tests may not work.
2. Return a simulated response. Then if the real response changes someday, the simulated response will have to change, too.

Which do you think is better?

tobiasschweizer:
> So far, the generation of test data doesn't actually update the repository. This is why there is no test response for creating a value. So I guess there are two options:
> 
> 1. Actually create a value and return the response. Then you have to be aware that your repository has been updated and other tests may not work.
> 2. Return a simulated response. Then if the real response changes someday, the simulated response will have to change, too.
> 
> Which do you think is better?

I think a simulated response would do it. Could you generate the simulated response from an instance of a case class that is then serialized to JSON-LD? If the original case class changes or its serialization in JSON-LD, the simulated response will too.

benjamingeer:
> Could you generate the simulated response from an instance of a case class that is then serialized to JSON-LD? If the original case class changes or its serialization in JSON-LD, the simulated response will too.

Good idea, I'll do that.

benjamingeer:
> Could you generate the simulated response from an instance of a case class that is then serialized to JSON-LD?

Done in e3e1727.

tobiasschweizer:
Thanks a lot for this! I will see if I can integrate this tomorrow, otherwise after my holiday.

benjamingeer:
> Could you make this constant?

Done in 673bff4.

webapi/src/main/twirl/queries/sparql/v2/changeLinkTarget.scala.txt
```diff
@@ -127,7 +128,7 @@ INSERT {
             knora-base:valueHasRefCount @linkUpdateForNewLink.newReferenceCount ;
             knora-base:valueHasOrder ?order ;
             knora-base:isDeleted false ;
-            knora-base:valueHasUUID "@{stringFormatter.base64EncodeUuid(UUID.randomUUID)}" ;
+            knora-base:valueHasUUID "@{stringFormatter.base64EncodeUuid(newLinkValueUUID)}" ;
```
tobiasschweizer: When the target of a link is changed, then the UUID is changed to? Is this because it has to be considered a new value (not a different version of the same value)?

benjamingeer: That's right. This was a design decision that Lukas and I made a long time ago. We thought that if you change the `rdf:object` of a reification, it wouldn't make sense to consider it a new version of the same reification, because it no longer reifies the same triple. So when you change a link's target, the old `LinkValue` is deleted, and a new one is created.

tobiasschweizer: I figure when you just change the comment, the uuid stays the same, right?

benjamingeer: Yes, if you just change the comment or any other metadata.

benjamingeer:
Thanks for asking for this and doing the review!

PR #1610, opened 2020-02-25
feat(api-v2): Remove client code generation

### LINKED ISSUE #1600: Make client tests not depend on incunabula
@tobiasschweizer Ivan asked me if we could remove the dependency on the Incunabula test data from the test requests/responses that are generated for `knora-api-js-lib`. I found only two things that depend on Incunabula:

- The test data for the search route uses `SharedTestDataADM.gravsearchComplexRegionsForPage`.
- The test for the `getProjectKeywordsTestResponse` function in the projects endpoint gets the keywords for the Incunabula project.

Can I change these so they use the `anything` test data instead?

### LINKED ISSUE #1607: Remove client code generation
Generating client code libraries would make sense if the libraries written in different languages were going to have the same structure. But it seems better for them to have different structures.

benjamingeer:
- [x] Remove client code generation (leaving client test data generation).
- [x] Remove external schema of admin ontology.
- [x] Update docs.
- [x] Remove client test data dependency on incunabula.
- [x] Change `JsonLDUtilSpec` so it doesn't start an `ActorSystem`.

Resolves #1607.
Resolves #1600.


benjamingeer:
@tobiasschweizer I don't know what needs to be changed in `knora-api-js-lib` for this, could you possibly help? I can probably do some of it if you can give me a list of things to do.

tobiasschweizer:
This PR will surely affect the way the generated test data is integrated into knora-api-js-lib.

The [integration scripts](https://github.com/dasch-swiss/knora-api-js-lib/tree/master/scripts) for the admin part will have to be adapted (see also https://github.com/dasch-swiss/knora-api-js-lib/blob/master/package.json and https://github.com/dasch-swiss/knora-api-js-lib/blob/master/Makefile). Also [E2E](https://github.com/dasch-swiss/knora-api-js-lib/blob/master/test-framework/e2e/src/app.e2e-spec.ts) test for the admin part will have to be written. For now, there are only unit tests for the admin part.

I can help you with that, but only after my holiday. I will be back to work on Monday, 9 March.

benjamingeer:
> I can help you with that, but only after my holiday. I will be back to work on Monday, 9 March.

OK no problem, let's do it then.

benjamingeer:
@subotic The JS Lib Tests are failing, do you know what I would need to change to make them work again? Or does this need help from Tobias?

benjamingeer:
@subotic Do you think this could be merged even though the JS Lib tests are failing? Then I wouldn't have to wait until Tobias gets back from holiday.

subotic:
Yeah sure.

benjamingeer:
Thanks for the review!

PR #1617, opened 2020-02-28
test: refactor to run against single knora-stack

### LINKED ISSUE #1611: Run tests against a single running knora-stack
Any tests that currently start `knora-api` should run against an externally started knora-stack.

Value proposition:
- allow using these tests as we use them now, but also to point them to any kind of knora-stack installation, and have it thoroughly checked (which we need but are currently missing for testing our "Infrastructure as Code")
- solve the intermittent BindException problem
- run tests a bit faster

subotic:
- [ ] ES: As a user, I would like for the Knora-API to auto initialize GraphDB at startup if the `KNORA_WEBAPI_TRIPLESTORE_AUTOINIT` flag is set. The DB will only be initialized if the triplestore check finds the triplestore in an uninitialized state.
  - [ ] S: As a user, I would like to have a library to connect to GraphDB and create a new repository
     - [ ] T: Add library
  - [ ] S: As a user, I would like to add custom rules (KnoraRules.pie) to a created repository
     - [ ] T: Add loading of KnoraRules.pie to a repository
  - [ ] T: Use library to create repository if it does not exists  

resolves #1611


subotic:
deprecated

PR #1628, opened 2020-03-10
build (ci): add node version

### LINKED ISSUE #1627: Failing github-ci JS lib
I had the same issue on the knora-api-lib-js repo. The solution was to run the JS lib tests with node version 11: https://github.com/dasch-swiss/knora-api-js-lib/blob/03e74aec56d7d005b866e6f48ddf6b771fa280a0/.github/workflows/main.yml#L14-L41

subotic:
resolves #1627

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -175,10 +175,11 @@ jobs:
         run: |
           make env-file
           make print-env-file
-      - name: run JS lib tests
+      - name: setup required node version
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
+      - name: run JS lib tests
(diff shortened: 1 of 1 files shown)


tobiasschweizer:
could not even parse it: "a step cannot have both the uses and run keys"

I had this problem too, but do not remember the solution right now

subotic:
It needs to run two separate steps. Should run now.

tobiasschweizer:
ok, let's see ;-)

tobiasschweizer:
looks good, I think this is going to work!

tobiasschweizer:
ok, great!

subotic:
great, thanks for the review :-)

PR #1633, opened 2020-03-31
feat(clientapi): Return a test text value as HTML

### LINKED ISSUE #1629: Resource Testding doesn't return a TextValueAsHtml
Can we update this resource to return a TextValueAsHtml?

benjamingeer:
User story [DSP-54](https://dasch.myjetbrains.com/youtrack/issue/DSP-54).

Resolves #1629.


benjamingeer:
This is not actually going to be feasible.

PR #1637, opened 2020-04-22
fix(OntologyResponderV2): Add a global ontology cache lock

benjamingeer:
Fixes https://dasch.myjetbrains.com/youtrack/issue/DSP-181.


benjamingeer:
@lrosenth Please try this branch. There was already an update lock on each ontology, but that doesn't help if you have concurrent updates to different ontologies, so I used a global update lock instead. See if this fixes the problem.

benjamingeer:
@subotic Thanks for the review!

PR #1642, opened 2020-04-28
build: upgrade akka

subotic:


subotic:
Thanks for the review. Yes, I would like to try to upgrade to Scala 2.13.2 in another PR.

PR #1643, opened 2020-04-30
feat(upgrade): Update repository on startup

benjamingeer:
User story:

https://dasch.myjetbrains.com/youtrack/issue/DSP-165

- [x] Move the code from the `upgrade` package into `webapi`, translating the Bash scripts into Scala, and incorporating the code into `ApplicationActor` and the new classes `RepositoryUpdater` and `RepositoryUpdatePlan`.
- [x] Implement the necessary HTTP operations (download and upload the whole repository) in `HttpTriplestoreConnector`.
- [x] Add an upgrade plugin for PR #1615, to remove the instance of `forbiddenResource` from the Knora system data. (That PR only removed the `ForbiddenResource` class, and GraphDB's consistency checker rejects the instance if the class doesn't exist anymore.)
- [x] In `ApplicationActor` and its messages, rename the parameter `skipLoadingOfOntologies` (i.e. don't load ontologies from the triplestore into the cache) to `ignoreRepository`. When Knora is started for a test, before the test data is loaded into the triplestore, this is `true`. We can also use it to skip checking whether we should update the repository.
- [x] Rename some things in `ApplicationMessages` and `TriplestoreMessages` to distinguish clearly between the triplestore and the repository.
- [x] Use a sealed trait instead of an `Enumeration` for `AppState`, so the compiler can check that all cases are matched (e.g. in `HealthRoute`).
- [x] Update the GitHub action for testing repository update.
  - [x] Add a script `webapi/scripts/wait-for-knora.sh` that polls Knora's `/health` route until Knora has finished starting up.
  - [x] Add a Makefile target `test-repository-update` that loads old repository test data into Knora, restarts Knora, and waits for the update to complete.
- [x] Update docs.


benjamingeer:
@subotic I've commented out the GitHub CI `upgrade-integration-tests`, because I removed the scripts that it was using. How do you want to do this now?

subotic:
We could add a new one based on the new feature, e.g., start the stack without api, load the v7 data, start the api and then check if it starts up.

For most of the steps there is a makefile target. Loading V7 data and checking if knora-api is running could be done with curl.

Would you like to try to do this?

benjamingeer:
@subotic 

> Would you like to try to do this?

I don't really understand how the Docker containers fit together, but I can try. How do I check whether it the API started without errors?

benjamingeer:
@subotic @nemmK Thank you both for the reviews!

PR #1645, opened 2020-05-07
default size for restricted view (DSP-218)

loicjaouen:
restricted view image size now defaults to thumbnail size

- adds a config item for default restricted view
- sets it to a percentage of the original image to a resolution limitation instead of a size limitation
- use it in pre_flight


loicjaouen:
giving up on the idea of setting a specific argument in sipi config because even though it's trivial, it has to be a coordinated release

benjamingeer:
@loicjaouen 

> even though it's trivial, it has to be a coordinated release

We're working on a plan for asynchronous releases, so this shouldn't be a problem. Will send you a link to the draft.

loicjaouen:
thanks :)

PR #1646, opened 2020-05-13
feat(api-v2): Specify custom IRIs when creating resources/values

benjamingeer:
User story: https://dasch.myjetbrains.com/youtrack/issue/DSP-159

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
@@ -90,17 +90,17 @@ object CreateValueRequestV2 extends KnoraJsonLDRequestReaderV2[CreateValueReques
 
         for {
             // Get the IRI of the resource that the value is to be created in.
-            resourceIri: SmartIri <- Future(jsonLDDocument.getIDAsKnoraDataIri)
+            resourceIri: SmartIri <- Future(jsonLDDocument.requireIDAsKnoraDataIri)
 
             _ = if (!resourceIri.isKnoraResourceIri) {
                 throw BadRequestException(s"Invalid resource IRI: <$resourceIri>")
             }
 
             // Get the resource class.
-            resourceClassIri: SmartIri = jsonLDDocument.getTypeAsKnoraTypeIri
+            resourceClassIri: SmartIri = jsonLDDocument.requireTypeAsKnoraTypeIri
 
             // Get the resource property and the value to be created.
-            createValue: CreateValueV2 <- jsonLDDocument.getResourcePropertyValue match {
+            createValue: CreateValueV2 <- jsonLDDocument.requireResourcePropertyValue match {
                 case (propertyIri: SmartIri, jsonLDObject: JsonLDObject) =>
                     if (jsonLDObject.value.get(JsonLDConstants.ID).nonEmpty) {
                         throw BadRequestException("The @id of a value cannot be given in a request to create the value")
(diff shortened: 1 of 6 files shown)


SepidehAlassi:
@benjamingeer looking at the code I see that: when a resource with multiple values is being created, the `creationDate` of the resource is also assigned to its values. Since a resource can be created with custom `creationData`, I believe that is enough for the checksum. Then the values will have the correct `creationDate` anyway. What do you think? 

In my opinion, custom `creationDate` should only be used for creating single values, i.e. `CreateValueRequestV2` should accept a custom `creationDate` for a value.

benjamingeer:
Yes, good analysis!

SepidehAlassi:
The following refactoring step 
- Refactor resource creation so that generating random resource IRIs, and checking new resource IRIs for uniqueness, happens in `ResourcesResponderV2` while holding a lock on the new IRI.

will happen in [DSP-355](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-30?p=110-4%2F111-29&issue=DSP-355)

SepidehAlassi:
@benjamingeer eventually the tests passed!

docs/src/paradox/03-apis/api-v2/editing-resources.md
```diff
@@ -185,6 +185,39 @@ than the requesting user only if the requesting user is an administrator of the
...
+   "@id" : "http://rdfh.ch/0001/a-custom-thing",
+   "@type" : "anything:Thing",
+   "knora-api:attachedToProject" : {
+     "@id" : "http://rdfh.ch/projects/0001"
+   },
+   "anything:hasInteger" : {
+       "@id" : "http://rdfh.ch/0001/a-thing/values/int-value-IRI",
+       "@type" : "knora-api:IntValue",
+       "knora-api:intValueAsInt" : 10,
+       "knora-api:valueHasUUID" : "IN4R19yYR0ygi3K2VEHpUQ"
```
benjamingeer, line 204: We should probably say that the value UUID needs to be a real Base64-encoded UUID, so it is mainly useful for transferring data from one Knora repository to another.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
```diff
@@ -117,12 +114,25 @@ object CreateValueRequestV2 extends KnoraJsonLDRequestReaderV2[CreateValueReques
                                 log = log
                             )
 
+                        // Get the custom value IRI if provided.
+                        maybeCustomValueIri: Option[SmartIri] = jsonLDObject.maybeIDAsKnoraDataIri
+                        // Get the custom value UUID if provided.
```
benjamingeer, line 119: Please add a blank line before a comment.

webapi/src/main/scala/org/knora/webapi/responders/v2/ValuesResponderV2.scala
```diff
@@ -354,27 +372,41 @@ class ValuesResponderV2(responderData: ResponderData) extends Responder(responde
...
-            newValueIri <- FastFuture.successful(stringFormatter.makeRandomValueIri(resourceInfo.resourceIri))
-            newValueUUID = UUID.randomUUID
-
-            currentTime: Instant = Instant.now
+            // Make an IRI and a UUID for the new value.
+            newValueIri: IRI <- maybeValueIri match {
+                case Some(customValueIri) => FastFuture.successful(customValueIri.toString)
+                case None => FastFuture.successful(stringFormatter.makeRandomValueIri(resourceInfo.resourceIri))
+            }
+            newValueUUID: UUID = maybeValueUUID match {
```
benjamingeer, line 402: Please add a blank line before any line that contains an opening brace (`{`).

docs/src/paradox/03-apis/api-v2/editing-resources.md
benjamingeer: I.e. you can't use `"my-custom-UUID"` or any random string like that.

SepidehAlassi:
> After thinking about this some more, I think we actually need to accept a custom value creation date for each value even when creating a new resource. 

In that case, if no custom creation date is provided for the value of a new resource, the creation date of the resource should be assigned to the value, right?

SepidehAlassi:
@benjamingeer last commit contains the implementation "create resources with values that have custom creation dates". Can you please tell me your opinion?

webapi/src/main/scala/org/knora/webapi/responders/v2/ValuesResponderV2.scala
```diff
@@ -588,30 +594,40 @@ class ValuesResponderV2(responderData: ResponderData) extends Responder(responde
...
-     * @param propertyIri    the IRI of the property that will point to the value.
-     * @param valueToCreate  the value to be created.
-     * @param valueHasOrder  the value's `knora-base:valueHasOrder`.
-     * @param creationDate   the timestamp to be used as the value creation time.
-     * @param requestingUser the user making the request.
+     * @param resourceIri           the IRI of the resource.
+     * @param propertyIri           the IRI of the property that will point to the value.
+     * @param valueToCreate         the value to be created.
+     * @param valueHasOrder         the value's `knora-base:valueHasOrder`.
+     * @param resourceCreationDate  the timestamp to be used as the value creation time.
```
benjamingeer, line 601: Comment needs fixing.

benjamingeer:
> In that case, if no custom creation date is provided for the value of a new resource, the creation date of the resource should be assigned to the value, right?

Yes.

I just talked to Ivan about some additional complexity in copying a repository. The result is that there's a new user story for additional functionality (https://dasch.myjetbrains.com/youtrack/issue/DSP-398). This would also require being able to specify:

- a custom value creation date when adding a new version of a value
- a custom delete date when marking a value as deleted

Do you want to do those in this PR?

SepidehAlassi:
> Do you want to do those in this PR?

No, I prefer to do it in a separate PR.

benjamingeer:
> I prefer to do it in a separate PR.

OK then I think this one is good to merge! You can approve it yourself. :)

PR #1649, opened 2020-05-26
feat(api-v2): Allow querying for rdfs:label in Gravsearch

benjamingeer:
User story: https://dasch.myjetbrains.com/youtrack/issue/DSP-271

- [x] Add `OntologyConstants.ResourceMetadataPropertyAxioms` specifying the object type of `rdfs:label`, so the type inspection mechanism knows it (by reading it in `IntermediateTypeInspectionResult.newInstance`).
- [x] Change `AbstractPrequeryGenerator` so it doesn't generate any extra statements for a variable that is the object of `rdfs:label`.
- [x] Add a `knora-api:matchLabel` function to `AbstractPrequeryGenerator`, so you can do Lucene full-text searches in labels.
  - [x] In `InferringGravsearchTypeInspector`, infer that the first argument of `knora-api:matchLabel` is a resource.
- [x] Support the `regex` function with `rdfs:label` in `AbstractPrequeryGenerator`.
- [x] Add tests:
  - [x] In `GravsearchTypeInspectorSpec` for type inspection
  - [x] In `NonTriplestoreSpecificGravsearchToPrequeryTransformerSpec` for generating the prequery
  - [x] In `SearchRouteV2R2RSpec` for searching for a resource by label:
    - [x] Using a literal in the simple schema
    - [x] Using a literal in the complex schema
    - [x] Using a variable and a `FILTER` in the simple schema
    - [x] Using a variable and a `FILTER` in the complex schema
    - [x] Using `knora-api:matchLabel` in the simple schema
    - [x] Using `knora-api:matchLabel` in the complex schema
    - [x] Using `regex` in the simple schema
    - [x] Using `regex` in the complex schema
- [x] Update docs (`query-language.md`).

### DIFF
--- a/docs/src/paradox/03-apis/api-v2/query-language.md
+++ b/docs/src/paradox/03-apis/api-v2/query-language.md
@@ -616,6 +616,39 @@ CONSTRUCT {
 }
 ```
 
+### Filtering on `rdfs:label`
+
+The `rdfs:label` of a resource is not a Knora value, but you can still search for it.
+This can be done in the same ways in the simple or complex schema:
+
+Using a string literal object:
+
+```
+?book rdfs:label "Zeitglöcklein des Lebens und Leidens Christi" .
+```
+
+Using a variable and a FILTER:
+
+```
+?book rdfs:label ?label .
+FILTER(?label = "Zeitglöcklein des Lebens und Leidens Christi")
+```
+
+Using the `regex` function:
+
+```
+?book rdfs:label ?bookLabel .
+FILTER regex(?bookLabel, "Zeit", "i")
+```
+
+To match words in an `rdfs:label` using the full-text search index, use the
+`knora-api:matchLabel` function, which works like `knora-api:matchText`,
+except that the first argument is a variable representing a resource:
+
+```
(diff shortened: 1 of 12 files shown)


tobiasschweizer:
I am looking forward to searching for labels!

benjamingeer:
Note to self: `apply` is better than `newInstance`.

benjamingeer:
Thanks for the review!

> do we still need to keep the separate route searchbylabel?

If I remember correctly, that route is mainly for populating a GUI text box representing a resource IRI in the advanced search widget, so the user can type “Euler” and get Euler’s IRI to use in the advanced search. I guess having a route for this is simpler than making a Gravsearch query, and keeping it means we don’t have to change existing client code.

tobiasschweizer:
> I guess having a route for this is simpler than making a Gravsearch query, and keeping it means we don’t have to change existing client code.

Maybe you could use Gravsearch internally at some point, and the client would still communicate with the same route. I think we cannot restrict by project yet in Gravsearch: HTTP GET to http://host/v2/searchbylabel/searchValue[limitToResourceClass=resourceClassIRI]
[limitToProject=projectIRI][offset=Integer]

PR #1650, opened 2020-05-28
test: Fix typo

benjamingeer:
In `SearchRouteV2R2RSpec`, `private val writeTestDataFiles = true` should be `false` so the test is actually run.


benjamingeer:
@SepidehAlassi So you can see the kind of silly mistakes I make. :)

SepidehAlassi:
> @SepidehAlassi So you can see the kind of silly mistakes I make. :)

:-D Still not as obviously silly as mine!

benjamingeer:
@SepidehAlassi Thanks!

PR #1654, opened 2020-06-10
Create List with custom IRI

SepidehAlassi:
Resolves [DSP-245](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-32?issue=DSP-245)

- [x] Create lists with custom IRI
- [x] add tests
- [x] update documentation

### DIFF
--- a/docs/src/paradox/03-apis/api-admin/lists.md
+++ b/docs/src/paradox/03-apis/api-admin/lists.md
@@ -64,6 +64,15 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
         "comments": []
     } 
     ```
+  Each list can have an optional custom IRI specified by the `listIri` in the request body as below:
+  ```jsonld
+    {
+        "projectIri": "http://rdfh.ch/projects/0001",
+        "listIri": "http://rdfh.ch/lists/0001/a-list-with-IRI",
+        "labels": [{ "value": "Neue Liste mit IRI", "language": "de"}],
+        "comments": []
+    }
+  ```
 
 ### Create new child node
 
--- a/webapi/src/main/scala/org/knora/webapi/SharedTestDataADM.scala
+++ b/webapi/src/main/scala/org/knora/webapi/SharedTestDataADM.scala
@@ -652,7 +652,7 @@ object SharedTestDataADM {
     val createListWithCustomIriRequest: String =
         s"""{
            |    "projectIri": "${SharedTestDataADM.ANYTHING_PROJECT_IRI}",
-           |    "listIri": "${SharedTestDataADM.customListIRI}"
+           |    "listIri": "${SharedTestDataADM.customListIRI}",
            |    "labels": [{ "value": "Neue Liste mit IRI", "language": "de"}],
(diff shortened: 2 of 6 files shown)


SepidehAlassi:
> From my understanding of [RFC2616](https://tools.ietf.org/html/rfc2616#section-9.6), when creating a new resource with a supplied ID then it should be a `PUT` request, e.g.,
> 
> * the client chooses the IRI: `PUT /admin/lists/IRI`
> * the server chooses the IRI: `POST /admin/lists`

I understand that RFC and also some documentation about the use HTTP  methods in Rest API support using PUT method for some creation scenarios. In my opinion, however, this is a debatable matter and depends on the design of the architecture of our API.

Let's think about it from a user's point of view. As a user, I know that I have to submit a POST request to create a new list. If we design the architecture so that for the creation of a list with supplied IRI, PUT method should be used, then the user has to think of which method to use every time creating a list depending on if she supplies a custom IRI or not in the request body. 

I advocate for simplicity, i.e. in both cases of supplying an IRI or not,  POST method should be used for the creation and PUT to replace/modify. @benjamingeer what do you think about this?

SepidehAlassi:
> If your only objection is that “having all requests for the admin API in one file is in my opinion not very readable“, why don’t you make a PR to split it into several files?

Well, in any case, if we decide to refactor the code as @subotic suggests or not, this has to happen in another PR, as it would be a major redesign.

benjamingeer:
> I advocate for simplicity, i.e. in both cases of supplying an IRI or not, POST method should be used for the creation and PUT to replace/modify. @benjamingeer what do you think about this?

I think that the traditional distinctions between `GET`, `POST`, `PUT`, and `DELETE` don't make sense in today's world. The W3C has recognised this: the SPARQL 1.1 protocol [accepts queries via `POST`](https://www.w3.org/TR/sparql11-protocol/#query-via-post-direct), because `GET` requests don't have a body, and SPARQL requests can be too long to include in the request URL. Similarly, Knora accepts delete requests via `POST`, because HTTP `DELETE` doesn't have a body, either.

I think most people are used to using `POST` for creating and `PUT` for updating, and that's what Knora does, too, so I suggest staying with that.

SepidehAlassi:
> I think most people are used to using `POST` for creating and `PUT` for updating, and that's what Knora does, too, so I suggest staying with that.

Exactly! I believe it would be the simplest case for users if we stick to this rule.

subotic:
I would like to set up a meeting for this afternoon to discuss these points. I will send you an invite.

benjamingeer:
16:00 is probably too late in the day for me today, I will try.

I would also like to point out that the SPARQL update protocol uses HTTP POST both for creating new data and for updating existing data:

https://www.w3.org/TR/2013/REC-sparql11-update-20130321/#terminology

SepidehAlassi:
@subotic is there any other point that you feel must be addressed regarding this PR?

SepidehAlassi:
> @SepidehAlassi Nope. LGTM. Thanks for implementing it :-)

Thanks for reviewing it!

PR #1655, opened 2020-06-10
Refactor entity creation to happen holding lock on the new IRI

SepidehAlassi:
Resolves [DSP-355](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-32?issue=DSP-355)

- [x] add the method `checkEntityIRI` to `Responder` class to check a custom IRI for uniqueness and to generate random IRI for entities without custom IRI using `stringFormatter.makeUnusedIri` so that generation of random IRIs repeats until hitting an unused one.

- - [x] refactor resource creation so that generating random resource IRIs, and checking new resource IRIs for uniqueness, happens in `ResourcesResponderV2` through `checkEntityIRI`  while holding a lock on the new IRI.

- - [x] refactor value creation so that generating random value IRIs, and checking new value IRIs for uniqueness, happens in `ValuesResponderV2` through `checkEntityIRI`  while holding a lock on the new IRI.

- - [x] refactor project creation so that generating random project IRIs, and checking new project IRIs for uniqueness, happens in `ProjectsResponderV2` through `checkEntityIRI`  while holding a lock on the new IRI.

- - [x] refactor list creation so that generating random list IRIs, and checking new list IRIs for uniqueness, happens in `ListsResponderV2` through `checkEntityIRI`  while holding a lock on the new IRI.


SepidehAlassi:
@benjamingeer in this PR, I have implemented the refactoring we talked about for having a method that must check the custom IRIs for the uniqueness, if not given it creates an unused random one. This function is used now for checking/generating all entity IRIs in create mode (resource, value, project, list). Later it can be used for creating other entities with custom Iris: user, group, etc. 

Can you please check if it fulfills all you had in mind? especially the part where resource Iri generation is moved to the `ResourceResponderV2` to run holding a lock on the IRI. 

I am going to add a couple more tests for checking the uniqueness of custom Iris given for resources and values.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -224,11 +226,13 @@ class ResourcesResponderV2(responderData: ResponderData) extends ResponderWithSt
                 throw ForbiddenException(s"User ${createResourceRequestV2.requestingUser.username} does not have permission to create a resource of class <${createResourceRequestV2.createResource.resourceClassIri}> in project <$projectIri>")
             }
 
+            resourceIri: IRI <- checkEntityIRI(createResourceRequestV2.createResource.resourceIri, stringFormatter.makeRandomResourceIri(createResourceRequestV2.createResource.projectADM.shortcode))
```
benjamingeer, line 229: This needs to happen inside `makeTaskFuture`, so it's done while we're holding a lock on the IRI. Otherwise, two concurrent requests could do the same check at the same time, and both find that same IRI is unused, and both try to use it.

benjamingeer:
Also, a minor stylistic convention: in the Knora code, `IRI` is written `Iri` if it's in a method name, like `checkEntityIri`.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
SepidehAlassi: Question: As far as I have got, the iri that must be locked should be passed to `runWithIriLock` function:
`def runWithIriLock[T](apiRequestID: UUID, iri: IRI, task: () => Future[T])`
Therefore I have formulated the resourceIri to pass it to this function, and also passed it as an argument to `makeTaskFuture`.  Wouldn't it do the trick? if I move the formulating Iri to `makeTaskFuture`, then how can I tell the locker which Iri to lock? The resourceIri member of `createResourceRequestV2` is only the Option[SmartIri].

SepidehAlassi:
@benjamingeer Thanks for your changes, and teaching me about scala especially the futures.

> There's one other thing, which I didn't do: `StringFormatter.checkIriExists` does the same thing as `Responder.checkEntityExists`. The SPARQL templates they use, `queries.sparql.admin.txt.checkIriExists` and `queries.sparql.v2.txt.checkEntityExists`, also do the same thing. I'm not sure what would be the best way to get rid of this redundancy, what do you think?

I think maybe it is better to keep the `StringFormatter.checkIriExists` and get rid of `Responder.checkEntityExists`. Stringformatter takes care of making unused Iris, therefore in my opinion it would make sense if checking the iri happens in the same place. 

Similarly, I believe `queries.sparql.admin.txt.checkIriExists` should be kept and `queries.sparql.v2.txt.checkEntityExists` should be removed.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
benjamingeer: You're right, good question! :)

If the user provided an IRI, we have to:

1. Get the lock by passing that IRI to `runWithIriLock`.
2. In `makeTaskFuture` (i.e. while holding the lock), check if the IRI is already used.

So maybe we need to change our design a little bit. How do you think we could do this?

If the user didn't provide an IRI, the situation is a bit more complicated, because we want to retry. But maybe in this case, we can skip getting the lock, what do you think? It seems incredibly unlikely that we would randomly generate two identical IRIs **at the same time**, unless something is seriously wrong with the random number generator. What do you think?

benjamingeer:
> I think maybe it is better to keep the `StringFormatter.checkIriExists` and get rid of `Responder.checkEntityExists`. Stringformatter takes care of making unused Iris, therefore in my opinion it would make sense if checking the iri happens in the same place. 
> 
> Similarly, I believe `queries.sparql.admin.txt.checkIriExists` should be kept and `queries.sparql.v2.txt.checkEntityExists` should be removed.

OK, sounds good to me, please do that.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
benjamingeer: Clarification:

> Therefore I have formulated the resourceIri to pass it to this function, and also passed it as an argument to makeTaskFuture. Wouldn't it do the trick?

Your code does this:

1. Check if the IRI is used.
2. Then lock it.

But this could happen:

1. Client A checks if the IRI is used, and finds that it is not.
2. Client B checks if the same IRI is used, and finds that it is not.
3. Client A gets the lock and creates the resource.
4. Client B then gets the lock and tries to create a resource with the same IRI.

This would probably result in data corruption in the triplestore.

That's why we need to do the check while holding the lock.

SepidehAlassi: > If the user provided an IRI, we have to:
> 
> 1. Get the lock by passing that IRI to `runWithIriLock`.
> 2. In `makeTaskFuture` (i.e. while holding the lock), check if the IRI is already used.
> 
> So maybe we need to change our design a little bit. How do you think we could do this?
> 

> If the user didn't provide an IRI, the situation is a bit more complicated, because we want to retry. But maybe in this case, we can skip getting the lock, what do you think? It seems incredibly unlikely that we would randomly generate two identical IRIs **at the same time**, unless something is seriously wrong with the random number generator. What do you think?

Well, the point of having `checkEntityIri` method was to have a consistent way of checking/generating Iris for new entities (resources, values, projects, etc). 

I would suggest keeping things simple and consistent by leaving the formulation of resourceIri by `checkEntityIri` before calling the `runWithIriLock` (as it is now), then passing the resourceIri to `runWithIriLock` and eventually adding an additional check inside `makeTaskFuture` to see if the IRI (customized or random) is already used while holding a lock on it. In this way, we check if the IRI already exists twice, once in `checkEntityIri` and once in `makeTaskFuture` while holding a lock. It seems paranoid and even redundant but makes things consistent and simple.

benjamingeer: OK, let's check it twice as you say. Probably the SPARQL query to do the check will be very fast, so it shouldn't be a performance problem to do it twice.

SepidehAlassi:
> OK, sounds good to me, please do that.

Done!

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
SepidehAlassi: @benjamingeer done! can you please finalize the review of this PR now. I also added some tests to check the uniqueness of custom iris of resources and values.

SepidehAlassi:
> This looks great now, thank you very much!

Thanks for reviewing it.

Issue #1620, opened 2020-03-02
typos in admin doc


loicjaouen:
incidentally browsing admin docs that fails to render nicely:
https://docs.knora.org/paradox/03-apis/api-admin/projects.html

![Capture d’écran 2020-03-02 à 10 34 35](https://user-images.githubusercontent.com/3933654/75663634-727bd780-5c71-11ea-90d6-e8c4652ca1e6.png)

SepidehAlassi:
resolved in PR #1656

PR #1656, opened 2020-06-11
Create a Project with Custom IRI

SepidehAlassi:
Resolves [DSP-242](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-32?issue=DSP-242)

- [x] allow the creation of a new project with a supplied IRI
- [x] add tests for a valid project IRI, and creation of the new project with assigned IRI
- [x] add test data for create-project-with-custom-IRI
- [x] update documentation
- [x] closes issue #1620 
- [x] add check for validity of custom list IRI

### DIFF
--- a/docs/src/paradox/03-apis/api-admin/projects.md
+++ b/docs/src/paradox/03-apis/api-admin/projects.md
@@ -80,7 +80,20 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
       "selfjoin": false
     }
     ```
-
+    Each project can have an optional custom IRI specified by the `projectIri` in the request body as below:
+    ```JSON
+    {
+        "projectIri": "http://rdfh.ch/projects/3333",
+        "shortname": "newprojectWithIri",
+        "shortcode": "3333",
+        "longname": "new project with a custom IRI",
+        "description": [{"value": "a project created with a custom IRI", "language": "en"}],
+        "keywords": ["projectWithIRI"],
+        "logo": "/fu/bar/baz.jpg",
+        "status": true,
+        "selfjoin": false
+    }   
+    ```
 ### Update project information:
 
   - Required permission: SystemAdmin / ProjectAdmin
--- a/docs/src/paradox/03-apis/api-admin/projects.md
+++ b/docs/src/paradox/03-apis/api-admin/projects.md
@@ -21,31 +21,37 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
 
 ## Endpoint Overview
 
-**Project Operations:**
+**Project Operations:**  
 
-- `GET: /admin/projects` : return all projects
-- `POST: /admin/projects` : create a new project
-- `GET: /admin/projects/[iri | shortname | shortcode]/<identifier>` : returns a single project identified either through iri, shortname, or shortcode
-- `PUT: /admin/projects/iri/<identifier>` : update a project identified by iri
-- `DELETE: /admin/projects/iri/<identifier>` : update project status to false
-- `GET: /admin/projects/iri/<identifier>/AllData` : returns a TriG file containing the project's data
+- `GET: /admin/projects` : return all projects  
 
-**Project Member Operations:**
+- `POST: /admin/projects` : create a new project  
 
-- `GET: /admin/projects/[iri | shortname | shortcode]/<identifier>/members` : returns all members part of a project identified through iri, shortname or shortcode
+- `GET: /admin/projects/[iri | shortname | shortcode]/<identifier>` : returns a single project identified either through iri, shortname, or shortcode  
 
-**Project Admin Member Operations:**
+- `PUT: /admin/projects/iri/<identifier>` : update a project identified by iri  
 
-- `GET: /admin/projects/[iri | shortname | shortcode]/<identifier>/admin-members` : returns all admin members part of a project identified through iri, shortname or shortcode
+- `DELETE: /admin/projects/iri/<identifier>` : update project status to false  
 
-**Project Keyword Operations:**
+- `GET: /admin/projects/iri/<identifier>/AllData` : returns a TriG file containing the project's data  
 
-- `GET: /admin/projects/Keywords` : returns all unique keywords for all projects as a list
-- `GET: /admin/projects/iri/<identifier>/Keywords` : returns all keywords for a single project
+**Project Member Operations:**  
 
-**Project Restricted View Settings Operations:**
+- `GET: /admin/projects/[iri | shortname | shortcode]/<identifier>/members` : returns all members part of a project identified through iri, shortname or shortcode  
 
-- `GET: /admin/projects/iri/<identifier>/RestrictedViewSettings` : returns the project's restricted view settings
+**Project Admin Member Operations:**  
+
+- `GET: /admin/projects/[iri | shortname | shortcode]/<identifier>/admin-members` : returns all admin members part of a project identified through iri, shortname or shortcode  
+
+**Project Keyword Operations:**  
+
+- `GET: /admin/projects/Keywords` : returns all unique keywords for all projects as a list  
+
+- `GET: /admin/projects/iri/<identifier>/Keywords` : returns all keywords for a single project  
+
+**Project Restricted View Settings Operations:**  
+
+- `GET: /admin/projects/iri/<identifier>/RestrictedViewSettings` : returns the project's restricted view settings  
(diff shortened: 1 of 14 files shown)


webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -762,46 +761,47 @@ class ProjectsResponderADM(responderData: ResponderData) extends Responder(respo
...
             // check if the optionally supplied shortcode is valid and unique
             shortcodeExists <- {
                 projectByShortcodeExists(validatedShortcode)
             }
             _ = if (shortcodeExists) {
-                throw DuplicateValueException(s"Project with the shortcode: '${createRequest.shortcode}' already exists")
+                throw DuplicateValueException(s"Project with the shortcode: '${createProjectRequest.shortcode}' already exists")
+            }
+            newProjectIRI: IRI = createProjectRequest.projectIri match {
+                case Some(customProjectIri) => customProjectIri.toString
```
subotic, line 782: Where do you validate the custom IRI? It needs to be a valid project IRI, or you will not be able to retrieve the project.

subotic: and you don't need the `.toString` since it is already a string.

subotic:
Sorry, but I really cannot find the place where you validate the custom IRI. But the e2e test fails, so it must be somewhere. Could you point me to the place?

SepidehAlassi:
> Sorry, but I really cannot find the place where you validate the custom IRI. But the e2e test fails, so it must be somewhere. Could you point me to the place?

I do not do an extra verification, the triplestore does not accept the given invalid IRI. I can, however, add an extra verification layer to enforce the given project IRI to be in the form `http://$IriDomain/projects/$shortcode`

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -762,46 +761,48 @@ class ProjectsResponderADM(responderData: ResponderData) extends Responder(respo
...
                 projectByShortcodeExists(validatedShortcode)
             }
             _ = if (shortcodeExists) {
-                throw DuplicateValueException(s"Project with the shortcode: '${createRequest.shortcode}' already exists")
+                throw DuplicateValueException(s"Project with the shortcode: '${createProjectRequest.shortcode}' already exists")
             }
 
-            newProjectIRI = stringFormatter.makeRandomProjectIri(validatedShortcode)
-
+            newProjectIRI: IRI = createProjectRequest.projectIri match {
```
SepidehAlassi, line 782: @subotic isn't it better to leave this here, because here it runs while holding a lock on the IRI?

subotic: In the route, i.e., the case class you check if the custom project IRI is valid. Here you would need to additionally check if the custom project IRI is not used already used. At this point, you only now that it is valid, but it doesn't mean that it is unique. This would also mean, that you need a test to check if you get the correct error if you supply an already existing custom project IRI.

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -778,11 +778,12 @@ class ProjectsResponderADM(responderData: ResponderData) extends Responder(respo
             _ = if (shortcodeExists) {
                 throw DuplicateValueException(s"Project with the shortcode: '${createProjectRequest.shortcode}' already exists")
             }
+
             newProjectIRI: IRI = createProjectRequest.projectIri match {
-                case Some(customProjectIri) => customProjectIri.toString
-                case None => stringFormatter.makeRandomProjectIri(validatedShortcode)
+              case Some(customProjectIri) => stringFormatter.toSmartIriWithErr(customProjectIri, throw BadRequestException(s"Invalid project IRI")).toString
```
benjamingeer, line 783: `StringFormatter.toSmartIriWithErr` checks that it's a valid IRI, but not that it's a valid project IRI. It looks like there's a method for that: `StringFormatter.validateProjectIri`, which is already used a lot in `ProjectsRouteADM`. In general, Knora does input format validation in routes rather than in responders.

SepidehAlassi: @benjamingeer thanks for reviewing this. I just added the validator as you suggested for the custom project IRI. Since the same validation was missing in the case of custom list IRI, I also added that in this PR. I hope it is OK. Is there anything else you believe must be added?

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -780,7 +779,7 @@ class ProjectsResponderADM(responderData: ResponderData) extends Responder(respo
             }
 
             newProjectIRI: IRI = createProjectRequest.projectIri match {
-              case Some(customProjectIri) => stringFormatter.toSmartIriWithErr(customProjectIri, throw BadRequestException(s"Invalid project IRI")).toString
+              case Some(customProjectIri) => stringFormatter.validateProjectIri(customProjectIri, throw BadRequestException(s"Invalid project IRI")).toString
```
benjamingeer, line 782: Could you do this in the route, rather than in the responder, to make things consistent?

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
SepidehAlassi: You are absolutely right, thanks for the heads up. I added validation to route, and checks for uniqueness to the responder + tests for uniqueness.

docs/src/paradox/03-apis/api-admin/projects.md
```diff
@@ -74,7 +80,20 @@ License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
       "selfjoin": false
     }
     ```
-
+    Each project can have an optional custom IRI specified by the `projectIri` in the request body as below:
+    ```JSON
+    {
+        "projectIri": "http://rdfh.ch/projects/3333",
+        "shortname": "newprojectWithIri",
```
subotic, line 87: I don't believe that the format of the custom IRI was documented somewhere in the past. Maybe the format that we expect and validate for could be mentioned somewhere in the docs?

benjamingeer: Should we require the normal project IRI format?

https://docs.knora.org/paradox/03-apis/api-v2/knora-iris.html#iris-for-data

subotic: I think so. This way, each project has it's own namespace that they can use however they believe it's best, without interfering with other projects. Also, if we allow any kinds of IRIs, then we would need to remove validation in all places and maybe even change the ARK server.

subotic: And thanks for the link. We could link to the page from here.

webapi/src/test/scala/org/knora/webapi/e2e/admin/ProjectsADME2ESpec.scala
```diff
@@ -146,6 +146,72 @@ class ProjectsADME2ESpec extends E2ESpec(ProjectsADME2ESpec.config) with Session
...
+                //check the rest of project info
+                result.shortcode should be ("3333")
+                result.shortname should be ("newprojectWithIri")
+                result.longname should be (Some("new project with a custom IRI"))
+                result.keywords should be (Seq("projectIRI"))
+                result.description should be (Seq(StringLiteralV2(value = "a project created with a custom IRI", language = Some("en"))))
+
+            }
+
+            "return 'BadRequest' if the supplied 'projectIri' is not a valid IRI" in {
```
SepidehAlassi, line 170: @subotic the optional `id` of the `project` is tested here.

subotic: This is an E2E test. I think that there is more value in testing things at a more granular level. Any particular reason why you need an E2E test in this case? E2E tests are more expensive to execute. So in this case, I would recommend a simple Spec.

PR #1658, opened 2020-06-16
Wip/dsp 168 refactor gravsearch

nemmK:
Task:
- [x] ConstructResponseUtilV2.splitMainResourcesAndValueRdfData: this method is too long and its nesting is too deep. Should be split up into several smaller, clearer methods.


webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -521,6 +416,183 @@ object ConstructResponseUtilV2 {
...
+        MainResourcesAndValueRdfData(
+            resources = mainResourcesNested,
+            hiddenResourceIris = mainResourceIrisNotVisible ++ dependentResourceIrisNotVisible
+        )
+    }
+
+    /**
+     * This method creates a map of property IRIs to sequences of value IRIs.
+     *
+     * @param assertionsExplicit
```
tobiasschweizer, line 448: Could you add a doc string comment for param`assertionsExplicit`?

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -521,6 +416,183 @@ object ConstructResponseUtilV2 {
...
+            case (pred: SmartIri, _: Seq[IriLiteralV2]) =>
+                // Turn the sequence of assertions into a Map of predicate IRIs to assertions.
+                pred
+        }.map {
+            case (pred: SmartIri, valueAssertions: Map[SmartIri, Seq[IriLiteralV2]]) =>
+                // Replace the assertions with their objects, i.e. the value object IRIs.
+                pred -> valueAssertions.values.flatten.map(_.value).toSeq
+        }
+
+        return valuePropertyToObjectIris
```
tobiasschweizer, line 474: In Scala, no return statement is needed. It's enough to write `valuePropertyToObjectIris` on line 474. Or you could even omit the assignment to `valuePropertyToObjectIris` but I think it's good for clarity because you are using explicit type annotations.

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -521,6 +416,183 @@ object ConstructResponseUtilV2 {
...
+    }
+
+    /**
+     * This method creates a map of property IRIs to sequences of value IRIs.
+     *
+     * @param assertionsExplicit
+     * @param valueObjectIris           set of all value object IRIs
+     * @return                          the sequence of IRIs
+     */
+    private def mapIriSequence(assertionsExplicit: ConstructPredicateObjects, valueObjectIris: Set[IRI] ) :  Map[SmartIri, Seq[IRI]] =  {
```
tobiasschweizer, line 452: Would `makePropertyIriToValueObjectIrisMap` be a clearer name for this method?

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -521,6 +416,183 @@ object ConstructResponseUtilV2 {
...
+    /**
+     * This method creates a map of IRI sequences which is used to create a map of Rdf Data
+     *
+     * @param valuePropertyToObjectIris     the map of sequence IRIs
+     * @param resourceIri                   the IRI of the resource.
+     * @param requestingUser                the user making the request.
+     * @param nonResourceStatements         the statements which got split up containing value objects and standoffs
+     * @return                              the map of IRI sequences
+     */
+    private def mapSequenceObjects(valuePropertyToObjectIris: Map[SmartIri, Seq[IRI]], resourceIri:IRI, requestingUser: UserADM,
```
tobiasschweizer, line 487: Could this method's name be made clearer? For example: `makePropertyIriToValueObjectsMap`?

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -521,6 +416,183 @@ object ConstructResponseUtilV2 {
...
+                // Replace the assertions with their objects, i.e. the value object IRIs.
+                pred -> valueAssertions.values.flatten.map(_.value).toSeq
+        }
+
+        return valuePropertyToObjectIris
+    }
+
+
+    /**
+     * This method creates a map of IRI sequences which is used to create a map of Rdf Data
```
tobiasschweizer, line 479: Since it returns `RdfPropertyValues`, you could say "A map of property IRIs to value RDF data.". (This is what the comment for `RdfPropertyValues` says.)

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -521,6 +416,183 @@ object ConstructResponseUtilV2 {
...
+                }
+
+                property -> valueRdfData
+        }.filterNot {
+            // filter out those properties that do not have value objects (they may have been filtered out because the user does not have sufficient permissions to see them)
+            case (_, valObjs: Seq[ValueRdfData]) =>
+                valObjs.isEmpty
+        }
+
+        return valuePropertyToValueObject
```
tobiasschweizer, line 584: no `return` statement needed in Scala, see comment above.

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -565,26 +637,10 @@ object ConstructResponseUtilV2 {
...
-                )
-
-                resourceIri -> transformedResource
-        }.toMap
-
-        MainResourcesAndValueRdfData(
-            resources = mainResourcesNested,
-            hiddenResourceIris = mainResourceIrisNotVisible ++ dependentResourceIrisNotVisible
-        )
+        return incomingLinksForResource
```
tobiasschweizer, line 640: no `return` statement needed, see comment above

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -521,6 +416,183 @@ object ConstructResponseUtilV2 {
...
+
+        return valuePropertyToValueObject
+
+    }
+
+
+    /**
+     * This method returns all the incoming link for each resource as a map of resource IRI to resources that link to it.
+     * @param visibleResources          the resources that the user has permission to see
+     * @param flatResourcesWithValues
```
tobiasschweizer, line 592: Could you add a docstring comment to param `flatResourcesWithValues`?

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
benjamingeer: `return` is never needed in Scala, and we don't use it in Knora:

https://tpolecat.github.io/2014/05/09/return.html

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -521,6 +416,183 @@ object ConstructResponseUtilV2 {
...
+     * @param resourceIri                   the IRI of the resource.
+     * @param requestingUser                the user making the request.
+     * @param nonResourceStatements         the statements which got split up containing value objects and standoffs
+     * @return                              the map of IRI sequences
+     */
+    private def mapSequenceObjects(valuePropertyToObjectIris: Map[SmartIri, Seq[IRI]], resourceIri:IRI, requestingUser: UserADM,
+                                   assertionsExplicit: ConstructPredicateObjects, nonResourceStatements: Statements)(implicit stringFormatter: StringFormatter): RdfPropertyValues ={
+
+        // Make a map of property IRIs to sequences of value objects.
+        val valuePropertyToValueObject: RdfPropertyValues = valuePropertyToObjectIris.map {
```
benjamingeer, line 491: This expression (`val valuePropertyToValueObject ...`, lines 491-582) is still much too long. It would be clearer to write it like this:

1. Assign the result of `valObjIris.map` to a variable with a descriptive name and a comment.
2. Call `filter` on the variable from (1), and assign the result to another variable with a descriptive name and a comment.
3. Call `flatMap` on the variable from (2), and assign the result to another variable with a descriptive name and a comment.

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -314,7 +320,7 @@ object ConstructResponseUtilV2 {
                                           requestingUser: UserADM)(implicit stringFormatter: StringFormatter): MainResourcesAndValueRdfData = {
 
         // An intermediate data structure containing RDF assertions about an entity and the user's permission on the entity.
-        case class RdfWithUserPermission(assertions: ConstructPredicateObjects, maybeUserPermission: Option[EntityPermission])
+        //case class RdfWithUserPermission(assertions: ConstructPredicateObjects, maybeUserPermission: Option[EntityPermission])
```
benjamingeer, line 323: It's better to delete code than to comment it out. It will still be there in the Git history if we ever need it again.

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
nemmK: > Please find the details of my review in my comments.
> 
> In case you have questions regarding semantics (i.e. to describe what the purpose of some parts of the code is), we can talk next week. I barely remember what we did. So I figure it must be difficult for someone else. ;-)
> 
> But the method `splitMainResourcesAndValueRdfData` looks less scary now which is great! :-)

Would be great. I was not always quite sure what some parts were doing so I thought I would be better to keep the descriptions short to be sure it is not wrong.
Also with the return statement I am so used to always use it I completely forgot that in scala it is not needed :-)
And of course I will do the changes you marked.

benjamingeer: @nemmK 

> I was not always quite sure what some parts were doing

Then please get @tobiasschweizer to explain it, because we really need to make it clear in the code/comments.

benjamingeer:
@nemmK Thanks, this is much better. I've tried to clarify some of the comments and method/variable names based on my own understanding. @tobiasschweizer can you check that what I wrote makes sense?

I think that if you have a method like

```scala
/**
  * Make a sequence of Foos.
  */
def makeFoos(): Seq[Foo] {
    // A sequence of foos.
    foos: Seq[Foo] = Seq(Foo(1), Foo(2))
    foos
}
```

it's actually clearer and simpler to write

```scala
/**
  * Make a sequence of Foos.
  */
def makeFoos(): Seq[Foo] {
    Seq(Foo(1), Foo(2))
}
```

You already have the return type in the method signature, so you don't need a variable to tell you what type you're returning. And you already have the ScalaDoc comment for the method, so you don't need another comment describing what's being returned.

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -441,20 +452,20 @@ object ConstructResponseUtilV2 {
...
-     * @param assertionsExplicit        assertion contains all non inferred statements
-     * @param valueObjectIris           set of all value object IRIs
-     * @return                          the sequence of IRIs
+     * @param assertionsExplicit all non-inferred statements.
+     * @param valueObjectIris    a set of all value object IRIs.
+     * @return a map of property IRIs to sequences of value IRIs.
      */
-    private def mapIriSequence(assertionsExplicit: ConstructPredicateObjects, valueObjectIris: Set[IRI] ) :  Map[SmartIri, Seq[IRI]] =  {
-        val valuePropertyToObjectIris: Map[SmartIri, Seq[IRI]] = assertionsExplicit.map {
+    private def mapIriSequence(assertionsExplicit: ConstructPredicateObjects, valueObjectIris: Set[IRI]): Map[SmartIri, Seq[IRI]] = {
```
tobiasschweizer, line 461: What about a clearer name for the method (https://github.com/dasch-swiss/knora-api/pull/1658#discussion_r442705413)?

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -722,7 +721,7 @@ object ConstructResponseUtilV2 {
                     case (propIri: SmartIri, values: Seq[ValueRdfData]) =>
 
                         // check if the property Iri already exists (there could be several instances of the same property)
-                        if (acc.get(propIri).nonEmpty) {
+                        if (acc.contains(propIri)) {
```
tobiasschweizer, line 724: I think this change could have been automated :-)

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
```diff
@@ -841,7 +840,7 @@ object ConstructResponseUtilV2 {
             case (acc: Set[IRI], (_: SmartIri, valObjs: Seq[ValueRdfData])) =>
                 val mappings: Seq[String] = valObjs.filter {
                     valObj: ValueRdfData =>
-                        valObj.valueObjectClass == OntologyConstants.KnoraBase.TextValue.toSmartIri && valObj.assertions.get(OntologyConstants.KnoraBase.ValueHasMapping.toSmartIri).nonEmpty
+                        valObj.valueObjectClass == OntologyConstants.KnoraBase.TextValue.toSmartIri && valObj.assertions.contains(OntologyConstants.KnoraBase.ValueHasMapping.toSmartIri)
```
tobiasschweizer, line 843: same here ;-)

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
benjamingeer: It was automated. :)

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
benjamingeer: Eventually IntelliJ will do the programming for us. :)

webapi/src/main/scala/org/knora/webapi/util/ConstructResponseUtilV2.scala
benjamingeer: Done in f883c02.

benjamingeer:
@tobiasschweizer OK now?

tobiasschweizer:
Ok, now you have to dismiss your own requested changes :-)

tobiasschweizer:
> @tobiasschweizer OK now?

ask @benjamingeer if he likes it :-)

PR #1666, opened 2020-06-29
Fix gren issue in github actions workflow

kilchenmann:
The gren (github release notes) was in the wrong place in github actions workflow. It should work in the next release.


kilchenmann:
🤞

PR #1672, opened 2020-07-08
test(clientapi): Return delete ontology response

benjamingeer:
https://dasch.myjetbrains.com/youtrack/issue/DSP-229#focus=streamItem-4-577.0-0

This PR includes a test data file `delete-ontology-response.json` in the test data returned by the `clientapitest` route.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
+++ b/webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
@@ -24,14 +24,13 @@ import java.time.Instant
 import java.util.UUID
 
 import akka.actor.ActorSystem
-import akka.http.scaladsl.client.RequestBuilding.{Get, Post}
 import akka.http.scaladsl.client.RequestBuilding._
 import akka.http.scaladsl.model.HttpEntity
 import akka.http.scaladsl.model.headers.BasicHttpCredentials
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.{PathMatcher, Route}
-import akka.stream.Materializer
 import akka.http.scaladsl.util.FastFuture
+import akka.stream.Materializer
 import org.knora.webapi._
 import org.knora.webapi.messages.v2.responder.ontologymessages._
 import org.knora.webapi.routing.{Authenticator, KnoraRoute, KnoraRouteData, RouteUtilV2}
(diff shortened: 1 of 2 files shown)


webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
```diff
@@ -948,6 +948,17 @@ class OntologiesRouteV2(routeData: KnoraRouteData) extends KnoraRoute(routeData)
         }
     }
 
+    private def deleteOntologyTestResponse: Future[TestDataFileContent] = {
+        val responseStr = SharedTestDataADM.successResponse("Ontology http://0.0.0.0:3333/ontology/00FF/foo/v2 has been deleted")
```
SepidehAlassi, line 952: @benjamingeer, I guess Andre does not want a dummy response. He wants the actual response that would be returned which is

`{
  "knora-api:result" : "Ontology http://0.0.0.0:3333/ontology/0001/testonto/v2 has been deleted",
  "@context" : {
    "knora-api" : "http://api.knora.org/ontology/knora-api/v2#"
  }
}`

SepidehAlassi:
@benjamingeer Andre also wants the response for `delete property` and `delete class`.

benjamingeer:
@SepidehAlassi [As the docs say](https://docs.knora.org/paradox/03-apis/api-v2/ontology-information.html#deleting-a-property), the response to deleting a property or a class is just the ontology metadata, and you already have test data for that.

SepidehAlassi:
> @SepidehAlassi [As the docs say](https://docs.knora.org/paradox/03-apis/api-v2/ontology-information.html#deleting-a-property), the response to deleting a property or a class is just the ontology metadata, and you already have test data for that.

yeah, I know that :-D.

webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
kilchenmann: Yes, it should be the same answer that I get from the API

benjamingeer: Why? Aren't you just testing whether you can parse it? Why does it need to be a real response?

I don't think that the `clientapitest` route should actually modify the respository, which would be the only way to get a real response, so we already have a lot of dummy responses for update operations.

kilchenmann:
Thanks for all the explanations. I will use the already existing data for delete-resource-class-response and delete-resource-property-response.

webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
kilchenmann: I have no idea. I'm not an expert in tests. I can try with the data in this PR. No worries.

kilchenmann: It's all fine. I get exactly what I need. I just generated the test-data and it seems to be good. Thanks

benjamingeer:
@kilchenmann great, thanks

@SepidehAlassi Can you approve this?

kilchenmann:
Just a question: As you know, we use this url <http://localhost:3333/clientapitest> to generate the test-data and to integrate into js-lib. This generates the foo ontology. But it doesn't delete the ontology with the new setup from this PR. So next time, when I run my script, the <http://localhost:3333/clientapitest> doesn't work anymore. I get the following error:

```
"error": "org.knora.webapi.ClientApiGenerationException: Failed to get test data: {\"knora-api:error\":\"org.knora.webapi.BadRequestException: Ontology http://0.0.0.0:3333/ontology/00FF/foo/v2 cannot be created, because it already exists\",\"@context\":{\"knora-api\":\"http://api.knora.org/ontology/knora-api/v2#\"}}"
```

So, we should re-init the database before running the script with the mentioned url. For me no problem, but some people had a lot of issues because of this. But we have to rethink the whole test-data config anyway.

benjamingeer:
> This generates the foo ontology.

I can change it so it doesn't generate the foo ontology, and just returns a dummy response.

> we have to rethink the whole test-data config anyway.

The other option would be to generate the test data as a side effect of running all the other knora-api tests and save it somewhere. Maybe this could be done in GitHub CI. But it would be a major refactoring.

kilchenmann:
> I can change it so it doesn't generate the foo ontology, and just returns a dummy response.

This would be helpful. Thanks

PR #1679, opened 2020-07-16
Gravsearch optimisations

SepidehAlassi:
resolves [DSP-478](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-36?issue=DSP-478), [DSP-479](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-46?issue=DSP-479), and [DSP-536](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-46?issue=DSP-536)

- [x] in the prequery generation (in AbstractPrequeryGenerator) we can simply not generate the check for `rdf:type knora-base:Resource`. I think this is not actually needed.
- [x] refactor the names of rules and comments to make the code and what every inference rule does clearer.

- [x]  implement an optimization to find and remove redundant `rdf:type` statements from the generated prequery. The main idea of how to do this is: go through all statements in the prequery, check if the predicate is `rdf:type`. If yes, check if the subject is inferred from a property (i.e subject is in the `entitiesInferredFromProperties`). If yes, remove the statement from the prequery. 

*Steps to do this*:
1. **create and populate the `entitiesInferredFromProperties` collection**:
- - [x] add a collection to `GravsearchTypeInspectionResult` in order to store the entities whose types are inferred from their use as subjects or objects of properties. Call the collection: `entitiesInferredFromProperties`
- - [x] If a rule infers the type of an entity from its use as a subject or an object, then while adding the inferred type to the `IntermediateResults`, it should have a flag `inferredFromProperty=True`. 

2. **Remove the redundant `rdf:types` statements from prequery if the type of subject could be inferred from a property.**  This step in optimization requires refining the types of resource entities: instead of the general main `Resource` type the actual resource class of the entity should be returned as its type. 
Steps to do this:
- - [x] change `NonPropertyTypeInfo` to store the actual resource class instead of the knora main "Resource" type, along with a boolean flag in its constructor to replace `isResourceType`. 
- - [x] change `PropertyTypeInfo` to store the actual resource class instead of the knora main "Resource" type, along with a boolean flag in its constructor to replace `ObjectIsResourceType`. **Reason**: `TypeOfObjectFromPropertyRule` finds the type of the resource used as the object form the `PropertyTypeInfo`, therefore if we want the `NonPropertyTypeInfo` to store the actual resource class instead of the knora main "Resource" type for such an object, `PropertyTypeInfo` should also contain the actual resource class of the object.
- - [x] change `InferringGravsearchTypeInspector` to return a `NonPropertyTypeInfo` containing the actual resource class.
- - - [x] for some of the entities it is possible to detect/infer multiples types from type annotations and other statements, then from the `Set[GravsearchEntityTypeInfo]]` of types associated with that `TypeableEntity` (i.e. IntermediateTypeInspectionResult.entities: Map[TypeableEntity, Set[GravsearchEntityTypeInfo]]), will contain multiple types. Go through all the types in this set, check the `classInfoMap` about each resource type, to choose the most specific class type. For example, consider the following query statements:

```
?letter rdf:type beol:WrittenSource .

?letter rdf:type beol:letter .

?letter beol:hasAuthor ?author .
```

The typeableEntity `?letter`will have types: `Set["knora-api:Resource", "beol:letter", "beol:WrittenSource"]`. Since `beol:letter` is `subClassOf beol:WrittenSource` which is in return `subClassOf knora-api:Resource`, then only `beol:letter` should be kept as the type of the `?letter`. 

**Note**: In this example a problem with a current approach becomes evident. Namely that, from the
`?letter beol:hasAuthor ?author .` 

using property `beol:hasAuthor`, the type of the`?letter` will be inferred as  `beol:WrittenSource` while from the above statements its types are annotated as`beol:letter`. As a result, `?letter` will be added to the set `entitiesInferredFromProperties`, hence the essential statement `?letter rdf:type beol:letter ` that identifies the correct and most specific type of the `?letter` will be removed from SPARQL query, and `?letter` will be stuck with the more general type `beol:WrittenSource` instead of `beol:letter`. 
- - [x] To avoid this, `entitiesInferredFromProperties` should be a map. This map should contain the entity and type of it inferred from a property. If a type of an entity is removed during the refining process, it should also be removed from this map.
- - [x] In the prequery Generator, remove the statements with RDF type annotations whose subject is in the `entitiesInferredFromProperties`

- [x] Add new tests and refactor the old tests to work with this optimization.
- [x] check if the [desired SPARQL](https://github.com/dasch-swiss/0113-ll-perf/issues/1#issuecomment-657590042) is generated after this optimization step. 
- [x] when the type of an entity must be inferred from a property that is specified in an `OPTIONAL` block, if we remove the rdfs:type statements, the triplestore cannot infer the type correctly. It considers the `OPTIONAL` statements to be none optional. Therefore, a check must be added to the code to remove the rdfs:type statements for entities whose type can be inferred from a property only if the entity is not a subject or object of an `OptionalQuery` statement. 
- [x] Some statements such as `{ ?writtenSource a beol:manuscript . } UNION { ?writtenSource a beol:letter . }` result in 2 different types inferred for an entity that are not related. Here for `?writtenSource` two types 'beol:letter' and 'beol:manuscript' will be inferred that are not related. (none of them is a base class of the other). This results in inconsistent types for an entity. A sanitizer `sanitizeInconsistentResourceTypes` must replace these types with a common base class. 

closes [0113-ll-perf/issue](https://github.com/dasch-swiss/0113-ll-perf/issues/1#issuecomment-656607297)


benjamingeer:
Looking forward to reviewing this when I get back!

SepidehAlassi:
> Looking forward to reviewing this when I get back!

Great! We should go through it together once, there are some points I need to explain.

SepidehAlassi:
Getting rid of redundant rdf:type statements solves the time-out problem @loicjaouen  was experiencing. With this optimization, the SPARQL query generated from [Lioic's gravsearch query](https://github.com/dasch-swiss/0113-ll-perf/issues/1#issuecomment-656607297) turns to be as [expected](https://github.com/dasch-swiss/0113-ll-perf/issues/1#issuecomment-657590042):

```sparql
PREFIX xml: <http://www.w3.org/2001/XMLSchema#> 
PREFIX lumieres-lausanne: <http://www.knora.org/ontology/0113/lumieres-lausanne#> 

SELECT DISTINCT ?BibliographicNotice 

WHERE { 
?httpwwwknoraorgontology0113lumiereslausannebibliographicNoticeHasTitle__subPropertyOf <http://www.w3.org/2000/01/rdf-schema#subPropertyOf>* <http://www.knora.org/ontology/0113/lumieres-lausanne#bibliographicNoticeHasTitle> . 

?BibliographicNotice ?httpwwwknoraorgontology0113lumiereslausannebibliographicNoticeHasTitle__subPropertyOf ?title . ?title <http://www.knora.org/ontology/knora-base#valueHasString> ?title__valueHasString . 

?title <http://jena.apache.org/text#query> "Test"^^<http://www.w3.org/2001/XMLSchema#string> . 

?BibliographicNotice <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> . 

?title <http://www.knora.org/ontology/knora-base#isDeleted> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 
} 
GROUP BY ?BibliographicNotice 
ORDER BY ASC(?BibliographicNotice) 
LIMIT 25

```

This query takes 23 seconds and returns three results. However, 23 seconds is still pretty high for such a query.

@benjamingeer As you suggested, by moving the 
```sparql
?title <http://jena.apache.org/text#query> "Test"^^<http://www.w3.org/2001/XMLSchema#string> . 
```
to the beginning of the generated SPARQL query, the same three results are returned in 49 ms! As discussed, I will work on that next in [DSP-479](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-42?issue=DSP-479).

benjamingeer:
We can also get rid of this, right?

```
?title <http://www.knora.org/ontology/knora-base#valueHasString> ?title__valueHasString . 
```

It looks like `?title__valueHasString` isn’t actually used.

SepidehAlassi:
> We can also get rid of this, right?
> 
> ```
> ?title <http://www.knora.org/ontology/knora-base#valueHasString> ?title__valueHasString . 
> ```
> 
> It looks like `?title__valueHasString` isn’t actually used.

We would never need this statement right?

benjamingeer:
That statement is needed with GraphDB, but it shouldn’t be generated with Fuseki. We can do this in another PR.

SepidehAlassi:
> That statement is needed with GraphDB, but it shouldn’t be generated with Fuseki. We can do this in another PR.

Well, what confuses me is that even though the statement with valueAsString is actually unnecessary, without it the query takes longer.

SepidehAlassi:
> This looks really great. A few notes from our discussion just now:

Thanks for summarizing all the discussed points here.

benjamingeer:
I found the relevant section of the Gravsearch paper:

Screenshot 2020-08-10 at 12 21 08
Screenshot 2020-08-10 at 12 21 15

🙂

webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/types/IntermediateTypeInspectionResult.scala
```diff
@@ -31,17 +31,51 @@ import org.knora.webapi.messages.{OntologyConstants, StringFormatter}
...
+     *  @param inferredFromProperty `true` if any of the types of this entity were inferred from its use with a property.
       * @return a new [[IntermediateTypeInspectionResult]] containing the additional type information.
       */
-    def addTypes(entity: TypeableEntity, entityTypes: Set[GravsearchEntityTypeInfo]): IntermediateTypeInspectionResult = {
+    def addTypes(entity: TypeableEntity, entityTypes: Set[GravsearchEntityTypeInfo], inferredFromProperty: Boolean = false): IntermediateTypeInspectionResult = {
         val newTypes = entities.getOrElse(entity, Set.empty[GravsearchEntityTypeInfo]) ++ entityTypes
-        IntermediateTypeInspectionResult(entities = entities + (entity -> newTypes))
+
+        val newEntitiesInferredFromProperties = if (inferredFromProperty && entityTypes.nonEmpty) {
+            entitiesInferredFromProperties + (entity -> entityTypes)
```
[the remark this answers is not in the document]
benjamingeer: See how `newTypes` merges the existing types with the new ones.

SepidehAlassi:
@benjamingeer In [last commit ](https://github.com/dasch-swiss/knora-api/pull/1679/commits/d0d726c74d8c5682bec43d84e3d91b5d25fbf913) I have added tests that clearly show the point with OPTIONAL statements.

benjamingeer:
Thanks, let me know when I should look at this again.

SepidehAlassi:
> Thanks, let me know when I should look at this again.

I think this inference from OPTIONAL statements problem has to do with Fuseki, but I am not entirely sure.

benjamingeer:
> I think this inference from OPTIONAL statements problem has to do with Fuseki

Why do you think so?

Tell me if I understand this correctly. With the following WHERE clause:

```sparql
WHERE {
  ?document rdf:type beol:writtenSource .

  OPTIONAL {
    ?document beol:hasRecipient ?recipient .
    ?recipient beol:hasFamilyName ?familyName .
    FILTER knora-api:matchText(?familyName, "Bernoulli")
  }
}
```

If we allowed to optimisation to remove the line `?document rdf:type beol:writtenSource .`, the generated SPARQL would be:

```sparql
WHERE {
  ?document knora-base:isDeleted false .

  OPTIONAL {
    # Same as above
  }
}
```

This should return everything in the triplestore that is not marked as deleted, right?

webapi/src/test/scala/org/knora/webapi/messages/util/search/gravsearch/prequery/NonTriplestoreSpecificGravsearchToPrequeryTransformerSpec.scala
```diff
@@ -1319,8 +1320,214 @@ class NonTriplestoreSpecificGravsearchToPrequeryTransformerSpec extends CoreSpec
...
+                    ),
+                    namedGraph = Some(IriRef(
+                        iri = "http://www.knora.org/explicit".toSmartIri,
+                        propertyPathOperator = None
+                    ))
+                ),
+                // This statement must not be removed by AbstractPrequeryGenerator.removeEntitiesInferredFromProperty
+                // because the property from which its type can be inferred is in an optional. Without this statement,
+                // the type beol:basicLetter (inferred from property beol:hasRecipient) would be considered for ?document.
+                StatementPattern(
```
SepidehAlassi, line 1372: @benjamingeer If we remove the rdf:type statement with object beol:writtenSource. I would have also expected it to consider any kind of resource, but Triplestore infers the type to be beol:basicLetter from the OPTIONAL statement.

benjamingeer: > Triplestore infers the type to be beol:basicLetter from the OPTIONAL statement.

Do you mean that all the results are instances of `beol:basicLetter`? How many results does it return? Is there more than one page?

SepidehAlassi: Ok, let me add an end to end test for this.

benjamingeer: I just ran this query in the Fuseki dashboard (http://localhost:3030/dataset.html?tab=query&ds=/knora-test):

```sparql
PREFIX beol: <http://www.knora.org/ontology/0801/beol#>

SELECT ?document WHERE {
  ?document knora-base:isDeleted false .

  OPTIONAL {
    ?document beol:hasRecipient ?recipient .
    ?recipient beol:hasFamilyName ?familyName .
  }
}
```

It returns 30,435 results, including resources of different types (e.g. `incunabula:book`) as well as values.

SepidehAlassi: When I run this, I get 371,400 results including values!
But if you run it without '?document knora-base:isDeleted false .' you get only one result. I mean following query returns one result.
```
PREFIX beol: <http://www.knora.org/ontology/0801/beol#>

SELECT ?document WHERE {
  OPTIONAL {
    ?document beol:hasRecipient ?recipient .
    ?recipient beol:hasFamilyName ?familyName .
  }
}
```

SepidehAlassi: When we remove the rdf:type statements for entities that can be inferred from properties, the isDeleted statements are not generated either.

benjamingeer: OK, I think I understand it now.

This query returns nothing:

```sparql
SELECT ?document WHERE { }
```

That's because the WHERE clause doesn't bind anything to `?document`. Now we add an `OPTIONAL`:

```sparql
SELECT ?document WHERE {
  OPTIONAL {
    ?document beol:hasRecipient ?recipient .
  }
}
```

The only thing an `OPTIONAL` can do is **add** data to the results. So the result of this query should be the same as the first one, plus whatever the `OPTIONAL` found. Does that make sense?

SepidehAlassi: > The only thing an `OPTIONAL` can do is **add** data to the results. So the result of this query should be the same as the first one, plus whatever the `OPTIONAL` found. Does that make sense?

yes, it does.

SepidehAlassi: > The only thing an `OPTIONAL` can do is **add** data to the results. So the result of this query should be the same as the first one, plus whatever the `OPTIONAL` found. Does that make sense?

yes, it does. So if we want the query to return something meaningful, we should not remove the rdfs:type statement in this case.

webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/types/InferringGravsearchTypeInspector.scala
```diff
@@ -793,14 +836,149 @@ class InferringGravsearchTypeInspector(nextInspector: Option[GravsearchTypeInspe
...
+        def getIsResourceFlags(typeInfo: GravsearchEntityTypeInfo): Boolean = {
+            typeInfo match {
+                case propertyTypeInfo: PropertyTypeInfo => propertyTypeInfo.objectIsResourceType
+                case nonPropertyTypeInfo: NonPropertyTypeInfo=> nonPropertyTypeInfo.isResourceType
+                case _ => throw GravsearchException(s"There is an invalid type")
+            }
+        }
+
+        // This method finds the common base class.
+        def findCommonBaseClass(typesToBeChecked: Set[GravsearchEntityTypeInfo]): SmartIri = {
```
benjamingeer, line 870: What if there is no common base class? For example, what if `typesToBeChecked` contains `beol:letter` and `knora-api:TextValue`?

webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/types/InferringGravsearchTypeInspector.scala
```diff
@@ -793,14 +836,149 @@ class InferringGravsearchTypeInspector(nextInspector: Option[GravsearchTypeInspe
...
+                        val baseClassesOfType: Set[SmartIri] = entityInfo.classInfoMap.get(iriOfGravsearchTypeInfo(aType)) match {
+                            case Some(classDef: ReadClassInfoV2) =>
+                                classDef.allBaseClasses
+                            case _ => Set.empty[SmartIri]
+                        }
+                        //find the common base classes
+                        baseClassesOfType.intersect(acc)
+                }
+                if (commonBaseClasses.nonEmpty) {
+                    // since the base classes are sorted, first one would be the most specific one.
```
benjamingeer, line 889: `commonBaseClasses` is a `Set`, and a `Set` can't be sorted because it has no order.

SepidehAlassi: The classDef.allBaseClasses has an order of base classes, i.e. the most specific one is first. When I do the intersect, it keeps the order.

webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/types/InferringGravsearchTypeInspector.scala
SepidehAlassi: well, if there is no common base class, it returns knora-base:Resource.
findCommonBaseClass is only used if all typesToBeChecked are resource types.

webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/types/InferringGravsearchTypeInspector.scala
benjamingeer: Per our discussion: you're right, but we'll make it a `Seq` in another PR so we can rely on the order.

SepidehAlassi:
@benjamingeer I have updated the Gravsearch design document, please let me know if there is anything missing. This PR can be completely reviewed now.

SepidehAlassi:
> Many thanks for this great work! 🎉

Thank you for reviewing this and all your help.

SepidehAlassi:
@loicjaouen with this PR your query takes 63ms on my old laptop. :-)

loicjaouen:
@SepidehAlassi, @benjamingeer , that looks promising, thanks, I will look at it this afternoon

PR #1686, opened 2020-08-11
feat(api-v2): Accept custom timestamps in update/delete requests

benjamingeer:
https://dasch.myjetbrains.com/youtrack/issue/DSP-399

- [x] Accept a custom timestamp when:
  - [x] updating a value
  - [x] marking a value as deleted
  - [x] marking a resource as deleted
- [x] Add docs.
- [x] Add tests:
  - [x] updating a value
  - [x] marking a value as deleted
  - [x] marking a resource as deleted

This also fixes a bug: the custom timestamp submitted when creating a value should be `knora-api:valueCreationDate`, not `knora-api:creationDate` (still works, but now deprecated in this context). **TODO: mention this in the next release notes.**


SepidehAlassi:
@benjamingeer I will review it this afternoon.

benjamingeer:
Thanks!

docs/03-apis/api-v2/editing-values.md
```diff
@@ -431,7 +436,10 @@ the resource to the value, and the value's ID and type. For example:
 ```
 
 The optional property `knora-api:deleteComment` specifies a comment to be attached to the
-value, explaining why it has been marked as deleted.
+value, explaining why it has been marked as deleted
+
+The optional property `knora-api:deleteDate` (an [xsd:dateTimeStamp](https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp))
+specifies a custom timestamp indicating when the value was deleted. If not specified, the current time is used.å
```
SepidehAlassi, line 442: what does the Swedish character å doing at the end of this line? :-D I actually like it giving knora a scandinavian flavor. Did you know this single character means a river?

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
```diff
@@ -606,15 +605,22 @@ object CreateResourceRequestV2 extends KnoraJsonLDRequestReaderV2[CreateResource
...
                                     log = log
                                 )
+
                                 maybeCustomValueIri: Option[SmartIri] = valueJsonLDObject.maybeIDAsKnoraDataIri
-                                maybeCustomValueUUID: Option[UUID] = valueJsonLDObject.maybeUUID
+                                maybeCustomValueUUID: Option[UUID] = valueJsonLDObject.maybeUUID(OntologyConstants.KnoraApiV2Complex.ValueHasUUID)
 
-                                // Get the values's creation date.
+                                // Get the value's creation date.
+                                // TODO: creationDate for values is a bug, and will not be supported in future. Use valueCreationDate instead.
```
SepidehAlassi, line 613: I guess you can remove this comment and TODO from here. As far as I see, you have fixed it already.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
```diff
@@ -120,23 +120,29 @@ object CreateValueRequestV2 extends KnoraJsonLDRequestReaderV2[CreateValueReques
                         maybeCustomValueIri: Option[SmartIri] = jsonLDObject.maybeIDAsKnoraDataIri
 
                         // Get the custom value UUID if provided.
-                        maybeCustomUUID: Option[UUID] = jsonLDObject.maybeUUID
+                        maybeCustomUUID: Option[UUID] = jsonLDObject.maybeUUID(OntologyConstants.KnoraApiV2Complex.ValueHasUUID)
 
                         // Get the value's creation date.
+                        // TODO: creationDate for values is a bug, and will not be supported in future. Use valueCreationDate instead.
```
SepidehAlassi, line 126: same here

webapi/src/main/scala/org/knora/webapi/responders/v2/ValuesResponderV2.scala
```diff
@@ -853,6 +852,11 @@ class ValuesResponderV2(responderData: ResponderData) extends Responder(responde
                 _ = if (currentValue.valueContent.valueType != submittedExternalValueType.toOntologySchema(InternalSchema)) {
                     throw BadRequestException(s"Value <$valueIri> has type <${currentValue.valueContent.valueType.toOntologySchema(ApiV2Complex)}>, but the submitted type was <$submittedExternalValueType>")
                 }
+
+                // If a custom value creation date was submitted, make sure it's later than the date of the current version.
```
SepidehAlassi, line 856: this check did not come to my mind back then when I worked on custom creation date. I trust clients too much I guess that I had not thought they might give a date in the past!

docs/03-apis/api-v2/editing-values.md
benjamingeer: I often find those "rivers" in my code, I think it happens when I accidentally press Option-A instead of Command-S. 😀

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
benjamingeer: The way I wrote it, it will accept both. The TODO is to remove support for creationDate later, leaving only the support for valueCreationDate. I wrote it this way because I thought we weren't allowed to make breaking changes anymore. :) Do you think that's right, or should I should just make it a breaking change?

webapi/src/main/scala/org/knora/webapi/responders/v2/ValuesResponderV2.scala
benjamingeer: I thought that if the user did make that mistake, it would totally mess up the version history of the value, because the version history query depends on the timestamps being in the right order.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
SepidehAlassi: > The way I wrote it, it will accept both. The TODO is to remove support for creationDate later, leaving only the support for valueCreationDate. I wrote it this way because I thought we weren't allowed to make breaking changes anymore. :) Do you think that's right, or should I just make it a breaking change?

ooh, tough question. I believe we must be able to do breaking changes, but in agreement with front-end/JS-lib team so that they can immediately adapt their code.

benjamingeer: OK then let's leave it like this, as a deprecation, to give them time to adapt. Then I'll make the breaking change later.

benjamingeer:
Does this look OK now?

benjamingeer:
Thanks for the review, see you in a week!

subotic:
Maybe a few remarks, regarding breaking changes.

We are allowed to make braking changes to features, provided we have a sufficiently long deprecation timeframe. We talked about a timeframe of 12 months so that external projects using the backend API have sufficient time to react before their application will potentially break. The deprecation timer starts running when we make an official release and deploy it into production. This also requires that the release notes mention the deprecation prominently.

If we see that no one is using a deprecated feature anymore, which will require to log usage of deprecated features, then we can remove it sooner, or even immediately.

In the case of the mentioned deprecation in this PR,  the deprecation timer will start as soon as we make a non-RC release of `v13` and deploy it into production.

PR #1693, opened 2020-08-24
fix(api-v2): Fix generated SPARQL for updating property comment

benjamingeer:
https://dasch.myjetbrains.com/youtrack/issue/DSP-561

- [x] In the `WHERE` clause of `changePropertyLabelsOrComments.scala.txt`, move `BIND` statements inside the `GRAPH` block, so that the `WHERE` clause will return results even if the property doesn't have an `rdfs:comment`. **TODO: ask the Jena developers why this makes a difference.**
- [x] Fix a bug in `OntologyResponderV2.changePropertyLabelsOrComments` which would incorrectly compare a saved link value property with the expected result.
- [x] Add a test for adding an `rdfs:comment` to a link property that doesn't have one.


SepidehAlassi (approved):
Thanks for the quick fix!

benjamingeer:
Thanks for the quick review. :)

PR #1696, opened 2020-08-28
feat(api-v2): Make inference optional in Gravsearch

benjamingeer:
https://dasch.myjetbrains.com/youtrack/issue/DSP-534

- [x] In `AbstractPrequeryGenerator`, parse `knora-api:GravsearchOptions knora-api:useInference` with a boolean object.
- [x] Ignore `knora-api:GravsearchOptions` and `knora-api:useInference` in type inspection.
- [x] In `SparqlTransformer`, if that option is set, change the way the generated prequery is processed:
  - [x] With Fuseki, don't simulate inference by expanding statements.
  - [x] With GraphDB, add `FROM <http://www.ontotext.com/explicit>`.
- [x] Add tests.
- [x] Add docs.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/OntologyConstants.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/OntologyConstants.scala
@@ -656,13 +656,28 @@ object OntologyConstants {
         val KnoraApiPrefix: String = KnoraApiOntologyLabel + ":"
 
         /**
-         * Returns `true` if the specified IRI is `knora-api:Resource` in Knora API v2, in the simple
-         * or complex schema.
+         * The IRIs representing `knora-api:Resource` in Knora API v2, in the simple and complex schemas.
          */
-        def isKnoraApiV2Resource(iri: SmartIri): Boolean = {
-            val iriStr = iri.toString
-            iriStr == OntologyConstants.KnoraApiV2Simple.Resource || iriStr == OntologyConstants.KnoraApiV2Complex.Resource
-        }
+        val KnoraApiV2ResourceIris: Set[IRI] = Set(
+            OntologyConstants.KnoraApiV2Simple.Resource,
+            OntologyConstants.KnoraApiV2Complex.Resource
+        )
+
+        /**
+         * The IRIs representing `knora-api:GravsearchOptions` in Knora API v2, in the simple and complex schemas.
+         */
+        val GravsearchOptionsIris: Set[IRI] = Set(
+            OntologyConstants.KnoraApiV2Simple.GravsearchOptions,
+            OntologyConstants.KnoraApiV2Complex.GravsearchOptions
+        )
+
+        /**
+         * The IRIs representing `knora-api:useInference` in Knora API v2, in the simple and complex schemas.
+         */
+        val UseInferenceIris: Set[IRI] = Set(
+            OntologyConstants.KnoraApiV2Simple.UseInference,
+            OntologyConstants.KnoraApiV2Complex.UseInference
+        )
 
         /**
          * Returns the IRI of `knora-api:subjectType` in the specified schema.
@@ -926,6 +941,9 @@ object OntologyConstants {
         val MatchTextInStandoffFunction: IRI = KnoraApiV2PrefixExpansion + "matchTextInStandoff"
         val MatchLabelFunction: IRI = KnoraApiV2PrefixExpansion + "matchLabel"
         val StandoffLinkFunction: IRI = KnoraApiV2PrefixExpansion + "standoffLink"
+
+        val GravsearchOptions: IRI = KnoraApiV2PrefixExpansion + "GravsearchOptions"
+        val UseInference: IRI = KnoraApiV2PrefixExpansion + "useInference"
     }
 
     object SalsahGuiApiV2WithValueObjects {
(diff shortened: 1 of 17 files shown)


webapi/src/main/scala/org/knora/webapi/messages/OntologyConstants.scala
```diff
@@ -655,13 +655,28 @@ object OntologyConstants {
...
         /**
-         * Returns `true` if the specified IRI is `knora-api:Resource` in Knora API v2, in the simple
-         * or complex schema.
+         * The IRIs representing `knora-api:Resource` in Knora API v2, in the simple and complex schemas.
          */
-        def isKnoraApiV2Resource(iri: SmartIri): Boolean = {
-            val iriStr = iri.toString
-            iriStr == OntologyConstants.KnoraApiV2Simple.Resource || iriStr == OntologyConstants.KnoraApiV2Complex.Resource
-        }
+        lazy val KnoraApiV2ResourceIris: Set[IRI] = Set(
```
SepidehAlassi, line 660: @benjamingeer question: why are using `lazy val` here in the following three cases? The initialization does not need any evaluation of an expression, it is just a set. In this case, it would not make any difference if the initialization happens immediately or is postponed to the first time the value is accessed.

benjamingeer: Great question! The problem is that `KnoraApiV2ResourceIris` is a static value (it's defined in an `object` rather than in a `class`), which refers to static values in another `object`. The order of initialisation of static values in the JVM depends on which `object` is loaded first by the JVM's class loader at run time. Without the `lazy`, if `object KnoraApi` is initialised before `object KnoraApiV2Simple`, `KnoraApiV2ResourceIris` will just contain `null` values. It took me a little while to figure this out yesterday. A test was failing, and it turned out that it was because `KnoraApiV2ResourceIris` did not contain `OntologyConstants.KnoraApiV2Complex.Resource`, because it hadn't been initialised properly.

In the earlier code, there was a method, which didn't access those contents until it was called, so it was fine.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -381,22 +389,28 @@ class SearchResponderV2(responderData: ResponderData) extends ResponderWithStand
...
             // Convert the preprocessed query to a non-triplestore-specific query.
+
             triplestoreSpecificPrequery = QueryTraverser.transformSelectToSelect(
                 inputQuery = nonTriplestoreSpecificPrequery,
                 transformer = triplestoreSpecificQueryPatternTransformerSelect
             )
 
-            // _ = println(triplestoreSpecificPrequery.toSparql)
+            triplestoreSpecificPrequerySparql = triplestoreSpecificPrequery.toSparql
+            _ = log.debug(triplestoreSpecificPrequerySparql)
```
SepidehAlassi, line 411: please remove this log

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -470,19 +484,17 @@ class SearchResponderV2(responderData: ResponderData) extends ResponderWithStand
...
                     transformer = triplestoreSpecificQueryPatternTransformerConstruct
                 )
 
                 // Convert the result to a SPARQL string and send it to the triplestore.
-                val triplestoreSpecificSparql: String = triplestoreSpecificQuery.toSparql
-
-                // println("++++++++")
-                // println(triplestoreSpecificSparql)
+                val triplestoreSpecificMainQuerySparql: String = triplestoreSpecificMainQuery.toSparql
+                log.debug(triplestoreSpecificMainQuerySparql)
```
SepidehAlassi, line 494: please remove log

webapi/src/main/scala/org/knora/webapi/messages/util/search/SparqlQuery.scala
```diff
@@ -573,19 +573,32 @@ case class OrderCriterion(queryVariable: QueryVariable, isAscending: Boolean) ex
     }
 }
 
+/**
+ * Represents a FROM clause.
+ *
+ * @param namedGraph the named graph to be used in the query.
+ */
+case class FromClause(namedGraph: IriRef) extends SparqlGenerator {
+    override def toSparql: String = s"FROM ${namedGraph.toSparql}\n"
```
SepidehAlassi, line 582: we have a named graph here, shouldn't we use `FROM NAMED`? Or are you using `FROM` because the namedGraph is anyway the default graph? In that case, please rename the `nameGraph` to `defaultGraph`.

SepidehAlassi (changes requested):
Thanks for this, please see my comments.

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
benjamingeer: Why? It should have no effect unless debug logging is enabled. Using `log` seems better than using `println`, because with `println`, you have to uncomment it and recompile if you want to use it. With `log`, you just have to change the logging config file.

webapi/src/main/scala/org/knora/webapi/messages/util/search/SparqlQuery.scala
benjamingeer: Sorry, this is confusing. For me, any graph in the triplestore that has a name is a "named graph". The pseudo-graph `<http://www.ontotext.com/explicit>` is a "named graph" in that sense:

http://graphdb.ontotext.com/documentation/standard/query-behaviour.html#how-to-query-explicit-and-implicit-statements

I've clarified this in a7f3a1cc.

SepidehAlassi (approved):

PR #1699, opened 2020-09-04
fix(gravsearch): When link property compared in filter, don't compare link value property, too

benjamingeer:
https://dasch.myjetbrains.com/youtrack/issue/DSP-594


tobiasschweizer (approved):

benjamingeer:
Thanks for the bug report and the review!

PR #1707, opened 2020-09-10
chore: Update ci workflow

kilchenmann:
We still use gren which generates release notes on each release automatically. This PR resolves the issue we had with `release candidates`.


subotic (approved):
Thanks, LGTM!

PR #1709, opened 2020-09-12
fix(OntologyResponderV2): Fix ontology cache update when ontology metadata changed

benjamingeer:
https://dasch.myjetbrains.com/youtrack/issue/DSP-626

- [x] When ontology metadata is changed, update the ontology cache correctly.
- [x] Add test.


SepidehAlassi (approved):

benjamingeer:
Thanks for the quick review!

PR #1711, opened 2020-09-17
update developer and create resource docs 

### LINKED ISSUE #1541: Docs v2 Resource Creation
V2 docs about "Editing Resources" say:

> The body of the request is a JSON-LD document in the complex API schema, specifying the resource’s IRI, type, and rdfs:label, along with its Knora resource properties and their values. 

the resource's Iri is created by the system

SepidehAlassi:
resolves [DSP-540](https://dasch.myjetbrains.com/youtrack/agiles/110-4/111-52?issue=DSP-540)
close #1541


benjamingeer (changes requested):
This is much better, but it creates some redundancy between `building-and-running.md` and `bazel.md`. Do we still need `bazel.md`?

SepidehAlassi:
> This is much better, but it creates some redundancy between `building-and-running.md` and `bazel.md`. Do we still need `bazel.md`?

We need the `bazel.md` still because it contains some basic bazel info. I believe at some point we should get rid of it, but for now if someone does not use IntelliJ, would need to have the info in this documentation to run things with command line. 

However, I removed the redundent parts about starting knora-stack from `bazel.md`

benjamingeer:
OK but now `make stack-up-fast` isn't documented anywhere. I think `make stack-up-fast` rebuilds the `knora-api` container if necessary (after code changes), while `make stack-restart-api` doesn't, so it's important for developers to know about `make stack-up-fast`.

SepidehAlassi:
> OK but now `make stack-up-fast` isn't documented anywhere. I think `make stack-up-fast` rebuilds the `knora-api` container if necessary (after code changes), while `make stack-restart-api` doesn't, so it's important for developers to know about `make stack-up-fast`.



> OK but now `make stack-up-fast` isn't documented anywhere. I think `make stack-up-fast` rebuilds the `knora-api` container if necessary (after code changes), while `make stack-restart-api` doesn't, so it's important for developers to know about `make stack-up-fast`.

added the explanation of make stack-up-fast

benjamingeer (approved):
That's great, thanks!

SepidehAlassi:
> That's great, thanks!

Thanks for your review.

PR #1717, opened 2020-09-24
enhancement(gravsearch): Optimise queries by moving up statements with resource IRIs

benjamingeer:
https://dasch.myjetbrains.com/youtrack/issue/DSP-696

- [x] Add optimisation in `SparqlTransformer.scala`.
- [x] Always put `BIND` expressions at the beginning of a block, regardless of other optimisations.
- [x] Add tests.


benjamingeer:
@loicjaouen Could you please try this branch?

benjamingeer:
E2E test failing because of:

```
akka.pattern.AskTimeoutException: Ask timed out on [Actor[akka://E2ETest/user/applicationManager#-1727947013]]
after [5000 ms]. Message of type [org.knora.webapi.messages.admin.responder.usersmessages.UserGetADM]
was sent by [Actor[akka://E2ETest/user/applicationManager#-1727947013]].
A typical reason for `AskTimeoutException` is that the recipient actor didn't send a reply.
```

benjamingeer:
@SepidehAlassi I'm trying to fix the test error...

loicjaouen (approved):
from the end user point of view, it works like a charm, requests don't have to be re-written, thanks @benjamingeer

benjamingeer:
@SepidehAlassi The tests pass now, so I might have fixed the timeout error by setting `akka.ask.timeout = 10 seconds` in `test.conf`. What I don't understand is why it was timing out after 5 seconds, which isn't configured anywhere as far as I can tell. I think it should be using `app.default-timeout`. Maybe @subotic knows more and can have a look when he gets back.

SepidehAlassi (approved):
Looks good, thanks!

subotic:
Here we have a few things timing out after 5 seconds: https://github.com/dasch-swiss/knora-api/blob/9e28e5bdbacc3422af088028b0b2ad635a10627d/webapi/src/test/scala/org/knora/webapi/E2ESpec.scala#L127

subotic:
@benjamingeer If you add DSP-XXX to the title of the PR, then the PR will be automatically linked to the youtrack issue.

benjamingeer:
@subotic

> Here we have a few things timing out after 5 seconds

Yes, but the thing that timed out wasn't an HTTP request in an E2E test, it was a `UserGetADM` message sent by the application actor. Shouldn't that be using the ask timeout, which should be `app.default-timeout`?

> If you add DSP-XXX to the title of the PR, then the PR will be automatically linked to the youtrack issue.

Thanks, will do.

PR #1719, opened 2020-09-25
change possibly conflictual env var USERNAME (DSP-706)

loicjaouen:
I don't know what sets the environment variable `USERNAME`, but it appear to be set not only on my set-up:
https://unix.stackexchange.com/questions/76354/who-sets-user-and-username-environment-variables

So we may consider `USERNAME` a common environment variable and we should therefore not use it because it is probable that it is set by default even without us specifying it in the command argument list.


loicjaouen:
ultimately, in my case, it halted the command `make stack-up` because the `wait-for-db.sh` script did not return, not because the db was not up, but because my `USERNAME` and the default db user name differs.

subotic:
@loicjaouen Thanks for finding this out. Do we have a YouTrack issue for this?

loicjaouen:
@subotic : I was not sure we needed one ;)  
here it is: https://dasch.myjetbrains.com/youtrack/issue/DSP-706

subotic (approved):
Great, thanks :-)

LukasStoeckli (commented):
the following scripts wil have the same issue:
* fuseki-upload-repository.sh
* fuseki-functions.sh
* fuseki-empty-repository.sh
* fuseki-dump-repository.sh  

@loicjaouen did you just use wait-for-db?

loicjaouen:
@LukasStoeckli : I, indeed, only used  `wait-for-knora.sh`, I'll change the others, thanks for mentionning

loicjaouen:
here it is @LukasStoeckli

LukasStoeckli (approved):
great, thank you

PR #1722, opened 2020-09-30
feat(api-v1): Change API v1 file uploads to work like API v2 (DSP-41, PR 3)

### LINKED ISSUE #535: Don't require Knora and Sipi to share a directory for image upload
At the moment, the file upload through the ``webapi`` requires that there is a shared directory with ``sipi``. This limits the way how the ``webapi`` and ``sipi`` can be deployed. Either they need to be running on the same server or both have access to the same network share.

benjamingeer:
This PR changes API v1 so that Knora and Sipi no longer need to share a filesystem.

- [x] Port #1667 to the current code structure:
  - [x] Change API v1 to use the same Sipi scripts as API v2.
  - [x] Remove the API v1 Sipi scripts.
  - [x] Change `salsah1` to use `upload.lua`, sending it a JWT token.
  - [x] Update other API v1 tests.
  - [x] Update API v1 docs.
- [x] Add necessary support for text files in API v1.
- [x] Update `KnoraSipiIntegrationV1ITSpec` to create `XSLTransformation` resources using `upload.lua`.

**Breaking API v1 change:** To enable Knora and Sipi to work without sharing a filesystem, the procedure for uploading files in API v1 has changed; see `docs/03-apis/api-v1/adding-resources.md`, sections "Adding Resources with Image Files" and "Bulk Import with Image Files".

Moved from #1667.
Resolves #535.
Resolves [DSP-41](https://dasch.myjetbrains.com/youtrack/issue/DSP-41).
Resolves [DSP-42](https://dasch.myjetbrains.com/youtrack/issue/DSP-42).


subotic (approved):
Thanks, LGTM!

benjamingeer:
Thank you for the review! So glad to see this finally done.
🎉     🥳

PR #1725, opened 2020-10-02
docs: Update required mkdocs package

kilchenmann:
I figured out that the Github actions script for deploying the docs uses the latest version of mkdocs-material that contains a breaking change. So I had to update the configuration in documentation.


benjamingeer (approved):
Thank you!

PR #1733, opened 2020-10-15
DSP-470 Intermittent bind errors

subotic:
resolves [DSP-470](https://dasch.myjetbrains.com/youtrack/issue/DSP-470)

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/app/ApplicationActor.scala
+++ b/webapi/src/main/scala/org/knora/webapi/app/ApplicationActor.scala
@@ -473,7 +473,11 @@ class ApplicationActor extends Actor with Stash with LazyLogging with AroundDire
                         ex.getMessage,
                         retryCnt
                     )
-                    self ! AppStart(ignoreRepository, requiresIIIFService, retryCnt + 1)
+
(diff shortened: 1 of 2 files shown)


subotic:
Testing autolink to DSP-470

subotic:
@kilchenmann I just found out about autolink. Look at my previous comment. It is now possible to write `DSP-xyz` and Github will turn it into the defined link. The setting is here: https://github.com/dasch-swiss/knora-api/settings/key_links

kilchenmann:
> @kilchenmann I just found out about autolink. Look at my previous comment. It is now possible to write `DSP-123` and Github will turn it into the defined link. The setting is here: https://github.com/dasch-swiss/knora-api/settings/key_links

Ah great. When it now works like this 👍
DSP-470

subotic:
@benjamingeer @SepidehAlassi It seems to work, at least in this PR. We need to see, how it goes when used in different PRs. I think, that from time to time we get an old VM and slow VM. We are looking into having our own CI infrastructure.

benjamingeer (approved):
Looks good, let's try it!

subotic:
Thanks for the review 😀

PR #1737, opened 2020-10-21
DSP-842 Bump used Bazel version to newly released 3.7.0

subotic:
resolves DSP-842


BalduinLandolt (approved):
Looks good so far.  
(I'm guessing the `insert_final_newline = true` does not directly contribute to the aim of bumping bazel version. But ok.)

To mark DSP-842 as done, this would also need doing in knora-py.

subotic:
> To mark [DSP-842](https://dasch.myjetbrains.com/youtrack/issue/DSP-842) as done, this would also need doing in knora-py.

Yes, and probably we should have created two tasks und the user story. Maybe next time ;-)

PR #1739, opened 2020-10-21
enhancement(api-v2): Optimise checking isDeleted (DSP-848)

benjamingeer:
This PR adds an optimisation to improve the filtering of resources and values marked as deleted in Jena.

- [x] With Jena, use `FILTER NOT EXISTS` to check `knora-base:isDeleted` in queries that could return many results.
- [x] Improve the optimisation from #1717 by not moving statements whose predicate is `rdf:subject` or `rdf:object` (meaning that the subject is a `LinkValue` which should be referred to first).
- [x] Add test.

Resolves DSP-848.


benjamingeer:
@SepidehAlassi Loïc says this change improves performance a lot: https://dasch.myjetbrains.com/youtrack/issue/DSP-848#focus=Comments-4-1575.0-0

SepidehAlassi (approved):
Looks good, thanks!

benjamingeer:
Thanks for the review!

PR #1742, opened 2020-10-26
feat: Add feature toggles (DSP-910)

benjamingeer:
- [x] Implementation framework in `FeatureFactory.scala`, `KnoraSettings.scala`, and `KnoraRoute.scala`
- [x] Proof of concept: replace the implementation of `ListsRouteADM` using a feature toggle (to be used for DSP-597)
- [x] Update all `v2` and `admin` routes so they can use feature toggles
- [x] Fix `R2RSpec` so tests can override the application configuration
- [x] Tests: `FeatureToggleR2RSpec`
- [x] Docs:
  - [x] Design/development docs
  - [x] API docs

Resolves DSP-910.


benjamingeer:
Doing some redesign after discussion with @SepidehAlassi, will re-request reviews when done.

webapi/src/main/scala/org/knora/webapi/settings/KnoraSettings.scala
```diff
@@ -258,4 +329,37 @@ object KnoraSettings extends ExtensionId[KnoraSettingsImpl] with ExtensionIdProv
...
+    case class FeatureToggleBaseConfig(featureName: String,
+                                       description: String,
+                                       availableVersions: Seq[Int],
+                                       defaultVersion: Int,
+                                       enabledByDefault: Boolean,
+                                       overrideAllowed: Boolean,
+                                       expirationDate: Option[Instant],
+                                       developerEmails: Set[String])
+
 }
```
SepidehAlassi, line 365: please empty line here

webapi/src/main/scala/org/knora/webapi/settings/KnoraSettings.scala
```diff
@@ -258,4 +329,37 @@ object KnoraSettings extends ExtensionId[KnoraSettingsImpl] with ExtensionIdProv
...
+     * @param featureName       the name of the feature.
+     * @param description       a description of the feature.
+     * @param availableVersions the available versions of the feature.
+     * @param defaultVersion    the version of the feature that should be enabled by default.
+     * @param enabledByDefault  `true` if the feature should be enabled by default, `false` if it should be
+     *                          disabled by default.
+     * @param overrideAllowed   `true` if this configuration can be overridden, e.g. by per-request feature
+     *                          toggle configuration.
+     * @param expirationDate    the expiration date of the feature.
+     * @param developerEmails   one or more email addresses of developers who can be contacted about the feature.
```
SepidehAlassi, line 354: I am not sure why we need the developer's email? since feature toggles last a year, maybe until it is removed a developer quits, then his email address would be a useless piece of information.

webapi/src/test/scala/org/knora/webapi/e2e/FeatureToggleR2RSpec.scala
```diff
@@ -0,0 +1,328 @@
...
+
+        "not enable a nonexistent version of a toggle" in {
+            Get(s"/foo").addHeader(RawHeader(FeatureToggle.REQUEST_HEADER, "new-foo:3=on")) ~> fooRoute ~> check {
+                val responseStr = responseAs[String]
+                assert(status == StatusCodes.BadRequest, responseStr)
+                assert(responseStr.contains("Feature toggle new-foo has no version 3"))
+            }
+        }
+
+        "not accept a version number when disabling a toggle" in {
```
SepidehAlassi, line 312: why not?

SepidehAlassi (changes requested):
Thank you very much for implementing this. I just see the following problem about not being able to specify the version number when turning a toggle off: 
Imagine we have initial `newListRoute:1` and months later we add `newListRoute:2` both are by default off. After a while, we decide to make a release with `newListRoute:1` by default on, but `newListRoute:2` to be by default off because other subsystems still need time to catch up with changes of `newListRoute:2`.
Now, imagine a client pulls this release and is ready with support of `newListRoute:2` and wants to use it. He has to turn `newListRoute:1` off and turn `newListRoute:2` on. If he cannot specify the version of the toggle that must be off, as it is now to achieve what he wants he needs to do

```
 X-Knora-Feature-Toggles: new-foo=off, new-foo:2=on
```
that is kind of confusing. I would suggest specification of version for turning a toggle off to be possible, i.e. user must be able to set

```
 X-Knora-Feature-Toggles: new-foo:1=off, new-foo:2=on
```

benjamingeer:
> After a while, we decide to make a release with `newListRoute:1` by default on, but `newListRoute:2` to be by default off because other subsystems still need time to catch up with changes of `newListRoute:2`.

This would be done with a base configuration like this:

```
new-list-route {
    available-versions = [ 1, 2 ]
    default-version = 1
    enabled-by-default = yes
}

> Now, imagine a client pulls this release and is ready with support of `newListRoute:2` and wants to use it. He has to turn `newListRoute:1` off and turn `newListRoute:2` on.

No, he just has to request:

```
 X-Knora-Feature-Toggles: newListRoute:2=on
```

I think it wouldn't make sense to specify a version when disabling a feature, because you can't enable more than one version at a time. It's not possible to enable both `newListRoute:1` and `newListRoute:2`. So there are only two possibilities:

1. You turn on one version of the feature.
2. You turn off all versions of the feature.

To turn on one version of the feature, you just enable it, specifying the version you want. There's no need to disable the other versions, because you can only get one version.

File: webapi/src/main/scala/org/knora/webapi/app/ApplicationActor.scala [hunk 4]
```diff
@@ -122,11 +120,6 @@ class ApplicationActor extends Actor with Stash with LazyLogging with AroundDire
      */
     implicit protected val timeout: Timeout = knoraSettings.defaultTimeout
 
-    /**
-     * A user representing the Knora API server, used for initialisation on startup.
-     */
-    private val systemUser = KnoraSystemInstances.Users.SystemUser
```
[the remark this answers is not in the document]
benjamingeer: It was unused.

webapi/src/main/scala/org/knora/webapi/settings/KnoraSettings.scala
benjamingeer: I think it's useful to know who to contact if you have questions about a feature. If a developer quits, we can update the email address.

benjamingeer:
Your example is exactly what's happening in this test:

https://github.com/dasch-swiss/knora-api/pull/1742/files#diff-8395123710217ba2f8f6805becb91f2e60821112a89b46bb9e25cb5a86642439R287

benjamingeer:
@SepidehAlassi I've added docs to clarify things, and added a test. Is this better?

docs/05-internals/design/principles/feature-toggles.md
```diff
@@ -0,0 +1,281 @@
...
+without versions, and then get versions later, it would not be obvious
+what should happen if a client then requested the toggle without specifying
+a version number.) Version numbers must be an ascending sequence of
+consecutive integers starting from 1.
+
+If `expiration-date` is provided, it must be an [`xsd:dateTimeStamp`](http://www.datypic.com/sc/xsd11/t-xsd_dateTimeStamp.html). All feature toggles
+should have expiration dates except for long-lived ops toggles like `fast-baz` above.
+
+`KnoraSettingsFeatureFactoryConfig` reads this base configuration on startup. If
+a feature toggle has an expiration date in the past, the application will not start.
```
subotic, line 136: oh, this is a hard one. I wouldn't punish ops for developers not removing a toggle in time ;-) Also what should then happen, if the expiration date is passed while the app is deployed into production. I hope the app will not stop working? I think a log output would be sufficient.

SepidehAlassi:
> @SepidehAlassi I've added docs to clarify things, and added a test. Is this better?

yes, it is very clear now, thank you.

SepidehAlassi (approved):
Thank you for the changes, it looks good to me.

webapi/src/main/scala/org/knora/webapi/routing/admin/ListsRouteADMFeatureFactory.scala
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param routeData the [[KnoraRouteData]] to be used in constructing the routes.
+ */
+class ListsRouteADMFeatureFactory(routeData: KnoraRouteData) extends KnoraRouteFactory(routeData)
+    with FeatureFactory {
+
+    /**
+     * The old lists route feature.
+     */
+    private val oldListsRouteADMFeature = new OldListsRouteADMFeature(routeData)
```
subotic, line 37: Could we maybe move the factory and features to a sub-package, e.g., org.knora.webapi.routing.admin.lists?

subotic (approved):
Thanks for implementing this :-)

docs/05-internals/design/principles/feature-toggles.md
```diff
@@ -0,0 +1,281 @@
...
+- Feature implementations should be produced by factory classes,
+  so that the code using a feature does not need to know
+  about the toggling decision.
+  
+- Feature factories should use toggle configuration taken
+  from different sources, without knowing where the configuration
+  came from.
+  
+- An HTTP response should indicate which features are turned
+  on.
```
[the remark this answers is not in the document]
benjamingeer: Currently it only returns enabled toggles. I was thinking that if a toggle is turned off by default because the feature isn't ready to use yet, there wouldn't be much point in returning it. But maybe it would be clearer and simpler to return everything. What do you think?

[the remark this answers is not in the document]
SepidehAlassi: I think returning everything might not be a good idea. Since the lifetime of toggles is a year, we might have many (>10 even) toggles in place that are turned off but still available. IMO, Returning them would confuse the user.

SepidehAlassi:
@benjamingeer I keep thinking about how we are going to add non-breaking new features, can you please not merge this PR until we discuss this thoroughly. 

Please imagine the following scenario: 
We have`oldListRoute` (the original existing form), `newListRoute:1` (contains some breaking changes), and `newListRoute:2` (contains even more breaking changes). That would mean we have three versions of `ListsRoute.scala` file, 3 versions of its spec, and 3 versions of its documentation. At this point, we decide to add a totally new feature that is non-breaking (like a new route), how are we going to handle it? I see the following options, please correct me if I am missing something:

1. Are we going to define a new toggle for it `newListRoute:3` to wrap the new feature? That would mean the new feature would not be available to those who have not yet the support for `newListRoute:2`. That means clients cannot benefit from the new features for a long time because they do not have the support for the undelaying breaking changes.

2. Are we going to add the new non-breaking feature to all three existing versions of the code (`oldListRoute`, `newListRoute:1`, and `newListRoute:2`)? This would make the new non-breaking feature available to all clients no matter which version of toggle they support, even those who still use the good old `oldListRoute`. However, that would mean adding the new piece of code to all 3 versions of `ListsRoute.scala`, its tests have to be added to all 3 versions of the `ListsRouteSpec.scala`, and its documentation has to be added to all three versions of `lists.md`. That would mean modifying 9 files instead of 3.  The more toggles there are for routes, the more files to modify.

3. Are we going to add a completely new independent toggle for it, `newRouteforList:1`, with a very short lifetime? That means, if `newRouteforList:1=on` use the copy of `ListsRoutes.scala` that has the new non-breaking feature, otherwise use the `oldListRoute`. In this case, once the short-lived toggle `newRouteforList:1` is removed, we should add the same changes to the codes under toggles `newListRoute:1`, and `newListRoute:2` that are still alive.

I would appreciate knowing your opinion on this, it can well be that I am missing a fourth perfect option.

subotic:
Does it probably depend on the granularity of a feature toggle? In the case of `newListRoute:1` the whole route is under the toggle. Maybe in your example, it would make sense to only have parts of the route under the toggle?

Or, the implementation is broken down one more level, where each subroute goes into its separate file and the factory part is only referencing the separate implementations. Those separate implementations are then unit tested.

I'm sure, that there are a lot more questions open, that we didn't think of yet. We need to jump into the cold water and see what happens.

docs/05-internals/design/principles/feature-toggles.md
benjamingeer: I changed it to:

```
the application will log the error message "SHOW-STOPPER!!!" and the server will then self-destruct
```

Just kidding. 😃

benjamingeer:
I've made the request and response headers identical, because I think this will make the response header easier to read: you don't have figure out that something is turned off because it's not there, you can see that it's turned off.

I changed `KnoraSettings` so that if a feature toggle has expired, you just get a warning in the log on startup.

I moved the list factory into a sub-package.

@subotic you wrote:

> Ideally, there would also be something like `x-knora-api-version`

Doesn't the existing server version header include that?

@SepidehAlassi About non-breaking new features, I agree with Ivan. In general I think it wouldn't be bad to go with your option (2), especially if we do as Ivan suggests:

> the implementation is broken down one more level, where each subroute goes into its separate file and the factory part is only referencing the separate implementations. Those separate implementations are then unit tested.

That would reduce code duplication.

But maybe we can take a different approach in each case, depending on how urgent the new feature is, how much work is involved in implementing support for the breaking changes, etc.

If it's OK with both of you, I'll merge this now.

subotic:
Fine with me :-) Thanks!

SepidehAlassi:
@benjamingeer it is also fine with me, thanks for all your work.

benjamingeer:
Many thanks to both of you for the very helpful reviews!

PR #1748, opened 2020-11-03
DSP-756 Tests failing because Knora version header and route are incorrect

subotic:
resolves DSP-756


benjamingeer (changes requested):
What was the problem? Could you add some documentation to explain how this works now? I'd like to understand it in case I have to fix it someday.

subotic:
@benjamingeer this was a problem with Bazel, i.e., I've incorrectly implemented it. It only worked when run the first time after cleaning the cache, but failed on subsequent runs. Also, the test was never run on Github-CI, so I falsely assumed that everything was fine.

> What was the problem? Could you add some documentation to explain how this works now? I'd like to understand it in case I have to fix it someday.

sure, I will add a section about build stamping to the docs.

subotic:
@benjamingeer could you please take a look at the docs and see if it is described good enough?

benjamingeer (approved):
Excellent, thank you very much!

subotic:
Thanks for the review :-)

PR #1751, opened 2020-11-06
DSP-1022 SIPI_EXTERNAL_HOSTNAME doesn't contain the external hostname

subotic:
resolves DSP-1022

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,7 +29,7 @@ services:
     restart: unless-stopped
     environment:
       - SIPI_EXTERNAL_PROTOCOL=http
-      - SIPI_EXTERNAL_HOSTNAME=sipi
+      - SIPI_EXTERNAL_HOSTNAME=localhost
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -29,7 +29,7 @@ services:
     restart: unless-stopped
     environment:
       - SIPI_EXTERNAL_PROTOCOL=http
-      - SIPI_EXTERNAL_HOSTNAME=sipi
+      - SIPI_EXTERNAL_HOSTNAME=localhost
```
mpro7, line 32: Question. This will work also on other environments? So uploading something on production will also return localhost as a hostname and will this address work?

subotic: this is only the config for the local dev environment. Prod and Test have their own configuration.

mpro7 (approved):

docker-compose.yml
mpro7: I've checked locally and works as expected. Thanks.

PR #1755, opened 2020-11-09
DSP-1029: Add the missing dependency

SepidehAlassi:
resolves DSP-1029


benjamingeer (approved):
Thank you!

subotic:
@SepidehAlassi Bazel complains only about not needed compile time dependencies. This seems to me to be a runtime dependency, which has to be added to the correct property. Please make this correction in one of your future PRs or it is just a matter of time until it will happen again. I don't delete dependencies by hand, instead I use the `buildozer` tool and this tool doesn't know how to read (comments) 😉

PR #1761, opened 2020-11-19
DSP-1099 Update Bazel maven rules to see if it fixes problems with macOS Big Sur

subotic:
resolves DSP-1099


subotic:
As it turns out, the problem we experienced wasn't with Bazel, but the Cisco packet filter which was automatically activated in Big Sur. Removing the packet filters, solved the issue.

benjamingeer (changes requested):
> Removing the packet filters, solved the issue.

Can you document what we need to do, so others don't get the same problem?

subotic:
This is a local macOS Big Sur network issue. Where would you like me to document it? I wrote it inside the YouTrack issues and mentioned it here.

benjamingeer:
If I understand correctly, every developer who uses macOS will need to do some additional configuration to avoid this problem. So a good place to document it would be somewhere under "Development":

https://docs.knora.org/05-internals/development/

subotic:
I've added a note about macOS Big Sur to the Maven dependencies docs.

benjamingeer (approved):
OK great, thanks!

subotic:
Thanks for the review :-)

PR #1762, opened 2020-11-25
feat(rdf-api): Add a general-purpose SHACL validation utility (DSP-930)

benjamingeer:
resolves DSP-930.

- [x] Add SHACL validation API in `ShaclValidator.scala`.
- [x] Add implementations:
  - [x] `JenaShaclValidator`
  - [x] `RDF4JShaclValidator`
- [x] Return these from `RdfFeatureFactory`.
- [x] Add tests.
- [x] Update docs.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/util/rdf/AbstractShaclValidator.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/util/rdf/AbstractShaclValidator.scala
@@ -1,0 +1,91 @@
+/*
+ * Copyright © 2015-2019 the contributors (see Contributors.md).
+ *
+ * This file is part of Knora.
+ *
+ * Knora is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Knora is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with Knora.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.knora.webapi.messages.util.rdf
+
+import java.io.File
+import java.nio.file.Path
+
+import org.knora.webapi.exceptions.AssertionException
+
+/**
+ * An abstract base class for classes that validate RDF models using SHACL shapes.
+ *
+ * @param baseDir       the base directory that SHACL graphs are loaded from.
+ * @param rdfFormatUtil an [[RdfFormatUtil]].
+ * @tparam ShaclGraphT an implementation-specific representation of a graph of SHACL shapes.
+ */
+abstract class AbstractShaclValidator[ShaclGraphT](baseDir: File, private val rdfFormatUtil: RdfFormatUtil) extends ShaclValidator {
+
+    /**
+     * A map of relative paths to objects representing graphs of SHACL shapes.
+     */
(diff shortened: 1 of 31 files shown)


webapi/src/main/scala/org/knora/webapi/messages/util/rdf/AbstractShaclValidator.scala
```diff
@@ -0,0 +1,91 @@
...
+            rdfModel = rdfModel,
+            shaclGraph = shaclGraphs.getOrElse(shaclPath, throw AssertionException(s"SHACL graph $shaclPath not found"))
+        )
+    }
+
+    /**
+     * Recursively loads graphs of SHACL shapes from a directory and its subdirectories.
+     *
+     * @param baseDir the base directory that SHACL graphs are loaded from.
+     * @param dir     the base directory or a subdirectory.
```
SepidehAlassi, line 56: It seems like the Shacl shapes are actually loaded from `dir` directory and its subdirectories, and `baseDir` is only used to get its path to make the `relativePath`. Can you please change the comments for `baseDir` and `dir` in scaladocs? 

I also wonder, if only the path of the `baseDir` is needed here, why not just using its path as the parameter of this function, I mean
```
 private def  loadShaclGraphs(baseDirPath: Path, dir: File) : Map[Path, ShaclGraphT] = {...}
```  
two directory parameters are a bit confusing.

webapi/src/main/scala/org/knora/webapi/messages/util/rdf/jenaimpl/JenaShaclValidator.scala
```diff
@@ -0,0 +1,69 @@
...
+    import JenaConversions._
+
+    private val shaclValidator: jena.shacl.ShaclValidator = jena.shacl.ShaclValidator.get
+
+    override def validateWithShaclGraph(rdfModel: RdfModel, shaclGraph: jena.shacl.Shapes): ShaclValidationResult = {
+        // Get the jena.graph.Graph representing the model's default graph.
+        val graphToValidate: jena.graph.Graph = rdfModel.asJenaDataset.asDatasetGraph.getDefaultGraph
+
+        // Validate the data and get Jena's validation report.
+        val validationReport: jena.shacl.ValidationReport = shaclValidator.validate(shaclGraph, graphToValidate)
```
SepidehAlassi, line 49: Nice! Jena Shacl `validate` is like pyShacl getting the same parameters and returning report and conforms in the same way.

SepidehAlassi (approved):
This looks great! Thanks for implementing this.

webapi/src/main/scala/org/knora/webapi/messages/util/rdf/AbstractShaclValidator.scala
benjamingeer: > It seems like the Shacl shapes are actually loaded from `dir` directory and its subdirectories, and `baseDir` is only used to get its path to make the `relativePath`.... I also wonder, if only the path of the `baseDir` is needed here, why not just using its path as the parameter of this function

Actually, the SHACL shapes are loaded from `baseDir` and its subdirectories. `File` and `Path` are basically the same thing, but `File` is the old `java.io` API which I'm more used to, and `Path` is the newer `java.nio` API. The consensus in the Java community seems to be that seems that `Path` should be used instead of `File` for everything nowadays (https://stackoverflow.com/a/26658436). I guess we should do a PR sometime to use `java.nio` instead of `java.io` everywhere. For now, I've refactored `AbstractShaclValidator` to use `java.nio` to walk the directory tree, which seems clearer.

webapi/src/main/scala/org/knora/webapi/messages/util/rdf/jenaimpl/JenaShaclValidator.scala
benjamingeer: Standards are nice. :)

PR #1767, opened 2020-12-02
enhancement(triplestore): Use N-Quads instead of TriG for repository upgrade (DSP-1129)

benjamingeer:
resolves DSP-1129

- [x] Add N-Quads support to the RDF API.
- [x] Use N-Quads for uploading and downloading the repository in `HttpTriplestoreConnector` and `RepositoryUpdater`.
- [x] Fix a bug in `RepositoryUpdater` that caused plugins to be run out of order.
- [x] Have `RdfFormatUtil` close streams when possible, so the caller doesn't have to.
- [x] Correctly convert empty blank nodes from an `RdfModel` to JSON-LD.
- [x] Add tests.


SepidehAlassi (approved):
This looks great, thanks for implementing it.

benjamingeer:
Thanks for the review!

PR #1770, opened 2020-12-04
refactor: Use java.nio.file.Path instead of java.io.File (DSP-1124)

benjamingeer:
resolves DSP-1124


SepidehAlassi (approved):
This looks great, thank you.

PR #1772, opened 2020-12-08
chore(rdf-api): Use the Jena RDF API implementation by default (DSP-1153)

benjamingeer:
resolves DSP-1153

- [x] Set feature toggle `jena-rdf-library` to be on by default.
- [x] Fix SPARQL in tests to use named graphs correctly with Jena.
  - [x] Document why.
- [x] Fix the string representation of datatype values in SPARQL query results from an in-memory repository.
  - [x] Check that in a test.


SepidehAlassi (approved):
This looks good, thanks!

benjamingeer:
Thanks for reviewing!

PR #1773, opened 2020-12-10
DSP-667 CI is failing to test upgrade correctly

subotic:
- resolves DSP-667
- refactor: move IT tests to "test" folder
- remove `KnoraSipiScriptsV1ITSpec.scala` (see DSP-707 for explanation)


benjamingeer (changes requested):
Thanks very much, this is a big improvement. Could we also get rid of `salsah1/src/test/scala/org/knora/salsah/TestContainers.scala`?

subotic:
> Could we also get rid of `salsah1/src/test/scala/org/knora/salsah/TestContainers.scala`?

Hmm, the `salsah1` tree is completely independent of the `webapi` tree, i.e., nothing is shared. If we break this isolation, then it will be more work to maintain the whole thing. Any changes in `webapi` could then potentially break something in `salsah1`. Maybe it would be better to extract `salsah1` out of the `dsp-api` repository?

benjamingeer:
> Maybe it would be better to extract salsah1 out of the dsp-api repository?

You're right, that makes more sense. Would you rather make another PR for that?

benjamingeer (approved):

subotic:
> > Maybe it would be better to extract salsah1 out of the dsp-api repository?
> 
> You're right, that makes more sense. Would you rather make another PR for that?

yes, I can make another PR for that.

PR #1776, opened 2021-01-04
feat(api-v2): Optionally return file values in full-text search results (DSP-1191)

benjamingeer:
Resolves DSP-1191.

- [x] Accept parameter `returnFiles=true` in full-text search route, to request any file values attached to matching resources.
- [x] Implement this in SPARQL.
- [x] Support it in the search responder.
- [x] Add tests.
- [x] Add docs.


kilchenmann:
Did I understand that correctly: If I specify the parameter "returnImageFiles=true" in my search query, I'll get back the information about images. Will this be extended for other representations (moving images, audio, document) later? But then we will have a lot of parameters.

benjamingeer:
@kilchenmann 

> Did I understand that correctly: If I specify the parameter "returnImageFiles=true" in my search query, I'll get back the information about images.

Yes.

> Will this be extended for other representations (moving images, audio, document) later? But then we will have a lot of parameters.

Which other file types would you like to get? How would you do a thumbnail of an audio file?

kilchenmann:
> @kilchenmann
> 
> > Did I understand that correctly: If I specify the parameter "returnImageFiles=true" in my search query, I'll get back the information about images.
> 
> Yes.
> 
> > Will this be extended for other representations (moving images, audio, document) later? But then we will have a lot of parameters.
> 
> Which other file types would you like to get? How would you do a thumbnail of an audio file?

I will integrate a preview for `MovingImageRepresentation` and a pre-listening thing of an `AudioRepresentation`. I did some experiments in my master thesis how we can do a "preview" of an audio file. Additional it would be nice to have a preview of a `DocumentRepresentation`. Anyway it would be good to know, if a resource has a file representation and we can show already in the simple list view a corresponding icon to highlight resources with file representation (similar to eMails where we see if it has an attachment or not).

benjamingeer:
> Anyway it would be good to know, if a resource has a file representation

How about if I change the parameter to `returnFiles`, and it returns any representation it finds?

kilchenmann:
> How about if I change the parameter to returnFiles, and it returns any representation it finds?

That would be great 👍 Thanks

SepidehAlassi (approved):
this looks great, thanks for implementing it.

benjamingeer:
Thanks for the review!

PR #1781, opened 2021-01-13
feat(api-v2): Generate IIIF manifest (DSP-50)

benjamingeer:
resolves DSP-50


benjamingeer:
Redoing this in another way.

PR #1786, opened 2021-01-14
refactor: bring back stack-down-delete-volumes

mpro7:
This PR is due [that change](https://github.com/dasch-swiss/dsp-api/pull/1773/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L189) and brings back `stack-down-delete-volumes` stack target. It also updates copyright years and pr template.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,4 @@
 ===REMOVE===
 Important! Please follow the new guidelines for naming Pull Requests: https://docs.dasch.swiss/developers/dsp/contribution/#pull-request-guidelines
 ===REMOVE===
-
 resolves DSP-
--- a/Makefile
+++ b/Makefile
@@ -275,27 +275,27 @@ test: docker-build ## runs all test targets.
 #################################
 
 .PHONY: init-db-test
-init-db-test: stack-db-remove stack-db-only ## initializes the knora-test repository
+init-db-test: stack-down-delete-volumes stack-db-only ## initializes the knora-test repository
 	@echo $@
 	@$(MAKE) -C webapi/scripts fuseki-init-knora-test
 
 .PHONY: init-db-test-minimal
-init-db-test-minimal: stack-db-remove stack-db-only ## initializes the knora-test repository with minimal data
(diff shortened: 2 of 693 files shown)


subotic (approved):
great, thanks. All 693 changed files LGTM :-)

BalduinLandolt:
IMO some of the wording in the doc could be made a bit clearer here: https://docs.knora.org/04-publishing-deployment/getting-started

The order in which `make stack-up` and `make init-db-test` are presented to me suggests that one should execute `stack-up` first. But that doesn't work, as I had to learn the hard way. ;-)  
Especially the sentence "... after which you can start the stack again with `make stack-up`, recreate the repository and load the data with `make init-db-test`" is misleading to me.  
It might even be worth being explicit on that, and just write that `init-db-test` must be called first.

BalduinLandolt:
also, I would appreciate a make rule that calls `make init-db-test` and `make stack-up` consecutively, for lazy people like me.

SepidehAlassi (approved):

mpro7:
> IMO some of the wording in the doc could be made a bit clearer here: https://docs.knora.org/04-publishing-deployment/getting-started
> 
> The order in which `make stack-up` and `make init-db-test` are presented to me suggests that one should execute `stack-up` first. But that doesn't work, as I had to learn the hard way. ;-)
> Especially the sentence "... after which you can start the stack again with `make stack-up`, recreate the repository and load the data with `make init-db-test`" is misleading to me.
> It might even be worth being explicit on that, and just write that `init-db-test` must be called first.

Good point.
Another thing is, the `stack-down-delete-volumes` is also somehow redundant here, because we have `init-db-test` which contains it:
```
.PHONY: init-db-test
init-db-test: stack-db-remove stack-db-only ## initializes the knora-test repository
	@echo $@
	@$(MAKE) -C webapi/scripts fuseki-init-knora-test
```

And since I've notice it, I only use 2 commands for initialisation of DSP-API
1. `make init-db-test`
2. `make stack-uo`

PR #1788, opened 2021-01-18
fix(api-v2): Don't allow an invalid cardinality on a boolean property (DSP-1236)

benjamingeer:
resolves DSP-1236

- [x] Don't allow adding an invalid cardinality on a boolean property.
- [x] Allow startup even if ontologies are invalid.
- [x] Add test.


SepidehAlassi (approved):
This looks great, thanks for fixing the bug.

benjamingeer:
Thanks for the review!

PR #1791, opened 2021-01-21
fix(api-v1): Optimise link value queries for Fuseki (DSP-1243)

benjamingeer:
This PR improves the performance of the API v1 SPARQL queries that find link values when using Fuseki.

- [x] Use IRI literals instead of `BIND`.
- [x] Use `FILTER NOT EXISTS` to check `knora-base:isDeleted` (as in #1739).
- [x] Remove unnecessary graph patterns.

resolves DSP-1243

### DIFF
--- a/webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/v1/findLinkValueByIri.scala.txt
+++ b/webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/v1/findLinkValueByIri.scala.txt
@@ -45,44 +45,37 @@ SELECT ?linkValue ?directLinkExists ?targetResourceClass ?objPred ?objObj
     FROM <http://www.ontotext.com/explicit>
 }
 WHERE {
-    BIND(IRI("@subjectIri") AS ?subject)
-    BIND(IRI("@predicateIri") AS ?predicate)
-    BIND(IRI("@{predicateIri}Value") AS ?predicateForLinkValue)
-
-    @maybeObjectIri match {
-        case Some(objectIri) => {
-
-            BIND(IRI("@objectIri") AS ?object)
-
-        }
-
-        case None => {}
-    }
-
     BIND(IRI("@linkValueIri") AS ?linkValue)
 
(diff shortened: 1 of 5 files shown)


webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/v1/findLinkValueByIri.scala.txt
```diff
@@ -45,44 +45,37 @@ SELECT ?linkValue ?directLinkExists ?targetResourceClass ?objPred ?objObj
...
 WHERE {
-    BIND(IRI("@subjectIri") AS ?subject)
-    BIND(IRI("@predicateIri") AS ?predicate)
-    BIND(IRI("@{predicateIri}Value") AS ?predicateForLinkValue)
-
-    @maybeObjectIri match {
-        case Some(objectIri) => {
-
-            BIND(IRI("@objectIri") AS ?object)
+    BIND(IRI("@linkValueIri") AS ?linkValue)
```
SepidehAlassi, line 48: In line 50 and 72 below, you are using `<@linkValueIri>` not the `?linkValue`, is the bind here really necessary? 
I see that `?linkValue` is used in `Select`, so if for that purpose BIND is necessary, why not use `?linkValue` in the other statements?

SepidehAlassi (changes requested):
This looks great, I just have one point, please see the comment.

webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/v1/findLinkValueByIri.scala.txt
benjamingeer: The bind is there only so we can return `?linkValue` in `SELECT`. Using an IRI literal instead of a variable improves performance, so I changed all the variables to literals wherever possible in these queries.

SepidehAlassi (approved):
Thanks for this optimization!

benjamingeer:
Thanks for reviewing!

PR #1799, opened 2021-01-27
fix (permissions): fix getting builtin groups while creating a permission (DSP-1296 )

SepidehAlassi:
resolves DSP-1296


subotic (approved):
awesome, thanks :-)

PR #1804, opened 2021-01-29
chore(gh-ci): Bring back the client-test-data command to github actions

kilchenmann:
We have to be sure, that everything runs well before publishing new version of DSP-API. The last release was not successful because of failing `make client-test-data` command. This PR brings back this step to github actions. The release publish process depends on a successful generation of client-test-data.


subotic (approved):
thanks :-)

PR #1808, opened 2021-02-01
feat(sipi): add storing of original and sidecar (DSP-1318)

lrosenth:
resolves DSP-1318. The changed lua scripts now preserve the original uploaded file and create a sidecar file that contains the original filename, derivative filename and SHA256-checksums.

In SIPI "knora.json" has been changed to read the sidecar file if available and return also the original filename for non-image files. This the original filename can be preserved on the dsp-api backend

DSP-API now returns a URL ending in `/file` as the object of `knora-api:fileValueAsUrl` for non-image files.


subotic:
@benjamingeer I've updated the Sipi version.

benjamingeer:
@subotic I've been working on it this morning, it's almost done.

benjamingeer (approved):
Thanks very much, I think this is good to merge now.

PR #1810, opened 2021-02-03
chore: make targets for adding metadata (DSP-1289)

BalduinLandolt:
Resolves DSP-1289

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,6 +111,10 @@ stack-up: docker-build env-file ## starts the knora-stack: fuseki, sipi, redis, 
 	$(CURRENT_DIR)/webapi/scripts/wait-for-db.sh
 	docker-compose -f docker-compose.yml up -d
 	$(CURRENT_DIR)/webapi/scripts/wait-for-knora.sh
+
+.PHONY: stack-up-with-metadata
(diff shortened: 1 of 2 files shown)


mpro7:
Thanks @BalduinLandolt, I add my 2 Rappens here instead MD channel.

I think `stack-down-delete-volumes` is obsolete, because `init-db-test` is covering it.

I would also not include into `make metadata` command others than these related to metadata, because most of the time at least `init-db-test` would be obsolete here too. IMO into `make metadata` I would include just `make stack-up` and all possible dataset to be loaded.

BalduinLandolt:
@mpro7 thanks for your input!

Sadly, we can only ever have one metadata set at once, since I'm loading them into the same project, otherwise they would just overwrite one another.  
But yes, I'll change it right away.

mpro7 (approved):
This is it, thanks!

BalduinLandolt:
@mpro7 I moved the metadata stuff to a separate `stack-up-with-metadata` target. @subotic did not want me to clutter `stack-up` even more.

mpro7 (approved):
Yes, that makes sense.

subotic:
@BalduinLandolt I have changed the targets a bit. Could you please take a quick look and try it out?

PR #1811, opened 2021-02-03
fix(Lists): Repositioning the node when new position equals length of new parent's children (DSP-1322)

SepidehAlassi:
resolves DSP-1322


benjamingeer (approved):
Great, thank you!

PR #1813, opened 2021-02-04
feat(gravsearch): Optimise Gravsearch queries using topological sort (DSP-1327)

benjamingeer:
resolves DSP-1327
resolves DSP-1361


benjamingeer:
### Removing special cases for standoff classes

The inferring type inspector always returned `knora-api:StandoffTag` as the type of any standoff tag, even if a more specific type was provided in the input query. This would have required a workaround in `AbstractPrequeryGenerator.removeEntitiesInferredFromProperty` so that a statement like this wouldn't be removed:

```sparql
?standoffParagraphTag a standoff:StandoffParagraphTag .
```

Instead, I refactored things so that the inferring type inspector returns the actual types of standoff tags.

### Not overriding a property's real object type with an inferred one

Given something like this:

```sparql
?standoffDateTag knora-api:standoffTagHasStartAncestor ?standoffParagraphTag .
?standoffParagraphTag a standoff:StandoffParagraphTag .
```

The inferring type inspector would:

1. Infer that `?standoffParagraphTag` is a `knora-api:StandoffTag`, from the object type of `knora-api:standoffTagHasStartAncestor` in the ontology.
2. Infer that the object type of `knora-api:standoffTagHasStartAncestor` is `standoff:StandoffParagraphTag`, because it's used with an object of that type.
3. Infer that `?standoffParagraphTag` is a `standoff:StandoffParagraphTag`, because it's used as the object of `knora-api:standoffTagHasStartAncestor`, whose type was inferred in (2), and also record that type in `typeInspectionResult.entitiesInferredFromProperties`.

Therefore, `AbstractPrequeryGenerator.removeEntitiesInferredFromProperty` would incorrectly remove this statement pattern:

```sparql
?standoffParagraphTag a standoff:StandoffParagraphTag .
```

The error is in step (2) above: if we got the object type of a property from its definition in an ontology, we shouldn't infer a more specific object type for it from an object that it's used with. I solved this in the inferring type inspector as follows:

1. Make sure that `InferTypeOfPropertyFromItsIri` is always run before `InferTypeOfPredicateFromObject` in the rule pipelines (already the case).
2. In `InferTypeOfPredicateFromObject`, if we already have an object type for the property, don't infer a subclass of that type.

benjamingeer:
@SepidehAlassi In `NonTriplestoreSpecificGravsearchToPrequeryTransformerSpec`, in the test `reorder query patterns in where clause with union`, the input `UNION` looks like this:

```sparql
{
    ?thing anything:hasRichtext ?richtext .
    FILTER knora-api:matchText(?richtext, "test")
    ?thing anything:hasInteger ?int .
    ?int knora-api:intValueAsInt 1 .
}
UNION
{
    ?thing anything:hasText ?text .
    FILTER knora-api:matchText(?text, "test")
    ?thing anything:hasInteger ?int .
    ?int knora-api:intValueAsInt 3 .
}
```

After topological sort, it looks like this:

```sparql
{
    ?thing anything:hasRichtext ?richtext .
    ?int knora-api:intValueAsInt 1 .
    ?thing anything:hasInteger ?int .
    FILTER(knora-api:matchText(?richtext, "test"^^xsd:string))
} UNION {
    ?thing anything:hasText ?text .
    ?int knora-api:intValueAsInt 3 .
    ?thing anything:hasInteger ?int .
    FILTER(knora-api:matchText(?text, "test"^^<http://www.w3.org/2001/XMLSchema#string>))
}
```

I'm wondering if there's a way to make `?int knora-api:intValueAsInt 1 .` come before `?thing anything:hasRichtext ?richtext .`

SepidehAlassi (approved):
thanks for all your work on this!

benjamingeer:
Thanks for doing most of it and for reviewing!

SepidehAlassi (changes requested):
I just remembered we have not added any documentation about topological sorting and its feature toggle. Let's add that before merging this.

SepidehAlassi:
BTW,
> I'm wondering if there's a way to make `?int knora-api:intValueAsInt 1 .` come before `?thing anything:hasRichtext ?richtext .`

well, topological sorting works with the degree of nodes; here both `?thing` and `?int` are of 2nd degree, so both permutations `(?thing, ?int, ?richtext, 1)` and `(?thing, ?int, 1, ?richtext)` are valid results of topological sorting. We can add a logic to consider the degree of nodes with respect to the number of outgoing edges instead of the number of total edges. In that case, `?int` would be of 1st degree and `?thing` of 2nd degree. 
An easy approach to do this without changing the logic of `findAllTopologicalOrders` would be the following: 
after calculating all permutations of topological sorting and obtaining results of `findOrdersNotEndingWithObjectOfRdfType`, instead of choosing any of `ordersNotEndingWithObjectOfRdfType`, we choose the one ordered with the number of outgoing edges. 
That means; considering the example above, `ordersNotEndingWithObjectOfRdfType` would contain `o1=(?thing, ?int, ?richtext, 1)` and `o2=(?thing, ?int, 1, ?richtext)`.  Putting the root element, `?thing` aside (because it is not an object of any statement), we take each node in these orders, find the statement it is the object of, calculate the number of outgoing edges of the node representing the subject of this statement. That would give for o1 ->  (2, 2, 1) and for o2 -> (2,1,2). Then we choose the permutation which ends with the lowest number, which would be o1.

SepidehAlassi:
@benjamingeer So I adjusted the code to get all permutations of the topological order according to Kahn's algorithm. For the following query 
```sparql
{
    ?thing anything:hasRichtext ?richtext .
    FILTER knora-api:matchText(?richtext, "test")
    ?thing anything:hasInteger ?int .
    ?int knora-api:intValueAsInt 1 .
}
UNION
{
    ?thing anything:hasText ?text .
    FILTER knora-api:matchText(?text, "test")
    ?thing anything:hasInteger ?int .
    ?int knora-api:intValueAsInt 3 .
}
```

After the topological sort, now it looks like this:

```sparql
{
   ?int knora-api:intValueAsInt 1 .
    ?thing anything:hasRichtext ?richtext .
    ?thing anything:hasInteger ?int .
    FILTER(knora-api:matchText(?richtext, "test"))
} UNION {
    ?int knora-api:intValueAsInt 3 .
    ?thing anything:hasText ?text .
    ?thing anything:hasInteger ?int .
    FILTER(knora-api:matchText(?text, "test"))
}
```

As you can see with the corrected version of the code, `?int knora-api:intValueAsInt 1 .` comes before `?thing anything:hasRichtext ?richtext .`

For the first block of the union, there are two valid topological sorting permutations: O1=(?thing, ?richtext, ?int, 1) and O2=(?thing, ?int, ?richtext, 1). As can be seen above, O2 is chosen to sort the statements. However, I believe maybe we should define logic to prefer O1 instead to make the query even faster. What do you think?

benjamingeer:
> So I adjusted the code to get all permutations of the topological order according to Kahn's algorithm.

Excellent!

> However, I believe maybe we should define logic to prefer O1 instead to make the query even faster. What do you think?

Sounds good to me, how do we do that?

SepidehAlassi:
> Sounds good to me, how do we do that?

I wrote down the idea in [DSP-1327](https://dasch.myjetbrains.com/youtrack/issue/DSP-1327). This can be implemented in a separate PR.

SepidehAlassi:
@benjamingeer I added the documentation, can you please proof-read it and add any missing points?

benjamingeer:
This looks good to me, I proofread the docs. If you approve the PR we can merge it.

SepidehAlassi (approved):
@benjamingeer thank you.

PR #1820, opened 2021-02-15
fix(Lists): Shift the second of two children after deletion of the first (DSP-1353)

SepidehAlassi:
resolves DSP-1353


benjamingeer (approved):
One good thing about off-by-one errors is that they're easy to fix. :)

mdelez (approved):
Thank you!!

PR #1826, opened 2021-03-03
fix(OntologyResponderV2): Fix check when updating ontology label and comment (DSP-1390)

benjamingeer:
resolves DSP-1390
resolves DSP-1397


mdelez:
Looks good to me! Could you also add some logic to allow an empty string for a comment? I don't believe comments are required and currently there is no way to remove a comment once one is provided.

benjamingeer:
@mdelez Thanks!

> Could you also add some logic to allow an empty string for a comment?

We don't allow empty strings in the triplestore, so I guess we have two options:

1. Change the semantics of that route so that if you don't provide a comment, the existing comment is removed.
2. Add a separate route for deleting the comment.

Please open a separate YouTrack issue for this indicating which you prefer.

mdelez:
separate Youtrack issue has been made https://dasch.myjetbrains.com/youtrack/issue/DSP-1397

SepidehAlassi (approved):
This is great, thank you!

PR #1827, opened 2021-03-03
fix(OntologiesRouteV2): Reject internal ontology names in external schema (DSP-1394)

benjamingeer:
resolves DSP-1394


webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -3236,4 +3236,16 @@ class StringFormatter private (val maybeSettings: Option[KnoraSettingsImpl] = No
...
+
+  /**
+    * Throws [[BadRequestException]] if a Knora API v2 definition API has an ontology name that can only be used
+    * in the internal schema.
+    *
+    * @param iri the IRI to be checked.
+    */
+  def checkExternalOntologyName(iri: SmartIri): Unit = {
+    if (iri.isKnoraApiV2DefinitionIri && OntologyConstants.InternalOntologyLabels.contains(iri.getOntologyName)) {
+      throw BadRequestException(s"Invalid IRI: <$iri>")
```
SepidehAlassi, line 3248: don't you think it might be better to throw a more clear exception message indicating that internal ontologies cannot be served?

benjamingeer: Done: 05a7ba1

SepidehAlassi (approved):
Perfect, thanks!

PR #1833, opened 2021-03-16
fix(ontology): Escape Special Characters in Ontology Comment (DSP-1406)

SepidehAlassi:
resolves DSP-1406


benjamingeer (approved):
Thanks, and sorry about the bug. :)

SepidehAlassi:
> Thanks, and sorry about the bug. :)
No worries, thanks for the quick review!

PR #1835, opened 2021-03-23
fix(gravsearch): Keep rdf:type knora-api:Resource when needed (DSP-1407)

benjamingeer:
resolves DSP-1407

- [x] Don't assume that `rdf:type knora-api:Resource` is always a Gravsearch type annotation and remove it.
- [x] Add an optimisation that removes `rdf:type knora-api:Resource` if there's another statement at the same level of the `WHERE` clause with the same subject and predicate and a different type IRI.


SepidehAlassi (approved):
This is great, thanks.

benjamingeer:
Thanks for the review!

PR #1844, opened 2021-04-14
feat(api-v2): Return events describing version history of resources and values of a project ordered by data (DSP-1528)

SepidehAlassi:
resolves DSP-1528


webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/v2/getAllResourcesInProjectPrequeryStandard.scala.txt
```diff
@@ -0,0 +1,46 @@
...
+
+SELECT DISTINCT ?resource
+WHERE {
+    ?resource knora-base:attachedToProject <@projectIri> .
+    ?resourceType rdfs:subClassOf* knora-base:Resource .
+    ?resource rdf:type ?resourceType .
+    ?resource knora-base:creationDate ?creationDate.
+
+    FILTER NOT EXISTS {
+        ?resource knora-base:isDeleted true .
```
benjamingeer, line 43: Suppose this happens:

1. Resource A is created.
2. Resource B links to resource A.
3. Resource A is marked as deleted.

Now `getProjectResourcesWithHistoryV2` is run. It will not find resource A, because it is marked as deleted. So it will not generate an event for (1), but it will generate an event for (2). But then won't the event for (2) fail because resource A doesn't exist?

Unless I've misunderstood, I think that the events will have to create resource A and then mark it as deleted.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -2246,4 +2254,313 @@ class ResourcesResponderV2(responderData: ResponderData) extends ResponderWithSt
...
+    * @param featureFactoryConfig       the feature factory configuration.
+    * @param requestingUser             the user making the request.
+    * @return the full representation of the resource at the given version date.
+    */
+  private def getResourceAtGivenTime(resourceIri: IRI,
+                                     versionHist: ResourceHistoryEntry,
+                                     featureFactoryConfig: FeatureFactoryConfig,
+                                     requestingUser: UserADM): Future[(ResourceHistoryEntry, ReadResourceV2)] =
+    for {
+      resourceFullRepAtCreationTime: ReadResourcesSequenceV2 <- (responderManager ? ResourcesGetRequestV2(
```
benjamingeer, line 2361: Instead of the responder sending itself a message, you could just call `getResourcesV2` directly.

benjamingeer (changes requested):
How is the performance if you get the full history of a project with a lot of data?

webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/v2/getAllResourcesInProjectPrequeryStandard.scala.txt
SepidehAlassi: Ah sorry, you are right, I should remove this `Filter Not Exists` block.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
SepidehAlassi: yes, I am not consistent in that, sometimes I send a message, sometimes I just call the function. It was done for sake of the debugger tool, I will refactor this.

SepidehAlassi:
> How is the performance if you get the full history of a project with a lot of data?
For `anything` project that has only 52 resources, it returns 88 events in 50 seconds. Probably if a project had thousands of resources, it will take pretty long and crash as soon as the time threshold is reached. The task in DSP-1528 was to write a route that gets project IRIs and returns the entire history of resources as events, and this PR does that. The performance issue would be something that definitely will come up.

I have already thought about this issue though.  I have already written the responder, `getResourceHistoryEvents`, that gets the IRI of a resource and returns history events for it. I could change it a bit and add a route for it, i.e. `GET` request that returns the events for a given resource IRI.  There is also this template to get all resources of a project. I could also add a route, a `GET` request that returns a list of resource IRIs for a given project IRI. Then the DSP tools could first use the latter route to get the resource IRIs and then could iterate over them and call the first route to get the events of that each resource and store them somewhere.

benjamingeer (approved):
This looks good now, thanks very much!

PR #1856, opened 2021-05-11
fix(gravsearch): don't move the patterns with resource IRI after topological sorting (DSP-1620)

SepidehAlassi:
resolves DSP-1620

For optimization of Gravsearch queries, we define a graph out of a given query for topological sorting. A resource IRI given as the subject or object of a query pattern would be a node of the graph. Therefore the topological sorting algorithm would already order these patterns properly with respect to other patterns in the query. Extra use of `moveResourceIrisToBeginning` would mess with this order, therefore now that we have topological sorting of patterns in place, this method is obsolete and has to be removed.


benjamingeer (approved):
Thanks for taking care of this!

SepidehAlassi:
@benjamingeer thanks for reviewing this.

PR #1859, opened 2021-05-17
feat(ontology): Allow adding new property to a resource class in use (DSP-1629)

SepidehAlassi:
resolves DSP-1629


benjamingeer (approved):
This is great, thank you!

PR #1864, opened 2021-05-27
fix(sipi): create tmp folder

subotic:
resolves DSP-1679

During upload there is a check if the `tmp` folder exists. If not, then the script should not throw an error but create the `tmp` folder.


benjamingeer:
Tests failing: Sipi is returning Internal server error.

subotic:
> Tests failing: Sipi is returning Internal server error.

yes, unfortunately. Tests are only failing on Github-CI. Locally they are running through.

benjamingeer:
> Tests are only failing on Github-CI. Locally they are running through.

I don't understand why that's OK, but I guess you know what you're doing. 🙂

benjamingeer (approved):

subotic:
> > Tests are only failing on Github-CI. Locally they are running through.
> 
> I don't understand why that's OK, but I guess you know what you're doing. 🙂

It is not ;-) I'm not going to merge before the tests run through. At the moment, I have no idea why those tests fail. The change introduced in this PR is not what makes the tests fail. At least, I don't see how it can.

subotic:
> > > Tests are only failing on Github-CI. Locally they are running through.
> > 
> > 
> > I don't understand why that's OK, but I guess you know what you're doing. 🙂
> 
> It is not ;-) I'm not going to merge before the tests run through. At the moment, I have no idea why those tests fail. The change introduced in this PR is not what makes the tests fail. At least, I don't see how it can.

Found the problem. Lua is a very obscure language. My change was the problem. I have reverted it and added some comments for the next poor soul who has to deal with Lua.

PR #1875, opened 2021-06-09
feat(projects): Exclude shortcode from project IRI (DSP-1608)

SepidehAlassi:
resolves DSP-1608
resolves DSP-1555


SepidehAlassi:
This PR is divided into a few other ones.

PR #1889, opened 2021-07-14
fix(api-v2, api-admin): ontology name and project name should be URL safe (DSP-1749)

SepidehAlassi:
resolves DSP-1749


kilchenmann (approved):
LGTM

PR #1893, opened 2021-07-28
feat(resource-metadata): return resource metadata after metadata update request (DSP-1828)

irinaschubert:
resolves DSP-1828

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
@@ -901,14 +901,14 @@ object UpdateResourceMetadataRequestV2 extends KnoraJsonLDRequestReaderV2[Update
   *
   * @param resourceIri               the IRI of the resource.
   * @param resourceClassIri          the IRI of the resource class.
-  * @param maybeLastModificationDate the resource's last modification date, if any.
+  * @param lastModificationDate      the resource's last modification date.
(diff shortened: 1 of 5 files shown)


webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
```diff
@@ -897,6 +896,89 @@ object UpdateResourceMetadataRequestV2 extends KnoraJsonLDRequestReaderV2[Update
   }
 }
 
+/**
+  * Represents a response after updating a resource's metadata.
+  *
+  * @param resourceIri               the IRI of the resource.
+  * @param resourceClassIri          the IRI of the resource class.
+  * @param maybeLastModificationDate the resource's last modification date, if any.
```
kilchenmann, line 904: I think `lastModificationDate` should always be returned because when we edit the resource's metadata it will always have a `lastModificationDate`. In the GUI it's important to have this information. Otherwise it's not possible to update one field after the other.

irinaschubert: I agree, I changed it: the value `newModificationDate` is now taken instead of `maybeNewModificationDate`. `newModificationDate` is an `Instant` and not an `Option` and is never empty. If no `maybeNewModificationDate` was given with the request, `Instant.now` is taken as `newModificationDate`. I renamed our parameter accordingly. Let's see if the tests pass through.

kilchenmann (approved):
works well

subotic (approved):
LGTM, thanks!

PR #1898, opened 2021-08-16
fix(ontology V2): use internal iri when updating a property (DSP-1868)

irinaschubert:
resolves DSP-1868


irinaschubert:
Hi @subotic , in the end, we didn't manage to write a test to properly check the message creation but we testet the correct update behavior with the dsp-app (GUI) and the local Fuseki instance.

subotic (approved):
cool, great work. LGTM!

PR #1903, opened 2021-09-08
refactor(ontology)!: use patch instead of delete for deleting cardinalities (DSP-1700)

irinaschubert:
resolves DSP-1700

### DIFF
--- a/webapi/src/test/scala/org/knora/webapi/e2e/CORSSupportE2ESpec.scala
+++ b/webapi/src/test/scala/org/knora/webapi/e2e/CORSSupportE2ESpec.scala
@@ -36,8 +36,8 @@ object CORSSupportE2ESpec {
 }
 
 /**
-  * End-to-end test specification for testing [[CORSSupport]].
-  */
+ * End-to-end test specification for testing [[CORSSupport]].
+ */
 class CORSSupportE2ESpec extends E2ESpec(CORSSupportE2ESpec.config) {
 
   implicit def default(implicit system: ActorSystem) = RouteTestTimeout(settings.defaultTimeout)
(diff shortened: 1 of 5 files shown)


irinaschubert:
Hi @subotic , I used PATCH instead of DELETE for the deletion of cardinalities (PUT was already in use). It seems that PATCh wasn't used before so I had to enable it and also adjust the tests for these methods. I assigned the PR to you just that you can see what I did. You don't need to review it as you are on holidays.

kilchenmann (approved):
LGTM. Thank you

webapi/src/test/scala/org/knora/webapi/e2e/CORSSupportE2ESpec.scala
```diff
@@ -51,30 +51,31 @@ class CORSSupportE2ESpec extends E2ESpec(CORSSupportE2ESpec.config) {
...
-      responseToString(response) shouldEqual "CORS: invalid method 'PATCH'"
-      response.status shouldBe StatusCodes.BadRequest
-    }
+//    "reject requests with invalid method" in {
+//      val request = Options(baseApiUrl + s"/admin/projects") ~> Origin(exampleOrigin) ~> `Access-Control-Request-Method`(
+//        PATCH)
+//      val response: HttpResponse = singleAwaitingRequest(request)
+//      responseToString(response) shouldEqual "CORS: invalid method 'PATCH'"
+//      response.status shouldBe StatusCodes.BadRequest
+//    }
```
BalduinLandolt, line 72: can this be removed?

BalduinLandolt (approved):
looks good to me

webapi/src/test/scala/org/knora/webapi/e2e/CORSSupportE2ESpec.scala
irinaschubert: You are right. It's probably confusing in the future to have uncommented code blocks. I removed it.

PR #1904, opened 2021-09-08
docs: add username to changeable attributes (DSP-1895)

irinaschubert:
resolves DSP-1895

### DIFF
--- a/docs/03-apis/api-admin/users.md
+++ b/docs/03-apis/api-admin/users.md
@@ -111,7 +111,7 @@ specified by the `id` in the request body as below:
 ### Update basic user information**
 
   - Required permission: SystemAdmin / self
-  - Changeable information: email, given name, family name,
+  - Changeable information: username, email, given name, family name,
     password, status, SystemAdmin membership 
   - TypeScript Docs: userFormats - ChangeUserApiRequestADM
   - PUT: `/admin/users/iri/<userIri>/BasicUserInformation`
--- a/webapi/src/main/scala/org/knora/webapi/messages/admin/responder/usersmessages/UsersMessagesADM.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/admin/responder/usersmessages/UsersMessagesADM.scala
@@ -39,27 +39,29 @@ import spray.json._
 // API requests
 
 /**
-  * Represents an API request payload that asks the Knora API server to create a new user.
-  *
-  * @param id          the optional IRI of the user to be created (unique).
-  * @param username    the username of the user to be created (unique).
-  * @param email       the email of the user to be created (unique).
-  * @param givenName   the given name of the user to be created.
-  * @param familyName  the family name of the user to be created
-  * @param password    the password of the user to be created.
-  * @param status      the status of the user to be created (active = true, inactive = false).
-  * @param lang        the default language of the user to be created.
-  * @param systemAdmin the system admin membership.
-  */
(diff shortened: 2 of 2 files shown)


docs/03-apis/api-admin/users.md
```diff
@@ -111,7 +111,7 @@ specified by the `id` in the request body as below:
 ### Update basic user information**
 
   - Required permission: SystemAdmin / self
-  - Changeable information: email, given name, family name,
+  - Changeable information: username, email, given name, family name,
```
irinaschubert, line 114: This is the only change I've done in this file. The rest is reformatting.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/usersmessages/UsersMessagesADM.scala
```diff
@@ -77,29 +79,31 @@ case class CreateUserApiRequestADM(id: Option[IRI] = None,
...
-                                   email: Option[String] = None,
-                                   givenName: Option[String] = None,
-                                   familyName: Option[String] = None,
-                                   lang: Option[String] = None,
-                                   requesterPassword: Option[String] = None,
-                                   newPassword: Option[String] = None,
-                                   status: Option[Boolean] = None,
-                                   systemAdmin: Option[Boolean] = None) {
+ * Represents an API request payload that asks the Knora API server to update an existing user. Information that can
+ * be changed include the user's username, email, given name, family name, language, password, user status, and system admin
```
irinaschubert, line 83: This is the only change I've done in this file. The rest is reformatting.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/usersmessages/UsersMessagesADM.scala
```diff
@@ -624,28 +645,26 @@ case class UserADM(id: IRI,
...
-    */
+   * Given an identifier, returns true if it is the same user, and false if not.
+   */
   def isSelf(identifier: UserIdentifierADM): Boolean = {
 
-    val iriEquals = identifier.toIriOption.contains(id)
-    val emailEquals = identifier.toEmailOption.contains(email)
+    val iriEquals      = identifier.toIriOption.contains(id)
+    val emailEquals    = identifier.toEmailOption.contains(email)
     val usernameEquals = identifier.toUsernameOption.contains(username)
```
mpro7, line 656: Although most of re-formatting done in the file makes sense, the one marked above I don't personally like and it's kind of the same "alignment idea" to arguments formatted here before. I opt to keep fields as it was before :)

irinaschubert: OK, I suggest that we do the reformatting in a separate PR and then reformat everything (we already tried that but then the tests didn't run through - so we need to be careful with this). Besides, I don't know how to re-do this now... (as it was done automatically by saving the file)

mpro7: It's matter of the formatter configuration. It should be rather easy to revert unwanted things.

BalduinLandolt (approved):

mpro7 (approved):
As we agreed, we need to agree on the formatting rules etc. until then we should avoid reformat the code, but in this single file we can have it as a reference, which will be adjusted soon.

PR #1908, opened 2021-09-23
refactor: reformatting Scala files (DSP-1897)

mpro7:
resolves DSP-1897


BalduinLandolt (approved):
I can't reasonably review this PR, but it looks ok as far as I can tell... and if the tests pass, then nothing is broken

irinaschubert (approved):

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/usersmessages/UsersMessagesADM.scala
```diff
@@ -618,8 +618,8 @@ case class UserADM(
    */
   def isSelf(identifier: UserIdentifierADM): Boolean = {
 
-    val iriEquals      = identifier.toIriOption.contains(id)
```
subotic, line 621: @mpro7 Any clue why the equal signs after the variables are not indented anymore? In my opinion, this is one of the most important formatting settings.

subotic: it needs to be at least more

PR #1910, opened 2021-09-28
fix(candeletecardinalities): return correct response on route negative case (DEV-36)

mpro7:
resolves DEV-36

### DIFF
--- a/webapi/src/test/scala/org/knora/webapi/e2e/v2/OntologyV2R2RSpec.scala
+++ b/webapi/src/test/scala/org/knora/webapi/e2e/v2/OntologyV2R2RSpec.scala
@@ -2864,6 +2864,56 @@ class OntologyV2R2RSpec extends R2RSpec {
       assert(resourceIri.toSmartIri.isKnoraDataIri)
     }
 
+    // payload
+    val cardinalityCantBeDeletedPayload =
+      s"""
+         |{
+         |  "@id" : "${SharedOntologyTestDataADM.FREETEST_ONTOLOGY_IRI_LocalHost}",
+         |  "@type" : "owl:Ontology",
+         |  "knora-api:lastModificationDate" : {
+         |    "@type" : "xsd:dateTimeStamp",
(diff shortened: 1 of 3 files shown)


irinaschubert (approved):

webapi/src/test/scala/org/knora/webapi/e2e/v2/OntologyV2R2RSpec.scala
```diff
@@ -78,7 +78,20 @@ class OntologyV2R2RSpec extends R2RSpec {
   private val clientTestDataPath: Seq[String] = Seq("v2", "ontologies")
 
   // Collects client test data
-  private val clientTestDataCollector = new ClientTestDataCollector(settings)
+  // TODO: redefine below method somewhere else if can be reused over other test files
+  private def CollectClientTestData(filename: String, payload: String) = {
```
irinaschubert:
```suggestion
  private def collectClientTestData(filename: String, filecontent: String) = {
```

irinaschubert (commented):
I like the removal of duplicated code. Did you test if it is working? Because there is now a new `clientTestDataCollector` created in each call of `CollectClientTestData(...)`. Before, it was one `clientTestDataCollector` for all the tests in this file. Do we still get all test data in the end?

webapi/src/test/scala/org/knora/webapi/e2e/v2/OntologyV2R2RSpec.scala
mpro7: @irinaschubert  at the end it's both payload and the file content ;) but in the matter of consistency I accept this suggestion.

`val clientTestDataCollector = new ClientTestDataCollector(settings)` can be also defined outside the method, maybe this is better performance wise. Everything was working well, but I think that some tests failed after that commit 2d9c26d where we added 3 missing test data.

subotic (approved):
LGTM, thanks!

PR #1921, opened 2021-10-13
fix(groups): update test data and documentation to use language specific group descriptions (DEV-123)

irinaschubert:
resolves DEV-123


docs/03-apis/api-admin/groups.md
```diff
@@ -51,7 +51,10 @@ License along with DSP. If not, see <http://www.gnu.org/licenses/>.
 ```json
     {
       "name": "NewGroup",
-      "description": "NewGroupDescription",
+      "descriptions": [
+        {"value": "NewGroupDescription", "language": "en"},
+        {"value": "NeueGruppenBeschreibung", "language": "de"}
+      ],
```
kilchenmann, line 57: Just a comment: In fact this is a breaking change. But groups endpoint is not used yet (at least in dsp-app) but already implemented in js-lib.

PR #1932, opened 2021-11-03
refactor(lists+groups): improve value objects + remove new lists implementation (DEV-160)

mpro7:
resolves DEV-160 + removes new lists route implementation. Some functionalities were moved to old one, the documentation was updated too.


webapi/scripts/expected-client-test-data.txt
```diff
@@ -29,38 +29,11 @@ test-data/admin/lists/delete-list-response.json
 test-data/admin/lists/get-anything-project-lists-response.json
 test-data/admin/lists/get-image-project-lists-response.json
 test-data/admin/lists/get-list-info-response.json
-test-data/admin/lists/get-list-node-info-response.json
```
kilchenmann, line 32: This test-data-file is now missing in js-lib. What was the reason to delete the file?

kilchenmann: And what will be the equivalent test-data for all the deleted "feature_toggle" lists?

mpro7: Test data are generated in relation to E2E tests on DSP-API side, if the test was deleted, the test data file needs to be deleted too. In this case, the test was redundant. In the case of lists - whole new implementation is gone, so the related tests and test-data are.

subotic (approved):

PR #1949, opened 2021-11-24
fix(sipi): remove costly cleanup at each upload

subotic:
resolves DEV-237

### DIFF
--- a/sipi/scripts/upload.lua
+++ b/sipi/scripts/upload.lua
@@ -228,7 +228,7 @@ for file_index, file_params in pairs(server.uploads) do
 end
 
 -- Clean up old temporary files.
-clean_temp_dir()
+-- clean_temp_dir()
 -- Return the file upload data in the response.
 local response = {}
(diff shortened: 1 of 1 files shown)


subotic:
@irinaschubert I've removed the temp folder cleanup, which is triggered at each uploaded file. This should improve a bit the upload performance. I still see that the Sipi's memory consumption is getting bigger and bigger with each upload, so there is a memory leak. I will try to track it down. Maybe @jnussbaum can try this branch as it is now, and see if it helps a bit.

jnussbaum:
@subotic @irinaschubert : 
 - I tried it out, and it basically works. 
 - Without taking measurements, cannot tell if it's faster than before.
 - It still interrupts after some time, but this time with a slightly different error message (see my edit in DEV-237 / https://github.com/dasch-swiss/0820-lhtt-scripts/blob/main/Varia/2021-11-12%20SIPI%20crash%20report.md)

subotic:
deprecated by #1959

PR #1958, opened 2021-12-06
fix(ontology): cardinality of one can be added to classes as long as not used in data

irinaschubert:
resolves DEV-233

It is now possible to add a cardinality to a class even if it has a subclass. This is only allowed if that class or any sub-class (recursively) is not yet used in data, i.e. has no resource instantiated from it. Therefore I added a new check `isClassUsedInData`.


webapi/src/main/scala/org/knora/webapi/responders/Responder.scala
```diff
@@ -153,6 +178,24 @@ abstract class Responder(responderData: ResponderData) extends LazyLogging {
...
+    classIri: SmartIri,
+    errorFun: => Nothing
+  ): Future[Unit] =
+    for {
+      classIsUsed: Boolean <- isClassUsedInData(classIri)
+
+      _ = if (classIsUsed) {
+        errorFun
+      }
+    } yield ()
```
BalduinLandolt, line 197: would it be worth considering using one of scala's ways of avoiding exception throwing (like `Either`) here?

BalduinLandolt (approved):
looks good

webapi/src/main/scala/org/knora/webapi/responders/Responder.scala
```diff
@@ -131,6 +131,31 @@ abstract class Responder(responderData: ResponderData) extends LazyLogging {
 
     } yield isEntityUsedResponse.results.bindings.nonEmpty
 
+  /**
+   * Checks whether an instance of a class exists
+   *
+   * @param classIri                 the IRI of the class.
+   *
+   * @return `true` if the class is used.
+   */
```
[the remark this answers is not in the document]
irinaschubert: No reason, just copy-pasted it from another one, will fix it. Thanks for the tip about creating stubs, I will check it out.

mpro7 (approved):

webapi/src/main/scala/org/knora/webapi/responders/Responder.scala
irinaschubert: As discussed, I won't change it here because it conforms to the overall pattern of V2.

subotic (approved):

PR #1966, opened 2021-12-14
fix(candeletecardinalities): return canDoResponse of false instead of throwing an exception (DEV-314)

irinaschubert:
resolves DEV-314

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cardinalities.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cardinalities.scala
@@ -52,7 +52,6 @@ object Cardinalities {
   )(implicit ec: ExecutionContext, stringFormatter: StringFormatter, timeout: Timeout): Future[CanDoResponseV2] = {
     for {
       cacheData: Cache.OntologyCacheData <- Cache.getCacheData
-      ontology = cacheData.ontologies(internalOntologyIri)
 
       submittedClassDefinition: ClassInfoContentV2 =
         deleteCardinalitiesFromClassRequest.classInfoContent.toOntologySchema(InternalSchema)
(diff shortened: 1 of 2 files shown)


subotic (approved):

webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cardinalities.scala
```diff
@@ -250,9 +262,24 @@ object Cardinalities {
...
+              internalOntologyIri
+            )
+          } yield isDefined
+        }.toList)
+
+      _ = if (isDefinedOnClassList.contains(false)) {
+        throw BadRequestException(
+          "The cardinality is not defined directly on the class and cannot be deleted."
+        )
+      }
```
BalduinLandolt, line 282: this seems to be pretty much exactly the same as you do above in the `canDelete...` (except throwning an exception instead of returning a boolean), so it would probably be cleaner to extract it into a separate function

webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cardinalities.scala
```diff
@@ -521,30 +529,58 @@ object Cardinalities {
...
+          FastFuture.failed(
+            throw BadRequestException(
+              s"Submitted cardinality for property $propertyIri does not match existing cardinality."
+            )
+          )
+        }
+      case None =>
         throw BadRequestException(
-          s"Cardinality for property ${propertyIri} is not defined."
+          s"Submitted cardinality for property $propertyIri is not defined for class $internalClassIri."
```
BalduinLandolt, line 561: Maybe you could add a hint to the error message, that it's defined on the superclass?

BalduinLandolt (approved):
Looks good. Just to minor remarks that you can feel free to implement or ignore.  
Plus maybe reconsider the PR title scope to be more on the module level.

webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cardinalities.scala
irinaschubert: At that point, there could be other reasons why it is not defined (p.ex. just forgotten or so).

webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cardinalities.scala
irinaschubert: I agree that is quite a lot of code copied but I actually don't see right now how to simplify it. `isCardinalityDefinedOnClass` has to be called on the Map`cardinalitiesToDelete`. I don't think it helps much to have a separate function that just "loops" through (maps) the list and returns a Future of a list (then, of course, it would make more sense to implement a higher-order function for this kind of action but I'm not sure this makes sense now before the greater refactoring we do anyway). So I leave it for now.

PR #1973, opened 2022-01-05
fix(search): Return matching sub-nodes when searching for list label (DEV-158)

irinaschubert:
resolves DEV-158


webapi/src/test/scala/org/knora/webapi/responders/v2/SearchResponderV2SpecFullData.scala
```diff
@@ -2042,38 +2077,58 @@ class SearchResponderV2SpecFullData(implicit stringFormatter: StringFormatter) {
 
   val fulltextSearchForListNodeLabel: ReadResourcesSequenceV2 = ReadResourcesSequenceV2(
```
BalduinLandolt, line 2078: just a thought on variable naming: this is quite specific test data, so I wonder if that should be reflected in the variable names?  
it is a) specifically books in this example, and b) expected search results. So maybe something along the lines of `expectedFulltextSearchForBookListNodeLabel` would be more explicit? but maybe that's also too much.

Another organizing option (which then should be done in a separate refactoring PR) might be to have separate subclasses like `FulltextSearch` etc. to separate out the different scopes of the data held in this (quite big) file.

BalduinLandolt (approved):
looks good

mpro7 (approved):

webapi/src/test/scala/org/knora/webapi/responders/v2/SearchResponderV2SpecFullData.scala
irinaschubert: I see your point and adjusted the name of the variables. I think when looking at the test it's now easy to understand what it is.
Regarding your second point: There are only 7 values defined in this file, so maybe it even makes sense to split the file into several files: one for each test. I think about it!

BalduinLandolt: sure, whatever seems reasonable to you!  
just keep in mind that having a diverging structure from all other tests might not be ideal either. But we can always use it as an example for a later refactoring.

subotic (approved):

PR #1976, opened 2022-01-06
refactor(admin): refactor projects & users value objects (DEV-240)

mpro7:
resolves DEV-240


webapi/src/test/scala/org/knora/webapi/messages/admin/responder/valueObjects/ProjectsValueObjectsADMSpec.scala
```diff
@@ -0,0 +1,205 @@
...
+import org.knora.webapi.messages.admin.responder.projectsmessages.ProjectsErrorMessagesADM._
+import org.knora.webapi.messages.store.triplestoremessages.StringLiteralV2
+import zio.prelude.Validation
+
+/**
+ * This spec is used to test the [[ProjectsValueObjectsADM]] value objects creation.
+ */
+class ProjectsValueObjectsADMSpec extends UnitSpec(ValueObjectsADMSpec.config) {
+  "ProjectIRI value object" when {
+//    TODO: check string formatter project iri validation because passing just "http://rdfh.ch/projects/@@@@@@" works
```
irinaschubert, line 19: Do you want to resolve this in this PR? If not, I would delete it and create a JIRA issue instead (if you haven't already). Otherwise I would resolve it in this PR and then get rid of the TODO.
In any case, I think this check is important and should be done.

webapi/src/test/scala/org/knora/webapi/messages/admin/responder/valueObjects/ProjectsValueObjectsADMSpec.scala
```diff
@@ -0,0 +1,205 @@
...
+          Validation.fail(BadRequestException(PROJECT_IRI_INVALID_ERROR))
+        )
+      }
+    }
+    "created using valid value" should {
+      "not throw BadRequestExceptions" in {
+        ProjectIRI.make(validProjectIri) should not equal Validation.fail(
+          BadRequestException(PROJECT_IRI_INVALID_ERROR)
+        )
+      }
```
irinaschubert, line 39: Does it need this test? If the next test ("return value passed to value object") passes, it's enough I think.

irinaschubert (commented):
I added some suggestions, nothing very important (and I've always only added it the first time I saw something, so you need to check for other occurrences if you agree with the suggestion).
Something that is not clear to me: for value objects, why do we need 2 `make()` methods again?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/valueObjects/ProjectsValueObjectsADM.scala
```diff
@@ -0,0 +1,189 @@
...
+      }
+    }
+
+  def make(value: Option[String]): Validation[Throwable, Option[ProjectIRI]] =
+    value match {
+      case Some(v) => self.make(v).map(Some(_))
+      case None    => Validation.succeed(None)
+    }
+}
+
```
BalduinLandolt, line 37: To me, this doesn't fully make sense (as I messaged you separately).

I think `make(None)` should only be a success, if the property is optional.  
Also, the reason for the second `make()` method is to be able to both handle `String` and `Option[String]`elegantely, which makes sense to me. But wouldn't this be a good place to get rid of the option? If it's `Some(v)`, you just create it for `v` (and drop the Option); and if it's None, it should imo fail anyways.

But maybe I just don't understand the reasoning behind this.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/valueObjects/ProjectsValueObjectsADM.scala
```diff
@@ -0,0 +1,189 @@
...
+      val validatedValue: Validation[Throwable, String] = Validation(
+        stringFormatter.validateProjectShortcode(value, throw BadRequestException(SHORTCODE_INVALID_ERROR))
+      )
+      validatedValue.map(new Shortcode(_) {})
+    }
+
+  def make(value: Option[String]): Validation[Throwable, Option[Shortcode]] =
+    value match {
+      case Some(v) => self.make(v).map(Some(_))
+      case None    => Validation.succeed(None)
```
BalduinLandolt, line 58: same as above in all those cases...

BalduinLandolt (approved):
Looks good over all. I added some minor things. And I don't get the `validation.success(None)` part.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/valueObjects/ProjectsValueObjectsADM.scala
irinaschubert: I see your point regarding the second `make()`, but I thought the creation of value objects is at the outermost edge of the API where the actual interaction with the client takes place. And there I thought it's always Strings and Booleans, never Options. But I could be wrong of course.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/valueObjects/ProjectsValueObjectsADM.scala
```diff
@@ -0,0 +1,189 @@
...
+        )
+
+        validatedValue.map(new ProjectIRI(_) {})
+      }
+    }
+
+  def make(value: Option[String]): Validation[Throwable, Option[ProjectIRI]] =
+    value match {
+      case Some(v) => self.make(v).map(Some(_))
+      case None    => Validation.succeed(None)
```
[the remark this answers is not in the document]
mpro7: As discussed yesterday, for optional fields None is valid.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/valueObjects/ProjectsValueObjectsADM.scala
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * ProjectStatus value object.
+ */
+sealed abstract case class ProjectStatus private (value: Boolean)
+object ProjectStatus { self =>
+  def make(value: Boolean): Validation[Throwable, ProjectStatus] =
+    Validation.succeed(new ProjectStatus(value) {})
+
+  def make(value: Option[Boolean]): Validation[Throwable, Option[ProjectStatus]] =
```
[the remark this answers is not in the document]
mpro7: As discussed yesterday, this is handy for both optional and non-optional fields. Instead of different name of the other method I just added missing types annotations to make it more visible which method is used for certain field.

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -880,7 +878,6 @@ class ProjectsResponderADM(responderData: ResponderData) extends Responder(respo
         throw NotFoundException(s"Project '$projectIri' not found. Aborting update request.")
       }
       // we are changing the project, so lets get rid of the cached copy
-      // invalidateCachedProjectADM isn't clearing cache as expected
```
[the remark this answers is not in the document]
mpro7: That method didn't clear the cache properly - that was done in other PR - just forgot to remove unused method.

webapi/src/test/scala/org/knora/webapi/messages/admin/responder/valueObjects/ProjectsValueObjectsADMSpec.scala
mpro7: It's possible bug => new bug :)

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/usersmessages/UsersPayloadsADM.scala
```diff
@@ -0,0 +1,18 @@
...
+ * User creation payload
+ */
+final case class UserCreatePayloadADM(
+  id: Option[UserIRI] = None,
+  username: Username,
+  email: Email,
+  givenName: GivenName,
+  familyName: FamilyName,
+  password: Password,
+  status: UserStatus,
```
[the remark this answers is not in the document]
mpro7: This gonna trigger breaking change and all the chain of tasks to complete in already added projects, therefore thank you for suggestion, but not :)

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/valueObjects/ProjectsValueObjectsADM.scala
mpro7: Explained [here](https://github.com/dasch-swiss/dsp-api/pull/1976#discussion_r784655179) and [here](https://github.com/dasch-swiss/dsp-api/pull/1976#discussion_r784657235).

subotic (approved):

PR #1978, opened 2022-01-12
fix(ontology): DSP-API creates wrong partOfValue property (DEV-216)

irinaschubert:
resolves DEV-216


kilchenmann (approved):
works as expected 👍 Thank you

webapi/src/test/scala/org/knora/webapi/responders/v2/OntologyResponderV2Spec.scala
```diff
@@ -2366,6 +2366,221 @@ class OntologyResponderV2Spec extends CoreSpec() with ImplicitSender {
...
+
+      val partOfValuePropGetRequest = PropertiesGetRequestV2(
+        propertyIris = Set(partOfValuePropertyIri),
+        allLanguages = true,
+        requestingUser = anythingAdminUser
+      )
+
+      responderManager ! partOfValuePropGetRequest
+
+      expectMsgPF(timeout) { case msg: ReadOntologyV2 =>
```
mpro7, line 2568: If possible I would use more descriptive name instead of `expectMsgPF`

mpro7 (approved):

webapi/src/test/scala/org/knora/webapi/responders/v2/OntologyResponderV2Spec.scala
irinaschubert: I don't even know what PF means, I just copy-pasted that ;) I will change it.

irinaschubert: Ah, I just see now that this is a method from Akka TestKit. So, I can't change it. It's what it is called.

subotic (approved):

PR #1981, opened 2022-01-14
chore(license): update the license (DEV-374)

mpro7:
resolves DEV-374


subotic:
@mpro7 we also need the start year, e.g., @ 2021 - 2022

subotic (changes requested):
we also need the start year, e.g., @ 2021 - 2022

subotic (approved):

PR #1990, opened 2022-01-31
feat(UUID): add IRI validation that allows only to create IRIs using UUID version 4 and 5 (DEV-402)

mpro7:
resolves DEV-402


webapi/src/main/scala/org/knora/webapi/messages/admin/responder/permissionsmessages/PermissionsMessagesADM.scala
```diff
@@ -23,6 +23,23 @@ import java.util.UUID
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 // API requests
 
+/**
+ * Validates permission IRI
+ * @param iri to be validated.
+ */
+private case class validatePermissionIRI(iri: IRI) {
```
BalduinLandolt, line 30: class names should be upper case

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/valueObjects/GroupsValueObjectsADM.scala
```diff
@@ -22,8 +23,12 @@ object GroupIRI { self =>
     if (value.isEmpty) {
       Validation.fail(BadRequestException(GROUP_IRI_MISSING_ERROR))
     } else {
+      val isUUID: Boolean = sf.couldBeUuid(value.split("/").last)
```
BalduinLandolt, line 26: I find the function name `couldBeUuid()` confusing. Either it is or it's not, but *could* seems weird in this context

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
```diff
@@ -35,6 +36,14 @@ import org.knora.webapi.util._
 
 import scala.concurrent.{ExecutionContext, Future}
 
+private case class validateUUIDOfResourceIRI(iri: IRI) {
```
BalduinLandolt, line 39: class name should be upper case

BalduinLandolt (approved):
looks good, apart from a few minor naming things

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/permissionsmessages/PermissionsMessagesADM.scala
mpro7: Obsolete, I did some refactoring just before a while.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/resourcemessages/ResourceMessagesV2.scala
mpro7: Obsolete, I did some refactoring just before a while.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/valueObjects/GroupsValueObjectsADM.scala
mpro7: You're partially right, because it only checks the length of string, so there is possibility checked string isn't the UUID, but I think to rename it.

subotic (approved):
LGTM, thanks

PR #1995, opened 2022-02-04
chore(deps): upgrade Jena to v4.3.2 (DEV-473)

mpro7:
resolves DEV-473

### DIFF
--- a/third_party/dependencies.bzl
+++ b/third_party/dependencies.bzl
@@ -39,14 +39,7 @@ def dependencies():
 
             # Jena
             "org.apache.jena:apache-jena-libs:%s" % (JENA_VERSION),
-            maven.artifact(
-                group = "org.apache.jena",
-                artifact = "jena-text",
-                version = JENA_VERSION,
-                exclusions = [
-                    "org.slf4j:slf4j-log4j12",
-                ],
-            ),
+            "org.apache.jena:jena-text:%s" % (JENA_VERSION),
 
             # Logging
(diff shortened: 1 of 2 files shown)


subotic:
@mpro7 could you also please remove:

```
maven.artifact(
                group = "org.apache.jena",
                artifact = "jena-text",
                version = JENA_VERSION,
                exclusions = [
                    "org.slf4j:slf4j-log4j12",
                ],
            ),
```

and

```
"org.apache.logging.log4j:log4j:2.17.0", # needed by apache-jena-libs. explicitly setting latest version (log4shell) to override the one used by jena.
```
from `dsp-api/third_party/dependencies.bzl`?

Both are not needed anymore, now that the Jena library is updated and pulls in the correct log4j version.

subotic (approved):

mpro7:
@subotic are you sure about your request? I tried to do that and at least `jena-text` is still used:
```
webapi/src/main/scala/org/knora/webapi/store/triplestore/embedded/JenaTDBActor.scala:14: error: object text is not a member of package org.apache.jena.query
import org.apache.jena.query.text._
```

subotic:
@mpro7 yes, you are right. you would need to add `jena-text` in the same way the other jena libraries are added. What this bit of code does is to exclude the old verson of `org.slf4j:slf4j-log4j12`.  Now we don't need to exclude it anymore but we still need to add the dependency to `jena-text`.

mpro7:
@subotic everything is clear now, thanks.

PR #1996, opened 2022-02-09
feat(ontologies): make comments optional for property and class creation (DEV-342)

mpro7:
resolves DEV-342


subotic (approved):
LGTM, thanks!

PR #2004, opened 2022-02-17
fix(permissions): Update default object access permissions (DEV-514)

irinaschubert:
resolves DEV-514

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
@@ -1037,10 +1037,7 @@ class ProjectsResponderADM(responderData: ResponderData) extends Responder(respo
             forGroup = Some(OntologyConstants.KnoraAdmin.ProjectAdmin),
             hasPermissions = Set(
               PermissionADM.changeRightsPermission(OntologyConstants.KnoraAdmin.ProjectAdmin),
-              PermissionADM.deletePermission(OntologyConstants.KnoraAdmin.ProjectAdmin),
-              PermissionADM.modifyPermission(OntologyConstants.KnoraAdmin.ProjectAdmin),
-              PermissionADM.viewPermission(OntologyConstants.KnoraAdmin.ProjectAdmin),
-              PermissionADM.restrictedViewPermission(OntologyConstants.KnoraAdmin.ProjectAdmin)
+              PermissionADM.modifyPermission(OntologyConstants.KnoraAdmin.ProjectMember)
             )
           ).prepareHasPermissions,
           featureFactoryConfig = featureFactoryConfig,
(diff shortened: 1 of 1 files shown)


subotic (approved):

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -1037,10 +1037,7 @@ class ProjectsResponderADM(responderData: ResponderData) extends Responder(respo
             forGroup = Some(OntologyConstants.KnoraAdmin.ProjectAdmin),
             hasPermissions = Set(
               PermissionADM.changeRightsPermission(OntologyConstants.KnoraAdmin.ProjectAdmin),
-              PermissionADM.deletePermission(OntologyConstants.KnoraAdmin.ProjectAdmin),
-              PermissionADM.modifyPermission(OntologyConstants.KnoraAdmin.ProjectAdmin),
-              PermissionADM.viewPermission(OntologyConstants.KnoraAdmin.ProjectAdmin),
-              PermissionADM.restrictedViewPermission(OntologyConstants.KnoraAdmin.ProjectAdmin)
+              PermissionADM.modifyPermission(OntologyConstants.KnoraAdmin.ProjectMember)
```
mpro7, line 1040: Shouldn't the most privileged permission contain all the others? This line then wouldn't be necessary for too.

docs/03-apis/api-admin/permissions.md
```diff
@@ -116,27 +121,33 @@ default object access permission for a group of a project the request body would
...
+    - `V`: view permission
+    - `M` modify permission
+    - `D`: delete permission
+    - `CR`: change rights permission (most privileged)
+  - `permissionCode`: The code assigned to a permission indicating its hierarchical level. These codes are as below:
+    - `1`: for restricted view permission (least privileged)
+    - `2`: for view permission
+    - `6`: for modify permission
+    - `7`: for delete permission
+    - `8`: for change rights permission (most privileged)
```
mpro7, line 150: Why do we need both ways to describe permissions?

mpro7 (approved):

docs/03-apis/api-admin/permissions.md
irinaschubert: I have no idea why both are used, I stumbled upon it as well but it is not the scope of this PR.

subotic: `V1` and `V2`. `V1` use numbers and `V2` uses letters.

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
subotic: they differ in the group

PR #2005, opened 2022-02-17
fix(timeout): Increase timeouts (DEV-536)

irinaschubert:
resolves DEV-536


subotic (approved):

mpro7 (approved):
IMO these changes should be reverted once the data are in the place.

BalduinLandolt (approved):
I still think it's an anti-pattern to have a config that needs to be changed in the repo and only takes effect after doing a release... but for now, that's what we have, so the PR is ok

PR #2009, opened 2022-02-25
refactor: start on a functional domain design implementation for ontologies (DEV-227)

subotic:
resolves DEV-227


project/plugins.sbt
```diff
@@ -20,4 +20,6 @@ addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.11.0")
 
 // addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.6.1")
 // addSbtPlugin("com.github.cb372" % "sbt-explicit-dependencies" % "0.2.16")
-// addDependencyTreePlugin
+
+// please don't remove or merge uncommented to main
```
irinaschubert, line 24: What do you mean by "don't merge uncommented to main"? Do you mean that the following line should never be commented out (as it was before)?

dsp-schema-repo/src/main/scala/dsp/schema/repo/SchemaRepoTest.scala
```diff
@@ -0,0 +1,24 @@
+package dsp.schema.repo
+
+import zio._
+import dsp.schema.domain.SchemaDomain.{UserID, UserProfile}
+
+case class SchemaRepoTest() extends SchemaRepo {
+  private var map: Map[UserID, UserProfile] = Map()
```
irinaschubert, line 7: We agreed on never using `var`. Is there a good reason to use it here instead of `val`? If so, what is the reason?

irinaschubert (commented):
Cool, I'm looking forward to working on this. 

Is there a reason why everything is equally placed on the top level of the folder structure? I would create one folder for all the new code and inside it one folder for each compartment (project, schema, resource, list,...). But perhaps I don't overlook all the consequences.

project/plugins.sbt
subotic: yes, and I broke the rule ;-) will need to fix that first. This pulls in a bunch of libraries, which can slow down the build.

dsp-schema-repo/src/main/scala/dsp/schema/repo/SchemaRepoTest.scala
subotic: to every rule there are exceptions :-) This is the simplest way of implementing a key-value store. Since it is only for testing, it should be fine. I "stole" the code from John workshop.

subotic:
> Cool, I'm looking forward to working on this.
> 
> Is there a reason why everything is equally placed on the top level of the folder structure? I would create one folder for all the new code and inside it one folder for each compartment (project, schema, resource, list,...). But perhaps I don't overlook all the consequences.

for reasons of visibility. the modules are grouped by name. The `webapi` folder should at some point be renamed to something more fitting. take a look at something other larger scala repos:
- akka: https://github.com/akka/akka
- zio: https://github.com/zio/zio
- zio-http: https://github.com/dream11/zio-http

irinaschubert (approved):

dsp-schema-core/src/main/scala/dsp/schema/domain/SchemaDomain.scala
```diff
@@ -0,0 +1,279 @@
...
+    Ontology
+      .empty(ontoInfo)
+      .withClass(classOne)
+      .withClass(classTwo)
+      .withProperty(propertyOne)
+      //.withProperty(propertyOne)
+      //.withProperty(propertyTwo)
+      .withCardinality(cardOne)
+      .withCardinality(cardOne)
+
```
BalduinLandolt, line 238: I'm guessing this should eventually be moved to a separate spec file

BalduinLandolt (approved):
I do not yet fully understand everything, but it looks exciting

PR #2046, opened 2022-04-21
feat(error-handling): return status 504 instead of 500 for triplestore timeout exception (DEV-749)

BalduinLandolt:
Resolves DEV-749

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

When a triplestore request times out, a generic `500 Internal Server Error` is returned.  
The app can therefor not apply any specific behaviour for this case.

Issue Number: DEV-749

## What is the new behavior?

A specific `504 Gateway Timeout` is returned to enable the App to handle this case specifically.

- [ ] Yes
- [x] No


kilchenmann (approved):
This is not a code review.
I tested it with the tweaked config that @BalduinLandolt mentioned in DEV-751 and I was able to produce a 504 timeout-error with a simple fulltext-search. So, it seems to work. Thanks

Hm, but a gravsearch query returns a 500 server error.

webapi/src/main/scala/org/knora/webapi/store/triplestore/http/HttpTriplestoreConnector.scala
```diff
@@ -303,11 +314,22 @@ class HttpTriplestoreConnector extends Actor with ActorLogging with Instrumentat
...
+              TriplestoreTimeoutException(
+                "Triplestore timed out while sending a response, after sending statuscode 200.",
+                e,
+                log
+              )
+            )
+          } else {
+            log.error(
+              e,
+              s"Couldn't parse response from triplestore:$logDelimiter$turtleStr${logDelimiter}in response to SPARQL query:$logDelimiter$sparql"
```
irinaschubert, line 329: Why does `logDelimiter` once is wrapped into `${...}` and once not? Can it be unified? I think both `$...` and `${...}` are the same.

irinaschubert (approved):
LGTM

webapi/src/main/scala/org/knora/webapi/store/triplestore/http/HttpTriplestoreConnector.scala
BalduinLandolt: you're right, that's just how it is in the place where I copy-pasted it from... ;) I'll unify it

BalduinLandolt: ah, no... this can't be changed because right after the variable there is a word, so it would consider this part of the variable name, which then could not be found... I'll have to leave it as it is, even though it looks strange

PR #2057, opened 2022-05-05
chore: check that the expected Fuseki version is present (DEV-331)

irinaschubert:
Resolves DEV-331

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe:

Issue Number: DEV-331

- [ ] Yes
- [x] No


BalduinLandolt (approved):
LGTM.  
In terms of code style: I think such long comments should not be on the same line as the code, but on the line before. But that's cosmetics

BalduinLandolt:
> > LGTM. In terms of code style: I think such long comments should not be on the same line as the code, but on the line before. But that's cosmetics
> 
> OK, I understand. I thought if it is on the same line, it is more obvious that it should be read. If it is above, one could easily not see it and forget to udpate the versions in the other places. What is your opinion on that?

Also makes sense. And if it has a functional purpose, I'm happy to live with a bit of ugly long lines :)

subotic (approved):

mpro7 (approved):
I would rather place comments in new line instead of making lines so long.

irinaschubert:
> I would rather place comments in new line instead of making lines so long.

I leave it on the same line. I'm afraid that we could forget to update the version in ops-deploy if we put the comment above or below. Hope that's OK for you, @mpro7 . We can still change it in the future should it bother us too much.

mpro7:
> > I would rather place comments in new line instead of making lines so long.
> 
> I leave it on the same line. I'm afraid that we could forget to update the version in ops-deploy if we put the comment above or below. Hope that's OK for you, @mpro7 . We can still change it in the future should it bother us too much.

IMO it works the other way around. Longer lines are usually less readable and easy to omit the info. That's just general remark and one of good practices to break the lines if too long, but yeah... whatever ;)

PR #2088, opened 2022-06-27
fix(value): make impossible to set list root node as a value (DEV-973)

mpro7:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: DEV-973

- [ ] Yes
- [x] No


.gitignore
```diff
@@ -36,7 +36,7 @@ sipi/test
 *.bak
 *.rdb
 .sbtrc
-/*.trig
+**/*.trig
```
irinaschubert, line 39: Wouldn't this ignore all trig files? I think Balduin had the problem that the trig file for an update plugin was ignored. This shouldn't be the case.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
```diff
@@ -2820,11 +2820,65 @@ object HierarchicalListValueContentV2 extends ValueContentReaderV2[HierarchicalL
...
+          HierarchicalListValueContentV2(
+            ontologySchema = ApiV2Complex,
+            valueHasListNode = listValueAsListNode.toString,
+            comment = getComment(jsonLDObject)
+          )
+        )
+    } yield listValueContent
+  }
+
+  private def checkIfNodeIsRoot(
```
irinaschubert, line 2845: docstring is missing

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
```diff
@@ -2820,11 +2820,65 @@ object HierarchicalListValueContentV2 extends ValueContentReaderV2[HierarchicalL
       throw BadRequestException(s"List node IRI <$listValueAsListNode> is not a Knora data IRI")
     }
 
-    HierarchicalListValueContentV2(
-      ontologySchema = ApiV2Complex,
-      valueHasListNode = listValueAsListNode.toString,
-      comment = getComment(jsonLDObject)
-    )
+    for {
+      isRootNode <- checkIfNodeIsRoot(
```
irinaschubert, line 2824: This check seems to be request specific. Did you check if `fromJsonLDObject` should never allow a root node?

irinaschubert (commented):
You didn't add a test. Is there already one? Or is it the case you mentioned today in the daily that you don't know how to test it?

.gitignore
mpro7: Previous state made`v7.0.0-knora-test.trig` trig enable to stage, therefore I change it. Now newly added plugins are also visible for staging because they are stored one lever deeper.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
mpro7: In this package only objects have docstrings, but in this case name speaks by itself and all arguments are inherited from the main definition.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
mpro7: I don't get your question. In the case of list values, yes it shouldn't be allowed a root node as value. That's actually the clue of this bug.

subotic: I'm not sure, that this check is implemented in the correct place. It should be probably better to move it to here: https://github.com/dasch-swiss/dsp-api/blob/3a0902ea9b44860bcb73e4ef25888f5ac8a5dc21/webapi/src/main/scala/org/knora/webapi/responders/v2/ValuesResponderV2.scala#L195-L201

This way, you will also be able to write the missing tests.

subotic (changes requested):

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
mpro7: @subotic Good point, moved checking method to `ResourceUtilV2` and added test to `ValuesResponderV2Spec`.

webapi/src/main/scala/org/knora/webapi/responders/v2/ValuesResponderV2.scala
```diff
@@ -193,13 +193,23 @@ class ValuesResponderV2(responderData: ResponderData) extends Responder(responde
...
         _ <- submittedInternalValueContent match {
                case listValue: HierarchicalListValueContentV2 =>
                  ResourceUtilV2.checkListNodeExists(listValue.valueHasListNode, appActor)
                case _ => FastFuture.successful(())
              }
 
+        //  If this is a list value, check if list value is not root list node.
+        _ <- submittedInternalValueContent match {
+               case listValue: HierarchicalListValueContentV2 =>
+                 ResourceUtilV2.checkIfNodeIsRoot(
```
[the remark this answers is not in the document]
mpro7: How this name is misleading? It says what it does = checks if the list node is root. To make it more precise it might be adjusted to `checkIfNodeIsRootAndThrowsErrorIfNot`, but IMO this is not really necessary here.

On the other comment you're right, this two queries are not optimal solution. I was actually thinking to use the one I created  -  which uses inside `getListNode`. At the end it gets the node, so if it isn't in the database should also return that information.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourceUtilV2.scala
```diff
@@ -132,6 +141,45 @@ object ResourceUtilV2 {
...
+        Future(
+          org.knora.webapi.messages.twirl.queries.sparql.admin.txt
+            .getListNode(nodeIri = nodeIri)
+            .toString()
+        )
+
+      listNodeResponse <-
+        appActor
+          .ask(
+            SparqlExtendedConstructRequest(
```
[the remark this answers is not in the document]
mpro7: There is only one but. With one query for both checks, I can only return one error, this shouldn't be a problem but it's less information for the user.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourceUtilV2.scala
```diff
@@ -133,25 +134,25 @@ object ResourceUtilV2 {
     } yield defaultObjectAccessPermissionsResponse.permissionLiteral
 
   /**
-   * Checks whether a list node exists, and throws [[NotFoundException]] otherwise.
+   * Checks whether a list node exists and has a root node, otherwise throws [[NotFoundException]].
    *
    * @param listNodeIri the IRI of the list node.
    */
-  def checkListNodeExists(listNodeIri: IRI, appActor: ActorRef)(implicit
+  def checkListNodeExistsAndHasRootNode(listNodeIri: IRI, appActor: ActorRef)(implicit
```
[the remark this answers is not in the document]
subotic: also, I'm missing a test for this method.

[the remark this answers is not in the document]
mpro7: That wasn't we but you are trying to force me to do so :) My opinion was different on that. However I've tried and spent some time to find out, that this method has been extracted to `ResourceUtilV2` because it's used twice in `ValuesResponderV2` as well as once in `ResourcesResponderV2`, where actually the usage is slightly different and I couldn't make things compile. Changing it from Unit method which throws to returning Boolean will also create more code duplication.
All this only strengthen my feelings:
1. I'm against increasing the scope of tasks, especially bugs. If we have bugs first policy, they should be fixed ASAP. This task is about fixing the bug, which I did applying the convention in the package, I also wrote test.
2. Changing  the way how only `checkListNodeExistsAndHasRootNode` works w/o touching other methods in `ResourceUtilV2` doesn't bring any value, in fact it makes even more confusion. If this is so important - should be moved to refactor task which will solve the problem comprehensively.
4. At the end I don't mind to refactor code while fixing the bugs, which in fact we all do. But this should be rather only necessary changes which helps in fixing the bug, like in this PR changing `checkListNodeExists` to `checkListNodeExistsAndHasRootNode` in order to reduce number of queries and lines of code - one method instead of two.

.gitignore
BalduinLandolt: I just tested it: this will ignore all .trig files, so future upgrade plugin test files will be ignored too, which is not ideal

webapi/src/main/scala/org/knora/webapi/responders/v2/ValuesResponderV2.scala
```diff
@@ -331,6 +330,9 @@ class ValuesResponderV2(responderData: ResponderData) extends Responder(responde
                                             featureFactoryConfig = createValueRequest.featureFactoryConfig,
                                             requestingUser = createValueRequest.requestingUser
                                           )
+
+        // _ = println(111, createValueRequest.createValue.resourceIri, resourceInfo.resourceClassIri)
+
```
BalduinLandolt, line 335: should be removed before merging

webapi/src/test/scala/org/knora/webapi/responders/v2/ValuesResponderV2Spec.scala
```diff
@@ -1780,6 +1780,32 @@ class ValuesResponderV2Spec extends CoreSpec() with ImplicitSender {
...
+            valueHasListNode = valueHasListNode
+          )
+        ),
+        featureFactoryConfig = defaultFeatureFactoryConfig,
+        requestingUser = anythingUser1,
+        apiRequestID = UUID.randomUUID
+      )
+
+      expectMsgPF(timeout) { case msg: akka.actor.Status.Failure =>
+        assert(msg.cause.isInstanceOf[NotFoundException])
```
BalduinLandolt, line 1805: should this really be a "NotFoundException", seems more like a "BadRequest", right?

BalduinLandolt (approved):
LGTM, just some minor things. (Not sure if that will show up easily because I responded in a resolved conversation, but I also commented on the .gitignore thing)

webapi/src/test/scala/org/knora/webapi/responders/v2/ValuesResponderV2Spec.scala
mpro7: It has been improved with [58b4551](https://github.com/dasch-swiss/dsp-api/pull/2088/commits/58b4551f7321c27301650478108afac59f56ddec) and now depends of outcome I throw `BadRequestException` or `NotFoundException`.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourceUtilV2.scala
```diff
@@ -133,27 +140,47 @@ object ResourceUtilV2 {
     } yield defaultObjectAccessPermissionsResponse.permissionLiteral
 
   /**
-   * Checks whether a list node exists, and throws [[NotFoundException]] otherwise.
+   * Checks whether a list node exists and if is a root node.
```
subotic, line 143: could you maybe also add a description of the return type?

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -1161,9 +1160,31 @@ class ResourcesResponderV2(responderData: ResponderData) extends ResponderWithSt
...
+          for {
+            checkNode <- ResourceUtilV2.checkListNodeExistsAndIsRootNode(listNodeIri, appActor)
+
+            _ = checkNode match {
+                  // it doesn't have isRootNode property - it's a child node
+                  case Right(false) =>
+                    FastFuture.successful(())
+                  // it does have isRootNode property - it's a root node
+                  case Right(true) =>
+                    FastFuture.failed(
```
subotic, line 1173: I'm not sure if you need `FastFuture.failed` when you are throwing.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -1161,9 +1160,31 @@ class ResourcesResponderV2(responderData: ResponderData) extends ResponderWithSt
 
     Future
       .sequence(
-        listNodesThatShouldExist
-          .map(listNodeIri => ResourceUtilV2.checkListNodeExists(listNodeIri, appActor))
-          .toSeq
+        listNodesThatShouldExist.map { listNodeIri =>
```
subotic, line 1163: since you are not interested in any values, maybe a `foreach` is more clearer.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -1161,9 +1160,31 @@ class ResourcesResponderV2(responderData: ResponderData) extends ResponderWithSt
...
+          for {
+            checkNode <- ResourceUtilV2.checkListNodeExistsAndIsRootNode(listNodeIri, appActor)
+
+            _ = checkNode match {
+                  // it doesn't have isRootNode property - it's a child node
+                  case Right(false) =>
+                    FastFuture.successful(())
```
subotic, line 1170: you are not interessted in the value, so simply return `()`. No need to warp it into a Future only to discard it later.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourceUtilV2.scala
mpro7: Added `@return` type.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
mpro7: That was the most tricky part. W/o`FastFuture.failed()` it didn't compile, but strangely now it seems working. Suspect metals was broken at that time...

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
mpro7: Same as above - I went for the same solution as I implemented in `ValueResponderV2`. but w/o`FastFuture.successful(())` it didn't compile. Strangely now it seems working. Suspect metals was broken at that time...

irinaschubert (approved):
LGTM

PR #2089, opened 2022-06-27
fix(ontology): Don't accept list values without gui attribute (DEV-775)

irinaschubert:
Resolves DEV-775

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: DEV-775

- [ ] Yes
- [x] No


dsp-shared/src/main/scala/dsp/valueobjects/Schema.scala
```diff
@@ -0,0 +1,209 @@
...
+
+  /**
+   * Validates if gui elements that require pointing to a list (List, Radio, Pulldown) actually point to a list
+   *
+   * @param guiElement the gui element that needs to be validated
+   * @param guiAttributes the gui attributes that need to be validated
+   *
+   * @return either the validated list of gui attributes or a [[dsp.errors.BadRequestException]]
+   */
+  private[valueobjects] def validateGuiObjectsPointingToList(
```
BalduinLandolt, line 134: no idea if that is how it should be done... but cool that it works

dsp-shared/src/main/scala/dsp/valueobjects/Schema.scala
```diff
@@ -0,0 +1,209 @@
...
+   * @param v the value parameter of the gui attribute (an int, a list's IRI etc.)
+   */
+  sealed abstract case class GuiAttribute private (k: String, v: String) {
+    val value = k + "=" + v
+  }
+  object GuiAttribute {
+    def make(keyValue: String): Validation[Throwable, GuiAttribute] = {
+      val k: String = keyValue.split("=").head.trim()
+      val v: String =
+        keyValue.split("=").last.trim() // TODO also check the type of the value (integer, string etc.)
```
mpro7, line 98: Why not to do in this PR?

dsp-shared/src/test/scala/dsp/valueobjects/SchemaSpec.scala
```diff
@@ -0,0 +1,483 @@
...
+      "pass a gui element that points to a list but has a misfitting gui attribute and return an error"
+    ) {
+      val guiElement    = guiElementList
+      val guiAttributes = scala.collection.immutable.List(guiAttributeSize)
+      assertTrue(
+        Schema.validateGuiObjectsPointingToList(guiElement, guiAttributes) == Validation.fail(
+          BadRequestException(
+            "salsah-gui:guiAttribute for salsah-gui:guiElement GuiElement(http://www.knora.org/ontology/salsah-gui#List) has to be a list reference of the form 'hlist=<LIST_IRI>', but found GuiAttribute(size,80)."
+          )
+        )
```
BalduinLandolt, line 142: I tend to like the pattern of first declaring a variable with the expected result and then asserting against that, rather than writing it only in the assertion. In most cases I find it more readable. But that's of course 100% a matter of taste.

dsp-shared/src/test/scala/dsp/valueobjects/SchemaSpec.scala
```diff
@@ -0,0 +1,483 @@
...
+                         scala.collection.immutable.List(guiAttributeHlist),
+                         Some(guiElementList)
+                       )
+      } yield assertTrue(
+        guiObject.guiAttributes == scala.collection.immutable.List(guiAttributeHlist)
+      ) &&
+        assertTrue(
+          guiObject.guiElement == Some(guiElementList)
+        )).toZIO
+    },
```
BalduinLandolt, line 233: I don't understand why we need to use a for comprehension and toZIO here (and in the following suites) but didn't in the first one

dsp-shared/src/main/scala/dsp/valueobjects/Schema.scala
```diff
@@ -0,0 +1,209 @@
...
+ */
+
+package dsp.valueobjects
+
+import dsp.constants.SalsahGui
+import dsp.errors.BadRequestException
+import zio.prelude.Subtype
+import zio.prelude.Validation
+
+object Schema {
```
mpro7, line 13: Why `Schema`?

webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
```diff
@@ -800,6 +817,53 @@ class OntologiesRouteV2(routeData: KnoraRouteData) extends KnoraRoute(routeData)
...
+              // validate the gui attributes by creating value objects
+              guiAttributes = maybeGuiAttributes.map(guiAttribute => GuiAttribute.make(guiAttribute)).toList
+
+              validatedGuiAttributes = Validation.validateAll(guiAttributes)
+
+              // validate the combination of gui element and gui attribute by creating a GuiObject value object
+              guiObject: GuiObject = Validation
+                                       .validate(validatedGuiAttributes, validatedGuiElement)
+                                       .flatMap(values => GuiObject.make(values._1, values._2))
+                                       .fold(e => throw e.head, v => v)
```
BalduinLandolt, line 866: here again, we're dropping all but one error

webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
```diff
@@ -904,21 +968,76 @@ class OntologiesRouteV2(routeData: KnoraRouteData) extends KnoraRoute(routeData)
...
+              // validate the new gui attributes by creating value objects
+              guiAttributes = newGuiAttributes.map(guiAttribute => GuiAttribute.make(guiAttribute)).toList
+
+              validatedGuiAttributes = Validation.validateAll(guiAttributes)
+
+              // validate the combination of gui element and gui attribute by creating a GuiObject value object
+              guiObject: GuiObject = Validation
+                                       .validate(validatedGuiAttributes, validatedNewGuiElement)
+                                       .flatMap(values => GuiObject.make(values._1, values._2))
+                                       .fold(e => throw e.head, v => v)
```
BalduinLandolt, line 1030: as above

mpro7 (approved):
Looks good, however I feel a little bit lost. Maybe you could give some introduction to it by the next occasion.

dsp-shared/src/main/scala/dsp/constants/SalsahGui.scala
```diff
@@ -0,0 +1,119 @@
...
+    val Iri: Value     = Value(4, "iri")
+
+    val valueMap: Map[String, Value] = values.map(v => (v.toString, v)).toMap
+
+    def lookup(name: String): Value =
+      valueMap.get(name) match {
+        case Some(value) => value
+        case None        => throw InconsistentRepositoryDataException(s"salsah-gui attribute type not found: $name")
+      }
+  }
```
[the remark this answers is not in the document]
irinaschubert: This enumeration was there before, I didn't touch it. I removed the other Enumerations we discussed this week. I used Lists instead. I see if I can easily replace it by a simple List.

dsp-shared/src/main/scala/dsp/valueobjects/Iri.scala
```diff
@@ -137,6 +137,27 @@ object Iri {
...
+        Validation.fail(BadRequestException(IriErrorMessages.PropertyIriMissing))
+      } else {
+        // TODO all the following needs to be checked when validating a property iri (see string formatter for the implementations of these methods)
+        // if (
+        //   !(propertyIri.isKnoraApiV2EntityIri &&
+        //     propertyIri.getOntologySchema.contains(ApiV2Complex) &&
+        //     propertyIri.getOntologyFromEntity == externalOntologyIri)
+        // ) {
+        //   throw BadRequestException(s"Invalid property IRI: $propertyIri")
+        // }
```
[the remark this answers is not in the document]
irinaschubert: Yes, I would like to keep this for the TODO task. I added it as follow-up task, which I hope can be done soon.

dsp-shared/src/main/scala/dsp/constants/SalsahGui.scala
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * `IRI` is a synonym for `String`, used to improve code readability.
+   */
+  type IRI = String
+
+  val InternalOntologyStart = "http://www.knora.org/ontology"
+
+  val SalsahGuiOntologyLabel: String = "salsah-gui"
+  val SalsahGuiOntologyIri: IRI      = InternalOntologyStart + "/" + SalsahGuiOntologyLabel
+  val SalsahGuiPrefixExpansion: IRI  = SalsahGuiOntologyIri + "#"
```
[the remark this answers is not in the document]
irinaschubert: Because the two Strings are not IRIs. I add the type though to the one that is missing. 
I copied these constants from the webapi project, so it was like this before.

dsp-shared/src/main/scala/dsp/constants/SalsahGui.scala
```diff
@@ -0,0 +1,119 @@
...
+  val Geometry: IRI    = SalsahGuiPrefixExpansion + "Geometry"
+  val Colorpicker: IRI = SalsahGuiPrefixExpansion + "Colorpicker"
+  val List: IRI        = SalsahGuiPrefixExpansion + "List"
+  val Radio: IRI       = SalsahGuiPrefixExpansion + "Radio"
+  val Checkbox: IRI    = SalsahGuiPrefixExpansion + "Checkbox"
+  val Richtext: IRI    = SalsahGuiPrefixExpansion + "Richtext"
+  val Interval: IRI    = SalsahGuiPrefixExpansion + "Interval"
+  val TimeStamp: IRI   = SalsahGuiPrefixExpansion + "TimeStamp"
+  val Geonames: IRI    = SalsahGuiPrefixExpansion + "Geonames"
+  val Fileupload: IRI  = SalsahGuiPrefixExpansion + "Fileupload"
```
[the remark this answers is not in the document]
irinaschubert: Again, I've just copied these values from webapi. You are actually right with the inconsistencies, but the values are defined (and named) like this in the Salsah GUI ontology. So, I think it is good to keep them exactly as they are named there even if it is ugly.

dsp-shared/src/main/scala/dsp/valueobjects/Schema.scala
subotic: @BalduinLandolt Do you mean the package scoped private method?

dsp-shared/src/test/scala/dsp/valueobjects/SchemaSpec.scala
```diff
@@ -0,0 +1,483 @@
...
+
+  private val guiAttributeTest = suite("gui attribute")(
+    test("pass an empty value and return an error") {
+      assertTrue(
+        Schema.GuiAttribute.make("") == Validation.fail(BadRequestException(SchemaErrorMessages.GuiAttributeMissing))
+      )
+    },
+    test("pass an invalid value and return an error") {
+      assertTrue(
+        Schema.GuiAttribute.make(guiAttributeInvalidString) == Validation.fail(
```
subotic, line 68: In my opinion, crating an additional variable instead of adding the string here directly is not more readable. The text in the name of the unit test is more than clear enough what will be passed. If anything, now I need to scroll up all the time to see the string against which the test is run, to check if it is actually correct.

subotic (approved):
LGTM, thanks!

dsp-shared/src/main/scala/dsp/valueobjects/Schema.scala
BalduinLandolt: yes, I didn't know this construct... Scala still baffles me at times - but in a good way!  
The reason I wrote "not sure if that's how it should be done" is that we had a discussion earlier, to which I was referring, where we the three of us were unclear what best practices are regarding unit-testing private methods.

dsp-shared/src/test/scala/dsp/valueobjects/SchemaSpec.scala
mpro7: Very elegant solution @BalduinLandolt but in this particular case, when string is used only once, wrapping it into val doesn't make sense. Also tests tends to be more dirty in terms of repetitions anyway, just to make things more visible/readable.

dsp-shared/src/test/scala/dsp/valueobjects/SchemaSpec.scala
irinaschubert: Like this, the `.fold(e => throw e.head, v => v)` is not needed to get to the actual value object, instead everything is treated inside the for comprehension. But now that you mentioned it, I changed it to the `.fold(...)`. I think it's easier to understand.

dsp-shared/src/test/scala/dsp/valueobjects/SchemaSpec.scala
irinaschubert: I followed Ivan's and Marcin's opinion here and replaced the `val`s with the actual strings where it made sense to me. The unit tests here are very simple indeed but for more complex tests I like Balduin's idea of defining given and expected values outside of the asserts. 

I agree with Balduin's suggestion about the Exception and try to change that in the value objects.

dsp-shared/src/main/scala/dsp/valueobjects/Schema.scala
irinaschubert: I do this task in https://dasch.atlassian.net/browse/DEV-1082

webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
irinaschubert: I will try the multiple errors handling in https://dasch.atlassian.net/browse/DEV-1082 (becuase right now, there is never more than one error)

dsp-shared/src/main/scala/dsp/valueobjects/Schema.scala
```diff
@@ -0,0 +1,209 @@
...
+
+      if (needsGuiAttribute) {
+        if (guiAttributes.isEmpty) {
+          return Validation.fail(BadRequestException(SchemaErrorMessages.GuiAttributesMissing))
+        }
+        val validatedGuiAttributes: Validation[Throwable, List[GuiAttribute]] = guiElement match {
+          // gui element is a list, radio or pulldown, so it needs a gui attribute that points to a list
+          case Some(guiElement) if guiElementsPointingToList.contains(guiElement.value) =>
+            validateGuiObjectsPointingToList(guiElement, guiAttributes).fold(
+              e => Validation.fail(e.head),
```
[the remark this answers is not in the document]
irinaschubert: I will try this out in https://dasch.atlassian.net/browse/DEV-1082 (because right now there is never more than one error)

PR #2182, opened 2022-08-29
chore: Improve GitHub actions (DEV-1112)

irinaschubert:
Resolves DEV-1112

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] CI related changes
- [x] Documentation content changes

Issue Number: N/A

- [ ] Yes
- [x] No


subotic (approved):

.github/workflows/scala-steward.yml
```diff
@@ -1,4 +1,4 @@
-name: Launch Scala Steward
+name: Dependencies
```
BalduinLandolt, line 1: do we now have two workflows with the same name "Dependencies"? (this one and dependency-graph.yml)

dsp-user/core/src/test/scala/dsp/user/domain/UserDomainSpec.scala
```diff
@@ -17,7 +17,7 @@ import dsp.valueobjects.User._
  */
 object UserDomainSpec extends ZIOSpecDefault {
 
-  def spec = (compareUsersTest + createUserTest + updateUserTest)
+  def spec = suite("UserDomainSpec")(compareUsersTest, createUserTest, updateUserTest)
```
BalduinLandolt, line 20: any particular reason for this change? should this be done everywhere?

BalduinLandolt (approved):

.github/workflows/scala-steward.yml
irinaschubert: Yes, because they both relate to dependencies.

dsp-user/core/src/test/scala/dsp/user/domain/UserDomainSpec.scala
irinaschubert: According to Ivan it is a recommendation by John to do it like that. The pattern is: the "root suite" should be called the same as the name of the file that is checked. If you like it, we can introduce it everywhere.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,28 @@
...
+          token: ${{ secrets.GH_TOKEN }}
+          release-type: simple
+          package-name: dsp-api
+          changelog-types: '[
+            {"type": "feat", "section": "Enhancements", "hidden": false },
+            {"type": "fix", "section": "Bug Fixes", "hidden": false },
+            {"type": "build", "section": "Maintenance", "hidden": false },
+            {"type": "chore", "section": "Maintenance", "hidden": false },
+            {"type": "refactor", "section": "Maintenance", "hidden": false },
+            {"type": "docs", "section": "Documentation", "hidden": false }
```
mpro7, line 27: What about `test` option we have in template and not here?

mpro7 (approved):

.github/workflows/create-release.yml
irinaschubert: I don't know to be honest. @subotic  is there a reason why `test` (and also `style`) is missing in this list?

dsp-user/core/src/test/scala/dsp/user/domain/UserDomainSpec.scala
irinaschubert: We can discuss it in our API meeting. I add it to the list of topics.

.github/workflows/create-release.yml
mpro7: `Build` also is obsolete here, I actually added it because someone never used `chore` for the maintenance stuff, and it was missing in the change log later on.

There is no Test section, so we can either or keep it simple and have it under `Maintenance`, the same as refactor.

irinaschubert: I would say `style`, `test` and `refactor` should go under `Maintenance` (I guess the end user doesn't care about the differences). And I would delete `build` as it is not on "our" list of prefixes (https://docs.dasch.swiss/latest/developers/dsp/contribution/). But as you are dealing with the release notes now, you can decide and I will do whatever you think makes sense.

mpro7: So add these which are missing and let's keep the `build` there as an alias to `chore`, because it's used frequently by Ivan.

subotic: > I don't know to be honest. @subotic is there a reason why `test` (and also `style`) is missing in this list?

This stems from the idea behind release notes, which are for the end-user. They should only be informed about bug fixes, improvements and new features. The rest is not having any impact for the end-user and thus does not need to be included in the release notes. Since these release notes are not for the end-user but for us, we can do here whatever we like.

subotic: Also keep in mind that these are not our release-notes. This is a summary of the Github release. Github release does not equal DaSCH DSP release.

PR #2223, opened 2022-09-28
fix: API returns invalid file URLs, due to including the port

BalduinLandolt:
Issue Number: DEV-1378

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/config/AppConfig.scala
+++ b/webapi/src/main/scala/org/knora/webapi/config/AppConfig.scala
@@ -134,12 +134,10 @@ final case class Sipi(
   audioMimeTypes: List[String],
   archiveMimeTypes: List[String]
 ) {
-  val internalBaseUrl: String = "http://" + internalHost + (if (internalPort != 80)
-                                                              ":" + internalPort
-                                                            else "")
-  val externalBaseUrl: String = "http://" + externalHost + (if (externalPort != 80)
-                                                              ":" + externalPort
-                                                            else "")
+  lazy val internalBaseUrl: String =
+    "http://" + internalHost + (if (internalPort != 80) ":" + internalPort else "")
+  lazy val externalBaseUrl: String =
+    "http://" + externalHost + (if (externalPort != 80) ":" + externalPort else "")
   val timeoutInSeconds: duration.Duration = scala.concurrent.duration.Duration(timeout)
 
 }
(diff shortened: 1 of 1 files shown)


irinaschubert (approved):

webapi/src/main/scala/org/knora/webapi/config/AppConfig.scala
```diff
@@ -134,12 +134,10 @@ final case class Sipi(
...
-  val internalBaseUrl: String = "http://" + internalHost + (if (internalPort != 80)
-                                                              ":" + internalPort
-                                                            else "")
-  val externalBaseUrl: String = "http://" + externalHost + (if (externalPort != 80)
-                                                              ":" + externalPort
-                                                            else "")
+  val internalBaseUrl: String =
+    internalProtocol + "://" + internalHost + (if (internalPort != 80) ":" + internalPort else "")
+  val externalBaseUrl: String =
+    externalProtocol + "://" + externalHost + (if (externalPort != 80) ":" + externalPort else "")
```
mpro7, line 140:
```suggestion
  val internalBaseUrl: String = internalProtocol + "://" + internalHost
  val externalBaseUrl: String = externalProtocol + "://" + externalHost
```
Does the port is needed here at all? If not then it should be shortened as above.

mpro7 (approved):
Approved, but check my suggestion/section.

webapi/src/main/scala/org/knora/webapi/config/AppConfig.scala
BalduinLandolt: hard to say, it might break locally, or in tests... But even if so, I do suspect that `externalPort != 80` does not make sense, it should be `1024` if anything, because that's the config default, which should only happen when there is no environment variable set (and it's also what the docker-compose specifies, when ewe run it locally).  
But I am reluctant to change these things now: Then we might not know what actually fixed the but. So IMO it would be best to only change what we expect to be the core issue, and then do a separate cleanup of all this stuff in a dedicated task. The same should also be done with all other configs, not only SIPI - but then we'll have to do much more testing, so we can't deploy the fix to prod anytime soon... Does that make sense?

BalduinLandolt: yes... it's a guessing fix at that point (though there is a good chance this will actually fix it, we did do some testing). But the guessing thing only gets harder when we change more things, so I'd actually prefer it like this, I think.

PR #2230, opened 2022-10-02
refactor: app actor cleanup

subotic:
Issue Number: DEV-

### Basic Requirements

### PR Type

- [ ] Yes
- [ ] No

- [ ] No

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
+++ b/webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
@@ -58,8 +58,7 @@ final case class AppServer(
   private def upgradeRepository(requiresRepository: Boolean): ZIO[Any, Nothing, Unit] =
     for {
       _ <- state.set(AppState.UpdatingRepository)
-      _ <- if (requiresRepository) ru.maybeUpgradeRepository.flatMap(response => ZIO.logInfo(response.message))
-           else ZIO.unit
+      _ <- ru.maybeUpgradeRepository.flatMap(response => ZIO.logInfo(response.message)).when(requiresRepository)
       _ <- state.set(AppState.RepositoryUpToDate)
     } yield ()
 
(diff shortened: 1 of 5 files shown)


subotic:
Just some minor cleanup making the code (hopefully) simpler to read

webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
```diff
@@ -97,19 +95,15 @@ final case class AppServer(
...
+      _ <- iiifs
+             .getStatus()
+             .flatMap {
+               case IIIFServiceStatusOK =>
+                 ZIO.logInfo("IIIF service running")
+               case IIIFServiceStatusNOK =>
+                 ZIO.logError("IIIF service not running") *> ZIO.die(new Exception("IIIF service not running"))
+             }
+             .when(requiresIIIFService)
       _ <- state.set(AppState.IIIFServiceReady)
```
BalduinLandolt, line 107: I'm just wondering if it's actually reasonable to set the AppState to IIIF stuff, if we don't require the IIIF service. So maybe this method could be rewritten to:

```scala
private def checkIIIFService(requiresIIIFService: Boolean): ZIO[Any, Nothing, Unit] =
  if (requiresIIIFService) {
    for {
      _      <- state.set(AppState.WaitingForIIIFService)
      status <- iiifs.getStatus()
      _      <- status match {
        case CacheServiceStatusNOK =>
          ZIO.logError("Cache service not running.") *> 
            ZIO.die(new Exception("Cache service not running."))
        case CacheServiceStatusOK =>
          ZIO.unit
      }
      _      <- state.set(AppState.IIIFServiceReady)
    } yield ()
  } else ZIO.unit
```

optionally, you could also use ZIO.when to test the bool condition. And even ZIO.whenCase to check the status

BalduinLandolt (approved):
note my comment on IIIF: not sure if we should use the stati WaitingForIIIF and IIIFReady in instances where no IIIF service is required

mpro7 (commented):
Why the conditions (`appConfig.allowReloadOverHttp` and `requiresIIIFService`) were removed?

irinaschubert (approved):

webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
subotic: yes, good point. With ZIO I'm actually not sure if we need these states in such a detail at all, or if something waiting for startup is not good enough. Now that we have logging that works, we should quickly figure out where the startup is stuck. So I will merge it as is, and we can create a follow-up issue after we discuss it.

subotic:
> Why the conditions (`appConfig.allowReloadOverHttp` and `requiresIIIFService`) were removed?

they are now checked inside `when`

PR #2256, opened 2022-10-19
refactor: Introduce ZIO HTTP (DEV-1425)

irinaschubert:
Issue Number: DEV-1425

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No


dsp-user/interface/src/main/scala/dsp/user/route/UserRoutes.scala
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * The user related routes which need AppConfig in the environment
+   */
+  val routes: Http[AppConfig, ValidationException, Request, Response] = Http.collectZIO[Request] {
+    // POST /admin/users
+    case req @ (Method.POST -> !! / "admin" / "users") =>
+      CreateUser
+        .route(req, userHandler)
+        .catchAll { e =>
+          ZIO.succeed(Response.text(e.getMessage).setStatus(Status.BadRequest))
```
BalduinLandolt, line 29: just wondering: in case of a failure, do we really want to return text, or rather JSON? If so, we'd probably need an object that holds a failure state, and can serialize itself to JSON

dsp-user/interface/src/main/scala/dsp/user/route/UserRoutes.scala
```diff
@@ -0,0 +1,55 @@
...
+      GetUser.route(uuid, userHandler).catchAll { e =>
+        ZIO.succeed(Response.text(e.getMessage).setStatus(Status.BadRequest))
+      }
+
+  }
+}
+
+/**
+ * Companion object providing the layer with an initialized implementation of the UserHandler
+ */
```
BalduinLandolt, line 51: do we need this level of comments? It's correct, but it's a pattern that's everywhere and that every DSP-API developer should know

webapi/src/main/scala/org/knora/webapi/routing/HealthRouteWithZIOHttp.scala
```diff
@@ -0,0 +1,117 @@
...
+  private def createResponse(result: HealthCheckResult): UIO[Response] =
+    ZIO.succeed(
+      Response
+        .json(
+          JsObject(
+            "name"     -> JsString("AppState"),
+            "severity" -> JsString("non fatal"),
+            "status"   -> JsString(status(result.status)),
+            "message"  -> JsString(result.message)
+          ).toString()
```
BalduinLandolt, line 64: could we use ZioJson here instead?

webapi/src/main/scala/org/knora/webapi/routing/HealthRouteWithZIOHttp.scala
```diff
@@ -0,0 +1,117 @@
...
+ */
+final case class HealthRouteWithZIOHttp(state: State) extends HealthCheckWithZIOHttp {
+
+  /**
+   * Returns the route.
+   */
+  val route: Http[State, Nothing, Request, Response] =
+    Http.collectZIO[Request] { case Method.GET -> !! / "healthZ" =>
+      for {
+        //  ec    <- ZIO.executor.map(_.asExecutionContext) // TODO leave this for reference about how to get the execution context
```
BalduinLandolt, line 103: this is not really a TODO, I would say:

```suggestion
        //  ec    <- ZIO.executor.map(_.asExecutionContext) // leave this for reference about how to get the execution context
```

webapi/src/main/scala/org/knora/webapi/routing/HealthRouteWithZIOHttp.scala
```diff
@@ -0,0 +1,117 @@
...
+    for {
+      _        <- ZIO.logDebug("get application state")
+      state    <- state.getAppState
+      result   <- setHealthState(state)
+      _        <- ZIO.logDebug("set health state")
+      response <- createResponse(result)
+      _        <- ZIO.logDebug("getting application state done")
+    } yield response
+
+  private def setHealthState(state: AppState): UIO[HealthCheckResult] =
```
BalduinLandolt, line 31: the naming of this function is confusing (to me). `setXY` to me suggests changing a state (e.g. updating a Ref) but what it does is actually turning one representation of a health state into a different representation

webapi/src/main/scala/org/knora/webapi/routing/HealthRouteWithZIOHttp.scala
BalduinLandolt: also, using a hard-coded sting for name and severity, when we actually have the identical string in the object seems very redundant

BalduinLandolt (approved):
LGTM, some minor remarks

dsp-shared/src/main/scala/dsp/config/AppConfig.scala
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Configuration
+ */
+final case class AppConfig(
+  knoraApi: KnoraApi,
+  bcryptPasswordStrength: PasswordStrength
+)
+
+final case class KnoraApi(
```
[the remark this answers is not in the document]
irinaschubert: good point, I will adjust it

dsp-shared/src/main/scala/dsp/valueobjects/User.scala
```diff
@@ -8,14 +8,17 @@ package dsp.valueobjects
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder
 import org.springframework.security.crypto.scrypt.SCryptPasswordEncoder
 import zio._
+import zio.config.magnolia.Descriptor
```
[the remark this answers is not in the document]
irinaschubert: Magnolia Descriptor allows us to automatically create the AppConfig from the typesafe source, it's this part:
```
private val configFromSource: IO[ReadError[String], AppConfig] = read(
    descriptor[AppConfig].mapKey(toKebabCase) from source
  )
```
Without it, we would have to manually define each parameter in the configuration (what type it is, what name it has etc.). At least that's how I understood it.

dsp-shared/src/main/scala/dsp/valueobjects/User.scala
```diff
@@ -25,8 +28,14 @@ object User {
   /**
    * Username value object.
    */
-  sealed abstract case class Username private (value: String)
+  final case class Username private (value: String)
```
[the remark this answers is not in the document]
irinaschubert: Regarding `Username`: It is valid as one word "username" in English, so I think this is OK without camel-casing it. Is that OK with you?

dsp-shared/src/main/scala/dsp/valueobjects/User.scala
```diff
@@ -160,11 +190,25 @@ object User {
 
   }
   object PasswordHash {
+    // TODO: get the passwordstrength from appConfig instead (see CreateUser.scala as example)
+
+    // the decoder defines how to decode json to an object
```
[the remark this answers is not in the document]
irinaschubert: yes

dsp-shared/src/main/scala/dsp/valueobjects/User.scala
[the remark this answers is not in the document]
BalduinLandolt: sure, it just struck me because we were talking about this camel case/not camel case recently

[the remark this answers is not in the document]
irinaschubert: Yes, I know. We should talk about the naming thoroughly in the domain model initiative...

dsp-user/handler/src/main/scala/dsp/user/handler/UserHandler.scala
```diff
@@ -247,9 +291,7 @@ final case class UserHandler(repo: UserRepo) {
...
 object UserHandler {
   val layer: ZLayer[UserRepo, Nothing, UserHandler] =
-    ZLayer {
-      for {
-        repo <- ZIO.service[UserRepo]
-      } yield UserHandler(repo)
-    }.tap(_ => ZIO.logInfo(">>> User handler initialized <<<"))
+    ZLayer
+      .fromFunction(UserHandler.apply _)
+      .tap(_ => ZIO.logInfo(">>> User handler initialized <<<"))
```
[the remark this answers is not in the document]
irinaschubert: Hm, I see your point, but I think `*>` does something else. When I replace it, the compiler tells me that `*>` is not a member of `ZLayer`s. I think `*>` ignores the first effect and executes the thing that comes after it, whereas `tap` "returns an effect that effectfully 'peeks' at the success of this effect" but still returns it.

dsp-user/interface/src/main/scala/dsp/user/route/UserRoutes.scala
irinaschubert: What do you mean by "failure state"? The HTTP status code is returned, isn't that enough?

dsp-user/interface/src/main/scala/dsp/user/route/UserRoutes.scala
irinaschubert: I think it was there before (in multiple/other locations, too). I removed it everywhere.

webapi/src/main/scala/org/knora/webapi/routing/HealthRouteWithZIOHttp.scala
irinaschubert: I thought the same but I only touched the code in order to get it to compile - nothing more. I don't know why we need the "AppState" and "non fatal" if it is hard-coded and in both cases (healthy/unhealthy) the same anyway. Maybe @subotic  has an opinion about it?

dsp-api-main/src/main/scala/dsp/api/main/DspMiddleware.scala
```diff
@@ -0,0 +1,26 @@
...
+        http: Http[R1, E1, Request, Response]
+      ): Http[R1, E1, Request, Response] =
+        Http.fromOptionFunction[Request] { request =>
+          Random.nextUUID.flatMap { requestId =>
+            ZIO.logAnnotate("RequestId", requestId.toString) {
+              for {
+                result <- http(request)
+              } yield result
+            }
+          }
```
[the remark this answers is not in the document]
irinaschubert: Hm, I tried but couldn't figure out how. If you have a suggestion..? :)

webapi/src/main/scala/org/knora/webapi/routing/HealthRouteWithZIOHttp.scala
irinaschubert: I totally agree. We need to decide how far we go when refactoring the routes: only shallow refactoring in order for the route to work or deep refactoring / rethinking them (including naming, structure, style etc.)? @subotic what is your opinion on that?

In the meantime I added some docstrings to explain what the methods do.

webapi/src/main/scala/org/knora/webapi/routing/HealthRouteWithZIOHttp.scala
BalduinLandolt: I think generally it's more important to replace sprayJson with zioJson than getting rid of two hardcoded strings

dsp-shared/src/main/scala/dsp/valueobjects/Id.scala
```diff
@@ -80,6 +82,11 @@ object Id {
    */
   object UserId {
 
+    implicit val decoder: JsonDecoder[UserId] = JsonDecoder[Iri.UserIri].mapOrFail { case iri =>
+      UserId.fromIri(iri).toEitherWith(e => e.head.getMessage())
```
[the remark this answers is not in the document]
BalduinLandolt: hm... it's gone now... but previously codacy complained with all those decoders that they're not covered by test coverage. But not to worry about it!

dsp-user/handler/src/main/scala/dsp/user/handler/UserHandler.scala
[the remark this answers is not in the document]
irinaschubert: Still the "not a member of ZLayers" error. I leave it as is right now. We can replace it, as soon as we found a better solution.

dsp-shared/src/main/scala/dsp/valueobjects/User.scala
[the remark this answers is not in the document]
irinaschubert: About the inconsistencies in the definition of the value object classes: I unified them to `sealed abstract case class`es, as they were in the beginning. I don't remember exactly what the problem was in the beginning with defining JSON encoders/decoders in `abstract` classes, but now it seems to work.

dsp-user/interface/src/main/scala/dsp/user/route/UserRoutes.scala
seakayone: In the long run the api should return a common type of response for all 400 errors. This is a question for designing the api. For the task at hand I think the existing solution is fine.

@BalduinLandolt I like this approach and it is not overthinking it is necessary design work we need to do for the v3 API at some point. However, exposing a trace in a response is helpful when developing but could be an unnecessary attack vector in production.

seakayone (approved):

seakayone (approved):

dsp-user/interface/src/main/scala/dsp/user/route/UserRoutes.scala
irinaschubert: Alright, thank you both for your inputs. I leave it for now and we can introduce such an ErrorResponse type in the future.

webapi/src/main/scala/org/knora/webapi/routing/HealthRouteWithZIOHttp.scala
irinaschubert: I see and you are totally right. There is already an epic for refactoring the health route: https://dasch.atlassian.net/browse/DEV-1218, so I classify your input to "out of scope" here and add it to that issue.

dsp-shared/src/main/scala/dsp/util/UuidGenerator.scala
```diff
@@ -0,0 +1,83 @@
...
+    ZLayer.succeed(UuidGeneratorLive())
+}
+
+/**
+ * Test instance of the UuidGenerator which stores the created UUIDs, so that
+ * they can be queried / returned in tests
+ *
+ * @param listOfUuids the list of created UUIDs
+ */
+final case class UuidGeneratorTest(listOfUuids: Ref[List[UUID]]) extends UuidGenerator {
```
seakayone, line 42: I would probably rename it to something like `MockUuidGenerator` and also put it into the `src/test/scala`source set as this would prevent us from accidentially use it in production code.

dsp-shared/src/main/scala/dsp/util/UuidGenerator.scala
```diff
@@ -0,0 +1,83 @@
...
+   * Returns the list of created UUIDs of this instance
+   *
+   * @return the list of created UUIDs
+   */
+  def getCreatedUuids: UIO[List[UUID]] =
+    listOfUuids.get
+
+}
+object UuidGeneratorTest {
+  val listOfUuids = Ref.make(List.empty[UUID])
```
seakayone, line 66: This appears to be unused.

seakayone, line 66: Can be removed I guess, it appears to be unused.

dsp-user/interface/src/main/scala/dsp/user/route/CreateUser.scala
```diff
@@ -0,0 +1,94 @@
...
+   * @param userHandler the userHandler that handles user actions
+   * @return            a response with the user as json
+   */
+  def route(req: Request, userHandler: UserHandler): RIO[AppConfig & UuidGenerator, Response] =
+    for {
+      // get the appConfig from the environment
+      appConfig <- ZIO.service[AppConfig]
+
+      userCreatePayload <-
+        req.body.asString.map(_.fromJson[CreateUserPayload]).orElseFail(ValidationException("Couldn't parse payload"))
```
BalduinLandolt, line 31: for me personally, that's too much going on in on e line. It's a matter of taste, but I think I would have first extracted the body as sting in a variable, and then parsed the variable into the payload object from that string. It would result in the same amount of lines, but (at least to me) look more readable

dsp-user/interface/src/main/scala/dsp/user/route/CreateUser.scala
```diff
@@ -0,0 +1,94 @@
...
+
+      response <-
```
BalduinLandolt, line 33: again, stylistics: I'm not a huge fan of nesting for-comprehensions: They are syntactic sugar for un-nesting consecutive flatmaps, so I would rather keep them in a "imperative line-by-line do-this-ten-that" style.  
(feel free to ignore)

dsp-user/interface/src/test/scala/dsp/user/route/CreateUserSpec.scala
```diff
@@ -0,0 +1,153 @@
...
+                  )
+        response    <- CreateUser.route(request, userHandler).map(_.body)
+        uuid        <- UuidGeneratorTest.getCreatedUuids.head
+        responseStr <- response.asString
+      } yield assertTrue(
+        responseStr.startsWith(
+          s"""{"id":"http://rdfh.ch/users/$uuid","givenName":"Hans","familyName":"Muster","username":"hansmuster","email":"hans.muster@example.org","password":"$$2a"""
+        ) &&
+          responseStr.endsWith(
+            """"language":"en","status":true}"""
```
BalduinLandolt, line 147: is there a assertion to match a regex? that would be slightly more elegant.  
Otherwise I'd suggest comparing the `.asStringPretty` (or what it's actually called) to a nice-looking json string.

On a different note: how concerned should I be that the response to the "create user" request returns the password hash? (I know it's not productive code, but just to future-proof it)

BalduinLandolt (approved):
LGTM, only cosmetic and "XY came to my mind" comments

dsp-api-main/src/main/scala/dsp/api/main/DspMain.scala
```diff
@@ -0,0 +1,41 @@
...
+import zio.logging.backend.SLF4J
+import dsp.config.AppConfig
+import dsp.util.UuidGeneratorLive
+
+object DspMain extends ZIOAppDefault {
+
+  /**
+   * Configures Metrics to be run at a set interval, in this case every 5 seconds
+   */
+  val metricsConfig = ZLayer.succeed(MetricsConfig(5.seconds))
```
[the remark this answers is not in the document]
irinaschubert: Currently nowhere, it is referenced on line 37 but commented out. I thought it might be useful in the future. But you are right, it is confusing and I will just delete this.

dsp-shared/src/main/scala/dsp/util/UuidGenerator.scala
irinaschubert: I renamed it to `UuidGeneratorMock` according to our (inofficial?) naming pattern to put "Live", "Test", "Mock" etc. at the end.

dsp-shared/src/main/scala/dsp/util/UuidGenerator.scala
```diff
@@ -0,0 +1,83 @@
...
+    ZIO.service[UuidGeneratorTest].flatMap(_.getCreatedUuids)
+
+  val layer: ULayer[UuidGeneratorTest] =
+    ZLayer {
+      for {
+        listOfUuids <- Ref.make(List.empty[UUID])
+      } yield UuidGeneratorTest(listOfUuids)
+    }
+
+}
```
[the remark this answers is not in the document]
irinaschubert: Yes, I like it very much. It was Christian who provided this solution. I renamed it now to "Mock" instead of "Test" and moved it to the test package. I also added a test and updated it (again according to Christian's input) so that we can control storing and retrieving UUIDs in a more explicit/controlled way in tests (as developers).

dsp-user/interface/src/test/scala/dsp/user/route/CreateUserSpec.scala
irinaschubert: About your concern: you are certainly right (Ivan also pointed this out to me), but that would IMO require a proper API design which requires a definition of our information objects which requires a valid domain model... It's out of scope here.

dsp-shared/src/main/scala/dsp/valueobjects/LanguageCode.scala
```diff
@@ -15,6 +17,12 @@ import dsp.errors.ValidationException
 sealed abstract case class LanguageCode private (value: String)
 
 object LanguageCode { self =>
+  implicit val decoder: JsonDecoder[LanguageCode] = JsonDecoder[String].mapOrFail { case value =>
+    LanguageCode.make(value).toEitherWith(e => e.head.getMessage())
+  }
+  implicit val encoder: JsonEncoder[LanguageCode] =
+    JsonEncoder[String].contramap((languageCode: LanguageCode) => languageCode.value)
```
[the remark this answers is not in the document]
irinaschubert: You are right, I simplify it.

dsp-shared/src/main/scala/dsp/valueobjects/User.scala
```diff
@@ -74,15 +83,22 @@ object User {
    */
   sealed abstract case class Email private (value: String)
   object Email { self =>
+    // the codec defines how to decode/encode the object from/into json
+    implicit val codec: JsonCodec[Email] =
+      JsonCodec[String].transformOrFail(
+        value => Email.make(value).toEitherWith(e => e.head.getMessage()),
```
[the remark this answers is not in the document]
irinaschubert: Certainly a good idea, but as this code is not done yet and not used, I classify it as out of scope for now.

dsp-user/interface/src/main/scala/dsp/user/route/CreateUser.scala
irinaschubert: I agree that this would be more readable. Unfortunately, I need to close this PR as soon as possible because other tasks with high priority are waiting. So, I leave it for now.

dsp-user/interface/src/test/scala/dsp/user/route/CreateUserSpec.scala
irinaschubert: About the ugly way I do the test: I agree that it should be improved. Unfortunately, I need to close this PR as soon as possible because other tasks with high priority are waiting. So, I leave it for now.

PR #2267, opened 2022-10-24
fix: Ask timeouts with GetUserADM (DEV-1443)

irinaschubert:
Issue Number: DEV-1443

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No

### Other information
I set an increased timeout to the Authenticator. The rest is refactoring/clean-up.


webapi/src/main/scala/org/knora/webapi/routing/Authenticator.scala
```diff
@@ -792,7 +814,7 @@ object Authenticator extends InstrumentationSupport {
               userInformationTypeADM = UserInformationTypeADM.Full,
               requestingUser = KnoraSystemInstances.Users.SystemUser
             )
-          )
+          )(Duration(100, SECONDS))
```
irinaschubert, line 817: this is the actual bug fix

subotic (approved):

webapi/src/main/scala/org/knora/webapi/routing/Authenticator.scala
mpro7, line 817: Did you try something between former 5 and current 100 seconds? Also, is this timeout related to the ones we have defined in the `application.conf` and if this could be solved on that level?

irinaschubert: No, I can't try out the effect on prod on my current setup (or any other setup) as we currently don't have the measures to reproduce the exact same load. Also, it is not the same timeout as the default timeout in `application.conf` (which would be 120 minutes). We could introduce a `askTimeout` in `application.conf` and use that everywhere. But I'm not sure if that would introduce more performance issues. What do the others think of Marcin's suggestion, @subotic , @BalduinLandolt ?

subotic: In my opinion, the current fix is sufficient enough. In mid-term, we need to get rid of Akka Actors and thus the Ask completely.

mpro7: > We could introduce a `askTimeout` in `application.conf` and use that everywhere. But I'm not sure if that would introduce more performance issues. What do the others think of Marcin's suggestion, @subotic , @BalduinLandolt ?

I was exactly referring to this idea, if it's used in more than one place. However if this is the temporary idea, it's fine as it is. Unless this is going to be as usually, temp but staying forever ;)

BalduinLandolt: IMO it would be nicer to have a config than to hard code values... but as it's supposed to be a quick fix, and in the mid term we want to get rid of Akka anyways, it may not be worth the effort

BalduinLandolt (approved):
looks good

webapi/src/main/scala/org/knora/webapi/routing/Authenticator.scala
irinaschubert: OK, thanks for your thoughts. I will leave it as is for now and we hopefully can handle load issues properly when we introduce ZIO HTTP :)

PR #2268, opened 2022-10-26
fix!: partOf and sequenceOf properties are not marked as isEditable

BalduinLandolt:
Issue Number: DEV-1439

### Basic Requirements

### PR Type

- [x] Yes
- [ ] No

The PR introduces changes to knora-base, which requires an upgrade script to run, so it needs to be marked as breaking. However, it does not change any public API or break existing code.

- [x] No


mpro7 (approved):

knora-ontologies/knora-base.ttl
```diff
@@ -561,6 +561,7 @@
     rdfs:comment            "Indicates that this resource is part of another resource"@en ;
     :subjectClassConstraint :Resource ;
     :objectClassConstraint  :Resource ;
+    :isEditable             true ;
```
irinaschubert, line 564: Hm, I think I don't understand why here everything is set to `:isEditable` `true`. Shouldn't `knora-base` things not be editable?

irinaschubert (approved):
LGTM, I just haven't understood why `isEditable` now needs to be set to true. Wouldn't that mean that users can edit knora-base properties - which should not be the case? (I am sure that you did it right, I just want to understand :)

BalduinLandolt:
> LGTM, I just haven't understood why `isEditable` now needs to be set to true. Wouldn't that mean that users can edit knora-base properties - which should not be the case? (I am sure that you did it right, I just want to understand :)

Well... I suppose `isEditable` is a very badly named flag. The API never lets you edit knora-base (I think/hope). There isn't really any documentation on it either, so it's hard to reconstruct the reasoning behind the flag. But in knora-base itself it's defined with the comment `"Indicates whether a property's values can be updated via the Knora API."@en`, so it seems to be referring to the values instantiating the property, rather than the property as defined in knora-base. And it seems that "editing a value" is to be taken as equivalent to "create a value". So it probably should be `canBeInstantiated` - but we already have that and it can only be used on classes, not on properties.

Does that make it clearer?

irinaschubert:
> > LGTM, I just haven't understood why `isEditable` now needs to be set to true. Wouldn't that mean that users can edit knora-base properties - which should not be the case? (I am sure that you did it right, I just want to understand :)
> 
> Well... I suppose `isEditable` is a very badly named flag. The API never lets you edit knora-base (I think/hope). There isn't really any documentation on it either, so it's hard to reconstruct the reasoning behind the flag. But in knora-base itself it's defined with the comment `"Indicates whether a property's values can be updated via the Knora API."@en`, so it seems to be referring to the values instantiating the property, rather than the property as defined in knora-base. And it seems that "editing a value" is to be taken as equivalent to "create a value". So it probably should be `canBeInstantiated` - but we already have that and it can only be used on classes, not on properties.
> 
> Does that make it clearer?

Yes, thanks a lot!

subotic (approved):
I have a few questions, if `isSequenceOf` is only used for audio/video, then shouldn't `:objectClassConstraint ` be `:Representation` instead of `:Resource`?

BalduinLandolt:
> I have a few questions, if `isSequenceOf` is only used for audio/video, then shouldn't `:objectClassConstraint ` be `:Representation` instead of `:Resource`?

Yes and no: It's the same with `isPartOf` - we don't say it can *only* be on these kinds of resources, we merely say "if it's used on these resources in combination with certain other properties (i.e. sequence bounds or seqnum), the frontend *may* provide special behaviour"

This could of course be changed, but then we have to check the DB for other usages

PR #2272, opened 2022-10-31
feat(projectsADM): add possibility to get project and members by UUID (DEV-1408)

mpro7:
Issue Number: DEV-1488

### Basic Requirements

### PR Type

- [x] Feature: represents a new feature

- [ ] Yes
- [x] No

- [x] No

### Other information

### DIFF
--- a/dsp-shared/src/main/scala/dsp/valueobjects/Iri.scala
+++ b/dsp-shared/src/main/scala/dsp/valueobjects/Iri.scala
@@ -133,6 +133,22 @@ object Iri {
   }
 
   /**
+   * Base64Uuid value object.
+   * This is base64 encoded UUID version without paddings.
+   *
+   * @param value to validate.
+   */
+  sealed abstract case class Base64Uuid private (value: String)
+  object Base64Uuid {
+    def make(value: String): Validation[ValidationException, Base64Uuid] =
+      if (value.isEmpty) {
+        Validation.fail(ValidationException("UUID is missing."))
+      } else if (!V2UuidValidation.isUuidVersion4Or5(value)) {
+        Validation.fail(ValidationException(IriErrorMessages.UuidVersionInvalid))
+      } else Validation.succeed(new Base64Uuid(value) {})
+  }
+
+  /**
    * RoleIri value object.
    */
   sealed abstract case class RoleIri private (value: String) extends Iri
--- a/dsp-shared/src/main/scala/dsp/valueobjects/Iri.scala
+++ b/dsp-shared/src/main/scala/dsp/valueobjects/Iri.scala
@@ -142,7 +142,9 @@ object Iri {
   object Base64Uuid {
     def make(value: String): Validation[ValidationException, Base64Uuid] =
       if (value.isEmpty) {
-        Validation.fail(ValidationException("UUID is missing."))
+        Validation.fail(ValidationException(IriErrorMessages.UuidMissing))
+      } else if (!V2UuidValidation.hasUuidLength(value)) {
+        Validation.fail(ValidationException(IriErrorMessages.UuidInvalid(value)))
       } else if (!V2UuidValidation.isUuidVersion4Or5(value)) {
         Validation.fail(ValidationException(IriErrorMessages.UuidVersionInvalid))
       } else Validation.succeed(new Base64Uuid(value) {})
@@ -237,6 +239,8 @@ object IriErrorMessages {
   val RoleIriInvalid     = (iri: String) => s"Role IRI: $iri is invalid."
   val UserIriMissing     = "User IRI cannot be empty."
   val UserIriInvalid     = (iri: String) => s"User IRI: $iri is invalid."
+  val UuidMissing        = "UUID cannot be empty"
+  val UuidInvalid        = (uuid: String) => s"'$uuid' is not a UUID"
   val UuidVersionInvalid = "Invalid UUID used to create IRI. Only versions 4 and 5 are supported."
   val PropertyIriMissing = "Property IRI cannot be empty."
 }
--- a/dsp-shared/src/main/scala/dsp/valueobjects/V2.scala
+++ b/dsp-shared/src/main/scala/dsp/valueobjects/V2.scala
@@ -225,14 +225,21 @@ object V2UuidValidation {
     getUUIDVersion(s) == 4 || getUUIDVersion(s) == 5
 
   /**
-   * Gets the last segment of IRI, decodes UUID and gets the version.
-   * @param s the string (IRI) to be checked.
+   * Gets the last segment of IRI, decodes UUID and gets its version.
+   * @param s the IRI [[String]] to be checked.
    * @return UUID version.
    */
   def getUUIDVersion(s: String): Int = {
-    val encodedUUID = s.split("/").last
+    val encodedUUID = getUuidFromIri(s)
     decodeUuid(encodedUUID).version()
   }
+
+  /**
+   * Gets the last IRI segment - Base64 encoded UUID
+   * @param iri the IRI [[String]] to get the UUID from.
+   * @return Base64 encoded UUID as [[String]]
+   */
+  def getUuidFromIri(iri: String): String = iri.split("/").last
 
   /**
    * Calls `decodeUuidWithErr`, throwing [[InconsistentRepositoryDataException]] if the string cannot be parsed.
--- a/dsp-shared/src/main/scala/dsp/valueobjects/V2.scala
+++ b/dsp-shared/src/main/scala/dsp/valueobjects/V2.scala
@@ -225,17 +225,17 @@ object V2UuidValidation {
     getUUIDVersion(s) == 4 || getUUIDVersion(s) == 5
 
   /**
-   * Gets the last segment of IRI, decodes UUID and gets its version.
-   * @param s the IRI [[String]] to be checked.
+   * Decodes Base64 encoded UUID and gets its version.
+   * @param uuid the Base64 encoded UUID as [[String]] to be checked.
    * @return UUID version.
    */
-  def getUUIDVersion(s: String): Int = {
-    val encodedUUID = getUuidFromIri(s)
+  def getUUIDVersion(uuid: String): Int = {
+    val encodedUUID = getUuidFromIri(uuid)
     decodeUuid(encodedUUID).version()
   }
 
   /**
-   * Gets the last IRI segment - Base64 encoded UUID
+   * Gets the last IRI segment - Base64 encoded UUID.
    * @param iri the IRI [[String]] to get the UUID from.
    * @return Base64 encoded UUID as [[String]]
    */
--- a/webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
@@ -16,8 +16,9 @@ import java.nio.file.Path
 import java.util.UUID
 
 import dsp.errors.BadRequestException
-import dsp.errors.DataConversionException
 import dsp.errors.OntologyConstraintException
+import dsp.valueobjects.Iri.ProjectIri
+import dsp.valueobjects.Project
 import dsp.valueobjects.V2
 import org.knora.webapi.IRI
 import org.knora.webapi.messages.ResponderRequest.KnoraRequestADM
@@ -504,132 +505,140 @@ case class ProjectADM(
   }
 }
 
-/**
- * The ProjectIdentifierADM factory object, making sure that all necessary checks are performed and all inputs
- * validated and escaped.
- */
+sealed trait ProjectIdentifierADM
 object ProjectIdentifierADM {
-  def apply(maybeIri: Option[IRI] = None, maybeShortname: Option[String] = None, maybeShortcode: Option[String] = None)(
-    implicit sf: StringFormatter
-  ): ProjectIdentifierADM = {
-
-    val parametersCount: Int = List(
-      maybeIri,
-      maybeShortname,
-      maybeShortcode
-    ).flatten.size
-
-    // something needs to be set
-    if (parametersCount == 0) throw BadRequestException("Empty project identifier is not allowed.")
-
-    if (parametersCount > 1) throw BadRequestException("Only one option allowed for project identifier.")
-
-    new ProjectIdentifierADM(
-      maybeIri =
-        sf.validateAndEscapeOptionalProjectIri(maybeIri, throw BadRequestException(s"Invalid user project $maybeIri")),
-      maybeShortname = sf.validateAndEscapeOptionalProjectShortname(
-        maybeShortname,
-        throw BadRequestException(s"Invalid user project shortname $maybeShortname")
-      ),
-      maybeShortcode = sf.validateAndEscapeOptionalProjectShortcode(
-        maybeShortcode,
-        throw BadRequestException(s"Invalid user project shortcode $maybeShortcode")
-      )
-    )
-  }
+  case class Iri(value: ProjectIri)              extends ProjectIdentifierADM
+  case class Shortcode(value: Project.ShortCode) extends ProjectIdentifierADM
+  case class Shortname(value: Project.ShortName) extends ProjectIdentifierADM
+  case class Uuid(value: String)                 extends ProjectIdentifierADM
 }
 
-/**
- * Represents the project's identifier. It can be an IRI, shortcode or shortname.
- *
- * @param maybeIri       the project's IRI.
- * @param maybeShortname the project's shortname.
- * @param maybeShortcode the project's shortcode
- */
-class ProjectIdentifierADM private (
-  maybeIri: Option[IRI] = None,
-  maybeShortname: Option[String] = None,
-  maybeShortcode: Option[String] = None
-) {
+// /**
+//  * The ProjectIdentifierADM factory object, making sure that all necessary checks are performed and all inputs
+//  * validated and escaped.
+//  */
+// object ProjectIdentifierADM {
+//   def apply(maybeIri: Option[IRI] = None, maybeShortname: Option[String] = None, maybeShortcode: Option[String] = None)(
+//     implicit sf: StringFormatter
+//   ): ProjectIdentifierADM = {
 
-  // squash and return value.
-  val value: String = List(
-    maybeIri,
-    maybeShortname,
-    maybeShortcode
-  ).flatten.head
+//     val parametersCount: Int = List(
+//       maybeIri,
+//       maybeShortname,
+//       maybeShortcode
+//     ).flatten.size
 
-  def hasType: ProjectIdentifierType =
-    if (maybeIri.isDefined) {
-      ProjectIdentifierType.IRI
-    } else if (maybeShortcode.isDefined) {
-      ProjectIdentifierType.SHORTCODE
-    } else {
-      ProjectIdentifierType.SHORTNAME
-    }
+//     // something needs to be set
+//     if (parametersCount == 0) throw BadRequestException("Empty project identifier is not allowed.")
 
-  /**
-   * Tries to return the value as an IRI.
-   */
-  def toIri: IRI =
-    maybeIri.getOrElse(
-      throw DataConversionException(s"Identifier $value is not of the required 'ProjectIdentifierType.IRI' type.")
-    )
+//     if (parametersCount > 1) throw BadRequestException("Only one option allowed for project identifier.")
 
-  /**
-   * Returns an optional value of the identifier.
-   */
-  def toIriOption: Option[IRI] =
-    maybeIri
+//     new ProjectIdentifierADM(
+//       maybeIri =
+//         sf.validateAndEscapeOptionalProjectIri(maybeIri, throw BadRequestException(s"Invalid user project $maybeIri")),
+//       maybeShortname = sf.validateAndEscapeOptionalProjectShortname(
+//         maybeShortname,
+//         throw BadRequestException(s"Invalid user project shortname $maybeShortname")
+//       ),
+//       maybeShortcode = sf.validateAndEscapeOptionalProjectShortcode(
+//         maybeShortcode,
+//         throw BadRequestException(s"Invalid user project shortcode $maybeShortcode")
+//       )
+//     )
+//   }
+// }
 
-  /**
-   * Tries to return the value as an SHORTNAME.
-   */
-  def toShortname: String =
-    maybeShortname.getOrElse(
-      throw DataConversionException(s"Identifier $value is not of the required 'ProjectIdentifierType.SHORTNAME' type.")
-    )
+// /**
+//  * Represents the project's identifier. It can be an IRI, shortcode or shortname.
+//  *
+//  * @param maybeIri       the project's IRI.
+//  * @param maybeShortname the project's shortname.
+//  * @param maybeShortcode the project's shortcode
+//  */
+// class ProjectIdentifierADM private (
+//   maybeIri: Option[IRI] = None,
+//   maybeShortname: Option[String] = None,
+//   maybeShortcode: Option[String] = None
+// ) {
 
-  /**
-   * Returns an optional value of the identifier.
-   */
-  def toShortnameOption: Option[String] =
-    maybeShortname
+//   // squash and return value.
+//   val value: String = List(
+//     maybeIri,
+//     maybeShortname,
+//     maybeShortcode
+//   ).flatten.head
 
-  /**
-   * Tries to return the value as an SHORTCODE.
-   */
-  def toShortcode: String =
-    maybeShortcode.getOrElse(
-      throw DataConversionException(s"Identifier $value is not of the required 'ProjectIdentifierType.SHORTCODE' type.")
-    )
+//   def hasType: ProjectIdentifierType =
+//     if (maybeIri.isDefined) {
+//       ProjectIdentifierType.IRI
+//     } else if (maybeShortcode.isDefined) {
+//       ProjectIdentifierType.SHORTCODE
+//     } else {
+//       ProjectIdentifierType.SHORTNAME
+//     }
 
-  /**
-   * Returns an optional value of the identifier.
-   */
-  def toShortcodeOption: Option[String] =
-    maybeShortcode
+//   /**
+//    * Tries to return the value as an IRI.
+//    */
+//   def toIri: IRI =
+//     maybeIri.getOrElse(
+//       throw DataConversionException(s"Identifier $value is not of the required 'ProjectIdentifierType.IRI' type.")
+//     )
 
-  /**
-   * Returns the string representation
-   */
-  override def toString: IRI =
-    s"ProjectIdentifierADM(${this.value})"
+//   /**
+//    * Returns an optional value of the identifier.
+//    */
+//   def toIriOption: Option[IRI] =
+//     maybeIri
 
-}
+//   /**
+//    * Tries to return the value as an SHORTNAME.
+//    */
+//   def toShortname: String =
+//     maybeShortname.getOrElse(
+//       throw DataConversionException(s"Identifier $value is not of the required 'ProjectIdentifierType.SHORTNAME' type.")
+//     )
 
-/**
- * Project identifier types:
- *  - IRI
- *  - Shortcode
- *  - Shortname
- */
-sealed trait ProjectIdentifierType
-object ProjectIdentifierType {
-  case object IRI       extends ProjectIdentifierType
-  case object SHORTCODE extends ProjectIdentifierType
-  case object SHORTNAME extends ProjectIdentifierType
-}
+//   /**
+//    * Returns an optional value of the identifier.
+//    */
+//   def toShortnameOption: Option[String] =
+//     maybeShortname
+
+//   /**
+//    * Tries to return the value as an SHORTCODE.
+//    */
+//   def toShortcode: String =
+//     maybeShortcode.getOrElse(
+//       throw DataConversionException(s"Identifier $value is not of the required 'ProjectIdentifierType.SHORTCODE' type.")
+//     )
+
+//   /**
+//    * Returns an optional value of the identifier.
+//    */
+//   def toShortcodeOption: Option[String] =
+//     maybeShortcode
+
+//   /**
+//    * Returns the string representation
+//    */
+//   override def toString: IRI =
+//     s"ProjectIdentifierADM(${this.value})"
+
+// }
+
+// /**
+//  * Project identifier types:
+//  *  - IRI
+//  *  - Shortcode
+//  *  - Shortname
+//  */
+// sealed trait ProjectIdentifierType
+// object ProjectIdentifierType {
+//   case object IRI       extends ProjectIdentifierType
+//   case object SHORTCODE extends ProjectIdentifierType
+//   case object SHORTNAME extends ProjectIdentifierType
+// }
 
 /**
  * Represents the project's restricted view settings.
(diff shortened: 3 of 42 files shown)


BalduinLandolt (commented):
Tests are missing... as discussed, this is probably not testable in unit tests. but is there a ent-to-end test that you could use?

mpro7:
> Tests are missing... as discussed, this is probably not testable in unit tests. but is there a ent-to-end test that you could use?

This is testable in unit tests of the project route which we don't have. We only have e2e tests of the project responder, which would look the same as get by IRI tests, because making UUID the IRI to reuse existing stuff. But if these tests pass we can assume that it's good.

BalduinLandolt (approved):
Looks good. Nothing to add from my side, apart from what we already discussed yesterday

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteADM.scala
```diff
@@ -173,18 +175,23 @@ class ProjectsRouteADM(routeData: KnoraRouteData)
     }
 
   /**
-   * returns a single project identified through iri
+   * Returns a single project identified through IRI or UUID
    */
-  private def getProjectByIri(): Route =
-    path(projectsBasePath / "iri" / Segment) { value =>
+  private def getProjectByIriOrUuid(routeResolver: String): Route =
```
subotic, line 180: If you would like to implement it in this way, than you would need change a few things IMO:
1. create a proper type for the `routeResolver`.
2. Also, `routeResolver` is not the best name for it. Maybe something like IdSegment would be more descriptive. This new type would then be a sum type for UuidSegment and IriSegment.
3. To be consistent you would need to also add a `ShortcodeSegment` and a `ShortnameSegment`.

The point I want to emphasize is that the structure of the addition should be consistent with the rest of the implementation and not just a quick fix.

The other option would be to stick to the current design and simply add `getProjectByUuid`.

mpro7: Applied your suggestions in [f3a35b9](https://github.com/dasch-swiss/dsp-api/pull/2272/commits/f3a35b921b963eccc97171158ab646c29f652df5).

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +506,61 @@ case class ProjectADM(
...
-        maybeShortname,
-        throw BadRequestException(s"Invalid user project shortname $maybeShortname")
-      ),
-      maybeShortcode = sf.validateAndEscapeOptionalProjectShortcode(
-        maybeShortcode,
-        throw BadRequestException(s"Invalid user project shortcode $maybeShortcode")
-      )
-    )
+  case class Iri(value: String) extends ProjectIdentifierADM {
+    def apply(value: String): ProjectIri = ProjectIri.make(value).fold(e => throw e.head, v => v)
```
seakayone, line 514: tbh, when constructing a `case class` the last thing I would expect that an Exception is thrown.

How about adding the respective four factory methods to the trait which return the Validation and then throw the exception on the callee side of these factory methods, i.e. in the Responders?

Also IMHO the case class should not wrap a `String` but rater the typed identifier, maybe it would make even sense to move the ProjectIdentifier to the shared project?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +506,61 @@ case class ProjectADM(
...
+  case class Uuid(value: String) extends ProjectIdentifierADM
 
-  /**
-   * Tries to return the value as an SHORTNAME.
-   */
-  def toShortname: String =
-    maybeShortname.getOrElse(
-      throw DataConversionException(s"Identifier $value is not of the required 'ProjectIdentifierType.SHORTNAME' type.")
-    )
+  def asIriOption(id: ProjectIdentifierADM): Option[String] =
```
seakayone, line 527: These `as...Option` methods could become methods in the trait

```scala
sealed trait ProjectIdentifierADM {
  def asIri(): Option[IRI] =
    this match {
      case Iri(value) => Some(value)
      case _          => None
    }
}
```

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +506,61 @@ case class ProjectADM(
...
+    }
 
-  /**
-   * Returns an optional value of the identifier.
-   */
-  def toShortcodeOption: Option[String] =
-    maybeShortcode
+  def asUuidOption(id: ProjectIdentifierADM): Option[String] =
+    id match {
+      case Uuid(value) => Some(s"http://rdfh.ch/projects/${value}")
```
seakayone, line 547: This case does not return  `Uuid` as promised by the function. This returns something which looks like a `String` representation of a `ProjectIri`, isn't it?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +506,61 @@ case class ProjectADM(
...
-  case object IRI       extends ProjectIdentifierType
-  case object SHORTCODE extends ProjectIdentifierType
-  case object SHORTNAME extends ProjectIdentifierType
+  def getId(identifier: ProjectIdentifierADM): String =
+    identifier match {
+      case Iri(value)       => value
+      case Shortname(value) => value
+      case Shortcode(value) => value
+      case Uuid(value)      => s"http://rdfh.ch/projects/${value}"
+    }
```
seakayone, line 563: >    * Helper method that gets right identifier value.

What is a "right identifier value"? 
Is it the IRI representation of the given identifier?

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteADM.scala
```diff
@@ -353,7 +372,35 @@ class ProjectsRouteADM(routeData: KnoraRouteData)
     }
 
   /**
-   * returns all members part of a project identified through shortname
+   * Returns all members part of a project identified through UUID
+   */
+  private def getProjectMembersByUuid(): Route =
+    path(projectsBasePath / "uuid" / Segment / "members") { value =>
+      val iri: String = s"http://rdfh.ch/projects/$value"
```
seakayone, line 379: This `String` is repeated across you PR on multiple places. Essentially this is domain logic a we should aim at not repeating it.

IMHO this construct should be a single function maybe use the `ProjectIri.make` factory method instead?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: Is there more difference between these 2 ideas instead of less code (no param in methods). We keep methods in companion objects here...

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: Thanks for the suggestion, I think I messed up here anyway and things are not working as expected... I mean everything is working because validation is not working 🙈

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: You are right. This is kind of fake case class which as stated in the PR description reuses IRI methods. I think this would stay like that for now and should be changed once all IRI types will be transformed to new shape with UUID instead of shortcode.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: Right in the meaning the one was passed as param, because we currently support 3 + the new one UUID which is a part of IRI...

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteADM.scala
mpro7: This is good point, although this is temporary solution, which would be changed most probably when all IRI types will be changed to contain UUID instead of the shortcode. Currently this is only true for Project IRI.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +506,61 @@ case class ProjectADM(
...
-
-  def hasType: ProjectIdentifierType =
-    if (maybeIri.isDefined) {
-      ProjectIdentifierType.IRI
-    } else if (maybeShortcode.isDefined) {
-      ProjectIdentifierType.SHORTCODE
-    } else {
-      ProjectIdentifierType.SHORTNAME
-    }
+  case class Shortcode(value: String) extends ProjectIdentifierADM {
```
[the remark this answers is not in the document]
subotic: Also here you are doing something (for me) unexpected, which took me some time to understand. The constructor of `ProjectIdentifierADM.Shortcode` is actually returning a `ShortCode`, which I would never expect. I don't think that this is making the whole thing easier to understand, and also not what we want here (or what I was thinking when I tried to convey the idea). Also, I don't think the compiler will help out when matching (which was the whole point of the initial idea).

So you need to decide what you want to do here. IMO there are two possibilities:
(1) you put the different project identifier value objects (Project Iri, ShortCode, Shortname) in a type hierarchy where the value objects are implemented, and then you can reuse the parent value object type directly
(2) or, the more flexible approach, you create a separate type hierarchy (like we are trying to do here) and then package the value objects as values.

Which one you choose depends on how you would like to use it.

Having a mix of both is in my opinion neither the first nor the second approach and only unnecessarily more complex and confusing.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +506,61 @@ case class ProjectADM(
...
+      case Iri(value) => Some(value)
+      case _          => None
+    }
 
-  /**
-   * Returns an optional value of the identifier.
-   */
-  def toShortnameOption: Option[String] =
-    maybeShortname
+  def asShortcodeOption(id: ProjectIdentifierADM): Option[String] =
```
subotic, line 533: Please, no parameters here. You want this method to be part of the object, where you can ask the object to return itself as something else.

It should be more something like this:

```
object ProjectIdentifierADM { self =>
     val asShortcodeOption: Option[String] =
           self match {
                 case Shortcode(value) => Some(value)
                 case _                => None
      }
}
```

In the callsite you would then write something like: `identifier.asShortcodeOption`.

subotic (approved):

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +511,104 @@ case class ProjectADM(
...
-  ).flatten.head
-
-  def hasType: ProjectIdentifierType =
-    if (maybeIri.isDefined) {
-      ProjectIdentifierType.IRI
-    } else if (maybeShortcode.isDefined) {
-      ProjectIdentifierType.SHORTCODE
-    } else {
-      ProjectIdentifierType.SHORTNAME
+  def asShortcodeOption: Option[String] =
```
irinaschubert, line 523: Sometimes there is `Shortcode` and sometimes `ShortCode` (same for `Shortname` and `ShortName`). Could you check what is the right version and adjust the naming here, so everything is consistent?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +511,104 @@ case class ProjectADM(
...
   /**
-   * Returns an optional value of the identifier.
+   * Represents [[UuidIdentifier]] identifier.
+   *
+   * @param value that constructs the identifier in the type of [[Base64Uuid]] value object.
    */
-  def toShortcodeOption: Option[String] =
-    maybeShortcode
+  final case class UuidIdentifier(value: Base64Uuid) extends ProjectIdentifierADM
+  // TODO-mpro: UuidIdentifier isn't fully used yet since the value object needs
```
irinaschubert, line 589: I don't understand this sentence (I think it is incomplete?)

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +511,104 @@ case class ProjectADM(
...
-   * Tries to return the value as an SHORTNAME.
+   * Represents [[IriIdentifier]] identifier.
+   *
+   * @param value that constructs the identifier in the type of [[ProjectIri]] value object.
    */
-  def toShortname: String =
-    maybeShortname.getOrElse(
-      throw DataConversionException(s"Identifier $value is not of the required 'ProjectIdentifierType.SHORTNAME' type.")
-    )
+  final case class IriIdentifier(value: ProjectIri) extends ProjectIdentifierADM
```
irinaschubert, line 549: I don't fully understand why we need an `IriIdentifier`, `ShortcodeIdentifier` etc. if we have `ProjectIri`, `Shortcode` etc. already. To me this looks overengineered. Can't we just use the value objects `ProjectIri`, `Shortcode` etc. directly?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +511,104 @@ case class ProjectADM(
...
+  object IriIdentifier {
```
irinaschubert, line 550: I would call it `ProjectIriIdentifier` in order to not confuse it with other types of IRIs.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -505,130 +511,104 @@ case class ProjectADM(
...
- *  - IRI
- *  - Shortcode
- *  - Shortname
- */
-sealed trait ProjectIdentifierType
-object ProjectIdentifierType {
-  case object IRI       extends ProjectIdentifierType
-  case object SHORTCODE extends ProjectIdentifierType
-  case object SHORTNAME extends ProjectIdentifierType
+  def getId(identifier: ProjectIdentifierADM): String =
```
irinaschubert, line 605: To get an ID from an Identifier seems like non-sense to me. Can this be called differently, something like `getProjectIdType` and call the `ProjectIdentifierADM` `ProjectIdType` instead? I think this would make things easier to understand for others / in the future.

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteADM.scala
```diff
@@ -353,7 +381,35 @@ class ProjectsRouteADM(routeData: KnoraRouteData)
     }
 
   /**
-   * returns all members part of a project identified through shortname
+   * Returns all members part of a project identified through UUID
```
irinaschubert, line 384:
```suggestion
   * Returns all members of a project identified through the project's UUID
```

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteADM.scala
```diff
@@ -325,7 +353,7 @@ class ProjectsRouteADM(routeData: KnoraRouteData)
     }
 
   /**
-   * returns all members part of a project identified through iri
+   * Returns all members part of a project identified through IRI
```
irinaschubert, line 356:
```suggestion
   * Returns all members of a project identified through IRI
```

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteADM.scala
irinaschubert: This goes for all the other docstrings here that are similar as well. I won't repeat it each time.

webapi/src/main/scala/org/knora/webapi/store/cache/impl/CacheServiceInMemImpl.scala
```diff
@@ -131,20 +134,41 @@ case class CacheServiceInMemImpl(
    * @param id the project's IRI
    * @return an optional [[ProjectADM]].
    */
-  def getProjectByIri(id: String) = projects.get(id).commit
+  def getProjectByIri(id: Iri.ProjectIri) = projects.get(id.value).commit
+
+  /**
+   * Retrieves the project stored under a SHORTNAME.
```
irinaschubert, line 140:
```suggestion
   * Retrieves the project by its SHORTNAME.
```

webapi/src/main/scala/org/knora/webapi/store/cache/impl/CacheServiceInMemImpl.scala
```diff
@@ -131,20 +134,41 @@ case class CacheServiceInMemImpl(
...
+   */
+  def getProjectByShortname(shortname: Project.ShortName): UIO[Option[ProjectADM]] =
+    (for {
+      iri     <- lut.get(shortname.value).some
+      project <- projects.get(iri).some
+    } yield project).commit.unsome
 
   /**
-   * Retrieves the project stored under a SHORTCODE or SHORTNAME.
+   * Retrieves the project stored under a SHORTCODE.
```
irinaschubert, line 152:
```suggestion
   * Retrieves the project by its SHORTCODE.
```

webapi/src/main/scala/org/knora/webapi/store/cache/impl/CacheServiceInMemImpl.scala
```diff
@@ -131,20 +134,41 @@ case class CacheServiceInMemImpl(
    * @param id the project's IRI
    * @return an optional [[ProjectADM]].
    */
-  def getProjectByIri(id: String) = projects.get(id).commit
+  def getProjectByIri(id: Iri.ProjectIri) = projects.get(id.value).commit
```
irinaschubert, line 137:
```suggestion
  def getProjectByIri(iri: Iri.ProjectIri) = projects.get(iri.value).commit
```

webapi/src/main/scala/org/knora/webapi/store/cache/impl/CacheServiceInMemImpl.scala
```diff
@@ -131,20 +134,41 @@ case class CacheServiceInMemImpl(
...
-  def getProjectByShortcodeOrShortname(shortcodeOrShortname: String): UIO[Option[ProjectADM]] =
+  def getProjectByShortcode(shortcode: Project.ShortCode): UIO[Option[ProjectADM]] =
     (for {
-      iri     <- lut.get(shortcodeOrShortname).some
+      iri     <- lut.get(shortcode.value).some
       project <- projects.get(iri).some
     } yield project).commit.unsome
 
+  /**
+   * Retrieves the project stored under and extracted from the IRI.
```
irinaschubert, line 164:
```suggestion
   * Retrieves the project by its UUID.
```

webapi/src/main/scala/org/knora/webapi/store/cache/impl/CacheServiceInMemImpl.scala
```diff
@@ -131,20 +134,41 @@ case class CacheServiceInMemImpl(
...
+   *
+   * @param id the project's UUID
+   * @return an optional [[ProjectADM]].
+   */
+  def getProjectByUuid(id: Iri.Base64Uuid) =
```
irinaschubert, line 169:
```suggestion
  def getProjectByUuid(uuid: Iri.Base64Uuid) =
```

webapi/src/main/scala/org/knora/webapi/store/cache/impl/CacheServiceRedisImpl.scala
```diff
@@ -124,23 +127,65 @@ case class CacheServiceRedisImpl(pool: JedisPool) extends CacheService {
    * @param id the project's IRI
    * @return an optional [[ProjectADM]].
    */
-  def getProjectByIri(id: String): Task[Option[ProjectADM]] =
+  def getProjectByIri(id: Iri.ProjectIri): Task[Option[ProjectADM]] =
```
irinaschubert, line 130:
```suggestion
  def getProjectByIri(iri: Iri.ProjectIri): Task[Option[ProjectADM]] =
```

webapi/src/main/scala/org/knora/webapi/store/cache/impl/CacheServiceRedisImpl.scala
```diff
@@ -124,23 +127,65 @@ case class CacheServiceRedisImpl(pool: JedisPool) extends CacheService {
...
+        }
+    } yield project)
+
+  /**
+   * Retrieves the project stored under and extracted from the IRI.
+   *
+   * @param id the project's UUID
+   * @return an optional [[ProjectADM]].
+   */
+  def getProjectByUuid(id: Iri.Base64Uuid): Task[Option[ProjectADM]] =
```
irinaschubert, line 180:
```suggestion
  def getProjectByUuid(uuid: Iri.Base64Uuid): Task[Option[ProjectADM]] =
```

webapi/src/main/scala/org/knora/webapi/store/cache/impl/CacheServiceRedisImpl.scala
```diff
@@ -124,23 +127,65 @@ case class CacheServiceRedisImpl(pool: JedisPool) extends CacheService {
    * @param id the project's IRI
    * @return an optional [[ProjectADM]].
    */
-  def getProjectByIri(id: String): Task[Option[ProjectADM]] =
+  def getProjectByIri(id: Iri.ProjectIri): Task[Option[ProjectADM]] =
+    (for {
+      bytes <- getBytesValue(id.value)
```
irinaschubert, line 132: I don't understand why this has to be treated differently than shortcode and shortname and can't be treated as a String? (Same for `getProjectByUuid`.)

irinaschubert (approved):
I don't really understand the whole point in having e.g. a `Shortcode` and a `ShortcodeIdentifier` and not merge these two types into one (reuse the existing (new) value objects and let them extend the `ProjectIdentifierADM`). To me this looks overengineered but I probably don't see the whole picture. Apart from that it looks good to me, I added mainly suggestions to improve the readability of the code.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: Of course this would be the ideal solution, but the design of the responder and SPARQL quarry force to distinguish between identifier we store in the project. It could be improved but that forces even deeper refactoring which wasn't the scope of this task. I went at the beginning for the simplest solution, buy actually reusing getByIri route, but then decided to make it better and at least refactor the IdentifierADM, as an exercise and for a better quality code.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: `ShortCode` is value object, `asShortcodeOption` is accessor of `ShortcodeIdentifier` - there isn't `Shortcode` anymore, so maybe I will make accessor named euqally.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: It lives in `ProjectIdentifierADM` so isn't it enough to know that it's about `ProjectIri`? I like informative naming but we should decide if we go for last tail having all information, or make it a bit shorter if object where it lives already have it, so:
`ProjectIdentifierADM. IriIdentifier` vs `ProjectIdentifierADM. ProjectIriIdentifier`

mpro7: Especially there are no more `IriIdentifier`, the only one exists for the Project

dsp-shared/src/main/scala/dsp/valueobjects/V2.scala
```diff
@@ -225,15 +225,22 @@ object V2UuidValidation {
     getUUIDVersion(s) == 4 || getUUIDVersion(s) == 5
 
   /**
-   * Gets the last segment of IRI, decodes UUID and gets the version.
-   * @param s the string (IRI) to be checked.
+   * Gets the last segment of IRI, decodes UUID and gets its version.
+   * @param s the IRI [[String]] to be checked.
    * @return UUID version.
    */
   def getUUIDVersion(s: String): Int = {
```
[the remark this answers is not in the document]
mpro7: Not true anymore, it now gets the version directly from the last segment of IRI using `getUuidFromIri` method. I just forgot to update docs ;)

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: Not valid anymore, sorry forgot to remove it.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
irinaschubert: If it is used together with the object, as `ProjectIdentifierADM.IriIdentifier`, then it is enough.

BalduinLandolt (approved):
not much to add from my side

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
irinaschubert: Or maybe `getIdAsString`? Do whatever you think is best, you know best what the context is in which this method is used.

webapi/src/main/scala/org/knora/webapi/store/cache/impl/CacheServiceRedisImpl.scala
mpro7: Good point. I just repeated @subotic implementation, so maybe he can highlight us why this was implemented differently? However as we agreed this is not used anywhere thus going to be removed.

dsp-shared/src/main/scala/dsp/valueobjects/Iri.scala
```diff
@@ -221,6 +239,8 @@ object IriErrorMessages {
   val RoleIriInvalid     = (iri: String) => s"Role IRI: $iri is invalid."
   val UserIriMissing     = "User IRI cannot be empty."
   val UserIriInvalid     = (iri: String) => s"User IRI: $iri is invalid."
+  val UuidMissing        = "UUID cannot be empty"
+  val UuidInvalid        = (uuid: String) => s"'$uuid' is not a UUID"
   val UuidVersionInvalid = "Invalid UUID used to create IRI. Only versions 4 and 5 are supported."
```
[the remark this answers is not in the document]
mpro7: Partially true. Subject for value objects refactor, because it's also used in all IRIs as UUID validation. It should be extracted only to UUID value object which could be stored separately, but would trigger another refactor in this PR which wasn't a scope of it.

webapi/src/main/scala/org/knora/webapi/responders/admin/ListsResponderADM.scala
```diff
@@ -2267,9 +2270,9 @@ class ListsResponderADM(responderData: ResponderData) extends Responder(responde
       maybeProject <- appActor
```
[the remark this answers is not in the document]
mpro7: Well, I try to update the formatting whenever I don't forget ;)

webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/plugins/UpgradePluginPR2255.scala
```diff
@@ -61,6 +63,18 @@ class UpgradePluginPR2255(log: Logger) extends UpgradePlugin {
...
+  // KEEP it for reference, not a part of plugin and used only for replacement of
+  // IRIs existing in DSP-API repo
+  val onlyFoundInRepo: Map[String, String] = Map(
+    "0113" -> "bFURQB6zR1ugDFQL7EZy9Q",
+    "0666" -> "cXxbOh4xRce3uLUyXHujfg",
+    // beol in v1 test data
+    "0801" -> "bL0y8GRuTUiFmvF1oXbeFQ",
+    // dokubib in v1 test data
+    "00FE" -> "oIjhUsZmQLuJ0VMGvJ2pfg"
+  )
```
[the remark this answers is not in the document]
mpro7: What do you mean?

webapi/src/test/scala/org/knora/webapi/sharedtestdata/SharedTestDataADM.scala
```diff
@@ -202,7 +202,8 @@ object SharedTestDataADM {
   /**
    * **********************************
    */
-  val IMAGES_PROJECT_IRI = "http://rdfh.ch/projects/MTvoB0EJRrqovzRkWXqfkA"
+  val imagesProjectUuid  = "MTvoB0EJRrqovzRkWXqfkA"
+  val IMAGES_PROJECT_IRI = s"http://rdfh.ch/projects/$imagesProjectUuid"
```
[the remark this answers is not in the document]
mpro7: Indeed. There is already a task for clearance of the test data where it would suit better.

webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/plugins/UpgradePluginPR2255.scala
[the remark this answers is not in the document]
BalduinLandolt: something like that:

```scala
/*
 * "0113" -> "bFURQB6zR1ugDFQL7EZy9Q",
 * "0666" -> "cXxbOh4xRce3uLUyXHujfg",
 * // beol in v1 test data
 * "0801" -> "bL0y8GRuTUiFmvF1oXbeFQ",
 * // dokubib in v1 test data
 * "00FE" -> "oIjhUsZmQLuJ0VMGvJ2pfg"
*/
```

so that we definitely never get a `XY is never used` compiler warning... but it really doesn't matter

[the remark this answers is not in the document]
mpro7: Forgot to answer and change it to as you suggested. At the end I'm not sure if we should move it to some documentation in the repo on Confluence, but living here is most probably can't be omitted if someone touches this code. I have have change it in another PR. Also what was weird IDE didn't complain to it being unused...

PR #2281, opened 2022-11-07
chore: update dependencies

BalduinLandolt:
Issue Number: DEV-

### Basic Requirements

### PR Type

- [x] Chore: maintenance tasks (no production code change)

- [ ] Yes
- [x] No

- [x] No

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/Main.scala
+++ b/webapi/src/main/scala/org/knora/webapi/Main.scala
@@ -23,7 +23,7 @@ object Main extends ZIOApp {
    * and cleaned up when the Runtime is shutdown.
    */
   override def bootstrap: ZLayer[
-    ZIOAppArgs with Scope,
+    ZIOAppArgs,
     Any,
     Environment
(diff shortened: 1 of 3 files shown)


webapi/src/main/scala/org/knora/webapi/Main.scala
```diff
@@ -23,7 +23,7 @@ object Main extends ZIOApp {
    * and cleaned up when the Runtime is shutdown.
    */
   override def bootstrap: ZLayer[
-    ZIOAppArgs with Scope,
```
seakayone, line 26: Basically it lgtm, I just wonder what motivates this change in `Main.scala`?

BalduinLandolt: this seems to have changed in the new ZIO version, presumably the Scope layer is now baked into it, so that it doesn't have to be explicitly provided.  
In any case, I didn't actually change any code, just the type annotation which wasn't correct anymore (and caused a compilation error). But if this hadn't been annotated explicitly, no change would have been needed.

seakayone: Make sense, thanks for the explanation.

mpro7 (approved):

PR #2363, opened 2023-01-03
fix: API starts up and reports healthy despite failing to load ontologies

BalduinLandolt:
### Task Description/Number

Issue Number: DEV-1495

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
+++ b/webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
@@ -80,10 +80,15 @@ final case class AppServer(
    *
    * @param requiresRepository If `true`, calls the AppRouter to populate the ontology caches, otherwise returns ()
    */
-  private def populateOntologyCaches(requiresRepository: Boolean): UIO[Unit] =
+  private def populateOntologyCaches(requiresRepository: Boolean): IO[Throwable, Unit] =
     for {
       _ <- state.set(AppState.LoadingOntologies)
-      _ <- ar.populateOntologyCaches.when(requiresRepository)
+      _ <-
+        ar.populateOntologyCaches
+          .tapError(e =>
+            state.set(AppState.FailedToLoadOntologies(s"Failed to load ontologies from triplestore: ${e.getMessage()}"))
+          )
+          .when(requiresRepository)
       _ <- state.set(AppState.OntologiesReady)
     } yield ()
 
@@ -132,7 +137,7 @@ final case class AppServer(
   def start(
     requiresAdditionalRepositoryChecks: Boolean,
     requiresIIIFService: Boolean
-  ): UIO[Unit] =
+  ): IO[Throwable, Unit] =
     for {
       _ <- ZIO.logInfo("=> Startup checks initiated")
       _ <- checkTriplestoreService
@@ -184,7 +189,7 @@ object AppServer {
   val live: ZIO[AppServerEnvironment, Nothing, Unit] =
     for {
       appServer <- AppServer.init()
-      _         <- appServer.start(requiresAdditionalRepositoryChecks = true, requiresIIIFService = true)
+      _         <- appServer.start(requiresAdditionalRepositoryChecks = true, requiresIIIFService = true).orElse(ZIO.unit)
     } yield ()
 
   /**
@@ -194,7 +199,7 @@ object AppServer {
   val testWithSipi: ZIO[AppServerEnvironment, Nothing, Unit] =
     for {
       appServer <- AppServer.init()
-      _         <- appServer.start(requiresAdditionalRepositoryChecks = false, requiresIIIFService = true)
+      _         <- appServer.start(requiresAdditionalRepositoryChecks = false, requiresIIIFService = true).orDie
     } yield ()
 
   /**
@@ -204,6 +209,6 @@ object AppServer {
   val testWithoutSipi: ZIO[AppServerEnvironment, Nothing, Unit] =
     for {
       appServer <- AppServer.init()
-      _         <- appServer.start(requiresAdditionalRepositoryChecks = false, requiresIIIFService = false)
+      _         <- appServer.start(requiresAdditionalRepositoryChecks = false, requiresIIIFService = false).orDie
     } yield ()
 }
--- a/webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cache.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cache.scala
@@ -201,9 +201,10 @@ object Cache extends LazyLogging {
       exception match {
         case inconsistentRepositoryDataException: InconsistentRepositoryDataException =>
           log.error(inconsistentRepositoryDataException.message)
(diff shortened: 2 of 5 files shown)


webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
```diff
@@ -80,10 +80,15 @@ final case class AppServer(
    *
    * @param requiresRepository If `true`, calls the AppRouter to populate the ontology caches, otherwise returns ()
    */
-  private def populateOntologyCaches(requiresRepository: Boolean): UIO[Unit] =
+  private def populateOntologyCaches(requiresRepository: Boolean): IO[Throwable, Unit] =
```
irinaschubert, line 83:
```suggestion
  private def populateOntologyCaches(requiresRepository: Boolean): Task[Unit] =
```

webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
```diff
@@ -132,7 +137,7 @@ final case class AppServer(
   def start(
     requiresAdditionalRepositoryChecks: Boolean,
     requiresIIIFService: Boolean
-  ): UIO[Unit] =
+  ): IO[Throwable, Unit] =
```
irinaschubert, line 140:
```suggestion
  ): Task[Unit] =
```

webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cache.scala
```diff
@@ -201,9 +201,10 @@ object Cache extends LazyLogging {
       exception match {
         case inconsistentRepositoryDataException: InconsistentRepositoryDataException =>
           log.error(inconsistentRepositoryDataException.message)
-          SuccessResponseV2(
-            s"An error occurred when loading ontologies: ${inconsistentRepositoryDataException.message}"
-          )
+          // SuccessResponseV2(
+          //   s"An error occurred when loading ontologies: ${inconsistentRepositoryDataException.message}"
+          // )
```
mpro7, line 206: Could be removed.

irinaschubert, line 206: I guess this can be removed?

webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
```diff
@@ -80,10 +80,15 @@ final case class AppServer(
    *
    * @param requiresRepository If `true`, calls the AppRouter to populate the ontology caches, otherwise returns ()
    */
-  private def populateOntologyCaches(requiresRepository: Boolean): UIO[Unit] =
+  private def populateOntologyCaches(requiresRepository: Boolean): IO[Throwable, Unit] =
     for {
       _ <- state.set(AppState.LoadingOntologies)
-      _ <- ar.populateOntologyCaches.when(requiresRepository)
+      _ <-
+        ar.populateOntologyCaches
```
mpro7, line 87: What's the `ar`? :)

mpro7 (approved):
Looks good.

irinaschubert (commented):
Should we have a discussion about how we want to deal with unhealthy / fatal states tomorrow in the BE meeting? I think it would be worth thinking about it in a general way how we want to handle states.

seakayone (changes requested):
I have many question. The TLDR is:

The dsp-api will not start if Fuseki is not available, wouldn't it be better to fail starting up the dsp-api process if it cannot find the ontologies? 

Is the error state "missing ontologies" an expected error for which the dsp-api should regularly perform a check during a healtch-check at all?

An unavailable database is expected because e.g. network is unreliable. But IMHO it is sane to assume that once dsp-api found a database with ontologies that removing these ontologies is an unexpected error we should not really account for, should we?

BalduinLandolt:
@seakayone you raise some good points, and I actually had some doubts of similar nature. I think we should discuss it in the BE meeting, as @irinaschubert suggested.

I also realized that there is another Health route in the InstrumentalizationHttpServer, which would in principal need updating too. This further suggests to me that we need a more birds eye vision of what the health route should be.

I will apply the small suggestions though

webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
BalduinLandolt: the sound a pirate makes when he approves... obviously! :P  
Seriously, it stands for AppRouter, in this file there are a lot of those overly short variable names, and I got annoyed by it too:

```scala
final case class AppServer(
  state: State,
  ts: TriplestoreService,
  ru: RepositoryUpdater,
  as: ActorSystem,
  ar: AppRouter,
  iiifs: IIIFService,
  cs: CacheService,
  hs: HttpServer,
  appConfig: AppConfig
) 
```

But I considered changing this out of scope for my task at hand.

mpro7: Thanks for explanation argh!

webapi/src/it/scala/org/knora/webapi/responders/v2/LoadOntologiesRequestV2Spec.scala
```diff
@@ -16,6 +16,7 @@ import org.knora.webapi.messages.store.triplestoremessages.ResetRepositoryConten
 import org.knora.webapi.messages.util.KnoraSystemInstances
 import org.knora.webapi.messages.v2.responder.SuccessResponseV2
 import org.knora.webapi.messages.v2.responder.ontologymessages.LoadOntologiesRequestV2
+import akka.actor.Status
```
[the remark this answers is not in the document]
BalduinLandolt: For now, this is what the actor returns (I just made it explicit). I think this is easier tackled when we look into what we do with the akka test kit tests.

webapi/src/main/scala/org/knora/webapi/responders/v2/ontology/Cache.scala
```diff
@@ -201,9 +201,10 @@ object Cache extends LazyLogging {
...
+          throw inconsistentRepositoryDataException
```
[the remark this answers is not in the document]
BalduinLandolt: I'm tempted to say we don't even need to log this here (the exception would propagate out anyways), so we could just not touch the future and return it right away. What do you think?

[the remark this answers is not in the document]
seakayone: Yes, I agree, either log or re-throw an exceptions, but never do both. Otherwise the exception ends up getting logged multiple times in multiple layers.

irinaschubert (approved):
LGTM, I assume you've tested it locally and the stack behaves as expected when there is corrupt data?

BalduinLandolt:
> LGTM, I assume you've tested it locally and the stack behaves as expected when there is corrupt data?

yes :)

seakayone (approved):

mpro7 (approved):

PR #2377, opened 2023-01-10
test: Add tests for ZIO HTTP project routes

irinaschubert:
Co-Authored-By: Balduin Landolt <33053745+BalduinLandolt@users.noreply.github.com>
Co-Authored-By: Marcin Procyk <29251288+mpro7@users.noreply.github.com>

### Task Description/Number

Issue Number: DEV-

### Basic Requirements

### PR Type

- [x] Test: all about tests: adding, refactoring tests (no production code change)

- [ ] Yes
- [x] No

- [x] No


webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsService.scala
```diff
@@ -0,0 +1,71 @@
...
+import org.knora.webapi.messages.admin.responder.usersmessages.UserADM
+import org.knora.webapi.responders.ActorToZioBridge
+
+trait ProjectsService {
+  def getProjectsADMRequest(): Task[ProjectsGetResponseADM]
+  def getSingleProjectADMRequest(identifier: ProjectIdentifierADM): Task[ProjectGetResponseADM]
+  def createProjectADMRequest(
+    payload: ProjectCreatePayloadADM,
+    requestingUser: UserADM
+  ): Task[ProjectOperationResponseADM]
```
seakayone, line 20: Not entirely sure what the reasons for renaming were. IMHO, the signature of the methods suggest that this service should indeed be named `Rest…` service.

Indicators are: 
* The return values are all `*ResponseADM` instances. 
* The`createProjectADMRequest` takes a user as an argument

IMHO, we should aim at separation of concerns and reflect that in the names of our new components.
I.e. `*Route`s handle mostly presentation logic; `Rest*Service`s handle aggregation of business logic calls as well as permissions mapping; `*Service`s  contain business logic; `*Repo`s handle persistence.

Wdyt?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -511,6 +511,11 @@ sealed trait ProjectIdentifierADM { self =>
       case ShortnameIdentifier(value) => Some(value.value)
       case _                          => None
     }
+  def valueAsString: String = self match {
```
seakayone, line 514: What is the point of adding this method?

The exact same implementation is already present as `org.knora.webapi.messages.admin.responder.projectsmessages.ProjectIdentifierADM#getId`

webapi/src/test/scala/org/knora/webapi/responders/admin/ProjectsServiceLiveSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+      _ <- projectsService.provide(bridge)
+    } yield assertTrue(true)
+  }
+
+  val getProjectByIdSpec = suite("get project by identifier")(
+    test("get project by shortcode") {
+      val identifier = ProjectIdentifierADM.ShortcodeIdentifier
+        .fromString("0001")
+        .getOrElse(throw new Exception("invalid shortcode"))
+      val projectsService =
```
seakayone, line 75: This is not the service but rather a zio containing the response of `getSingleProjectADMRequest`, isn't it?

webapi/src/test/scala/org/knora/webapi/responders/admin/ProjectsServiceLiveSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+        ZIO
+          .serviceWithZIO[ProjectsService](_.getSingleProjectADMRequest(identifier))
+          .provideSome[ActorToZioBridge](layers)
+      val bridge = ActorToZioBridgeMock.AskAppActor
```
seakayone, line 79: Would be nice to have speaking name for this mock layer.

webapi/src/test/scala/org/knora/webapi/responders/admin/ProjectsServiceLiveSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+object ProjectsServiceLiveSpec extends ZIOSpecDefault {
+
+  private val expectNoInteraction = ActorToZioBridgeMock.empty
+
+  val layers = ZLayer.makeSome[ActorToZioBridge, ProjectsService](ProjectsService.live)
+
+  /**
+   * Creates a [[ProjectADM]] with empty content or optionally with a given ID.
+   */
+  private def getProjectADM(id: String = "") =
```
seakayone, line 41: The parameter is never used, this could be a `val`, which would make reading the references further down much clearer.

seakayone (changes requested):

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsService.scala
BalduinLandolt: This happened during the mob session, when we were all not sure, what the deal was with the `Rest*` name.

I fully agree that we should aim at separation of concerns and should reflect that in our naming. But for conventions to be meaningful, we should all know them. ;-)  
Also, as a personal matter of taste, I don't like `Rest*Service`. I can absolutely be happy about a prefix or a suffix based naming convention, but I don't like circumfix based naming conventions.

In any case, this sounds like a perfect case for a RFC, right?

But for now, this shouldn't block us from merging. Shall I revert the name change, or leave it as is for now?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
BalduinLandolt: true, overlooked this one. Will remove it

webapi/src/test/scala/org/knora/webapi/responders/admin/ProjectsServiceLiveSpec.scala
BalduinLandolt: right... it was used at some point, but then things changed

webapi/src/test/scala/org/knora/webapi/responders/admin/ProjectsServiceLiveSpec.scala
BalduinLandolt: done

webapi/src/test/scala/org/knora/webapi/responders/admin/ProjectsServiceLiveSpec.scala
BalduinLandolt: yes, technically this is not the ProjectsService itself but a ZIO that, provided an ActorToZioBridge, will call the method of the ProjectsService, that is under test. According to Fowler, that should be called `sut`, but I don't think that's more understandable, just more generic.  
But I'm open to naming suggestions.

seakayone (approved):

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsService.scala
seakayone: `Rest*Service` is my suggestion and it is what we have used so far and what I am currently using in my other PR.

PR #2394, opened 2023-01-19
feat: expose PUT /admin/projects/iri/{project_iri} as ZIO HTTP route (DEV-1587)

irinaschubert:
### Task Description/Number

Issue Number: DEV-1587

### Basic Requirements

### PR Type

- [x] Feature: represents a new feature

- [ ] Yes
- [x] No

- [x] No


webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -838,62 +819,67 @@ final case class ProjectsResponderADM(actorDeps: ActorDeps, cacheServiceSettings
...
 
-      _ = if (projectUpdatePayload.shortname.isDefined) {
-            val unescapedShortName: String = stringFormatter.fromSparqlEncodedString(projectUpdatePayload.shortname.get)
+      _ = projectUpdatePayload.shortname.map { shortname =>
+            val unescapedShortName: String = stringFormatter.fromSparqlEncodedString(shortname.value)
             if (updatedProject.shortname != unescapedShortName)
               throw UpdateNotPerformedException(
                 "Project's 'shortname' was not updated. Please report this as a possible bug."
               )
           }
```
seakayone, line 828: I support your effort in refactoring the code base into a more functional style.
However, when I read `map` I would generally expect a pure function to be used in its code block.
The `throw ...Exception` in the `map` code block is surprising at best. 

How about changing it to something like:
```scala
projectUpdatePayload.shortname
    .map(_.value)
    .map(stringFormatter.fromSparqlEncodedString)
    .filter(_ != updatedProject.shortname)
   .getOrElse(throw UpdateNotPerformedException("Project's 'shortname' was not updated. Please report this as a possible bug.")
```

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsService.scala
```diff
@@ -77,19 +84,48 @@ final case class ProjectsServiceLive(bridge: ActorToZioBridge) extends ProjectsS
...
+   *
+   * @param projectIri           the [[IRI]] of the project
+   * @param payload              a [[ProjectUpdatePayloadADM]] instance
+   * @param requestingUser       the user making the request.
+   * @return
+   *     '''success''': information about the project as a [[ProjectOperationResponseADM]]
+   *
+   *     '''failure''': [[dsp.errors.NotFoundException]] when no project for the given IRI can be found
+   */
+  def changeProject(
```
seakayone, line 114: `ProjectUpdatePayloadADM` vs the method name `changeProject`. 
I think we should aim to keep the naming consistent. If the necessity exists to use different names I would like to limit the inconsistency in naming in the api/rest layer if possible.
wdyt?

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteZ.scala
```diff
@@ -82,11 +86,21 @@ final case class ProjectsRouteZ(
...
       iriDecoded            <- RouteUtilZ.urlDecode(iriUrlEncoded, s"Failed to URL decode IRI parameter $iriUrlEncoded.")
-      projectDeleteResponse <- projectsService.deleteProject(iriDecoded, requestingUser)
+      projectIri            <- Iri.ProjectIri.make(iriDecoded).toZIO
+      projectDeleteResponse <- projectsService.deleteProject(projectIri, requestingUser)
     } yield Response.json(projectDeleteResponse.toJsValue.toString())
+
+  private def changeProject(iriUrlEncoded: String, request: Request, requestingUser: UserADM): Task[Response] =
+    for {
+      iriDecoded            <- RouteUtilZ.urlDecode(iriUrlEncoded, s"Failed to URL decode IRI parameter $iriUrlEncoded.")
+      projectIri            <- Iri.ProjectIri.make(iriDecoded).toZIO
```
seakayone, line 99: The `make` function returns `Validation[ValidationException, ProjectIri]`. Shouldn't this error be mapped to a `BadRequestException` - here and above in the `deleteProject` case?

As far as I can see these cases are not tested, are they?

seakayone (changes requested):

webapi/src/it/scala/org/knora/webapi/responders/admin/ProjectsResponderADMSpec.scala
```diff
@@ -370,24 +370,27 @@ class ProjectsResponderADMSpec extends CoreSpec with ImplicitSender {
...
       "UPDATE a project" in {
+        val iri             = Iri.ProjectIri.make(newProjectIri.get).getOrElse(throw BadRequestException(""))
+        val updatedLongname = Name.make("updated project longname").getOrElse(throw BadRequestException(""))
+        val updatedDescription = ProjectDescription
+          .make(Seq(V2.StringLiteralV2("""updated project description with "quotes" and <html tags>""", Some("en"))))
+          .getOrElse(throw BadRequestException(""))
+        val updatedKeywords = Keywords.make(Seq("updated", "keywords")).getOrElse(throw BadRequestException(""))
+        val updatedLogo     = Logo.make("/fu/bar/baz-updated.jpg").getOrElse(throw BadRequestException(""))
+        val projectStatus   = ProjectStatus.make(true).getOrElse(throw BadRequestException(""))
+        val selfJoin        = ProjectSelfJoin.make(true).getOrElse(throw BadRequestException(""))
```
mpro7, line 381: Should it be empty `throw BadRequestException("")`?

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteADM.scala
```diff
@@ -238,17 +239,34 @@ class ProjectsRouteADM(routeData: KnoraRouteData)
         entity(as[ChangeProjectApiRequestADM]) { apiRequest => requestContext =>
           val checkedProjectIri =
             stringFormatter.validateAndEscapeProjectIri(value, throw BadRequestException(s"Invalid project IRI $value"))
+          val projectIri: ProjectIri =
+            ProjectIri.make(checkedProjectIri).getOrElse(throw BadRequestException(s"Invalid Project IRI"))
```
mpro7, line 243:
```suggestion
            ProjectIri.make(checkedProjectIri).getOrElse(throw BadRequestException(s"Invalid Project IRI."))
```

mpro7 (approved):
Looks OK.

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
irinaschubert: I really like your suggestion and I will apply it. For the simple values it is easy. What about more complex ones like `descriptions` of type `Seq` where you have a sequence that needs to be processed inside a map again. I came up with the code below (see also https://github.com/dasch-swiss/dsp-api/pull/2394/commits/aa17b20c037d6992cbd98e0abf291978fc7184ad) but it seems very hard to understand (because of the map inside the map). Do you maybe have any suggestion to improve it? I tried to model it inside a for comprehension but it doesn't look better to me.

```scala
_ = projectUpdatePayload.description
    .map(_.value)
    .map(_.map(d => V2.StringLiteralV2(stringFormatter.fromSparqlEncodedString(d.value), d.language)))
    .filter(updatedProject.description.diff(_).nonEmpty)
    .getOrElse(
        throw UpdateNotPerformedException(
            "Project's 'description' was not updated. Please report this as a possible bug."
        )
    )
```

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsService.scala
irinaschubert: I totally agree, I planned to only use "update" in the new code but this one slipped through.. I adjusted the method name.

webapi/src/it/scala/org/knora/webapi/responders/admin/ProjectsResponderADMSpec.scala
irinaschubert: The `getOrElse` is only there to get the value out of the `Validation`. This is not supposed to fail in this test. But I agree that if something should fail in the future (because the definition / validation of the value objects changes), it would be more helpful to have meaningful error messages. I added them now.

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteADM.scala
irinaschubert: "Invalid Project IRI" is not a sentence, grammatically. So, I think the dot can be omitted. But I checked other error messages and there doesn't seem to be a consistenccy. So, I just leave it as it is.

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteZ.scala
irinaschubert: Absolutely.
Regarding tests, you are right, until now we don't have negative tests at all for the ported routes. I will add them.

irinaschubert: Aha, I've just realized that the methods that would have to be tested in order to test these unhappy cases are private. The `ProjectsRouteZSpec` tests the `ProjectService` call which is inside these private methods. @seakayone do you still think these cases need to be tested? I'm obviously still unsure about the unit of unit tests..

irinaschubert: I added negative tests for updating projects. I will add them for the other use cases (routes), too. But in a separate PR.

seakayone (approved):

PR #2405, opened 2023-01-26
refactor: Replace Cardinality isStricterThan with isIncludedIn

seakayone:
### Task Description/Number

A cardinality includes another cardinality if and only if the other cardinality is included in its entire range of possible values.

The previous method `isStricterThan` assumes/suggests a commutative property in the sense that when comparing different cardinalities `a` and `b` then if the comparison `a isStricterThan b == false` then `b isStricterThan a == true` must also hold. However, this is not the always the case as for example it is with `ZeroOrOne` and `AtLeastOne`. Both cardinalities are "stricter" than the other as their range of values only has a very small intersection and none contains the other in full. In the future with the introduction of other cardinalities e.g. ExactlyTwo an intersection might not even occur.

### Task Description/Number

Issue Number: DEV-1563

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/slice/ontology/domain/model/Cardinality.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/ontology/domain/model/Cardinality.scala
@@ -18,17 +18,31 @@ sealed trait Cardinality {
   def max: Option[Int]
 
   /**
-   * Checks whether a cardinality is stricter than another one.
+   * Checks whether a cardinality is included in another one.
+   *
+   * A [[Cardinality]] includes another cardinality if and only if the other cardinality
+   * is included in its entire range of possible values.
    *
    * @param other The cardinality to be compared against.
-   * @return `true` if this cardinality is stricter than the `other`, `false` otherwise.
+   * @return `true` if this cardinality is include in the `other`, `false` otherwise.
    */
-  def isStricterThan(other: Cardinality): Boolean =
-    (other.min, this.min, other.max, this.max) match {
-      case (otherMin, thisMin, _, _) if otherMin < thisMin => true
-      case (_, _, otherMax, Some(thisMax))                 => otherMax.forall(_ > thisMax)
-      case _                                               => false
+  def isIncludedIn(other: Cardinality): Boolean = {
+    val lowerBoundIsIncluded = this.min >= other.min && other.max.forall(this.min <= _)
+    lazy val upperBoundIsIncluded = (this.max, other.max) match {
+      case (_, None)                       => true
+      case (None, Some(_))                 => false
+      case (Some(thisMax), Some(otherMax)) => thisMax >= other.min && thisMax <= otherMax
     }
+    lowerBoundIsIncluded && upperBoundIsIncluded
+  }
+
+  /**
+   * Negated check for [[org.knora.webapi.slice.ontology.domain.model.Cardinality#isIncludedIn(org.knora.webapi.slice.ontology.domain.model.Cardinality)]]
+   *
+   * @param other The cardinality to be compared against.
+   * @return `true` if this cardinality is not include in the `other`, `false` otherwise.
+   */
+  def isNotIncludedIn(other: Cardinality): Boolean = !isIncludedIn(other)
 
   def getIntersection(other: Cardinality): Option[Cardinality] = {
     val lower = math.max(this.min, other.min)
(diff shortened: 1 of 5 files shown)


webapi/src/main/scala/org/knora/webapi/slice/ontology/domain/model/Cardinality.scala
```diff
@@ -18,17 +18,31 @@ sealed trait Cardinality {
...
   /**
-   * Checks whether a cardinality is stricter than another one.
+   * Checks whether a cardinality is included in another one.
+   *
+   * A [[Cardinality]] includes another cardinality if and only if the other cardinality
+   * is included in its entire range of possible values.
    *
    * @param other The cardinality to be compared against.
-   * @return `true` if this cardinality is stricter than the `other`, `false` otherwise.
+   * @return `true` if this cardinality is include in the `other`, `false` otherwise.
```
irinaschubert, line 27:
```suggestion
   * @return `true` if this cardinality is included in the `other`, `false` otherwise.
```

webapi/src/test/scala/org/knora/webapi/slice/ontology/domain/model/CardinalitySpec.scala
```diff
@@ -11,13 +11,23 @@ import zio.test._
...
+  object Generator {
+    def cardinalitiesGen(cardinalities: Cardinality*): Gen[Any, Cardinality] = Gen.fromZIO {
+      val candidates: Array[Cardinality] = if (cardinalities != Nil) {
+        cardinalities.toArray
+      } else {
+        allCardinalities
+      }
+      val length = candidates.length
+      if (length == 0) {
+        throw new IllegalArgumentException("The parameter cardinalities may not be an empty Array")
```
mpro7, line 26:
```suggestion
        throw new IllegalArgumentException("The parameter cardinalities may not be an empty Array.")
```

irinaschubert (approved):
LGTM

webapi/src/test/scala/org/knora/webapi/slice/ontology/domain/model/CardinalitySpec.scala
```diff
@@ -11,13 +11,23 @@ import zio.test._
...
+      } else {
+        Random.nextIntBetween(0, length).map(i => candidates(i))
```
mpro7, line 28:
```suggestion
        Random.nextIntBetween(0, length).map(candidates(_))
```

webapi/src/main/scala/org/knora/webapi/slice/ontology/domain/model/Cardinality.scala
```diff
@@ -18,17 +18,31 @@ sealed trait Cardinality {
...
+    lazy val upperBoundIsIncluded = (this.max, other.max) match {
+      case (_, None)                       => true
+      case (None, Some(_))                 => false
+      case (Some(thisMax), Some(otherMax)) => thisMax >= other.min && thisMax <= otherMax
     }
+    lowerBoundIsIncluded && upperBoundIsIncluded
+  }
+
+  /**
+   * Negated check for [[org.knora.webapi.slice.ontology.domain.model.Cardinality#isIncludedIn(org.knora.webapi.slice.ontology.domain.model.Cardinality)]]
```
mpro7, line 40: Does it makes sense to expose such a long class path here?

webapi/src/main/scala/org/knora/webapi/slice/ontology/domain/model/Cardinality.scala
```diff
@@ -18,17 +18,31 @@ sealed trait Cardinality {
...
+   *
+   * @param other The cardinality to be compared against.
+   * @return `true` if this cardinality is not include in the `other`, `false` otherwise.
+   */
+  def isNotIncludedIn(other: Cardinality): Boolean = !isIncludedIn(other)
```
mpro7, line 45: Maybe I don't get the context but, what this brings in instead of just negate the existing one `isIncludedIn`:
`!newCardinality. isIncludedIn`

mpro7 (approved):
Look OK, please check my comments and suggestions.

webapi/src/main/scala/org/knora/webapi/slice/ontology/domain/model/Cardinality.scala
seakayone: Readability, more expressiveness in your code.
Also convenience with autocompletion.

webapi/src/main/scala/org/knora/webapi/slice/ontology/domain/model/Cardinality.scala
seakayone: Maybe I do not get the scaladoc syntax yet.  Do you have a better suggestion in having a linked reference to the other method?

mpro7: I don't know what's the best practice here if links are very long, but it just bothers my eyes ;) Having part of the link as an import would not work right?

PR #2419, opened 2023-02-02
test: Add more tests for the ZIO HTTP routes (DEV-1695)

irinaschubert:
### Task Description/Number

Issue Number: DEV-1695

### Basic Requirements

### PR Type

- [x] Test: all about tests: adding, refactoring tests (no production code change)

- [ ] Yes
- [x] No

- [x] No


seakayone (approved):
LGTM.

However, as a small improvement you might want to consider when setting up the test data for your test which contain:

```scala
           .getOrElse(throw BadRequestException("Invalid …"))
```

I understand that the factory methods for the value objects and also the validation forces us to handle the failure. In the test setup for this data then we will have to do a `.get` or a `.getOrElse` but I think a `throw BadRequestException` is the "wrong" exception to throw in the test, `IllegalArgumentException` would probably be more appropriate.

Also I do think repeating this pattern for each test is making the test setup pretty noisy, as essentially we know the provided shortcode is valid but the test source code still contains all the handling of the failure case which should not occur.

Maybe it makes sense to extract method for the setup of test data like the `ProjectIri` into some share file and throw an `IllegalArgumentException` instead?

E.g.:

```scala
object TestDataFactory {
  def projectShortcode(shortcode: String): ShortcodeIdentifier =
    ProjectIdentifierADM.ShortcodeIdentifier
        .fromString(shortcode)
        .getOrElse(throw new IllegalArgumentException(s"Invalid value ${shortcode}"))
}
```

In the test it is then also easier to use, because creating e.g. a shortcode is a single line only:
```scala
val shortcode: ShortcodeIdentifier = projectShortcode("3333")
```

wdyt?

webapi/src/test/scala/org/knora/webapi/routing/admin/ProjectsRouteZSpec.scala
```diff
@@ -159,12 +118,23 @@ object ProjectsRouteZSpec extends ZIOSpecDefault {
...
+
+      for {
+        response     <- applyRoutes(request).provide(ProjectsServiceMock.empty)
+        bodyAsString <- response.body.asString
+      } yield assertTrue(response.status == Status.BadRequest) &&
+        assertTrue(bodyAsString == """{"error":"dsp.errors.BadRequestException: Project IRI is invalid."}""")
+    },
+    test("get a project by shortname") {
+      val shortname = "someProject"
+      val identifier: ProjectIdentifierADM = ProjectIdentifierADM.ShortnameIdentifier
```
mpro7, line 134: IMO either explicit type or class is enough. I've seen it in few places.

irinaschubert:
> LGTM.
> 
> However, as a small improvement you might want to consider when setting up the test data for the tests containing something like:
> 
> ```scala
>            .getOrElse(throw BadRequestException("Invalid …"))
> ```
> 
> I understand that the factory methods for the value objects and also the validation forces us to handle the failure. In the test setup for this data then we will have to do a `.get` or a `.getOrElse` but I think a `throw BadRequestException` is the "wrong" exception to throw in the test, `IllegalArgumentException` would probably be more appropriate.
> 
> Also I do think repeating this pattern for each test is making the test setup pretty noisy, as essentially we know the provided shortcode is valid but the test source code still contains all the handling of the failure case which should not occur.
> 
> Maybe it makes sense to extract method for the setup of test data like the `ProjectIri` into some share file and throw an `IllegalArgumentException` instead?
> 
> E.g.:
> 
> ```scala
> object TestDataFactory {
>   def projectShortcode(shortcode: String): ShortcodeIdentifier =
>     ProjectIdentifierADM.ShortcodeIdentifier
>         .fromString(shortcode)
>         .getOrElse(throw new IllegalArgumentException(s"Invalid value ${shortcode}"))
> }
> ```
> 
> In the test it is then also easier to use, because creating e.g. a shortcode is a single line only:
> 
> ```scala
> val shortcode: ShortcodeIdentifier = projectShortcode("3333")
> ```
> 
> wdyt?

I totally agree and think that a factory method is a very good idea. I will try to implement it.

mpro7 (approved):
Just one trivial remark.

PR #2426, opened 2023-02-03
fix: fix JVM metrics and logging DEV-1639

seakayone:
### Task Description/Number

closes: DEV-1639

* Simplify setup of  `Main` and `nstrumentationServer` - I am not entirely sure what the problem was. However with this simplification at least locally I do get JVM metrics reliably.
* Fix logging to always use slf4j by horizontally composing the log layers before starting up the `DspEnvironmentLive`.

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
+++ b/webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
@@ -181,7 +181,7 @@ object AppServer {
   /**
    * The live AppServer
    */
-  val live: ZIO[AppServerEnvironment, Nothing, Unit] =
+  val make: ZIO[AppServerEnvironment, Nothing, Unit] =
     for {
       appServer <- AppServer.init()
       _         <- appServer.start(requiresAdditionalRepositoryChecks = true, requiresIIIFService = true).orDie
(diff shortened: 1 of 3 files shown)


webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
```diff
@@ -181,7 +181,7 @@ object AppServer {
   /**
    * The live AppServer
    */
-  val live: ZIO[AppServerEnvironment, Nothing, Unit] =
+  val make: ZIO[AppServerEnvironment, Nothing, Unit] =
```
irinaschubert, line 184: What was the reason to change this name?

irinaschubert (approved):
I don't fully understand the whole instrumentation server but to me it looks good.

webapi/src/main/scala/org/knora/webapi/core/AppServer.scala
seakayone: Often we used `live` as something which returns a `ZLayer`, which is not the case here as this produces a `ZIO`. 

The renaming also aligns with the same in `InstrumentationServer` so that the startup is 
`InstrumentationServer.make *> AppServer.make *> ZIO.never`
rather than the more chaotic
`InstrumentationServer.make *> AppServer.live *> ZIO.never`.

irinaschubert: makes sense

BalduinLandolt (approved):
LGTM, eventually we might really consider moving the `dspLAyersLive` out of the bootstrap. But for now, cool that it works

mpro7 (approved):

PR #2438, opened 2023-02-10
chore: cleanup remove unused shacl and redundant StringFormatter setup

seakayone:
### Task Description/Number

* Remove unused shacl implementations 
* Remove redundant StringFormatter initialization
* Rename `AppConfig.layer` as only a single `AppConfig` is loaded anyway
* Remove unused `testing` boolean flag from `AppConfig` and `application.conf`

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No


mpro7 (approved):

irinaschubert (approved):
LGTM, I don't remember why we had a separate `test` layer for the `AppConfig` but I assume you've tested that it still works and the `test.conf` is loaded when running the tests. If this works, than this is a nice simplification.

seakayone:
> LGTM, I don't remember why we had a separate `test` layer for the `AppConfig` but I assume you've tested that it still works and the `test.conf` is loaded when running the tests. If this works, than this is a nice simplification.

It works and that is due to a feature of the underlying typelevel config. It will load the `application.conf` from the `main` source set and if in the `test` or `it` it finds another `application.conf` it will merge both overwriting any value from `main`. Since the `test.conf` is included in the `application.conf` from the `it` source set its values will also have precedence and hence overwrite existing values from the `main`. For more details please refer to the documentation: https://github.com/lightbend/config#standard-behavior

The config is designed this way so that library authors can even provide an `application.conf` with default values in their jars so that the user of these libraries are then able to provide their own values with their own conf files.

PR #2440, opened 2023-02-13
chore: Rename TestDataFactory -> ITTestDataFactory

seakayone:
In IntelliJ IDEA the compile does not separate the source sets of test and it so it would fail with a duplicate class being present.

### Task Description/Number

In IntelliJ IDEA the compile does not separate the source sets of test and it so it would fail with a duplicate class being present.
### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No

### DIFF
--- a/webapi/src/it/scala/org/knora/webapi/ITTestDataFactory.scala
+++ b/webapi/src/it/scala/org/knora/webapi/ITTestDataFactory.scala
@@ -8,7 +8,7 @@ import org.knora.webapi.messages.admin.responder.projectsmessages.ProjectIdentif
 /**
  * Helps in creating value objects for tests.
  */
-object TestDataFactory {
+object ITTestDataFactory {
   def projectShortcodeIdentifier(shortcode: String): ShortcodeIdentifier =
     ShortcodeIdentifier
       .fromString(shortcode)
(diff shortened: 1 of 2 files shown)


webapi/src/it/scala/org/knora/webapi/ITTestDataFactory.scala
```diff
@@ -8,7 +8,7 @@ import org.knora.webapi.messages.admin.responder.projectsmessages.ProjectIdentif
 /**
  * Helps in creating value objects for tests.
  */
-object TestDataFactory {
+object ITTestDataFactory {
```
irinaschubert, line 11:
```suggestion
object ItTestDataFactory {
```

irinaschubert: We once (before you were here) agreed on camelCasing acronyms / abbreviations, too.

irinaschubert (approved):
LGTM just a minor remark on the usage of camel case.

mpro7 (approved):

mpro7:
Please reflect both names in the PR title, just to make it easier to find if anyone needs to in the future.

webapi/src/it/scala/org/knora/webapi/ITTestDataFactory.scala
seakayone: We are apparently not very consistent with this decision as there are all the `ADM` responders.
There is also an `ITKnoraLiveSpec` class.

If this is bothering us I suggest to create a followup issue and rename everything to this standard.
Since I have already a couple of branches open with this exact change in there it would lead to conflicts I am not ready to resolve just yet. 

In future I will take care of this naming convention once it is documented in some coding guidelines.

PR #2441, opened 2023-02-13
fix!: Allow special characters in full-text search (DEV-1712)

irinaschubert:
### Task Description/Number

Issue Number: DEV-1712, marked as breaking change because the search behaviour is quite different than it was before.

### Basic Requirements

### PR Type

- [x] Yes
- [ ] No

- [x] No


docs/03-endpoints/api-v2/reading-and-searching-resources.md
```diff
@@ -501,24 +498,30 @@ The response to a count query request is an object with one predicate,
...
 
-Please note that the search
-terms have to be URL-encoded.
+The search index used by DSP transforms all text into lower case characters and splits text into tokens by whitespace. 
+For example, if a text value is: `The cake needs flour, sugar, and butter.`, 
+the tokens are `the`, `cake`, `needs`, `flour,`, `sugar,`, `and`, `butter.`.
+Note that punctuation marks like `,` and `.` are left with the word where they occurred.
+Therefore, if you search for `sugar` you would have to use `sugar*` or `sugar?` to get results that contain `sugar,` or `sugar.` as well.
+The reason for this kind of tokenization is that some users need to be able to search explicitly for special characters including
+punctuation marks.
```
BalduinLandolt, line 513: I know you have thought about this a lot, and I haven't, so correct or ignore me, if what I'm saying is stupid or irrelevant... But isn't it strange that a search index requires an exact match on the entire token? shouldn't `beet` not match both `beetle` and `beetroot`, rather than just matching german `Beet`?

BalduinLandolt (approved):
LGTM

docs/03-endpoints/api-v2/reading-and-searching-resources.md
irinaschubert: That's something RDU decided. Some projects needed the possibility to search for tokens with punctuation marks in it. And you can't have both. Either an exact match or one that eliminates those special characters.. It's always a trade-off. And it is impossible to satisfy all users here as long as we can't have a separate search index per project - but this is out of scope here :)

BalduinLandolt: definitely out of scope here - but it seems very weird to me (at least for a free text search - in structured search that would be different)... but yea, that's "nicht Problem meiner Stufe" :)

seakayone (approved):

PR #2452, opened 2023-02-15
chore(instrumentation): expose ZIO-HTTP metrics (DEV-1714)

BalduinLandolt:
### Task Description/Number

Issue Number: DEV-1417

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/core/HttpServerZ.scala
+++ b/webapi/src/main/scala/org/knora/webapi/core/HttpServerZ.scala
@@ -13,6 +13,7 @@ import zio.http.model.Method
 import org.knora.webapi.config.AppConfig
 import org.knora.webapi.routing.admin.ProjectsRouteZ
 import org.knora.webapi.slice.resourceinfo.api.ResourceInfoRoute
+import scala.annotation.tailrec
 
 object HttpServerZ {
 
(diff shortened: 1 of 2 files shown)


webapi/src/main/scala/org/knora/webapi/core/HttpServerZ.scala
```diff
@@ -30,19 +32,45 @@ object HttpServerZ {
         )
     } yield Middleware.cors(corsConfig)
 
+  private def metricsMiddleware() = {
+    // in order to avoid extensive amounts of labels, we should use the `pathLabelMapper`
+    // which maps paths containing slug values (`/project/iri/SOME-IRI`)
+    // to paths with a slug inserted instead (`/projects/iri/:iri`).
```
mpro7, line 38: Maybe the link to docs you added above would be enough?

mpro7 (approved):

webapi/src/main/scala/org/knora/webapi/core/HttpServerZ.scala
```diff
@@ -30,19 +32,45 @@ object HttpServerZ {
...
+    val slugs = Set(
+      "iri",
+      "shortcode",
+      "shortname"
+    )
+    @tailrec
+    def step(seg: List[Path.Segment], acc: Path): Path =
```
irinaschubert, line 45: The name of this function is very genetic, could you name it something like `replaceProjectSpecificPartsOfRoute`?

irinaschubert (approved):
LGTM, see my comment about naming

webapi/src/main/scala/org/knora/webapi/core/HttpServerZ.scala
BalduinLandolt: I don't think `projectSpecific` should be in there, because the same will go for all other entities (Users, Resources, etc.). Also, I want to have it short, so the recursive calls don't look too bad (and it's an internal function so the visibility is small enough to warrant a short name. How do you feel about `replaceSlugs()`?

irinaschubert: Yes, that makes sense.`replaceSlugs()` sounds good to me.

seakayone (approved):

SamuelBoerlin (approved):
I can't comment on the ZIO code, but the suggested metrics are looking good to me 👍

PR #2457, opened 2023-02-17
feat: Add new upload route to Sipi without processing (DEV-1700)

irinaschubert:
### Task Description/Number

Issue Number: DEV-1700

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No


webapi/src/main/scala/org/knora/webapi/messages/v2/routing/authenticationmessages/AuthenticationMessagesV2.scala
```diff
@@ -97,3 +101,8 @@ trait AuthenticationV2JsonProtocol extends DefaultJsonProtocol with NullOptions
     jsonFormat(LoginApiRequestPayloadV2, "iri", "email", "username", "password")
   implicit val SessionResponseFormat: RootJsonFormat[LoginResponse] = jsonFormat1(LoginResponse.apply)
 }
+
+trait AuthenticationV2Serialization {
+  implicit lazy val encLoginResponse: JsonEncoder[LoginResponse] = DeriveJsonEncoder.gen[LoginResponse]
+  implicit lazy val decLoginResponse: JsonDecoder[LoginResponse] = DeriveJsonDecoder.gen[LoginResponse]
+}
```
BalduinLandolt, line 108: is this common to have the codec in the trait rather than in the companion object? In any case, as a codec, this could be derived as one, rather than deriving encoder and decoder separately.

BalduinLandolt (approved):
LGTM, not sure though if I follow everything that happens in the lua script.

One thing you may consider is adding more log statements there, as we found it's very hard to see what's happening in SIPI. But if you think this suffices, then I'm absolutely happy with it. (Also, you could use the log util that adds the date and time in front of the log statements, but again, doesn't really matter that much.)

sipi/config/sipi.docker-config.lua
```diff
@@ -195,6 +195,11 @@ routes = {
         method = 'DELETE',
         route = '/delete_temp_file',
         script = 'delete_temp_file.lua'
+    },
+    {
+        method = 'POST',
+        route = '/upload_without_transcoding',
```
jnussbaum, line 201: In DSP-TOOLS, I always refer to this as "preprocessing", rather than "transcoding". For me, "preprocessing" is more telling than "transcoding", but perhaps this is just me.

irinaschubert: To me, transcoding doesn't sound natural, either. But it comes from Ivan. @subotic , do you have a specific reason or strong opinion about this? I wouldn't call it "preprocessing", because it is not prior to something. But just "processing".

webapi/src/main/scala/org/knora/webapi/messages/v2/routing/authenticationmessages/AuthenticationMessagesV2.scala
irinaschubert: This comes from Ivan. @subotic , do you have any reason why you put it into the trait instead of the companion objects? So far, we have all other codecs in the companion objects.

I replaced the encoder/decoder with a single codec.

sipi/config/sipi.docker-config.lua
subotic: Nope, you can call it however you like. Preprocessing / processing both sounds fine to me. I would lean towards `processing` just because it is shorter.

subotic (approved):
LGTM, thanks 👍

PR #2471, opened 2023-02-22
feat: Add context for can set cardinalities failed responses (DEV-1768)

seakayone:
### Task Description/Number

Issue Number: DEV-1768

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No


BalduinLandolt (changes requested):
I don't really have the overview over the cardinality stuff, to be honest. On first glance, most of it looks good.

One thing needs changing though: The new property you define has to be added to knora-base too (and if you have added properties previously, those would need adding too).  
(And when changing the knora-base ontology, you also need to bump the knora-base version, and the DSP-API major version [i.e. breaking change], and you have to run a no-op plugin to ensure that all servers load the changed ontology)

build.sbt
```diff
@@ -79,6 +79,10 @@ lazy val root: Project = Project(id = "root", file("."))
   )
 
 addCommandAlias("fmt", "; all root/scalafmtSbt root/scalafmtAll; root/scalafixAll")
+addCommandAlias(
+  "headerCreateAll",
+  "; all webapi/headerCreate webapi/Test/headerCreate webapi/IntegrationTest/headerCreate"
+)
```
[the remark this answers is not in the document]
seakayone: This allows us to call `sbt headerCreateAll` and in the webapi project the copyright headers are added where missing.

webapi/src/test/scala/org/knora/webapi/slice/ontology/api/service/RestCardinalityServiceLiveSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+                   )
+                 )
+            response     <- RestCardinalityService.canSetCardinality(classIri, propertyIri, "1", userWithAccess)
+            responseJson <- renderResponseJson(response)
+          } yield assertTrue(
+            responseJson ==
+              """
+                |{
+                |  "knora-api:canDo": false,
+                |  "knora-api:cannotDoReason": "The new cardinality is not included in the cardinality of a super-class. Please fix super-classes first: http://0.0.0.0:3333/ontology/0801/biblio/v2#somePublicationInstance,http://0.0.0.0:3333/ontology/0801/biblio/v2#someArticleInstance. The new cardinality does not include the cardinality of a subclass. Please fix subclasses first: http://0.0.0.0:3333/ontology/0801/biblio/v2#someJournalArticleInstance.",
```
irinaschubert, line 81: I think this is a bit hard to read. Is the FE going to format it or is it handing it over to the end-user 1:1? In case of the latter, it might help to separate the different reasons with a line break.

irinaschubert (approved):
LGTM, just a minor suggestion

webapi/src/main/scala/org/knora/webapi/messages/OntologyConstants.scala
```diff
@@ -676,6 +676,7 @@ object OntologyConstants {
     val Error: IRI              = KnoraApiV2PrefixExpansion + "error"
     val CanDo: IRI              = KnoraApiV2PrefixExpansion + "canDo"
     val CannotDoReason: IRI     = KnoraApiV2PrefixExpansion + "cannotDoReason"
+    val CannotDoContext: IRI    = KnoraApiV2PrefixExpansion + "cannotDoContext"
```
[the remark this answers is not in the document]
seakayone: I have spoken also with @irinaschubert and it turns out that none of the existing other `knora-api:canDo…` properties we ever defined in an ontology. I was also not able to find other ontology models for api responses which do not contain data from the triplestore.

In the context of this PR I will resolve this conversation but we will have to discuss how to approach this situation in the future, possibly with a new issue.

webapi/src/test/scala/org/knora/webapi/slice/ontology/api/service/RestCardinalityServiceLiveSpec.scala
seakayone: The `cannotDoReason` was meant to be human readable single string. 
The `cannotDoContext` contains the details about the individual reasons in machine-readable form. 

I cannot add a linebreak in the multiline `String` of this test, because then the `assert` will fail as it expects a line break character which is not in the response.

irinaschubert: I thought adding it not only to the test but also to the response if there are multiple reasons. But this is probably not possible?

mpro7 (approved):

domsteinbach:
For the FE having the context is obviously great. LGTM.

domsteinbach (approved):

webapi/src/test/scala/org/knora/webapi/slice/ontology/api/service/RestCardinalityServiceLiveSpec.scala
seakayone: The `cannotDoReason` is a single String only, changing it would mean breaking the API. Hence, I have introduced the `cannotDoContext` which contains the individual reasons in machine-readable form.

PR #2530, opened 2023-03-15
fix: deleted properties in cardinalities should be included in the count query (DEV-1878)

seakayone:
### Task Description/Number

Issue Number: DEV-1878

Adds a test for the case that the object of a property was deleted.
Given a deleted property the count query should count instances with deleted properties with a count of zero instead of removing the instance from the result set entirely.

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No


mpro7 (approved):

BalduinLandolt (approved):
LGTM, I assumed the implications have been discussed in the project group? (If I understand correctly, this change makes it possible to change a cardinality which is used by a deleted resource. This means that un-deleting the resource no longer necessarily is possible because the resource might violate cardinality constraints. Hence we need to be sure that this is how we want our system to behave; and it also requires a cardinality check on un-deleting resources, which I don't know if it's already in place.)

PR #2559, opened 2023-03-30
refactor: ziofy ValuesRouteV1 (DEV-2012)

seakayone:
### Task Description/Number

Issue Number: DEV-2012

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No


webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -577,7 +578,7 @@ class StringFormatter private (
   initForTest: Boolean = false
 ) {
 
-  import StringFormatter._
+  import org.knora.webapi.messages.StringFormatter._
```
irinaschubert, line 581: Could that be moved to the beginning of the file as Codacy reports?

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilV1.scala
```diff
@@ -127,17 +130,14 @@ object RouteUtilV1 {
...
-        textWithStandoffTagV1 <-
-          ZIO.attempt(
-            StandoffTagUtilV2.convertXMLtoStandoffTagV2(xml, mappingResponse, acceptStandoffLinksToClientIDs, log)
-          )
-      } yield textWithStandoffTagV1
-    }
+  ): ZIO[MessageRelay, Throwable, TextWithStandoffTagsV2] =
+    for {
+      mappingResponse <- MessageRelay.ask[GetMappingResponseV2](GetMappingRequestV2(mappingIri, userProfile))
+      textWithStandoffTagV1 <-
```
irinaschubert, line 136:
```suggestion
      textWithStandoffTagV2 <-
```

irinaschubert: Not sure if V1 is on purpose or not.. (not that it matters a lot as it is only a variable name)

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilV1.scala
```diff
@@ -273,4 +273,34 @@ object RouteUtilV1 {
     } else {
       throw BadRequestException(s"MIME type ${fileMetadataResponse.internalMimeType} not supported in Knora API v1")
     }
+
+  def verifyNumberOfParams[A](
```
irinaschubert, line 277: Where are this and the following functions coming from..?

irinaschubert (approved):
I couldn't properly review the changes in `ValuesRouteV1` as there were that many changes. But I guess it is OK. There are a lot of Codacy complaints about missing tests and bad code (e.g. get() on an option) but I guess it's out of scope here..

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilV1.scala
seakayone: I have extracted duplicate code in the Route itself in order to make it easily reusable within the other routes.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
seakayone: done during the merge with main.

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilV1.scala
```diff
@@ -273,4 +274,32 @@ object RouteUtilV1 {
...
+    params: Seq[A],
+    errorMessage: String,
+    length: Int
+  ): IO[BadRequestException, Option[Nothing]] =
+    ZIO.fail(BadRequestException(errorMessage)).when(params.length != length)
+
+  def toSparqlEncodedString(str: String, errorMsg: String): ZIO[StringFormatter, BadRequestException, IRI] =
```
mpro7, line 285: If this is here defined, is the function named the same in SF ` def toSparqlEncodedString(s: String): Option[String]` used at all?

mpro7 (approved):
It's really hard to read diff of `ValuesRouteV1.scala`, but at the end the outcome looks way cleaner. Thanks! Below few remarks:

seakayone:
> I couldn't properly review the changes in `ValuesRouteV1` as there were that many changes. But I guess it is OK. There are a lot of Codacy complaints about missing tests and bad code (e.g. get() on an option) but I guess it's out of scope here..

The complaints about the `.get` on an Option value were present before. Probably we have to fail with a `BadRequest` in these cases as well. In those cases I guess the assumption is that the value must be filled as part of that particular request. Before the user would see a 500 in those cases. It would be an improvement if this 

wdyt, @irinaschubert ?

irinaschubert:
> > I couldn't properly review the changes in `ValuesRouteV1` as there were that many changes. But I guess it is OK. There are a lot of Codacy complaints about missing tests and bad code (e.g. get() on an option) but I guess it's out of scope here..
> 
> The complaints about the `.get` on an Option value were present before. Probably we have to fail with a `BadRequest` in these cases as well. In those cases I guess the assumption is that the value must be filled as part of that particular request. Before the user would see a 500 in those cases. wdyt, @irinaschubert ?

Yes, a Bad Request sounds reasonable in that case.

seakayone:
> > > I couldn't properly review the changes in `ValuesRouteV1` as there were that many changes. But I guess it is OK. There are a lot of Codacy complaints about missing tests and bad code (e.g. get() on an option) but I guess it's out of scope here..
> > 
> > 
> > The complaints about the `.get` on an Option value were present before. Probably we have to fail with a `BadRequest` in these cases as well. In those cases I guess the assumption is that the value must be filled as part of that particular request. Before the user would see a 500 in those cases. wdyt, @irinaschubert ?
> 
> Yes, a Bad Request sounds reasonable in that case.

see PR https://github.com/dasch-swiss/dsp-api/pull/2561

PR #2585, opened 2023-04-12
Dev 1996 remove errorfun from iri related methods

mpro7:
### Task Description/Number

Issue Number: DEV-

### Basic Requirements

### PR Type

- [ ] Yes
- [ ] No

- [ ] No

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
@@ -2134,11 +2134,14 @@ class StringFormatter private (
    *                 username.
    * @return the same string.
    */
+  @deprecated("Use validateUsername(String) instead.")
   def validateUsername(value: String, errorFun: => Nothing): String = // V2 / value objects
-    UsernameRegex.findFirstIn(value) match {
-      case Some(username) => username
-      case None           => errorFun
-    }
+    validateUsername(value).getOrElse(errorFun)
+
+  def validateUsername(value: String): Validation[Throwable, String] =
+    Validation
+      .fromOption(UsernameRegex.findFirstIn(value))
+      .mapError(_ => ValidationException(s"The username '${value}' contains invalid characters."))
 
   /**
    * Check that the string represents a valid username and escape any special characters.
@@ -2148,11 +2151,12 @@ class StringFormatter private (
    *                 username.
    * @return the same string with escaped special characters.
    */
+  @deprecated("Use validateAndEscapeUsername(String) instead.")
   def validateAndEscapeUsername(value: String, errorFun: => Nothing): String = // V2 / value objects
-    UsernameRegex.findFirstIn(value) match {
-      case Some(username) => toSparqlEncodedString(username, errorFun)
-      case None           => errorFun
-    }
+    validateAndEscapeUsername(value).getOrElse(errorFun)
+
+  def validateAndEscapeUsername(value: String): Option[String] =
+    UsernameRegex.findFirstIn(value).flatMap(toSparqlEncodedString)
 
   /**
    * Check that an optional string represents a valid username.
@@ -2162,14 +2166,15 @@ class StringFormatter private (
    *                    username.
    * @return the same optional string.
    */
+  @deprecated("Use validateAndEscapeOptionalUsername(Option[String]) instead.")
   def validateAndEscapeOptionalUsername(
     maybeString: Option[String],
     errorFun: => Nothing
   ): Option[String] = // V2 / value objects
-    maybeString match {
-      case Some(s) => Some(validateAndEscapeUsername(s, errorFun))
-      case None    => None
-    }
+    maybeString.map(validateAndEscapeUsername(_).getOrElse(errorFun))
+
+  def validateAndEscapeOptionalUsername(maybeString: Option[String]): Option[String] =
+    maybeString.flatMap(validateAndEscapeUsername)
 
   /**
    * Generates an ARK URL for a resource or value, as per [[https://tools.ietf.org/html/draft-kunze-ark-18]].
--- a/webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
@@ -2337,16 +2337,18 @@ class StringFormatter private (
    * @param errorFun   a function that throws an exception. It will be called if the string cannot be parsed.
    * @return the decoded UUID.
    */
-  def validateBase64EncodedUuid(base64Uuid: String, errorFun: => Nothing): UUID = { // V2 / value objects
-    val decodeTry = Try {
-      base64DecodeUuid(base64Uuid)
-    }
+  @deprecated("Use validateBase64EncodedUuid(String) instead.")
+  def validateBase64EncodedUuid(base64Uuid: String, errorFun: => Nothing): UUID = // V2 / value objects
+    validateBase64EncodedUuid(base64Uuid).getOrElse(errorFun)
 
-    decodeTry match {
-      case Success(uuid) => uuid
-      case Failure(_)    => errorFun
-    }
-  }
+  def validateBase64EncodedUuid(base64Uuid: String): Validation[Throwable, UUID] =
+    Validation.fromTry(
+      Try(base64DecodeUuid(base64Uuid))
+        .fold(
+          _ => Failure(ValidationException(s"String: '$base64Uuid' is not a valid base 64 UUID.")),
+          uuid => Success(uuid)
+        )
+    )
 
   /**
    * Encodes a [[UUID]] as a string in one of two formats:
@@ -2384,7 +2386,7 @@ class StringFormatter private (
    * @param errorFun a function that throws an exception. It will be called if the string cannot be parsed.
    * @return the decoded [[UUID]].
    */
-  @deprecated("Use decodeUuidWithErr(String) instead")
+  @deprecated("Use decodeUuidWithErr(String) instead.")
   def decodeUuidWithErr(uuidStr: String, errorFun: => Nothing): UUID = // V2 / value objects
     decodeUuidWithErr(uuidStr).getOrElse(errorFun)
 
--- a/webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
@@ -2384,15 +2384,19 @@ class StringFormatter private (
    * @param errorFun a function that throws an exception. It will be called if the string cannot be parsed.
    * @return the decoded [[UUID]].
    */
+  @deprecated("Use decodeUuidWithErr(String) instead")
   def decodeUuidWithErr(uuidStr: String, errorFun: => Nothing): UUID = // V2 / value objects
+    decodeUuidWithErr(uuidStr).getOrElse(errorFun)
+
+  def decodeUuidWithErr(uuidStr: String): Validation[Throwable, UUID] = // V2 / value objects
     if (uuidStr.length == CanonicalUuidLength) {
-      UUID.fromString(uuidStr)
+      Validation.succeed(UUID.fromString(uuidStr))
     } else if (uuidStr.length == Base64UuidLength) {
-      base64DecodeUuid(uuidStr)
+      Validation.succeed(base64DecodeUuid(uuidStr))
     } else if (uuidStr.length < Base64UuidLength) {
-      base64DecodeUuid(uuidStr.reverse.padTo(Base64UuidLength, '0').reverse)
+      Validation.succeed(base64DecodeUuid(uuidStr.reverse.padTo(Base64UuidLength, '0').reverse))
     } else {
-      errorFun
+      Validation.fail(InconsistentRepositoryDataException(s"Invalid UUID: $uuidStr"))
     }
 
   /**


webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -1822,24 +1823,32 @@ class StringFormatter private (
    *                  valid for a Knora XML import namespace.
    * @return the corresponding project-specific internal ontology IRI.
    */
+  @deprecated("Use xmlImportNamespaceToInternalOntologyIriV1(String) instead.")
   def xmlImportNamespaceToInternalOntologyIriV1(namespace: String, errorFun: => Nothing): SmartIri =
+    xmlImportNamespaceToInternalOntologyIriV1(namespace).getOrElse(errorFun)
+
+  def xmlImportNamespaceToInternalOntologyIriV1(namespace: String): Option[SmartIri] =
```
BalduinLandolt, line 1830: I think I would prefer it to use Validation rather than Option in such cases: here `None` represents an "illegal" state, but in many places we have `None` represent a "legal" state

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -1946,11 +1959,14 @@ class StringFormatter private (
    * @return the same string but escaped.
    */
   def validateAndEscapeProjectIri(iri: IRI, errorFun: => Nothing): IRI = // V2 / value objects
```
BalduinLandolt, line 1961: should be deprecated?

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -1960,6 +1976,7 @@ class StringFormatter private (
    *                    project IRI.
    * @return the same optional string but escaped.
    */
+  @deprecated("Use validateAndEscapeOptionalProjectIri(Option[String]) instead.")
   def validateAndEscapeOptionalProjectIri(
```
BalduinLandolt, line 1980: could use the non-deprecated implementation internally

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -1998,6 +2024,7 @@ class StringFormatter private (
    *                    project shortname.
    * @return the same optional string.
    */
+  @deprecated("Use validateAndEscapeOptionalProjectShortname(Option[String]) instead.")
   def validateAndEscapeOptionalProjectShortname(
```
BalduinLandolt, line 2028: could use the new implementation internally

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2007,49 +2034,63 @@ class StringFormatter private (
...
    */
+  @deprecated("Use def validateProjectShortcode(String) instead.")
   def validateProjectShortcode(shortcode: String, errorFun: => Nothing): String = // V2 / value objects
-    ProjectIDRegex.findFirstIn(shortcode.toUpperCase) match {
-      case Some(value) => value
-      case None        => errorFun
-    }
+    validateProjectShortcode(shortcode).getOrElse(errorFun)
+
+  def validateProjectShortcode(shortcode: String): Option[String] =
```
BalduinLandolt, line 2055: None represents an illegal state, so maybe validation would be better here too?

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -1946,11 +1959,14 @@ class StringFormatter private (
...
-      errorFun
-    }
+    validateAndEscapeProjectIri(iri).getOrElse(errorFun)
+
+  def validateAndEscapeProjectIri(iri: IRI): Validation[ValidationException, IRI] =
+    if (isKnoraProjectIriStr(iri))
+      Validation
+        .fromOption(toSparqlEncodedString(iri))
+        .mapError(_ => ValidationException(s"Invalid IRI: $iri"))
+    else Validation.fail(ValidationException(s"Invalid IRI: $iri"))
```
seakayone, line 1969: IMHO it would be totally ok to simply return an `Option` for all the functions which have taken a `errorFun: => Nothing`. This is simply because the function throwing the exception did not provide any additional information about the cause of the error case to the `errorFun`.

However, if now the method returns a `Validation` in I would expect the reason being precise. In the example of the above function in both cases the same `ValidationException` with the exact same message is returned for the three different cases of why the `iri` could not be 1. validated, 2. escaped/encoded or 3. is not a project iri.

Also on the caller side currently we do not use the `ValidationException` in those cases. So even if we provide this information it is ignored.

The same applies to the other methods you have added with a `Validation[E,A]` return type instead of using an `Option[A]`.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2007,49 +2034,63 @@ class StringFormatter private (
...
     // TODO: I leave this for now to avoid merge conflicts. Should be moved to the ValuesValidator as soon as possible. (depends on toSparqlEncodedString())
-    maybeString match {
-      case Some(s) =>
-        Some(toSparqlEncodedString(s, errorFun))
-      case None => None
-    }
+    maybeString.map(toSparqlEncodedString).getOrElse(errorFun)
+
+  def escapeOptionalString(maybeString: Option[String]): Option[String] =
+    maybeString.flatMap(toSparqlEncodedString)
```
BalduinLandolt, line 2064: this doesn't look right: here, in the response None is an illegal state (so should be Validation IMO), but in the input it's legal, so when passing in None, it would wrongly fail, I think.  
I think this should return `Validation[ValidationException. Option[String]]` to make it clear what's legal and what not

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -1969,6 +1986,14 @@ class StringFormatter private (
       case None    => None
     }
 
+  def validateAndEscapeOptionalProjectIri(
+    maybeString: Option[String]
```
seakayone, line 1990: As above with the `validateAndEscapeOptionalIri` I do think it is worth to consider removing these methods which operate on an `Option[String]` for the same reasoning.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2103,11 +2153,12 @@ class StringFormatter private (
...
+  @deprecated("Use validateEmailAndThrow(String) instead.")
   def validateEmailAndThrow(email: String, errorFun: => Nothing): String = // V2 / value objects
-    EmailAddressRegex.findFirstIn(email) match {
-      case Some(value) => value
-      case None        => errorFun
-    }
+    validateEmailAndThrow(email).getOrElse(errorFun)
+
+  def validateEmailAndThrow(email: String): Option[String] =
+    EmailAddressRegex.findFirstIn(email)
```
BalduinLandolt, line 2161: return validation here?

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2117,14 +2168,15 @@ class StringFormatter private (
...
     maybeString: Option[String],
     errorFun: => Nothing
   ): Option[String] = // V2 / value objects
-    maybeString match {
-      case Some(s) => Some(toSparqlEncodedString(validateEmailAndThrow(s, errorFun), errorFun))
-      case None    => None
-    }
+    maybeString.map(validateEmailAndThrow(_).getOrElse(errorFun))
+
+  def validateAndEscapeOptionalEmail(maybeString: Option[String]): Option[String] =
```
BalduinLandolt, line 2178: return validation?

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2134,11 +2186,12 @@ class StringFormatter private (
...
    */
+  @deprecated("Use validateUsername(String) instead.")
   def validateUsername(value: String, errorFun: => Nothing): String = // V2 / value objects
-    UsernameRegex.findFirstIn(value) match {
-      case Some(username) => username
-      case None           => errorFun
-    }
+    validateUsername(value).getOrElse(errorFun)
+
+  def validateUsername(value: String): Option[String] =
```
BalduinLandolt, line 2193: return validation?

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2148,11 +2201,12 @@ class StringFormatter private (
...
    */
+  @deprecated("Use validateAndEscapeUsername(String) instead.")
   def validateAndEscapeUsername(value: String, errorFun: => Nothing): String = // V2 / value objects
-    UsernameRegex.findFirstIn(value) match {
-      case Some(username) => toSparqlEncodedString(username, errorFun)
-      case None           => errorFun
-    }
+    validateAndEscapeUsername(value).getOrElse(errorFun)
+
+  def validateAndEscapeUsername(value: String): Option[String] =
```
BalduinLandolt, line 2208: return validation?

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2162,14 +2216,15 @@ class StringFormatter private (
...
     errorFun: => Nothing
   ): Option[String] = // V2 / value objects
-    maybeString match {
-      case Some(s) => Some(validateAndEscapeUsername(s, errorFun))
-      case None    => None
-    }
+    maybeString.map(validateAndEscapeUsername(_).getOrElse(errorFun))
+
+  def validateAndEscapeOptionalUsername(maybeString: Option[String]): Option[String] =
+    maybeString.flatMap(validateAndEscapeUsername)
```
BalduinLandolt, line 2227: I think this mixes legal and illegal state again

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2337,16 +2392,18 @@ class StringFormatter private (
...
+  @deprecated("Use validateBase64EncodedUuid(String) instead.")
+  def validateBase64EncodedUuid(base64Uuid: String, errorFun: => Nothing): UUID = // V2 / value objects
+    validateBase64EncodedUuid(base64Uuid).getOrElse(errorFun)
 
-    decodeTry match {
-      case Success(uuid) => uuid
-      case Failure(_)    => errorFun
-    }
-  }
+  def validateBase64EncodedUuid(base64Uuid: String): Validation[Throwable, UUID] =
```
BalduinLandolt, line 2399: can you narrow the error type?

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2337,16 +2392,18 @@ class StringFormatter private (
...
+    Validation.fromTry(
+      Try(base64DecodeUuid(base64Uuid))
+        .fold(
+          _ => Failure(ValidationException(s"String: '$base64Uuid' is not a valid base 64 UUID.")),
+          uuid => Success(uuid)
+        )
```
BalduinLandolt, line 2405: does it need the `uuid => Success(uuid)`? I think that's just identity, right?  
If so, you could do something like `orElseFailWith` or `mapError`

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2007,49 +2034,63 @@ class StringFormatter private (
       case None    => None
     }
 
+  def validateAndEscapeOptionalProjectShortname(
+    maybeString: Option[String]
+  ): Validation[ValidationException, Option[String]] =
+    maybeString match {
+      case None        => Validation.succeed(None)
+      case Some(value) => validateAndEscapeProjectShortname(value).map(Some(_))
```
seakayone, line 2042: Again, consider removing those two methods entirely.

If you want to retain them, then the `validateAndEscapeOptionalProjectShortname(maybeString: Option[String],  errorFun: => Nothing)` should point to the `validateAndEscapeOptionalProjectShortname(maybeString: Option[String])` otherwise the logic which assumes the validity of a `None` input is duplicated.

I suppose same applies to the similar methods above, please check.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2007,49 +2034,63 @@ class StringFormatter private (
...
   def escapeOptionalString(maybeString: Option[String], errorFun: => Nothing): Option[String] = // --
     // TODO: I leave this for now to avoid merge conflicts. Should be moved to the ValuesValidator as soon as possible. (depends on toSparqlEncodedString())
-    maybeString match {
-      case Some(s) =>
-        Some(toSparqlEncodedString(s, errorFun))
-      case None => None
-    }
+    maybeString.map(toSparqlEncodedString).getOrElse(errorFun)
+
+  def escapeOptionalString(maybeString: Option[String]): Option[String] =
```
seakayone, line 2063: Consider removing this method and inline the original `escapeOptionalString` instead.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2007,49 +2034,63 @@ class StringFormatter private (
...
    * Given the optional group IRI, checks if it is in a valid format.
    *
    * @param maybeIri the optional group's IRI to be checked.
    * @return the same optional IRI.
    */
+  @deprecated("Use validateOptionalGroupIri(Option[IRI]) instead.")
   def validateOptionalGroupIri(maybeIri: Option[IRI], errorFun: => Nothing): Option[IRI] = // V2 / value objects
+    validateOptionalGroupIri(maybeIri).getOrElse(errorFun)
+
+  def validateOptionalGroupIri(maybeIri: Option[IRI]): Validation[ValidationException, Option[IRI]] =
```
seakayone, line 2090: Consider removing this method and inline the original `validateOptionalGroupIri` instead.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2058,12 +2099,13 @@ class StringFormatter private (
...
+  @deprecated("Use validatePermissionIri(IRI) instead.")
   def validatePermissionIri(iri: IRI, errorFun: => Nothing): IRI = // V2 / value objects
-    if (isKnoraPermissionIriStr(iri)) {
-      iri
-    } else {
-      errorFun
-    }
+    validatePermissionIri(iri).getOrElse(errorFun)
+
+  def validatePermissionIri(iri: IRI): Validation[ValidationException, IRI] =
```
seakayone, line 2106: Please, consider using `Option` return type.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2073,12 +2115,16 @@ class StringFormatter private (
...
+  @deprecated("Use validateAndEscapeUserIri(IRI) instead.")
   def validateAndEscapeUserIri(iri: IRI, errorFun: => Nothing): String = // V2 / value objects
-    if (isKnoraUserIriStr(iri)) {
-      toSparqlEncodedString(iri, errorFun)
-    } else {
-      errorFun
-    }
+    toSparqlEncodedString(iri).getOrElse(errorFun)
+
+  def validateAndEscapeUserIri(iri: IRI): Validation[ValidationException, String] =
```
seakayone, line 2122: Please, consider using `Option` return type.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2088,13 +2134,17 @@ class StringFormatter private (
...
    * @return the same optional string.
    */
+  @deprecated("Use validateAndEscapeOptionalUserIri(Option[String]) instead.")
   def validateAndEscapeOptionalUserIri(
     maybeString: Option[String],
     errorFun: => Nothing
   ): Option[String] = // V2 / value objects
+    validateAndEscapeOptionalUserIri(maybeString).getOrElse(errorFun)
+
+  def validateAndEscapeOptionalUserIri(maybeString: Option[String]): Validation[ValidationException, Option[String]] =
```
seakayone, line 2144: Consider removing this method and inline the original `validateAndEscapeOptionalUserIri` instead.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
seakayone, line 2178: Consider removing this method and inline the original `validateAndEscapeOptionalEmail` instead.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2162,14 +2216,15 @@ class StringFormatter private (
...
     maybeString: Option[String],
     errorFun: => Nothing
   ): Option[String] = // V2 / value objects
-    maybeString match {
-      case Some(s) => Some(validateAndEscapeUsername(s, errorFun))
-      case None    => None
-    }
+    maybeString.map(validateAndEscapeUsername(_).getOrElse(errorFun))
+
+  def validateAndEscapeOptionalUsername(maybeString: Option[String]): Option[String] =
```
seakayone, line 2226: Consider removing this method and inline the original `validateAndEscapeOptionalUsername` instead.

BalduinLandolt (changes requested):
I think there are some cases where the logic isn't correct.

Apart from that I suggested in many places that you could return `Validation` instead of `Option`. I used to use Option myself in the ValuesValidator, I think... but I developed the feeling that it's unclear to have `None` sometimes represent a valid state and sometimes a invalid state. Validation makes this very clear.

I'm also wondering if test coverage on the stringformatter should be improved, seeing that the broken logic apparently often didn't make tests fail.

Finally, I noticed you didn't add any docstrings to the non-deprecated methods. Is this on purpose?

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
seakayone, line 2399: Please, consider using `Option` return type.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2384,15 +2441,19 @@ class StringFormatter private (
    * @param errorFun a function that throws an exception. It will be called if the string cannot be parsed.
    * @return the decoded [[UUID]].
    */
+  @deprecated("Use decodeUuidWithErr(String) instead.")
   def decodeUuidWithErr(uuidStr: String, errorFun: => Nothing): UUID = // V2 / value objects
+    decodeUuidWithErr(uuidStr).getOrElse(errorFun)
+
+  def decodeUuidWithErr(uuidStr: String): Validation[Throwable, UUID] = // V2 / value objects
```
seakayone, line 2448: Please, consider using `Option` return type.

seakayone (changes requested):
I would prefer 
1. Not adding more methods which accept a single `Option` as an argument unless we have a deep understaning about why a `None` should always be valid in those cases
2. Align the return type of the newly introduced methods as replacements for `errorFun` mehtods. Currently it is a mix of `Validation` and `Option` and the `Validation` does not provide any value on the caller side as the reasons are ignored and also not detailed enough.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
seakayone: `None` is neither "legal" nor "illegal", it simply has a value or not. Used as a return type of a function it simply states "the function is able to return a value for some of its input values". An `Option` does not allow to include any reason of why this is.

Before we had a partial function which would throw and exception provided by the caller. The closest return type is an `Option`. That is because before - for values the function was not defined on - the function did not tell anything about the reasons of why it was not defined, i.e. it did not anything to the externally provided excepetion.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: All "optional" methods are gone now 🔥

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: In keep it as an `Option`.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: I keep it as an `Option`.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: All "optional" methods are gone now 🔥

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: I keep it as an `Option`.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: I keep it as an `Option`.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: All "optional" methods are gone now 🔥

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: It is an `Option` now instead.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: fixed

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -1493,10 +1493,14 @@ class StringFormatter private (
    *                    IRI.
    * @return the same optional string.
    */
+  @deprecated("Use validateAndEscapeOptionalIri(Option[String]) instead.")
   def validateAndEscapeOptionalIri(maybeString: Option[String], errorFun: => Nothing): Option[IRI] =
+    validateAndEscapeOptionalIri(maybeString).getOrElse(errorFun)
+
+  def validateAndEscapeOptionalIri(maybeString: Option[String]): Validation[ValidationException, Option[String]] =
```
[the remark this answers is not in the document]
mpro7: All "optional" methods are gone now 🔥

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: Changed to `Option`.

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: All "optional" methods are gone now 🔥

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: All "optional" methods are gone now 🔥

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: All "optional" methods are gone now 🔥

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: All "optional" methods are gone now 🔥

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: All "optional" methods are gone now 🔥

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
mpro7: All "optional" methods are gone now 🔥

BalduinLandolt (approved):
LGTM

seakayone (approved):

webapi/src/main/scala/org/knora/webapi/messages/StringFormatter.scala
```diff
@@ -2337,16 +2220,12 @@ class StringFormatter private (
...
+  @deprecated("Use validateBase64EncodedUuid(String) instead.")
+  def validateBase64EncodedUuid(base64Uuid: String, errorFun: => Nothing): UUID = // V2 / value objects
+    validateBase64EncodedUuid(base64Uuid).getOrElse(errorFun)
 
-    decodeTry match {
-      case Success(uuid) => uuid
-      case Failure(_)    => errorFun
-    }
-  }
+  def validateBase64EncodedUuid(base64Uuid: String): Option[UUID] =
```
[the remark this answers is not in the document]
mpro7: Good point. To get this merged now and don't wait next few rounds I will note it and reflect in next possible SF task.

PR #2593, opened 2023-04-14
refactor: ziofy OntologiesRouteV2 (DEV-2052)

seakayone:
### Task Description/Number

Issue Number: DEV-2052
### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No


webapi/src/main/scala/org/knora/webapi/messages/v2/responder/ontologymessages/OntologyMessagesV2.scala
```diff
@@ -80,34 +80,17 @@ case class CreateOntologyRequestV2(
 /**
  * Constructs instances of [[CreateOntologyRequestV2]] based on JSON-LD requests.
  */
-object CreateOntologyRequestV2 extends KnoraJsonLDRequestReaderV2[CreateOntologyRequestV2] {
+object CreateOntologyRequestV2 {
```
BalduinLandolt, line 83: can you elaborate on this removal on our next backend meeting? seems nice to get rid of this, but it makes me wonder where else we should do this, now?

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilZ.scala
```diff
@@ -39,6 +39,35 @@ object RouteUtilZ {
         )
       )
 
+  def ensureExternalOntologyName(iri: SmartIri): ZIO[StringFormatter, BadRequestException, SmartIri] =
+    ZIO.serviceWithZIO[StringFormatter] { sf =>
```
BalduinLandolt, line 43: is this the ontology name or IRI? reads more like IRI, so it would seem the method name is misleading

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilZ.scala
BalduinLandolt, line 1: For now I'm happy with these changes. But we should be cautious not to create a new monster here, it feels a bit like we're putting many different things in this util file, which can be a slippery slope

webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
```diff
@@ -919,107 +710,69 @@ final case class OntologiesRouteV2(
...
-          requestingUser <- getUserADM(requestContext)
+        val requestMessageTask = for {
+          propertyIri <-
+            ZIO
+              .succeed(externalPropertyIris)
+              .filterOrFail(_.size == 1)(BadRequestException(s"Only one property can be deleted at a time"))
+              .map(_.head)
+              .flatMap(iri => RouteUtilZ.toSmartIri(iri, s"Invalid property IRI: $iri"))
+              .flatMap(RouteUtilZ.ensureExternalOntologyName)
+              .flatMap(RouteUtilZ.ensureApiV2ComplexSchema)
```
BalduinLandolt, line 765: multiple flatmaps within a for-comprehension make me suspicious... could this be re-written to a more direct syntax along those lines?

```suggestion
    iris <- ZIO.succeed(externalPropertyIris)
    iri  <- iris.filterOrFail(_.size == 1)(BadRequestException(s"Only one property can be deleted at a time")).map(_.head)
    smartIri    <- RouteUtilZ.toSmartIri(iri, s"Invalid property IRI: $iri")
    _           <- RouteUtilZ.ensureExternalOntologyName(smartIri)
    _           <- RouteUtilZ.ensureApiV2ComplexSchema(smartIri)
```

BalduinLandolt: (same goes for the following similar instances... could this maybe even be factored out?)

BalduinLandolt (approved):
Over all looking good, thanks! But not an easy one to to review...

I added some comments, but that's all marginalia.

webapi/src/it/scala/org/knora/webapi/e2e/v2/OntologyV2R2RSpec.scala
```diff
@@ -401,6 +401,10 @@ class OntologyV2R2RSpec extends R2RSpec {
 
           Get(httpGetTest.urlPath).addHeader(Accept(mediaType)) ~> ontologiesPath ~> check {
             val responseStr: String = responseAs[String]
+            val isOkResponse        = response.status == StatusCodes.OK
+            if (!isOkResponse) {
+              println(httpGetTest)
+            }
```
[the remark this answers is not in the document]
seakayone: yes in this case it is very helpful. That is because this particular test is a kind of "parameterized" test, i.e. it is executing the same test for a lot of requests. It failed during development and I did not now which request was the perpetrator. This println will give the information in the case of a failure.

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilZ.scala
seakayone: I am open for suggestions on how to structure the code. For now I am collecting "duplicate" code in all of our `Routes` here. Some functions are regarding extracting params from the request context, some are methods validating and parsing those values into "domain" objects by wrapping the infamous `StringFormatter` & `throw` combination approach into a collection of functions returning a `ZIO[StringFormatter, BadRequestException, A]`.

webapi/src/main/scala/org/knora/webapi/routing/RouteUtilZ.scala
seakayone: The original check was based on the `checkExternalOntologyName` of the `StringFormatter` which checks that the iri is a `SmartIri.isKnoraApiV2DefinitionIri` and that the `SmartIri.getOntologyName` is one of `knora-base` or `knora-admin`.

So probably the newly extracted method `org.knora.webapi.messages.StringFormatter#isKnoraOntologyIri` which contains above check should have a better name, e.g. `isKnoraOntologyName`?

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/ontologymessages/OntologyMessagesV2.scala
seakayone: The `KnoraJsonLDRequestReaderV2` trais has tied the `fromJsonLD` factory methods to a very strict form with parameters which are not needed in most places and this is the only definition in that `trait`. These factory methods where always used directly and therefore the interface did not provide any value other than tying the signature to something unnecessary.

Bit by bit I will get rid if when tackling the other nine places where it currently used.

webapi/src/main/scala/org/knora/webapi/routing/v2/OntologiesRouteV2.scala
seakayone: IMHO this reads more nicely because it is entirely clear for which iri the `ensure` functions are used.

In your example you have to read and understand that `smarIri` is passed into the function:
```
   smartIri    <- RouteUtilZ.toSmartIri(iri, s"Invalid property IRI: $iri")
    _           <- RouteUtilZ.ensureExternalOntologyName(smartIri)
    _           <- RouteUtilZ.ensureApiV2ComplexSchema(smartIri)
```

whereas in the original code it is immediately obvious:
```
   smartIri    <- RouteUtilZ.toSmartIri(iri, s"Invalid property IRI: $iri")
                         .flatMap(RouteUtilZ.ensureExternalOntologyName)
                         .flatMap(RouteUtilZ.ensureApiV2ComplexSchema)
```
 An improvement could be to extract the whole thing into a a method. I have considered that but decided against it for now. This is because it is hard to name this method also because there are so many different combinations of check in the other routes and I liked to be able to directly see which `ensure` checks are done on an iri passed in from the route.

mpro7 (approved):

PR #2763, opened 2023-07-26
Add metric

seakayone:
### Task Description/Number

Issue Number: DEV-

### Basic Requirements

### PR Type

- [ ] Yes
- [ ] No

- [ ] No


webapi/src/main/scala/org/knora/webapi/routing/AroundDirectives.scala
```diff
@@ -18,13 +27,38 @@ trait AroundDirectives extends InstrumentationSupport {
...
+  private val iriPattern = Pattern.compile(".*?/(https?:%2F%2F[^/]+)")
+  private def replaceIris(path: String) = {
+    val matcher = iriPattern.matcher(path)
+    if (matcher.find()) {
+      val iri = matcher.group(1)
+      path.replace(iri, "__IRI__")
+    } else {
+      path
     }
   }
```
BalduinLandolt, line 63: I assume this is done to keep the number of distinct paths as small as possible? In that case it probably needs to be done for the other possible identifiers (shortnames, shortcodes and UUIDs) too.  
I had to do a similar thing in the ZIO HTTP metric.

BalduinLandolt (changes requested):
Please see my comment regarding identifier substitution.

Also, I'm not sure you are aware of this: We don't include all metrics in DataDog for cost reasons (mainly because Traefik generates lots of them), but manual exceptions can be put in place, if we want something to show up. Best to ask Sämi in which cases this is relevant.

seakayone:
IMHO this approach does not really make sense. This PR was simply meant to be a quick fix for easily adding some http metrics. However, it is clear that the chosen approach is suboptimal and hard to extend.

It makes more sense to use Route specific metrics like https://github.com/RustedBones/akka-http-metrics for Akka. However this would require to label each individual endpoint.

Since we are moving away from Akka Http this is too much work. Honestly, I was only interested in a single route and it is much easier to add a dedicated Metric for this route than trying to retrofit http metrics for all endpoints.

Closing this PR without followup.

PR #2786, opened 2023-08-14
fix: query-patterns-should-not-be-separated DEV-2473

seakayone:
### Task Description/Number

https://github.com/dasch-swiss/dsp-api/commit/e3515925da45e2c10cf357ce25d1bcdd03fb39b7 introduces a changed behaviour in how query patterns are transformed/optimized.

The create slow down reported in DEV-2473 is caused by this.
The gravsearch query which is used for retrieving the information of the resource to which a value should be created is containing a different optional block whilst the rest of the query still was the same.

The difference in runtime is enormous:
407 results in 0.083 seconds (fixed, old behaviour):
```
OPTIONAL {
  ?standoffNode <http://www.knora.org/ontology/knora-base#standoffTagHasInternalReference> ?targetStandoffTag .
  ?targetStandoffTag <http://www.knora.org/ontology/knora-base#standoffTagHasOriginalXMLID> ?targetOriginalXMLID .
}
```

and 

88198 results in 5.43 seconds (faulty behaviour):
```
OPTIONAL {
  ?standoffNode <http://www.knora.org/ontology/knora-base#standoffTagHasInternalReference> ?targetStandoffTag .
}
OPTIONAL {
  ?targetStandoffTag <http://www.knora.org/ontology/knora-base#standoffTagHasOriginalXMLID> ?targetOriginalXMLID .
}
```
In the faulty example if a match is found for the first triple pattern, it will return results that include that match, even if there is no match for the second triple pattern, and vice versa.

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No


BalduinLandolt (approved):
Thanks for spotting and fixing!

At some point we had a convention to name tests that use ZIO-Test `...ZSpec.scala` - not sure if we still do that? But if we do, you may want to rename the files.

webapi/src/test/scala/org/knora/webapi/messages/util/search/gravsearch/transformers/ConstructTransformerSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+object ConstructTransformerSpec extends ZIOSpecDefault {
+
+  private implicit val sf: StringFormatter = StringFormatter.getInitializedTestInstance
+
+  private def constructTransformerTransform(q: ConstructQuery, limit: Option[Set[SmartIri]] = None) =
+    ZIO.serviceWithZIO[ConstructTransformer](_.transform(q, limit))
+
+  val spec = suite("ConstructTransformerLive")(
+    test(
+      """Given an optional pattern in the Where clause it should transform the inner but not split the pattern"""
```
mpro7, line 28:
```suggestion
      "Given an optional pattern in the Where clause it should transform the inner but not split the pattern"
```

seakayone:
> Thanks for spotting and fixing!
> 
> At some point we had a convention to name tests that use ZIO-Test `...ZSpec.scala` - not sure if we still do that? But if we do, you may want to rename the files.

I am happily ignoring this naming convention. Eventually all of our spec should be migrated to zio-test anyway. Also in the webapi/test source set zio-test should be the only available test framework.

mpro7 (approved):

PR #2794, opened 2023-08-15
feat: Add route that sets projectRestrictedViewSetting size (DEV-2304)

mpro7:
### Task Description/Number

Issue Number: DEV-2304

### Basic Requirements

### PR Type

- [ ] Yes
- [x] No

- [x] No


seakayone (changes requested):
User input must always be sanitized and validated. You cannot just simply take the users `String` value and put it into a query template. This opens the door for all kind of bad things, e.g. attack and invalid data.

Make sure the user input is:
1. _sanitized_ : the new endpoint is not open for a malicious SPARQL injection 
2. _validated_: the `size` value is valid. AFAIU this must be a IIIF size expression

Add documentation for the `size` parameter. Also check your current documentation its example is `GET` whilst the endpoint is `POST`.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -538,6 +563,20 @@ trait ProjectsADMJsonProtocol extends SprayJsonSupport with DefaultJsonProtocol
...
   import org.knora.webapi.messages.admin.responder.usersmessages.UsersADMJsonProtocol._
 
+  implicit val restrictedViewSizeFormat: JsonFormat[RestrictedViewSize] =
+    new JsonFormat[RestrictedViewSize] {
+      override def read(json: JsValue): RestrictedViewSize = json match {
+        case JsString(str) =>
+          Try(RestrictedViewSize.make(str)) match {
+            case Success(result)    => result.asInstanceOf[RestrictedViewSize]
+            case Failure(exception) => deserializationError(s"Could not parse $str", exception)
+          }
```
seakayone, line 573: I wonder how this should work…
`RestrictedViewSize.make` never throws and always returns a `Validation`. IMHO this will always fail at runtime with a `ClassCastException` on the `.asInstanceOf` call.

Where is the `restrictedViewSizeFormat` even used?

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -436,18 +453,47 @@ final case class ProjectsResponderADMLive(
...
+             .flatMap(ZIO.fromOption(_))
+             .orElseFail(NotFoundException(s"Project '${getId(id)}' not found."))
+      _ <- ZIO
+             .fail(ForbiddenException("User is neither system nor project admin."))
+             .unless(user.permissions.isSystemAdmin || user.permissions.isProjectAdmin(iri.value))
+
+      query = twirl.queries.sparql.admin.txt
+                .setProjectRestrictedViewSettings(iri.value, size.value, None)
+
+      _ <- triplestoreService.sparqlHttpUpdate(query.toString)
```
seakayone, line 493: Your PR has conflicts with main, please resolve. 

You will also find that the `sparqlHttpUpdate` method does not exist anymore, you need to use `query(Update("<your query>")`.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: Using `JsonFormat1` didn't work in long term. This is the only way I found it working also making possible to set `RestrictedViewSize` value object as `sealed abstract`. As for what I understand, if there are no `JsonFormat` instances provided, the only way to get things work is to provide own implementation of `write` and `read` methods. If I comment out `implicit val restrictedViewSizeFormat: JsonFormat[RestrictedViewSize]` implementation, test are failing with following error:
```
/Users/mpro/repos/dsp-api/webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala:647:15
could not find implicit value for evidence parameter of type ProjectsADMJsonProtocol.this.JF[dsp.valueobjects.RestrictedViewSize] (Cannot find JsonWriter or JsonFormat type class for dsp.valueobjects.RestrictedViewSize)
    jsonFormat(ProjectRestrictedViewSizeResponseADM, "size")
```

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -438,18 +455,47 @@ final case class ProjectsResponderADMLive(
...
+             .flatMap(ZIO.fromOption(_))
+             .orElseFail(NotFoundException(s"Project '${getId(id)}' not found."))
+      _ <- ZIO
+             .fail(ForbiddenException("User is neither system nor project admin."))
+             .unless(user.permissions.isSystemAdmin || user.permissions.isProjectAdmin(iri.value))
+
+      query = twirl.queries.sparql.admin.txt
+                .setProjectRestrictedViewSettings(iri.value, size.value, None)
+
+      _ = Update(query.toString)
```
seakayone, line 495:
```suggestion
      _ <- triplestoreService.query(Update(query.toString))
```

seakayone: Otherwise it does not persist.

seakayone (changes requested):

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -204,6 +219,8 @@ final case class ProjectsResponderADMLive(
       projectRestrictedViewSettingsGetADM(identifier)
     case ProjectRestrictedViewSettingsGetRequestADM(identifier) =>
       projectRestrictedViewSettingsGetRequestADM(identifier)
+    case ProjectRestrictedViewSettingsSetRequestADM(iri, user, size) =>
+      setProjectRestrictedViewSettings(iri, user, size)
```
seakayone, line 223: The `ProjectRestrictedViewSettingsSetRequestADM` is not used anywhere else and the pattern here will never match.

IMHO, we should delete `ProjectRestrictedViewSettingsSetRequestADM` and this match.

docs/03-endpoints/api-admin/projects.md
```diff
@@ -813,6 +813,42 @@ Example response:
...
+a square, so that the width and height of the returned image are not greater than the requested value.
+Example: `!512,512` means the image's bigger side will be set to 512 pixels, setting the other side respectively to
+image aspect ratio.
+The percentage pattern looks like: `pct:X`, where X is the number between 1-100 representing the percentage the image
+will be scaled to. Example: `pct:1` means the image will be scaled to 1% of the original image size.
+
+Permissions: ProjectAdmin/SystemAdmin
+
+Request definition:
+- `POST /admin/projects/iri/{iri}/RestrictedViewSettings`
```
seakayone, line 829: The `GET` api supports _iri_, _shortname_ and _shortcode_.

Should we also support all three possible identifiers for the `POST`?

webapi/src/main/scala/org/knora/webapi/slice/admin/api/service/ProjectsADMRestService.scala
```diff
@@ -232,13 +239,28 @@ final case class ProjectsADMRestServiceLive(
...
+   * @param user requesting user,
+   * @param size value to be set,
+   * @return [[ProjectRestrictedViewSettingsResponseADM]].
+   */
+  override def setProjectRestrictedViewSettings(
+    iri: Iri.ProjectIri,
+    user: UserADM,
+    size: RestrictedViewSize
+  ): Task[ProjectRestrictedViewSizeResponseADM] =
+    responder.setProjectRestrictedViewSettings(iri, user, size)
```
seakayone, line 262: Here we should check whether a project really exists with the given iri. Currently your code would simply persist a triple also if it is a valid `iri` but the project does not exist in the db.

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -438,18 +455,47 @@ final case class ProjectsResponderADMLive(
...
+  override def setProjectRestrictedViewSettings(
+    iri: Iri.ProjectIri,
+    user: UserADM,
+    size: RestrictedViewSize
+  ): Task[ProjectRestrictedViewSizeResponseADM] = {
+    val id = ProjectIdentifierADM.IriIdentifier(iri)
+    for {
+      _ <- getProjectFromCacheOrTriplestore(id)
+             .flatMap(ZIO.fromOption(_))
+             .orElseFail(NotFoundException(s"Project '${getId(id)}' not found."))
```
seakayone, line 487: Instead of `.flatMap(ZIO.fromOption(_)).orElseFail(` ZIO provides `.someOrElseFail(`

docs/03-endpoints/api-admin/projects.md
```diff
@@ -813,6 +813,42 @@ Example response:
 }
 ```
 
+### Set Restricted View Settings
```
BalduinLandolt, line 816: Please also update the overview table at the beginning of the file, for consistency

BalduinLandolt (commented):
See my comment on the documentation. Apart from that I don't have much to add beyond what @seakayone already commented.

Regarding tests: Since there are no unit tests except for the value objects, should this be added to an integration test?

Another thing I noticed: This route now only exists in the ZIO implementation of the admin API, but we still have the old one running. How should we handle this circumstance?

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
```diff
@@ -438,18 +455,47 @@ final case class ProjectsResponderADMLive(
...
+      _ <- ZIO
+             .fail(ForbiddenException("User is neither system nor project admin."))
+             .unless(user.permissions.isSystemAdmin || user.permissions.isProjectAdmin(iri.value))
```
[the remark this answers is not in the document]
seakayone: Maybe it would make sense to not have this method in the responder but add it to the `ProjectADMService` instead, wdyt?

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
seakayone: guess I did.

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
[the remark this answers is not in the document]
mpro7: This actually make sense, because responder method looks like repetition to the one in service with additional triplestore query handling. Only one thing is that I will need to call this method also in another place in responder when the default value will be decided (the TODO I left). This will sets the value while creating new project. So it's feasible, but would require importing `ProjectADMRestService` into responder. This shouldn't make any concerns, right?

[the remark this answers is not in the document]
seakayone: The responder should not have a dependency on `ProjectADMRestService`. Actually we want to refactor the code so that we can remove the `Responders` entirely. 

The goal here is separating the concerns. The `ProjectADMService` knows nothing about authorization and when called it trusts that the caller has done its part. That is, you should move the method to `ProjectADMService` (without Rest) and inject this service into the responder. The responder currently handles authorization by itself and can reuse the service to perform the business logic.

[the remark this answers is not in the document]
seakayone: See also [our discussion on refactoring](https://www.notion.so/draft-RFC-0001-Package-Structure-Webapi-Refactoring-V3-56eee56e3e6e4d569ace273e611b696b?pvs=4)

BalduinLandolt (approved):

webapi/src/main/scala/org/knora/webapi/responders/admin/ProjectsResponderADM.scala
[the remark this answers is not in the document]
mpro7: Done in https://github.com/dasch-swiss/dsp-api/pull/2794/commits/43b5a0e6183ef96b19eb588dd825f693f2f113e3

docs/03-endpoints/api-admin/projects.md
```diff
@@ -814,6 +815,49 @@ Example response:
...
+
+Both routes take String parameter which sets restricted view size in one of two formats: as an image dimensions or a 
+percentage. The dimensions pattern looks like: `!X,X`, where X is the number representing scaled image dimensions in
+a square, so that the width and height of the returned image are not greater than the requested value.
+Example: `!512,512` means the image's bigger side will be set to 512 pixels, setting the other side respectively to
+image aspect ratio. The percentage pattern looks like: `pct:X`, where X is the number between 1-100 representing the
+percentage the image will be scaled to. Example: `pct:1` means the image will be scaled to 1% of the original image
+size. **Both routes are implemented in `ZIO-HTTP`, thus can be reached only under port `5555`.**
```
seakayone, line 826: > Both routes are implemented in `ZIO-HTTP`

Using zio an implementation detail and can be omitted from the end users documentation, also it is likely that this detail changes but the route will still remain.

> , thus can be reached only under port `5555`.
This is hidden from the user by our load balancer. 

Come to think of it:
`GET /admin/projects/iri/{iri}/RestrictedViewSettings` is on port 3333
`POST /admin/projects/iri/{iri}/RestrictedViewSettings` is on port 5555

The same is true for the `/admin/projects/shortcode/{shortcode}/RestrictedViewSettings` routes. Have you clarified with infrastructure that we can expose both routes on the same port with our loadbalancer.

docs/03-endpoints/api-admin/projects.md
```diff
@@ -814,6 +815,49 @@ Example response:
...
+
+```bash
+curl --request POST 'http://0.0.0.0:5555/admin/projects/shortcode/0001/RestrictedViewSettings' \
+--header 'Authorization: Basic cm9vdEBleGFtcGxlLmNvbTp0ZXN0' \
+--data '{"size": "!512,512"}
+```
+
+Example response:
+
+```jsonc
```
seakayone, line 855: I would have used `json`. I guess `jsonc` stands for json with comments, but you are not using comments.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -338,12 +341,18 @@ case class ProjectKeywordsGetResponseADM(keywords: Seq[String]) extends KnoraRes
  *
  * @param settings the restricted view settings.
  */
-case class ProjectRestrictedViewSettingsGetResponseADM(settings: ProjectRestrictedViewSettingsADM)
+case class ProjectRestrictedViewSettingsResponseADM(settings: ProjectRestrictedViewSettingsADM)
     extends KnoraResponseADM
     with ProjectsADMJsonProtocol {
   def toJsValue: JsValue = projectRestrictedViewGetResponseADMFormat.write(this)
```
seakayone, line 347: Given that we have a `zio-json` codec now it would be nice to look into removing the spray json format. Could you please look into that.

webapi/src/main/scala/org/knora/webapi/routing/admin/ProjectsRouteZ.scala
```diff
@@ -234,6 +246,31 @@ final case class ProjectsRouteZ(
...
+      response   <- projectsService.setProjectRestrictedViewSettings(id, user, size)
+    } yield Response.json(response.toJson)
+
+  private def setProjectRestrictedViewSizeByShortcode(
+    shortcode: String,
+    body: Body,
+    user: UserADM
+  ): Task[Response] = for {
+    id       <- ShortcodeIdentifier.fromString(shortcode).toZIO.mapError(e => BadRequestException(e.msg))
+    body     <- body.asString
```
seakayone, line 269: From this line to end of the function it is a duplicate from the function above. 
You could extract the shared code into a function"
```scala
private def setProjectRestrictedViewSizeByIri(
   id: ProjectIdentifier,
    body: Body,
    user: UserADM
  ): Task[Response]
```

webapi/src/main/scala/org/knora/webapi/slice/admin/api/service/ProjectsADMRestService.scala
```diff
@@ -61,10 +63,16 @@ trait ProjectADMRestService {
...
+  ): Task[ProjectRestrictedViewSettingsResponseADM]
+  def setProjectRestrictedViewSettings(
+    id: ProjectIdentifierADM,
+    user: UserADM,
+    size: RestrictedViewSize
+  ): Task[ProjectRestrictedViewSizeResponseADM]
 }
 
 final case class ProjectsADMRestServiceLive(
+  triplestore: TriplestoreService,
```
seakayone, line 75: This dependency is unused, please remove it.

webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/admin/setProjectRestrictedViewSettings.scala.txt
```diff
@@ -0,0 +1,49 @@
...
+ * Sets project restricted view settings.
+ *
+ * @param iri project's IRI.
+ * @param size project's restricted view size.
+ * @param watermark project's watermark.
+ *
+ *@
+@(iri: IRI,
+  size: String,
+  watermark: Option[String])
```
seakayone, line 19: The `watermark` argument is always passed in as `None`.
Do we really need this parameter? Seem to me like it is unused and untested code.

docs/03-endpoints/api-admin/projects.md
mpro7: It is an implementation detail, but it doesn't hurt to highlight it for us, because all other routes have both implementations.

`GET` routes are exposed on both ports and `POST` routes are exposed to `5555` only. Thought this could be set on the APP side - @SamuelBoerlin what do you think?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: No, this is out of scope. I renamed this class to reuse also with `POST` routes I implemented in this PR, but it ended up that it's not reuse, so I will bring old name back.

webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/admin/setProjectRestrictedViewSettings.scala.txt
mpro7: Good question. I made it the way to have one template for both fields of `RestrictedViewSettings`, however watermark is not yet supported. So now either:
1. I leave it as it is.
2. Remove unused field in the template only.
3. Above + rename entire route, methods and classes from  set `RestrictedViewSettings` to `RestrictedViewSize`, because it is only about the size now.

What do you think? 1, 2 or 3?

seakayone: I would prefer the second option. We can always add the `watermark` support to the route later if we need to.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
seakayone: `ProjectRestrictedViewSettingsResponseADM` and `ProjectRestrictedViewSettingsGetResponseADM` should be exactly the same response. I do not think having two classes for the identical response has some value for us.

I get your point, but I think we should explore the option of removing spray-json further. It seems like it could be a relatively simple task. Can you clarify why you think it's out of scope? It might not be as complicated as it initially sounds.

docs/03-endpoints/api-admin/projects.md
seakayone: The `docs/03-endpoints/api-admin/projects.md` is part of the official end user documentation on docs.dasch.swiss. 
The use of zio will not be helpful for our enduser and might possibly be confusing. It is worthy to point out that is is only available on the 5555 port.

> "but it doesn't hurt to highlight it for us"

We have the "is using zio" information in the code. Shouldn't this be sufficient?

SamuelBoerlin: > It is an implementation detail, but it doesn't hurt to highlight it for us, because all other routes have both implementations.
> 
> `GET` routes are exposed on both ports and `POST` routes are exposed to `5555` only. Thought this could be set on the APP side - @SamuelBoerlin what do you think?

We don't set port 5555 in the app config.  
The DSP APP we host on [app.dasch.swiss](app.dasch.swiss) prepends /zio/ to ZIO routes and those routes are then internally (i.e. transparent to the end user) redirected to port 5555 by Traefik (and non-ZIO to 3333).  
So, there are two cases that might be relevant for end users of the API depending on whether we talk about [api.dasch.swiss](https://api.dasch.swiss) or DSP-API in general:
1) On [api.dasch.swiss](https://api.dasch.swiss) all routes are exposed over port 443/https, so the ZIO routes are not accessed via port 5555 but instead over port 443 and by prefixing routes with /zio/ (e.g.: https://api.dasch.swiss/zio/admin/projects -> port 5555 vs https://api.dasch.swiss/admin/projects -> port 3333)
2) If you self host DSP API then you access the ZIO routes via port 5555 by default (e.g.: http://localhost:5555/admin/projects vs http://localhost:3333/admin/projects)

mpro7: Thanks @SamuelBoerlin for the explanation and confirmation!
I simply removed that sentence from docs.

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
mpro7: There is only one class left, now called as it was before `ProjectRestrictedViewSettingsGetResponseADM`. This PR is about adding new routes, not refactoring entire package.

seakayone: I value your perspective, and I'd like to emphasize that my goal wasn't to kickstart a significant package refactoring. Instead, I aimed to make a minor yet valuable enhancement that could enhance code simplicity and long-term maintainability.

I'm sincerely disappointed that the suggestion to investigate removing spray-json wasn't taken into account. It appeared to be a manageable task with potential benefits for our codebase. While I'm disappointed, I believe that it's crucial to consider even smaller improvements in the spirit of maintaining a clean and efficient codebase. Despite this, I still approve this PR.

seakayone (approved):

PR #2902, opened 2023-10-24
feat(Gravsearch): Enable ORDER BY external link (DEV-2704)

mpro7:
### Task Description/Number

Issue Number: DEV-2704

### PR Type

- [ ] Yes

- [ ] Yes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ version: '3.7'
 services:
 
   app:
-    image: daschswiss/dsp-app:v10.23.5
+    image: daschswiss/dsp-app:v10.24.2
     ports:
       - "4200:4200"
     networks:
--- a/integration/src/test/scala/org/knora/webapi/e2e/v2/SearchRouteV2R2RSpec.scala
+++ b/integration/src/test/scala/org/knora/webapi/e2e/v2/SearchRouteV2R2RSpec.scala
@@ -216,6 +216,31 @@ class SearchRouteV2R2RSpec extends R2RSpec {
     /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
     // Queries without type inference
 
+    "perform a Gravsearch query using simple schema which allows to sort the results by external link" in {
+      val query =
+        """PREFIX knora-api: <http://api.knora.org/ontology/knora-api/simple/v2#>
+          |PREFIX anything: <http://0.0.0.0:3333/ontology/0001/anything/simple/v2#>
+          |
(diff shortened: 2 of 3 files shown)


BalduinLandolt (approved):
LGTM

seakayone:
I'm uncertain about how to interpret this PR. The underlying issue is related to an error in a specific query. You've added tests for two similar queries, but it appears that no changes were made to the production code. This raises some questions for me:

1. Does the error still exists in our stage environment where @mdelez has reported the problem? If it does, how does this PR address that issue? If not, do you know what resolved the original problem?
2. Your tests currently only assert the HTTP status code. Are we confident that the ORDER BY clause is functioning correctly? Would it be a good idea to include assertions to ensure that the returned results are ordered as expected? I would even suggest having a test that checks if `ORDER BY DESC(?prop0)` returns the reverse order.

mpro7:
> I'm uncertain about how to interpret this PR. The underlying issue is related to an error in a specific query. You've added tests for two similar queries, but it appears that no changes were made to the production code. This raises some questions for me:
> 
> 1. Does the error still exists in our stage environment where @mdelez has reported the problem? If it does, how does this PR address that issue? If not, do you know what resolved the original problem?
> 2. Your tests currently only assert the HTTP status code. Are we confident that the ORDER BY clause is functioning correctly? Would it be a good idea to include assertions to ensure that the returned results are ordered as expected? I would even suggest having a test that checks if `ORDER BY DESC(?prop0)` returns the reverse order.
> 
> Edit: I initially misunderstood the change to the `AbstractPreQueryGenerator` as a simple property ordering adjustment. However, upon closer inspection, I see that you've added two mappings: `Xsd.Uri` and `KnoraApiV2Complex.UriValue`. This addresses my initial question nr. 1.
> 
> Upon further reflection, and this may be a question for the issue creator, @mdelez: If I understand correctly, the 'uri' property being used for sorting contains the value of the actual URI. Is this really suitable for creating a meaningful order?

This small PR just enables something in already working sorting feature. I find the idea of checking the order in this case simply as the over-testing.

mdelez:
@seakayone Whether or not the ordering of a URI value is meaningful or not, I don't see a reason why we wouldn't allow a user to do it. It may be meaningful for the user in some way if they have very specific URIs that they know can be sorted alphabetically to help them find their data faster.

seakayone (approved):

mpro7:
I found the `searchextended` route order being already checked in one of the `SearchResponderV2Spec` tests.

PR #2914, opened 2023-11-02
fix(performance): Reverse order of topological sorting in Gravsearch queries

seakayone:
### Task Description/Number

I have tested this change with a couple of queries, e.g. from DEV-2675, DEV-2699.
It seems like these queries benefit drastically from the change.
     
However, I am not sure about the affect of all kinds of queries. This might even slow down other searches.

### PR Type

- [ ] Yes

- [ ] Yes

### DIFF
--- a/integration/src/test/scala/org/knora/webapi/messages/util/search/gravsearch/prequery/GravsearchToCountPrequeryTransformerSpec.scala
+++ b/integration/src/test/scala/org/knora/webapi/messages/util/search/gravsearch/prequery/GravsearchToCountPrequeryTransformerSpec.scala
@@ -7,8 +7,6 @@ package org.knora.webapi.messages.util.search.gravsearch.prequery
 
 import zio._
 
-import scala.collection.mutable.ArrayBuffer
-
 import dsp.errors.AssertionException
 import org.knora.webapi.CoreSpec
 import org.knora.webapi.core.MessageRelay
@@ -102,7 +100,7 @@ class GravsearchToCountPrequeryTransformerSpec extends CoreSpec {
       groupBy = Nil,
       orderBy = Nil,
       whereClause = WhereClause(
-        patterns = ArrayBuffer(
+        patterns = Vector(
           StatementPattern(
             subj = QueryVariable(variableName = "thing"),
             pred = IriRef(
@@ -201,7 +199,7 @@ class GravsearchToCountPrequeryTransformerSpec extends CoreSpec {
   val transformedQueryWithDecimalOptionalSortCriterionAndFilterComplex: SelectQuery =
     SelectQuery(
       fromClause = None,
-      variables = Vector(
+      variables = List(
         Count(
           outputVariableName = "count",
           distinct = true,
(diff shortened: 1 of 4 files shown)


integration/src/test/scala/org/knora/webapi/messages/util/search/gravsearch/prequery/GravsearchToCountPrequeryTransformerSpec.scala
```diff
@@ -201,7 +199,7 @@ class GravsearchToCountPrequeryTransformerSpec extends CoreSpec {
   val transformedQueryWithDecimalOptionalSortCriterionAndFilterComplex: SelectQuery =
     SelectQuery(
       fromClause = None,
-      variables = Vector(
+      variables = List(
```
mpro7, line 202: Isn't `Vector` faster than `List`? Did you check if only changing `ArrayBuffer` to `Vector` above improves the results?

mpro7 (commented):
Please check one question below. Due to auto-merge mode on I only comment now instead of approving.

integration/src/test/scala/org/knora/webapi/messages/util/search/gravsearch/prequery/GravsearchToCountPrequeryTransformerSpec.scala
seakayone: > Isn't Vector faster than List? 

Not in all cases, and does it really matter in the tests. `Vector` is a more efficient data structure for random access and updates, while a `List` is more efficient for sequential access and prepending elements.

I had started using `List` because that was the `Seq` actually returned. However, the scalatest matchers do not mind the type of the collection, only the contents.

seakayone:
> Please check one question below. Due to auto-merge mode on I only comment now instead of approving.

"auto-merge" is not enabled.

mpro7:
> > Please check one question below. Due to auto-merge mode on I only comment now instead of approving.
> 
> "auto-merge" is not enabled.

True, my bad 😄

mpro7 (approved):

BalduinLandolt (approved):
I'm not sure how to feel about this PR: I of course like the fact that it improves performance; but just inverting the order of the topological sort raises some red flags for me... was the assumption of the topological sorting correct but the order just inverted all along? or has only the behaviour of the type statement changed so drastically that it has more influence than the entire topological sorting? do we even need topological sorting, in that case (if it doesn't have the intended effect, why maintain the code for it)?  
I'm happy to see this merged because it has a positive impact for the queries we know. But I would like us to continue the conversation of how to proceed with regards to Gravsearch on a higher level.

seakayone:
> I'm not sure how to feel about this PR: I of course like the fact that it improves performance; but just inverting the order of the topological sort raises some red flags for me... was the assumption of the topological sorting correct but the order just inverted all along? or has only the behaviour of the type statement changed so drastically that it has more influence than the entire topological sorting? do we even need topological sorting, in that case (if it doesn't have the intended effect, why maintain the code for it)? I'm happy to see this merged because it has a positive impact for the queries we know. But I would like us to continue the conversation of how to proceed with regards to Gravsearch on a higher level.

I appreciate your feedback and agree that we have to have a broader discussion about the future of the use of SPARQL for search. Gravsearch is already only exposing a SPARQL subset and tries to address and optimize some problems which come with exposing SPARQL. However, a SPARQL/Gravsearch endpoint has its [limitations due to the general approach of using a query language for end user use cases](https://daverog.wordpress.com/2013/06/04/the-enduring-myth-of-the-sparql-endpoint/).

PR #2916, opened 2023-11-03
refactor: Load ontologies when querying for KnoraProjects

seakayone:
### Task Description/Number

Instead of loading each ontology for every project in separate queries load the ontology when querying for `KnoraProject`s.

This speeds up loading the dsp-app start page.

### PR Type

- [ ] Yes

- [ ] Yes


webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/admin/getProjects.scala.txt
```diff
@@ -21,23 +21,31 @@
...
 PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
 PREFIX knora-admin: <http://www.knora.org/ontology/knora-admin#>
+PREFIX knora-base: <http://www.knora.org/ontology/knora-base#>
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
 
-CONSTRUCT { ?s ?p ?o . }
-
+CONSTRUCT {
+  ?project ?p ?o .
+  ?project knora-admin:belongsToOntology ?ontology .
```
BalduinLandolt, line 29: Do I understand correctly that this property is added here and is namespaced as `knora-admin`, although it is not actually in the ontology? Should we be doing that? I know it works, but it somehow feels wrong.

BalduinLandolt (approved):
looks good, just one question

webapi/src/main/twirl/org/knora/webapi/messages/twirl/queries/sparql/admin/getProjects.scala.txt
seakayone: It is a _virtual_ property just for this particular query for building up the graph of the CONSTRUCT response. Basically it is the inverse of `knora-base:attachedToProject`. But since the subject and object are swapped I thought it is better to have a more speaking name.

If you have a more fitting property I am happy to use it.

BalduinLandolt: I don't have a better one. I'm just wondering if it is a good idea to use virtual properties namespaced as non-virtual ontologies because `knora-admin:xyz` (at least to me) suggests that `xyz` can actually be found in this ontology.  
Would it maybe make sense to define a namespace for virtual properties that we use by convention (something like `dsp-virtual:xyz`), just so we avoid this kind of confusion. But I also don't mind just doing it the way you did, if you feel that's simpler/better.

seakayone: Fun fact aside: `knora-api:belongsToOntology` exists.

mpro7 (approved):
Smart idea, thanks!

PR #2923, opened 2023-11-09
refactor: Move project related value objects to admin.domain.model package and cleanup code

seakayone:
### Task Description/Number

[from 2.13.2](https://github.com/scala/scala/releases/tag/v2.13.2), the `sealed abstract case class private` can be reduced  by [enabling `-Xsource:3`](https://docs.scala-lang.org/scala3/guides/migration/tooling-tour.html#the-scala-213-compiler). From [scala/scala#7702](https://github.com/scala/scala/pull/7702):

> Backport from dotty:
> 
> * If a case class constructor is `private` or `private[foo]`: the synthesized `copy` and `apply` methods will have the same access modifier.
>  * If a case class constructor is `protected` or `protected[foo]`: the synthesized `copy` method will have the same access modifier. The synthesized `apply` method will remain public, because `protected` doesn't make sense in an object. 
>
> Obviously, if a user defines a custom `copy` or `apply` method, that one—including its access modifier—will take precedence.

see also https://tuleism.github.io/blog/2020/scala-smart-constructors/

This way it is also possible to reduce boxing for [value classes which can extend `AnyVal`](https://docs.scala-lang.org/overviews/core/value-classes.html).

However this PR does not include the activation of that flag as some existing code needs to be updated as well, this is done in https://github.com/dasch-swiss/dsp-api/pull/2924

### PR Type

- [ ] Yes

- [ ] Yes


BalduinLandolt (approved):

webapi/src/test/scala/dsp/valueobjects/KnoraProjectSpec.scala
```diff
@@ -0,0 +1,182 @@
...
+        Gen.stringBounded(0, 2)(Gen.printableChar) ++ Gen.stringBounded(257, 300)(Gen.printableChar)
+      check(invalidNames) { name =>
+        assertTrue(
+          Longname.from(name) == Validation.fail(ValidationException("Longname must be 3 to 256 characters long."))
+        )
+      }
+    },
+    test("pass a valid value and successfully create value object") {
+      val validNames = Gen.stringBounded(3, 256)(Gen.printableChar)
+      check(validNames) { name =>
```
mpro7, line 107:
```suggestion
  private val nameTest = suite("Longname")(
    test("pass invalid Longname and expect an error to be returned") {
      val invalidLongnames =
        Gen.stringBounded(0, 2)(Gen.printableChar) ++ Gen.stringBounded(257, 300)(Gen.printableChar)
      check(invalidLongnames) { name =>
        assertTrue(
          Longname.from(name) == Validation.fail(ValidationException("Longname must be 3 to 256 characters long."))
        )
      }
    },
    test("pass a valid value and successfully create value object") {
      val validLongnames = Gen.stringBounded(3, 256)(Gen.printableChar)
      check(validLongnames) { name =>
```

mpro7 (approved):
Looks good, apart of places where I suggest changes.

Also general question, why to not keep store error messages in object vals instead of keep repeating them?

seakayone:
> Also general question, why to not keep store error messages in object vals instead of keep repeating them?

The error message is only used once in production code and is encapsulated in the function returning the error. This approach keeps it close to the actual usage of the code responsible for validation.

I have introduced redundancy in the test to ensure that a change in the production code will cause the test to fail if we unintentionally alter the desired return value. Previously, one could change any of the `Project.ErrorMessages` in the production code to anything, and the test would always pass because the did not test the actual desired return value but rather that the exact same, changed `String` was returned.

webapi/src/test/scala/dsp/valueobjects/KnoraProjectSpec.scala
seakayone: I have added d60a54f7aadc9b475bf4f0b9100799436f1cf785 to the other PR, which should clarify the naming a bit.

PR #2857, opened 2023-09-29
fix: Improve performance for Gravsearch queries

seakayone:
This is an attempt at a minimally invasive fix for some of the performance issue we face with Gravsearch.

Gravsearch queries contain `FILTER NOT EXISTS` blocks which are used to only retrieve resources which were not marked as deleted. For some reason this slows down the query substantially. Replacing this block with a `MINUS` makes simple queries faster.

For the difference between those two statements in the official SPARQL documentation:
https://www.w3.org/TR/sparql11-query/#neg-notexists-minus

> `NOT EXISTS` and `MINUS` represent two ways of thinking about negation, one based on testing whether a pattern exists in the data, given the bindings already determined by the query pattern, and one based on removing matches based on the evaluation of two patterns.

### Task Description/Number

Issue Number: DEV-2714

### Basic Requirements

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

### PR Type

What kind of change does this PR introduce?

- [x] fix: represents bug fixes
- [ ] refactor: represents production code refactoring
- [ ] feat: represents a new feature
- [ ] docs: documentation changes (no production code change)
- [ ] chore: maintenance tasks (no production code change)
- [ ] test: all about tests: adding, refactoring tests (no production code change)
- [ ] other... Please describe:

### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No
- [ ] Maybe (not 100% sure => check with FE)

### Does this PR change client-test-data?

- [ ] Yes (don't forget to update the JS-LIB team about the change)
- [x] No

PR #2945, opened 2023-11-17
fix: BEOL timeouts

SepidehAlassi:
This reverts commit 86cc4f26f0ce93f7a0b51121e26a15890b7f815c.
Revert "fix: Improve performance for Gravsearch queries (#2857)"

### Task Description/Number

Replacing FILTER NOT EXISTS with MINUS in Gravsearch to remove deleted triples introduces query timeouts for BEOL. 

### PR Type

- [ ] Yes

- [ ] Yes


seakayone (approved):
IMHO we can give this a try. Back then (https://github.com/dasch-swiss/dsp-api/pull/2857) was an attempt on improving the performance and it worked for projects which did not have many deleted resources. However, the actual improvements were not very big and given that the queries in BEOL now timeout the revert seems reasonable. 

Though I wonder whether BEOL is so much different, are there so many deleted resources?

BalduinLandolt (approved):
I'm of course ok with reverting a contraproductive optimization. However in my eyes it stresses the point again, that optimizing gravsearch will be never-ending, as there are too many parameters out of our control (so far we discussed mostly the possibility of arbitrarily many different query structures - but of course data structures can vary too, as we see here)

Regarding the large number of deleted resources in BEOL - what numbers are we talking? Could it have to do with the fact that historically, BEOL data was bulk imported, and with each import, all existing data was essentially over-written? I always assumed that the old data was erased entirely in these cases, but I'm not sure.

SepidehAlassi:
@BalduinLandolt and @seakayone, BEOL does not have a large number of deleted resources or values. The data on DaSCH is the result of one bulk import of all BEOL data at the end of the project to an empty database. 

To answer the question of why the queries with MINUS take longer than those with FILTER NOT EXISTS on BEOL:

Both operations are valid negation forms with different performances depending on the size of the queried data and the triplestore. Sometimes, one operation may perform better on smaller datasets, while the other may be more efficient on larger datasets. Given these factors, it's recommended to test the performance of both MINUS and FILTER NOT EXISTS on specific datasets and SPARQL engines to determine which one performs better.

Having a look at Fuseki's [documentation](https://jena.apache.org/documentation/query/negation.html):

> One of the key differences between MINUS and NOT EXISTS is that it is a child graph pattern and so breaks the graph pattern and so the result of the query can change depending where the MINUS is placed. This is unlike the earlier NOT EXISTS examples, where moving the position of the FILTER resulted in equivalent queries.

Since the placement of MINUS in the query affects the performance, in Fuseki, the MINUS operation seem not to be as optimized as FILTER NOT EXISTS.

I just tried queries with MINUS and FILTER NOT EXISTS on Fuseki with BEOL data, here is what I found:

**with MINUS in the prequery to remove deleted items**
If you query for letters (huge dataset), the pre-query has to collect thousands of resources and then remove those marked as deleted or have a deleted property matching a query pattern. Since BEOL has thousands of letters and hundreds of thousands of letter properties, this query with MINUS **times-out**.  
If you query for a smaller set of resources (e.g., manuscript transcription with only 200 resources in the database), a similar query with MINUS in the pre-query is sufficiently fast.
Hence, depending on how many resources there are in graphs matching the main resource of your query, the performance of MINUS differs. 

**with FILTER NOT EXISTS  to remove deleted items**
Performing the same query explained above for letters with FILTER NOT EXISTS is much faster (ca 2 seconds). 

FILTER NOT EXISTS is, in general, a more flexible negation system and can contain complex patterns compared to MINUS, which only can contain simple patterns; that is why FILTER NOT EXISTS is more widely used. Fuseki offers a better optimization for FILTER NOT EXISTS as well.

As far as I remember, back then, we used FILTER NOT EXISTS to remove the deleted items from query results because the property isDeleted was not added to the resources and values of old databases. It was only added with the object value `true` if the resource or value was deleted. With FILTER NOT EXISTS, we could catch the resources/values that had a triple with the property isDeleted with object value "true".  In April 2022, @BalduinLandolt made the isDeleted property mandatory (minCardinality 1); I believe you have written the migration task to add the `isDeleted "False"` to old resources and values. If so, you can then get rid of FILTER NOT EXISTS and MINUS clauses for checking removed resources/values and simply add query patterns `?variable knora-base:isDeleted "false"^^xsd:boolean` (for all variables of the query) to prequeries exclude deleted resources/values from the query results.


To check the performance analysis mentioned above, you can use the following query:

```
PREFIX beol: <http://0.0.0.0:3333/ontology/0801/beol/v2#>
PREFIX knora-api: <http://api.knora.org/ontology/knora-api/v2#>

CONSTRUCT {
    ?letter ?linkingProp1  ?person1 .
    ?letter ?linkingProp2  ?person2 .

} WHERE {

   ?gnd1 knora-api:valueAsString "(DE-588)118531379" .
   ?person1 beol:hasIAFIdentifier ?gnd1 .

   ?gnd2 knora-api:valueAsString "(DE-588)118696149" .
   ?person2 beol:hasIAFIdentifier ?gnd2 .

    ?letter ?linkingProp1 ?person1 .
    FILTER(?linkingProp1 = beol:hasAuthor || ?linkingProp1 = beol:hasRecipient )

    ?letter ?linkingProp2 ?person2 .
    FILTER(?linkingProp2 = beol:hasAuthor || ?linkingProp2 = beol:hasRecipient )

?letter beol:letterHasTranslation ?translation .


} ORDER BY ?date

            OFFSET 0
```

BalduinLandolt:
@SepidehAlassi thank you for the detailed explanation, that is very helpful!

SepidehAlassi:
@seakayone or @BalduinLandolt can you please fix the problem with tests not running on PR from fork? I need this fix to be merged a.s.a.p. Thanks

seakayone:
> @seakayone or @BalduinLandolt can you please fix the problem with tests not running on PR from fork? I need this fix to be merged a.s.a.p. Thanks

I think the problem is that the special "actuated" runners are not available in forks. 
I have pushed your changes to https://github.com/dasch-swiss/dsp-api/pull/2948, if the tests are successful there IMHO we can merge it.

seakayone:
Tests passed: https://github.com/dasch-swiss/dsp-api/pull/2948

seakayone:
@SepidehAlassi Thank you very much for the contribution and the finding. In our internal issue tracker I have created a followup issue in which will look into your suggestion to replace the `FILTER NOT EXISTS` with a `?s knora-base:isDeleted false` triple => DEV-2987

PR #2980, opened 2024-01-08
refactor: Unify user value objects and validation (DEV-3155)

BalduinLandolt:
### Task Description/Number

Issue Number: DEV-

### PR Type

- [ ] Yes

- [ ] Yes


webapi/src/test/scala/org/knora/webapi/slice/admin/domain/model/UserSpec.scala
```diff
@@ -0,0 +1,161 @@
...
+  private val validPassword           = "pass-word"
+  private val validGivenName          = "John"
+  private val validFamilyName         = "Rambo"
+  private val pwStrength              = PasswordStrength.unsafeMake(12)
+
+  private val userSuite = suite("User")()
+
+  private val usernameSuite = suite("Username")(
+    test("Username must not be empty") {
+      assertTrue(Username.from("") == Left(UserErrorMessages.UsernameMissing))
```
seakayone, line 25: When testing please do not reuse production values for the assertion and rather add some redundancy to the test.

For example if someone changes the `UsernameMissing` automatically the test will pass even if the change was unintentional. 

```suggestion
      assertTrue(Username.from("") == Left("Username cannot be empty."))
```

Please also check the other cases below  ↓

webapi/src/test/scala/org/knora/webapi/slice/admin/domain/model/UserSpec.scala
```diff
@@ -0,0 +1,161 @@
...
+    test("Username has to be at most 50 characters long") {
+      assertTrue(
+        Username
+          .from("123456789012345678901234567890123456789012345678901")
+          .left
+          .exists(_.equals(UserErrorMessages.UsernameInvalid))
+      )
+    },
+    test("Username must not contain other characters") {
+      assertTrue(
```
seakayone, line 42: how about using check?

```
val invalid = List("a_2.3!", …)
check ( Gen.fromIterable(invalid)){ i =>
   assertTrue(Username.from(i) == Left("Username is invalid.")
}
```

The advantage is that this way the test will verify all of the values in `invalid` and it would not fail for the first and not check the others.

seakayone (changes requested):
Please do not reuse the `UserErrorMessages` constants in the test code. See my comment above for an explanation.

webapi/src/test/scala/org/knora/webapi/slice/admin/domain/model/UserSpec.scala
BalduinLandolt: I think I copied this pattern over from the existing tests... but happy to improve it

seakayone (approved):

PR #3003, opened 2024-01-23
feat: Add watermark support (DEV-2993) (DEV-2991)

seakayone:
### Task Description/Number

* Adds a `watermark.tif` to knora-sipi container which is used for watermarking.
* Changes `knora-admin:projectRestrictedViewWatermark` property to a `boolean`.
* Extends the `POST admin/projects/[shortcode|iri]/RestrictedViewSettings` endpoints with an optional `watermark` boolean value.
* Extends the `GET admin/files/[shortcode]/filename` and  `GET admin/projects/[shortcode|iri]/RestrictedViewSettings` endpoints to return the `watermark` boolean.
* Adds lua code to handle the watermark and lets Sipi return watermarked images.
* Adds documentation.
* Fixes the urls to the servers in the swagger-ui.
* Removes `zio-mock` dependency and related tests.

Depends on https://github.com/dasch-swiss/sipi/pull/405 being released in a new Sipi version. 

Updating the knora base version is done in a followup PR  #3005

### PR Type

- [ ] Yes

- [ ] Yes


sipi/scripts/sipi.init.lua
```diff
@@ -123,32 +123,27 @@ function pre_flight(prefix, identifier, cookie)
...
-                log("pre_flight - restricted view settings - size: " .. tostring(restrictedViewSize),
-                    server.loglevel.LOG_DEBUG)
-            else
-                log("pre_flight - using default restricted view size", server.loglevel.LOG_DEBUG)
-                restrictedViewSize = config.thumb_size
+        -- watermark and/or size (set per project, managed by DSP-API)
+        local settings_from_api = permission_info.restrictedViewSettings
+
+        local settings_for_sipi = {}
+        settings_for_sipi['type'] = "restrict"
```
mpro7, line 130: Should this be `restrict` or  `restricted`?

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
```diff
@@ -427,7 +428,7 @@ object ProjectIdentifierADM {
  * @param size      the restricted view size.
  * @param watermark the watermark file.
  */
-case class ProjectRestrictedViewSettingsADM(size: Option[String] = None, watermark: Option[String] = None)
+case class ProjectRestrictedViewSettingsADM(size: Option[String] = None, watermark: Boolean)
```
mpro7, line 431: How about to apply `watermark` default value here instead of everywhere else?

mpro7 (commented):
Tests are failing and have 2 questions:

sipi/scripts/sipi.init.lua
seakayone: It is actually https://github.com/dasch-swiss/sipi/pull/404

webapi/src/main/scala/org/knora/webapi/messages/admin/responder/projectsmessages/ProjectsMessagesADM.scala
seakayone: IMHO the default should only be decided in the cases where it makes sense, i.e. on the api level should the user not provide a value. 

I have even removed the size default here as it was unused. => cd952cd40f5b004ef02ef7f164e76e5c8c7ae8bd

mpro7 (commented):
I think the documentation needs to be updated and final question - what about the knora-ontology change? Should then the version be bumped and plugin created? If so, maybe we shouldn't include this PR to tomorrows release?

seakayone:
> I think the documentation needs to be updated 

It is part of this PR … [c28e711](https://github.com/dasch-swiss/dsp-api/pull/3003/commits/c28e71134a3ac84ba5b96461e0c062dc97988546)

seakayone:
> what about the knora-ontology change?

Either I will create a followup which lets us update the single statement in the db or it will have to be a manual step.

Since we are unsure about if the `UpgradePlugin`s are working I will need more time to find a solution.

project/Dependencies.scala
```diff
@@ -13,7 +13,7 @@ object Dependencies {
 
   val fusekiImage =
     "daschswiss/apache-jena-fuseki:2.1.4" // should be the same version as in docker-compose.yml, also make sure to use the same version when deploying it (i.e. version in ops-deploy)!
-  val sipiImage = "daschswiss/sipi:3.8.8" // base image the knora-sipi image is created from
+  val sipiImage = "daschswiss/sipi:3.8.8-3-g39dc0e8" // base image the knora-sipi image is created from
```
seakayone, line 16:
```suggestion
  val sipiImage = "daschswiss/sipi:3.8.9" // base image the knora-sipi image is created from
```

subotic (approved):
LGTM 👍

PR #3035, opened 2024-02-08
chore: Remove `knora-ontologies` symlink (DEV-3236)

siers:
### Task Description/Number

Issue Number: DEV-

### PR Type

- [ ] Yes

- [ ] Yes


webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/RepositoryUpdatePlan.scala
```diff
@@ -71,23 +71,23 @@ object RepositoryUpdatePlan {
...
       name = "http://www.knora.org/ontology/salsah-gui"
     ),
     RdfDataObject(
-      path = "knora-ontologies/standoff-onto.ttl",
+      path = "webapi/src/main/resources/knora-ontologies/standoff-onto.ttl",
       name = "http://www.knora.org/ontology/standoff"
     ),
     RdfDataObject(
-      path = "knora-ontologies/standoff-data.ttl",
+      path = "webapi/src/main/resources/knora-ontologies/standoff-data.ttl",
```
seakayone, line 90: Please dont update here, this is loading from the class path

siers: ✔️

siers:
@seakayone The resetting from docker now works, but I don't think it _ever_ worked while running directly from SBT, at least, I can't see how it could have, because it was relying specifically on the docker mounted paths.

siers:
@seakayone Some advice would be helpful, here's the problem:

The `knora-ontologies` and `test_data` are loaded in local development in `TriplestoreServiceLive.insertDataIntoTriplestore` with cwd = `.../dsp-api/integration`, so from there the old paths were `../knora-ontologies` (symlink) and `../test_data`.

From the container, cwd = `/opt/docker` which is the unpacked jar. Both resource directories are mounted from the project directory with docker compose, making both `../knora-ontologies` and `../test_data` work, which is a lucky accident.

The DefaultRdfData contains a list of paths and introducing a path distinction for the deployment type would be weird.

Because of this having the symlink is a wonderful fix, however, if we do indeed want to get rid of it, I would suggest loading both of these resource directories as resources, have them baked in the jar for docker and accessible via regular resource loading mechanism in local dev. It would add 45M of content, however, but at least the access would be uniform.

What do you think? Other options are probably possible.

seakayone:
> @seakayone Some advice would be helpful, here's the problem:
> 
> The `knora-ontologies` and `test_data` are loaded in local development in `TriplestoreServiceLive.insertDataIntoTriplestore` with cwd = `.../dsp-api/integration`, so from there the old paths were `../knora-ontologies` (symlink) and `../test_data`.
> 
> From the container, cwd = `/opt/docker` which is the unpacked jar. Both resource directories are mounted from the project directory with docker compose, making both `../knora-ontologies` and `../test_data` work, which is a lucky accident.
> 
> The DefaultRdfData contains a list of paths and introducing a path distinction for the deployment type would be weird.
> 
> Because of this having the symlink is a wonderful fix, however, if we do indeed want to get rid of it, I would suggest loading both of these resource directories as resources, have them baked in the jar for docker and accessible via regular resource loading mechanism in local dev. It would add 45M of content, however, but at least the access would be uniform.
> 
> What do you think about the proposition?
> 
> The non-uniform option would be to create a configuration after all, which would be 2x2 choices – local/container × `knora-ontologies`/`test_data` and then find the data according to the configuration table.

As discussed, I would not like to have the `test_data` packaged in the container. Probably you can use the path to determine how to load. If it starts with `knora-ontologies` load it from the classpath, otherwise it remains a relative path as it is not.

siers:
I was just once again reminded that there's the third scenario where these files get loaded, so to sum them up:
* if you run `sbt integration/test`, then cwd = `dsp-api/integration`
* if you run it from the container, then cwd = `/opt/docker` (= unpacked jar), which has the data because of docker compose volumes
* if you run `sbt webapi/run` with the env variable to allow reseting the triplestore content, then cwd = `dsp-api`.

The third one never worked, so I haven't fixed it.

seakayone (approved):

webapi/src/main/scala/org/knora/webapi/store/triplestore/defaults/DefaultRdfData.scala
```diff
@@ -83,5 +65,5 @@ object DefaultRdfData {
       path = "test_data/project_ontologies/webern-onto.ttl",
       name = "http://www.knora.org/ontology/0806/webern"
     )
-  )
+  ) ++ Chunk(RepositoryUpdatePlan.builtInNamedGraphs.toSeq: _*)
```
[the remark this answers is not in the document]
siers: I am not entirely sure what would be best. It could be defined in DefaultRdfData indeed, if you wish me to move it.

mpro7 (approved):
Don't forget to assign yourself to PR and to go through the checkboxes. If PR doesn't correspond to any issue/ticket, add short description which tells what is the motivation behind it.

siers:
@mpro7 I added the ticket number to the PR title, I'd truly forgotten it. However, I was wondering what the benefits of adding myself as an assignee are?

mpro7:
> @mpro7 I added the ticket number to the PR title, I'd truly forgotten it. However, I was wondering what the benefits of adding myself as an assignee are?

As I mentioned already in one of your former PRs, it's easier to serach through PRs if one is looking for something - it's possible to filter by assignee field too.

siers:
@mpro7 I'm not sure there's much use since I'm already the author, but ok.

mpro7:
> @mpro7 I'm not sure there's much use since I'm already the author, but ok.

You're right, author works too, but it became our habit so why to change it? I find useful to find recent PRs in history of closed ones.

siers:
@mpro7 Fair enough.

PR #3036, opened 2024-02-08
fix: OWL property type of knora-api:valueAsString

Nora-Olivia-Ammann:
### Task Description/Number

 DEV-3270

The owl property type for the knora-api:valueAsString is listed as: owl:ObjectProperty.
owl:ObjectProperty are for properties that have a class in object position of the triple.
This property has a xsd:string in object position. Therefore it is a owl:DatatypeProperty.

### PR Type

- [ ] Yes

- [ ] Yes


BalduinLandolt (commented):
looks good, thanks!

@mpro7 @seakayone, what do you think, should the ontology version be bumped and a plugin be added to update it on the servers? Or are we ok with this change not being applied right away? (The changes are just RDF correctness, but make no functional difference in the API)  
Depending on that, it can either be merged as is, or needs some more things to happen.

mpro7 (commented):
IMO this should also be updated across the servers by writing update plugin.

mpro7:
> looks good, thanks!
> 
> @mpro7 @seakayone, what do you think, should the ontology version be bumped and a plugin be added to update it on the servers? Or are we ok with this change not being applied right away? (The changes are just RDF correctness, but make no functional difference in the API) Depending on that, it can either be merged as is, or needs some more things to happen.

What would the be the advantage on doing the plugin in separate PR? It should rather be added within this PR.

seakayone:
Shouldn't we have a change in `knora-base` ontology for a plugin update?

I haven't checked myself if there is even a change necessary in `knora-base`, or if this is only present in `knora-api`?

seakayone (changes requested):
Please check if this change is compatible with the knora-base ontology.

Nora-Olivia-Ammann:
> Please check if this change is compatible with the knora-base ontology.

The corresponding property in knora-base (which is named differently) is correct: https://github.com/dasch-swiss/dsp-api/blob/1b7f88b58eb8263481c819d7f81fd25d59db5ee0/webapi/src/main/resources/knora-ontologies/knora-base.ttl#L1261

I think I have looked at knora-base in regard with these issues and found no mistakes. knora-api still has four more but they are wrong in a different less straight forward way this is. (three `ObjectProperties` that are declared as literals but refer to knora-base or knora-admin classes) And a reference to a `knora-api:User`, which does not at all exist in the ontology but does in knora-admin.

I cannot however say how these changes could cause problems in the conversion.

seakayone (approved):

BalduinLandolt (approved):

PR #3069, opened 2024-02-28
refactor: Move code from UsersResponder to UserService and UserRestService

seakayone:
- **Move add groups methods from UsersResponder to RestService**
- **Move validation business logic from RestService to UserService**
- **Merge UserResponseADM with UserOperationResponseADM**
- **fmt**

### Task Description/Number

Issue Number: DEV-

### PR Type

- [ ] Yes

- [ ] Yes

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/UserService.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/UserService.scala
@@ -11,6 +11,8 @@ import zio.ZIO
 import zio.ZLayer
 
 import dsp.valueobjects.LanguageCode
+import org.knora.webapi.messages.admin.responder.groupsmessages.GroupADM
+import org.knora.webapi.messages.admin.responder.projectsmessages.ProjectADM
 import org.knora.webapi.responders.admin.GroupsResponderADM
 import org.knora.webapi.responders.admin.PermissionsResponderADM
 import org.knora.webapi.slice.admin.domain.model.Email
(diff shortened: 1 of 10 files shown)


mpro7 (approved):

webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/UserService.scala
```diff
@@ -84,15 +88,49 @@ case class UserService(
...
+      isInProject = update.projects.getOrElse(kUser.isInProject).distinct,
+      isInProjectAdminGroup = update.projectsAdmin.getOrElse(kUser.isInProjectAdminGroup).distinct,
+      isInGroup = update.groups.getOrElse(kUser.isInGroup).distinct,
       isInSystemAdminGroup = update.systemAdmin.getOrElse(kUser.isInSystemAdminGroup)
     )
     userRepo.save(updatedUser)
   }
 
-  private def toUser(kUser: KnoraUser): Task[User] = for {
+  def addGroupToUserIsInGroup(user: KnoraUser, group: GroupADM): IO[UserServiceError, KnoraUser] = for {
```
BalduinLandolt, line 99: **suggestion (non-blocking):** Rename these functions.  
I know, this has been like that previously. But I find the naming of these functions unclear and confusing at best.

I would propose something along the lines of:

- `addGroupToUserIsInGroup` -> `addUserToGroup`
- `addProjectToUserIsInProject` -> `addUserToProject` or `makeProjectMember`
- `addProjectToUserIsInProjectAdminGroup` -> `addUserToProjectAdmin` or `makeUserProjectAdmin`

BalduinLandolt (approved):
LGTM, added one suggestion (non-blocking)

webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/UserService.scala
seakayone: The naming of these methods are actually clearer than your proposals. It reads a bit clunky but follows the model and naming.

Let me explain:
1. `Group`'s do not have members attached to them, the relationship is the other way round.
2. A user has an `IsInGroup` property which contains all `GroupIri`s the `User` belongs to.

So the current name does exactly what it says: 
* `addGroupToUserIsInGroup`
* add a `Group` to the `User`'s `IsInGroup` property.

BalduinLandolt: That was clear to me, but I don't like it. A function name IMO should be clear and easy to understand, and tell me what the function does on a high level, and not be concerned with the implementation details of how it is modelled under the hood.

But notice the `non-blocking` - if you want to keep it as is, that's ok.

siers (approved):

webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/UserService.scala
seakayone: You have convinced me:  a cleaner interface for the `UserService` is actually welcome. I will tackle this in my next PR.

BalduinLandolt: cool, looking forward to it :)

PR #3083, opened 2024-03-05
feat!: Remove isSequenceOf and instead add Segment to DSP-API (DEV-3326)

BalduinLandolt:
### Task Description/Number

Issue Number: DEV-3326

This PR:

- Breaking: Removes `isSequenceOf` and `hasSequenceBounds` from the `knora-base` ontology
- Adds `Segment` etc. as a replacement of sequences

The following will be done in follow-up PRs:

- Add documentation on the new feature
- Add to the persisted test data (so we have examples in `make init-db-test`)
- write non-E2E tests

### PR Type

- [x] Yes

- [ ] Yes


seakayone:
As discussed: 

I understand that the removed properties of the ontology are not used in any of our projects and it is therefore safe to delete them from the ontology.

Nonetheless, I would appreciate if we are handling breaking changes in a more not-so-surprising way for our users. Until we have a process for handling this I am indifferent on whether or not we should remove them in this PR. Now I will shutup and write a proposal on a proper depreciation procedure for implementing breaking changes.

seakayone (approved):

siers (approved):

mpro7 (approved):
I didn't find anything suspicious. Should this PR bump the major api version?

BalduinLandolt:
> I didn't find anything suspicious. Should this PR bump the major api version?

Yes, that was the idea (because removing the sequence thing is a breaking change), I guess the exclamation mark should do the trick, right?

mpro7:
> > I didn't find anything suspicious. Should this PR bump the major api version?
> 
> Yes, that was the idea (because removing the sequence thing is a breaking change), I guess the exclamation mark should do the trick, right?

Indeed. I think there were some braking changes in between last major version update and now which didn't bump the number, therefor I needed to ask. Actually we should maybe analyze this maybe on the BE meeting.

PR #3089, opened 2024-03-05
test: Make test ontologies more diffable

siers:
## Pull Request Checklist

* Please see the commit without the test data.
* Since we're using the n-quads format and saving it sorted, we can find a better difference between the datasets if we exclude blank notes (`git diff -I_:B`), however, any changes that are in the blank note triples would then not be visible, so maybe this is not the way to go.
* Size over the turtle format has tripled.

### Task Description/Number

Created while investigating DEV-3341

Issue Number: DEV-

### PR Type

- [ ] Yes

- [ ] Yes


siers:
Discussed in person, this RP would have to be improved:
* reduce the number of stored formats (but keep asserting) 
* move into its own unit
* see if json-ld checks can be united with other formats

PR #3093, opened 2024-03-05
refactor(/admin/groups/): Move parameter check from responder to endpoint

mpro7:
### Task Description/Number

Small PR which relates to DEV-3292 and two questions.
1. Should we move checks which could be done ASAP to the endpoint or handler side instead of having them in responder/service?
2. Does it make sense to test it on the endpoint side?

### PR Type

- [ ] Yes

- [ ] Yes

### DIFF
--- a/integration/src/test/scala/org/knora/webapi/responders/admin/GroupsResponderADMSpec.scala
+++ b/integration/src/test/scala/org/knora/webapi/responders/admin/GroupsResponderADMSpec.scala
@@ -186,20 +186,6 @@ class GroupsResponderADMSpec extends CoreSpec {
           s"Group with the name '${groupName.value}' already exists.",
         )
       }
-
-      "return 'BadRequest' if nothing would be changed during the update" in {
-        val exit = UnsafeZioRun.run(
-          GroupsResponderADM.updateGroup(
-            GroupIri.unsafeFrom(newGroupIri.get),
-            GroupUpdateRequest(None, None, None, None),
-            UUID.randomUUID,
-          ),
-        )
-        assertFailsWithA[BadRequestException](
-          exit,
-          "No data would be changed. Aborting update request.",
-        )
-      }
     }
 
     "used to query members" should {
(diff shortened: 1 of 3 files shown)


webapi/src/main/scala/org/knora/webapi/slice/admin/api/GroupsEndpoints.scala
```diff
@@ -96,8 +96,17 @@ object GroupsRequests {
...
-    implicit val jsonCodec: JsonCodec[GroupUpdateRequest] = DeriveJsonCodec.gen[GroupUpdateRequest]
+    implicit val jsonCodec: JsonCodec[GroupUpdateRequest] = DeriveJsonCodec
+      .gen[GroupUpdateRequest]
+      .transformOrFail(
+        request =>
+          if (List(request.name, request.descriptions, request.status, request.selfjoin).flatten.isEmpty)
+            Left("At least one field must be present.")
+          else Right(request),
+        identity,
+      )
```
seakayone, line 109: **issue**: A serializer/deserializer contains validation logic for the value type

IMHO this behavior of a codec is surprising in a bad way. The product type `GroupUpdateRequest` allows for all combinations of params and that includes all being `None`. A serializer/deserializer should then also be able to produce a `GroupUpdateRequest` json reflecting the all `None` state.

I would rather like to see this done programatically in the rest service or in a smart constructor. A codec to me feels like the last place I would assume such logic exists.

integration/src/test/scala/org/knora/webapi/responders/admin/GroupsResponderADMSpec.scala
```diff
@@ -186,20 +186,6 @@ class GroupsResponderADMSpec extends CoreSpec {
           s"Group with the name '${groupName.value}' already exists.",
         )
       }
-
-      "return 'BadRequest' if nothing would be changed during the update" in {
```
seakayone, line 190: *suggestion*: Maybe move this test to the `GroupsADME2ESpec`?

seakayone (changes requested):

mpro7:
@seakayone I will address your comments in another (follow-up) PR.

PR #3109, opened 2024-03-11
refactor(knora-admin): Remove unused Institution class and its properties (DEV-3365)

seakayone:
### Task Description/Number

Issue Number: DEV-

### PR Type

- [ ] Yes

- [ ] Yes


webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/plugins/UpgradePluginPR3110.scala
```diff
@@ -0,0 +1,41 @@
...
+    Queries
+      .MODIFY()
+      .prefix(Vocabulary.KnoraAdmin.NS, RDF.NS)
+      .delete(s.has(p, o))
+      .where(
+        s.isA(Rdf.iri(KnoraAdminPrefixExpansion, "Institution"))
+          .andHas(p, o)
+          .from(Vocabulary.NamedGraphs.knoraAdminIri),
+      )
+  }
```
BalduinLandolt, line 40: **Question (non-blocking):** Just for my understanding: This would remove any institution from an object which is linked to an institution (like `?someProject :belongsToInstitution ?institution`). But it would not remove the institution itself from the DB, right? Do we have any instances of institutions in our DB? Do we want to keep them?

BalduinLandolt (approved):
Nice!  
Would you mind adding this to the migrations RFC? I think that's already a bit more ergonomical than it used to be - and it should be fast since it works only on one graph.

webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/plugins/UpgradePluginPR3110.scala
seakayone: You are raising a good point. 

This query removes a triples belonging to an `Institution` class.
I have missed to remove the `?project knora-admin:belongsTo ?s` triples as well. Will add it to the PR.

BalduinLandolt: well... I wrongly read it the other way around... but you're right, and I'm glad that my confusion helped to identify a real issue :)

seakayone: see [3c86d84](https://github.com/dasch-swiss/dsp-api/pull/3109/commits/3c86d8475f05e2250fbc45cad9040a3a6d0aa5c2)

BalduinLandolt (approved):
Looks good now

PR #3115, opened 2024-03-13
test: Improve test for serving ontologies in different formats

BalduinLandolt:
### Task Description

This PR attempts to make the approval tests more ergonomic and feel more like "real" approval tests.  
It also removes redundant test cases.

- The `knora-api` can be retrieved through two separate routes; instead of checking both against the same files, the test now ensures that both routes return the same content.
- The number of test cases was radically reduced
  - testing exactly the following things (both in simple and complex schema; and in the formats turtle, JSON-LD and RDF-XML):
    - the built-in ontologies `knora-api`, `salsah-gui` and `standoff`
    - a class definition from a built-in ontology
    - a property definition from a built-in ontology
    - metadata of a project ontology
    - a project ontology
    - a project ontology returning labels in all languages
    - a class definition of a project ontology
    - a property definition of a project ontology
  - the following test cases were removed because they were redundant according to the above systematic
    - `allOntologyMetadata`
    - `incunabulaOntologyMetadata`
    - `beolOntologyMetadata`
    - `incunabulaOntologySimple`
    - `incunabulaOntologyWithValueObjects`
    - `knoraApiSimpleHasColor`
    - `knoraApiWithValueObjectsHasColor`
    - `incunabulaSimplePubDate`
    - `incunabulaWithValueObjectsPubDate`
    - `incunabulaPageAndBookWithValueObjects`
    - `boxOntologyWithValueObjects`
    - `minimalOntologyWithValueObjects`
    - `imagesBild`
    - `incunabulaBook`
    - `incunabulaPage`
    - `anythingHasDate`
    - `imagesTitel`
    - `incunabulaPartOf`
  - the following test cases have been added to fulfil the above systematic
    - `anythingOntologySimple`
    - `anythingHasListItemSimple`
    - `anythingHasListItem`
    - `anythingThingSimple`
    - `anythingThing`
  - the following test cases are no longer available in the client test data
    - `all-ontology-metadata-response`
    - `get-ontologies-project-incunabula-response`
    - `get-ontologies-project-beol-response`
    - `incunabula-ontology`
    - `minimal-ontology`
    - `get-class-image-bild-response`
    - `get-class-incunabula-book-response`
    - `get-property-DateValue-response`
    - `get-property-textValue-response`
    - `get-property-linkvalue-response`

### PR Type

- [ ] Yes

- [x] Yes


siers (approved):
Judging by the description sounds good, however one question – could it be that some routes now have less integration test coverage?

BalduinLandolt:
> Judging by the description sounds good, however one question – could it be that some routes now have less integration test coverage?

this test (mostly) only tests one route, so it is fairly confined where we might have lost coverage. The only thing I can think of that might have lost test coverage now, is if we have any ontology features that were used in some of the no-longer-requested project ontologies, but are not used in the anything ontology (which is still tested here). But this would be limited to things like different value types or stuff like that, so if we lose any coverage here at all, it should only be stuff that is covered by other tests. But I can't say 100% for sure.

mpro7 (approved):
I guess corresponding tests tests need to be removed on the dsp-js side... :)

seakayone (approved):

PR #3166, opened 2024-04-03
test: Move E2EZSpecs to separate SBT project (RFC-013)

BalduinLandolt:
### Task Description/Number

Move e2ez specs to separate SBT project

### PR Type

- [ ] Yes

- [ ] Yes

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -61,6 +61,7 @@ lazy val root: Project = Project(id = "root", file("."))
     webapi,
     sipi,
     integration,
+    e2e,
   )
   .settings(
     // values set for all sub-projects
@@ -287,7 +288,6 @@ lazy val integration: Project = Project(id = "integration", base = file("integra
     Test / fork               := true, // run tests in a forked JVM
     Test / testForkedParallel := false,
     Test / parallelExecution  := false,
-    Test / javaOptions += "-Dkey=" + sys.props.getOrElse("key", "pekko"),
     Test / testOptions += Tests.Argument("-oDF"), // show full stack traces and test case durations
     libraryDependencies ++= Dependencies.webapiDependencies ++ Dependencies.webapiTestDependencies ++ Dependencies.integrationTestDependencies,
   )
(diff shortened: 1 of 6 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -89,6 +89,47 @@ jobs:
...
+    name: Build and e2e-test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ "21" ]
+    env:
+      IS_NOOP: ${{ github.ref_name == 'main' || startsWith(github.ref_name, 'release-') }}
+    steps:
+      - name: Inform on no-op
+        run: echo "Running this as a no-op job... ${{ env.IS_NOOP}}"
```
seakayone, line 102: **issue (non-blocking)**: The `env.IS_NOOP` workaround is really a pain to read and understand

I understand this is currently required, because on main we do not (want?) to have a proper build. Maybe it is time to think about review this setup in a followup?

seakayone:
**issue**: Huge amount of duplication necessary for setting up the tests.

I would not have expected so many lines of code which we have to duplicate in order to have it in a separate module. TBH I think this is a bit too much to understand an maintain. Maybe we really have to consider creating a shared project or make e2e and integration dependent.

seakayone (changes requested):

.github/workflows/build-and-test.yml
BalduinLandolt: sure, if you can think of a better way, I'm all ears!

BalduinLandolt:
> **issue**: Huge amount of duplication necessary for setting up the tests.
> 
> I would not have expected so many lines of code which we have to duplicate in order to have it in a separate module. TBH I think this is a bit too much to understand an maintain. Maybe we really have to consider creating a shared project or make e2e and integration dependent.

Personally, I don't feel it's really all that much.  
And I invested quite a bit of effort in streamlining the existing setup, so it's not strictly duplicated, I would say it's largely improved (and surely more improvement could be done!).  
I'm open to having a shared test project, but I would vote for merging this PR as is, and carefully extracting shared logic out into the shared project, simply because I wouldn't like to lose the streamlining efforts I put into it; but largely taking over the new code would probably break old tests in annoying ways.

If you don't want this merged as is, feel free to take the PR over and give it a shot. I'm not working on it before Monday.

PR #3179, opened 2024-04-10
refactor: Align rest services naming & make layers more readable

mpro7:
### Task Description/Number

Issue Number: DEV-

### PR Type

- [ ] Yes

- [ ] Yes


integration/src/test/scala/org/knora/webapi/core/LayersTest.scala
```diff
@@ -85,16 +85,58 @@ object LayersTest {
...
-      AuthorizationRestService & CacheServiceRequestMessageHandler & CardinalityHandler & ConstructResponseUtilV2 &
-      DspIngestClient & GravsearchTypeInspectionRunner & GroupsResponderADM & GroupsRestService & GroupService &
-      HttpServer & IIIFRequestMessageHandler & InferenceOptimizationService & IriConverter & KnoraUserToUserConverter & ListsResponder &
-      ListsResponderV2 & MessageRelay & OntologyCache & OntologyHelpers & OntologyInferencer & OntologyRepo &
-      OntologyResponderV2 & PermissionUtilADM & PermissionsResponder & PermissionsRestService & ProjectExportService &
-      ProjectExportStorageService & ProjectImportService & ProjectService & ProjectRestService & QueryTraverser &
-      RepositoryUpdater & ResourceUtilV2 & ResourcesResponderV2 & RestCardinalityService & SearchApiRoutes &
-      SearchResponderV2 & StandoffResponderV2 & StandoffTagUtilV2 & State & TestClientService & TriplestoreService &
-      UserService & UsersResponder & UsersRestService & ValuesResponderV2
+    AdminApiEndpoints &
```
seakayone, line 88: I am not a big fan of reformatting long type definitions like this in every other pull request. We should actually agree on a format in our team before this.

I remember I had a type definition on a single line and you have reformatted it to this block style. Now this block style explodes to single types per row. 

All formats have their pros and cons. I would just like to have a team agreement on this before changing it again.

seakayone (changes requested):

integration/src/test/scala/org/knora/webapi/core/LayersTest.scala
seakayone: Other than that the pull request looks good to me. Thanks for cleaning this up.

siers: Per-line is diffable and I am for it. Just like 30 list parameters will get reformatted to one per line.

mpro7: This is something I came to while was adding that definition - it was cumbersome to add it in the right place. Of course, it is possible to add new at the end or wherever, but this is probably not the desired approach. I don't remember reformatting one liner into this, but if I did that I basically see the right direction:
- all in one line - too long, a lot of scrolling and not easy to read/edit,
- grouped into max line length - better to read, but hard to edit,
- one per line - both reading and editing seems to be the easiest.

The last one in fact would be also most probably done automatically by the formatter if instead of `&` there would be `,` in between. So this is IMO the most natural and as above the most efficient way of formatting them.

I also recall @BalduinLandolt comment on one of your PRs to make them like I did now. So I guess we have 2 votes in.

BalduinLandolt (approved):

integration/src/test/scala/org/knora/webapi/core/LayersTest.scala
BalduinLandolt: I see the point that we should agree on these things as a team. But as pointed out, I have suggested that before, so it was indeed obvious that at least parts of the team find that a valuable change.  
And just as a side note, we had similar discussions regarding to the formatting of imports (because Intellij doesn't support scalafix), and while I have always been a proponent of aligning our output as a team, so that back-and-forth changes don't get out of hand, we also agreed as a team, that we don't want to have code style to get in our way of being productive.

So, long story short: I'd be happy to see this merged! :)

BalduinLandolt: (also, @seakayone, just FYI: github branch protection works in such a way that once someone has "requested changes", it is not possible to merge that PR until the same person approves. So if you can be persuaded to approve of this PR, you will have to let Github know, otherwise it will remain blocked.)

seakayone: Happy to see that we all agree on the one type per line approach!

seakayone (approved):

PR #3206, opened 2024-04-23
refactor: Refine scaffolding for maintenance actions (DEV-3532)

BalduinLandolt:
### Task Description/Number

This PR is mostly strictly refactoring: Moving code to separate file, etc. The most significant changes are:
- removal of an unused dependency (`PredicateObjectMapper`) and its transitive dependencies
- introducing a trait for maintenance actions (let me know if you find that useful at all!)
- using the actual string values of the path variables for the different maintenance actions endpoints as examples in the OpenAPI docs.

### PR Type

- [ ] Yes

- [ ] Yes

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/model/MaintenanceAction.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/model/MaintenanceAction.scala
@@ -30,5 +30,5 @@ trait MaintenanceAction[A] {
    *
    * @param params optional parameters as provided by the caller
    */
-  def execute(params: Option[A]): Task[Unit]
+  def execute(params: Option[A] = None): Task[Unit]
 }
--- a/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/model/MaintenanceAction.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/model/MaintenanceAction.scala
@@ -30,5 +30,5 @@ trait MaintenanceAction[A] {
    *
    * @param params optional parameters as provided by the caller
    */
-  def execute(params: Option[A] = None): Task[Unit]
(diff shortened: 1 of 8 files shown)


webapi/src/main/scala/org/knora/webapi/slice/admin/domain/model/MaintenanceAction.scala
```diff
@@ -0,0 +1,34 @@
...
+   * @return true if the action should be executed, false otherwise
+   */
+  def shouldExecute: Task[Boolean] = ZIO.succeed(true)
+
+  /**
+   * Executes the maintenance action.
+   *
+   * @param params optional parameters as provided by the caller
+   */
+  def execute(params: Option[A] = None): Task[Unit]
```
seakayone, line 33: If we would type the `params` as `A` you could use `Any` for `A` to indicate that no params are necessary. This saves us from double checking in the `execute` method if params is  a `Some`.

seakayone (approved):

mpro7 (approved):

webapi/src/main/scala/org/knora/webapi/slice/admin/domain/model/MaintenanceAction.scala
BalduinLandolt: done in [a26c378](https://github.com/dasch-swiss/dsp-api/pull/3206/commits/a26c3788f270efb086898d023a71dc82b678d96b), thanks for the hint, looks much nicer.

We can then type it as `Any` or `Nothing`, both seem to work.  
The only downside is that I cannot provide a default argument anymore, so instead of calling `FooAction(null).execute()`, I have to do `FooAction.execute(())`. That's slightly less intuitive, but I think it's ok.  
One option to get this to work might be to have a trait `NoParamMaintenanceAction` that constraints the type... but I don't think it's worth it.

PR #3216, opened 2024-04-25
feat: Upgrade built-in graphs automatically (DEV-3552)

seakayone:
The last releases where updating only the built in graphs.

The built in graphs should be upgraded automatically added in certain cases.

An upgrade is necessary under these conditions:

* If the server version (defined in knora-base.ttl and org.knora.webapi#KnoraBaseVersion) is higher than the version in the database. 
* If there is no version in the database all plugins must be run
* The application startup should fail if the database version is higher than the server version 

When collecting the necessary plugins results in an empty list of plugins then a `PluginForKnoraBaseVersion(versionNumber = serverVersion, plugin = new MigrateOnlyBuiltInGraphs)` should be run.

When collecting the necessary plugins to run results in a list which contains at least one plugin nothing has to be done because every UpgradePlugin must provide a MigrateSpecificGraphs object which in all cases contains the built in graphs.

### Task Description/Number

Issue Number: [DEV-3552](https://linear.app/dasch/issue/DEV-3552/run-migrateonlybuiltingraphs-upgradeplugin-on-demand)

### PR Type

- [ ] Yes

- [ ] Yes


webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/RepositoryUpdater.scala
```diff
@@ -44,41 +39,22 @@ final case class RepositoryUpdater(triplestoreService: TriplestoreService) {
...
-      // Is the repository up to date?
-      repositoryUpToDate <- ZIO.succeed(foundRepositoryVersion.contains(requiredRepositoryVersion))
-
+      repositoryVersion <- getRepositoryVersion
       repositoryUpdatedResponse <-
-        if (repositoryUpToDate) {
-          // Yes. Nothing more to do.
-          ZIO.succeed(RepositoryUpdatedResponse(s"Repository is up to date at $requiredRepositoryVersion"))
-        } else {
+        if (repositoryVersion.contains(KnoraBaseVersion)) {
```
siers, line 44: shouldn't it check that it's strictly smaller? or are you anyway relying on the fact that only the plugins that are bigger are filtered?

webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/RepositoryUpdater.scala
```diff
@@ -97,57 +73,33 @@ final case class RepositoryUpdater(triplestoreService: TriplestoreService) {
...
-            plugin.versionNumber > pluginForRepositoryVersion.versionNumber
-          }
-
-        case None =>
-          // The repository has no version string. Include all updates.
-          plugins
-      }
-    }.orDie
+  private def selectPluginsForNeededUpdates(maybeRepositoryVersion: Option[String]): Seq[PluginForKnoraBaseVersion] = {
+    val repositoryVersion = maybeRepositoryVersion.flatMap(_.replace("knora-base v", "").toIntOption).getOrElse(-1)
```
siers, line 96: so if this is not found, then all plugins are executed?

seakayone: Yes, if the db does not have a valid version we run all migrations.

webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/RepositoryUpdater.scala
siers: Yes, I think it would make most sense to not do anything, but maybe since it's just the built in graph updates, it should be fine.

siers (approved):

webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/RepositoryUpdatePlan.scala
```diff
@@ -47,12 +47,6 @@ object RepositoryUpdatePlan {
       PluginForKnoraBaseVersion(versionNumber = 29, plugin = new UpgradePluginPR3110()),
       PluginForKnoraBaseVersion(versionNumber = 30, plugin = new UpgradePluginPR3111()),
       PluginForKnoraBaseVersion(versionNumber = 31, plugin = new UpgradePluginPR3112()),
-      PluginForKnoraBaseVersion(versionNumber = 32, plugin = new MigrateOnlyBuiltInGraphs()),
-      PluginForKnoraBaseVersion(versionNumber = 33, plugin = new MigrateOnlyBuiltInGraphs()),
-      PluginForKnoraBaseVersion(versionNumber = 34, plugin = new MigrateOnlyBuiltInGraphs()),
-      PluginForKnoraBaseVersion(versionNumber = 35, plugin = new MigrateOnlyBuiltInGraphs()),
-      PluginForKnoraBaseVersion(versionNumber = 36, plugin = new MigrateOnlyBuiltInGraphs()),
-      PluginForKnoraBaseVersion(versionNumber = 37, plugin = new MigrateOnlyBuiltInGraphs()),
```
mpro7, line 55: I don't quite understand why these `MigrateOnlyBuiltInGraphs` where removed, but the the others above are still presented. This would mean gaps between knora-base version/plugins here, but since the current version is `37` the next plugin will create the gap anyway, because it will bump the knora-base version to `38` and the last version presented here is `31`.  So IMO make sense to either remove or leave all so far created `MigrateOnlyBuiltInGraphs`. Unless I miss something...

webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/RepositoryUpdater.scala
BalduinLandolt: This happens regularly on dev servers, when the API version deployed there is lower than the one on prod, and then prod data gets mirrored to that dev server.

And previously, in this scenario the start-up would fail because of the exception raised in L139 which killed the fiber. I think we should keep that behaviour, unless we have good reasons to change it.

webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/RepositoryUpdatePlan.scala
seakayone: I removed the needless plugins: https://github.com/dasch-swiss/dsp-api/pull/3216/commits/c883a517faa57dff8ee4f5bc7ccd44c645a0c45b

webapi/src/main/scala/org/knora/webapi/store/triplestore/upgrade/RepositoryUpdater.scala
seakayone: This commit add the behaviour back again: 
https://github.com/dasch-swiss/dsp-api/pull/3216/commits/bf97820d833ba44e39ef74a9ee2ed278d4e93e4f

seakayone:
`if (repositoryVersion.contains(KnoraBaseVersion)) `

> Would this case be even possible?

Yes, every time the application is restarting.

seakayone:
Seems to be a fluke. org.knora.webapi.it.v2.KnoraSipiIntegrationV2ITSpec pass locally and after a rerun.

PR #3231, opened 2024-05-03
chore: Port to Scala 3 (DEV-3543)

seakayone:
- **Remove unused non-unit return type compiler warnings**
- **remove needless braces**
- **set scala version to 3.3.3**
- **remove unsupport removeUnused from scalafix.conf**
- **fix unit tests**
- **fix non-unit return value discarded warning**

### Task Description/Number

Issue Number: DEV-

### PR Type

- [x] build/chore: maintenance tasks (no production code change)

- [ ] Yes


integration/src/test/scala/org/knora/webapi/core/LayersTest.scala
```diff
@@ -85,56 +84,16 @@ object LayersTest {
...
+    AdminModule.Provided & AdminApiEndpoints & ApiRoutes & ApiV2Endpoints & AppRouter & AssetPermissionsResponder &
+      Authenticator & AuthorizationRestService & CardinalityHandler & ConstructResponseUtilV2 & DspIngestClient &
+      GravsearchTypeInspectionRunner & GroupRestService & HttpServer & IIIFRequestMessageHandler &
+      InferenceOptimizationService & InvalidTokenCache & IriConverter & JwtService & ListsResponder & ListsResponderV2 &
+      MessageRelay & OntologyCache & OntologyHelpers & OntologyInferencer & OntologyRepo & OntologyResponderV2 &
+      PermissionRestService & PermissionUtilADM & PermissionsResponder & ProjectExportService &
+      ProjectExportStorageService & ProjectImportService & ProjectRestService & QueryTraverser & RepositoryUpdater &
+      ResourceUtilV2 & ResourcesResponderV2 & RestCardinalityService & SearchApiRoutes & SearchResponderV2 &
+      SipiService & StandoffResponderV2 & StandoffTagUtilV2 & State & StringFormatter & TestClientService &
+      TriplestoreService & UserRestService & ValuesResponderV2
```
mpro7, line 96: Is there a way to prevent grouping these classes? As we agreed before, keeping them one per line looks like the easiest to read and maintain.

webapi/src/main/scala/org/knora/webapi/routing/v2/StandoffRouteV2.scala
```diff
@@ -69,8 +69,6 @@ final case class StandoffRouteV2()(
                 throw BadRequestException("part of HTTP multipart request has no name")
 
               case b: BodyPart => throw BadRequestException(s"multipart contains invalid name: ${b.name}")
-
-              case _ => throw BadRequestException("multipart request could not be handled")
```
mpro7, line 73: What is the story behind so many default cases being removed in this PR?

integration/src/test/scala/org/knora/webapi/core/LayersTest.scala
BalduinLandolt, line 96: nitpick: I preferred these to be on separate lines for friendlier diffs

project/Dependencies.scala
```diff
@@ -87,31 +87,27 @@ object Dependencies {
...
+  val scalaGraph =
+    "org.scala-graph" %% "graph-core" % "2.0.1" cross CrossVersion.for3Use2_13 // Should be Scala 3 compatible
+  val titaniumJSONLD = "com.apicatalog" % "titanium-json-ld" % "1.4.0"
+  val xmlunitCore    = "org.xmlunit"    % "xmlunit-core"     % "2.10.0"
 
   // test
   val pekkoHttpTestkit   = "org.apache.pekko" %% "pekko-http-testkit"   % PekkoHttpVersion  // Scala 3 incompatible
   val pekkoStreamTestkit = "org.apache.pekko" %% "pekko-stream-testkit" % PekkoActorVersion // Scala 3 compatible
   val pekkoTestkit       = "org.apache.pekko" %% "pekko-testkit"        % PekkoActorVersion // Scala 3 compatible
   val scalaTest          = "org.scalatest"    %% "scalatest"            % "3.2.18"          // Scala 3 compatible
```
BalduinLandolt, line 107: suggestion (non-blocking): I think it would make sense to remove all the comments relating to Scala 3 compatibility

BalduinLandolt (approved):
Great job, thanks!

.scalafmt.conf
```diff
@@ -1,5 +1,5 @@
 version = "3.8.1"
-runner.dialect = Scala213Source3
```
mpro7, line 2: Is this the core of this PR or could/should be actually moved to another PR including some discussion around the dialect topic? This is the also most probably the reason some formatting failed on some files. I'm also not sure if this change will trigger more formatting noise in the upcoming PRs. This is also most probably the reason of grouping combines classes mentioned in another comment. 
At the end we would like start to use Scala 3 in this repo, but the Scala 2 legacy code will be still here, so I'm not sure if we should change just like that. What do you think?

mpro7 (commented):
Overall looks fine however I have few questions, see below.

webapi/src/main/scala/org/knora/webapi/routing/v2/StandoffRouteV2.scala
seakayone: Compiler warnings in case you have an exhaustive pattern match. The "default" case could only have been reached by matching on a null.

```scala
sealed trait A
case object X extends A
case object Y extends A

val foo: A = ???
foo match {
   case X => ???
   case Y => ???
   case _ => ??? // creates a compiler warning
   // case null => ??? // would be fine
 }}
 ```

Comment (mpro7) [reply webapi/src/main/scala/org/knora/webapi/routing/v2/StandoffRouteV2.scala:73]: That was my guess after checking few cases in detail. Thanks for explanation.

File: .scalafmt.conf (hunk 4 above)
Comment (seakayone) [reply .scalafmt.conf:2]: `Scala213Source3` to `Scala3` does not do anything to our formatting since we already used the Scala 3 syntax.

In `.scalafix.conf` I have left the import on `Scala 2` syntax so that the wildcard imports are still `_` instead of `*`. Will change that in a followup, this PR is already too noise.

File: integration/src/test/scala/org/knora/webapi/core/LayersTest.scala (hunk 1 above)
Comment (siers) [reply integration/src/test/scala/org/knora/webapi/core/LayersTest.scala:96]: I tried finding the scalafmt rule, it's a bit hard to google for, I tried a few ```binPack.parentConstructors```, but that doesn't help. It's also slow to reformat, which doesn't help.

File: project/Dependencies.scala (hunk 3 above)
Comment (seakayone) [reply project/Dependencies.scala]: https://github.com/dasch-swiss/dsp-api/pull/3231/commits/a768fcfd65aa4a31b806d73e2f60c341993ce0df

File: integration/src/test/scala/org/knora/webapi/core/LayersTest.scala (hunk 1 above)
Comment (siers) [reply integration/src/test/scala/org/knora/webapi/core/LayersTest.scala:96]: https://github.com/dasch-swiss/dsp-api/pull/3231#discussion_r1590741879

Comment (seakayone) [reply integration/src/test/scala/org/knora/webapi/core/LayersTest.scala:96]: If scalafmt supports it I'd be happy.

Comment (siers) [reply integration/src/test/scala/org/knora/webapi/core/LayersTest.scala:96]: First one to find the option wins a medal (chocolate).

File: .scalafmt.conf (hunk 4 above)
Comment (siers) [reply .scalafmt.conf:2]: Is `Scala3` the only choice for Scala 3?

Comment (mpro7) [reply .scalafmt.conf:2]: @siers all available dialects are listed in the link above. There is really not much info about them but it seems so there is only one `scala3` related. I guess things could be adjusted with more in detail settings and you can also override dialects for particular paths, but I'm not sure the second is the way to go.

**Review by mpro7 — APPROVED** [review]

File: .scalafmt.conf (hunk 4 above)
Comment (siers) [reply .scalafmt.conf:2]: Although, @mpro7, as Christian says it only changes one line in formatting if we set it back to Scala213Source3. Can it stay as it is then?

File: integration/src/test/scala/org/knora/webapi/core/LayersTest.scala (hunk 1 above)
Comment (mpro7) [reply integration/src/test/scala/org/knora/webapi/core/LayersTest.scala:96]: I think I found something, not the neatest but might be sufficient for all these multilines - https://scalameta.org/scalafmt/docs/configuration.html#-format-off
I have the issues with building the project in this branch, could you please try it out @seakayone or I will continue to check that after lunch. Postpone the merge pelase.

Comment (mpro7) [reply integration/src/test/scala/org/knora/webapi/core/LayersTest.scala]: I confirm, on-off works. Could you please revert formatting in all places where classes were groups and apply above rule? Thanks!

Comment (seakayone) [reply integration/src/test/scala/org/knora/webapi/core/LayersTest.scala]: done

File: .scalafmt.conf (hunk 4 above)
Comment (seakayone) [reply .scalafmt.conf:2]: The target dialect in `scalafix.conf` will be changed in https://github.com/dasch-swiss/dsp-api/pull/3236/files

File: integration/src/test/scala/org/knora/webapi/core/LayersTest.scala (hunk 1 above)
Comment (mpro7) [reply integration/src/test/scala/org/knora/webapi/core/LayersTest.scala]: Great, thank you!
@siers did I just win the chocolate? :D

File: webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala [hunk 5]
```diff
@@ -1595,12 +1595,10 @@ case class TextValueContentV2(
...
-            startParentIri = startParentIndex.map(parentIndex =>
-              startIndexesToStandoffNodeIris(parentIndex),
-            ), // If there's a start parent index, get its IRI, otherwise None
-            endParentIri = endParentIndex.map(parentIndex =>
-              startIndexesToStandoffNodeIris(parentIndex),
-            ), // If there's an end parent index, get its IRI, otherwise None
+            // If there's a start parent index, get its IRI, otherwise None
+            startParentIri = startParentIndex.map(startIndexesToStandoffNodeIris(_)),
+            // If there's an end parent index, get its IRI, otherwise None
+            endParentIri = endParentIndex.map(startIndexesToStandoffNodeIris(_)),
```
siers, line 1601: pointless commentary

integration/src/test/scala/org/knora/webapi/core/LayersTest.scala
siers: You do!

siers (approved):

PR #3240, opened 2024-05-13
chore: Remove .env file requirement from setup

BalduinLandolt:
### Task Description/Number

With this PR, make no longer interferes with local .env files

### PR Type

- [x] build/chore: maintenance tasks (no production code change)

- [ ] Yes


seakayone (approved):

Makefile
```diff
@@ -70,38 +68,24 @@ docker-publish: docker-publish-dsp-api-image docker-publish-sipi-image ## publis
...
 
 .PHONY: stack-up-fast
-stack-up-fast: docker-build-dsp-api-image env-file ## starts the dsp-stack by skipping rebuilding most of the images (only api image is rebuilt).
+stack-up-fast: docker-build-dsp-api-image ## starts the dsp-stack by skipping rebuilding most of the images (only api image is rebuilt).
 	docker-compose -f docker-compose.yml up -d
 
 .PHONY: stack-up-ci
-stack-up-ci: KNORA_DB_REPOSITORY_NAME := knora-test-unit
-stack-up-ci: docker-build env-file print-env-file ## starts the dsp-stack using 'knora-test-unit' repository: fuseki, sipi, api.
+stack-up-ci: KNORA_WEBAPI_TRIPLESTORE_FUSEKI_REPOSITORY_NAME := knora-test-unit
```
seakayone, line 87:
```suggestion
stack-up-ci: KNORA_DB_REPOSITORY_NAME := knora-test-unit
```

**question**: Shouldn't this remain the old environment variable name?

BalduinLandolt: I would hope that no, but I'm glad to have someone checking my logic.  
Here's my reasoning:

Previously, it worked as follows:

- in `vars.mk`, we had `KNORA_DB_REPOSITORY_NAME := knora-test`
- in any other make target case, this remained unchanged; here in `stack-up-ci`, this was overwritten with `KNORA_DB_REPOSITORY_NAME := knora-test-unit`; if the stack was started without make, `KNORA_DB_REPOSITORY_NAME` was not set.
- in `docker-compose.yml` we had `KNORA_WEBAPI_TRIPLESTORE_FUSEKI_REPOSITORY_NAME=${KNORA_DB_REPOSITORY_NAME}`, so the same applied, the variable there could be "knora-test", "knora-test-unit" or undefined.
- in `application.conf` we have another default `repository-name = "knora-test"` `repository-name = ${?KNORA_WEBAPI_TRIPLESTORE_FUSEKI_REPOSITORY_NAME}`

My thinking now was:
- the default "knora-test" in vars.mk is redundant, in that case we can just make sure that it remains undefined so that the same default is used in application.conf.
- in that case, the only value that `KNORA_DB_REPOSITORY_NAME` can ever have, is "knora-test-unit" and the only point where it's read is in the docker-compose file where it is written into `KNORA_WEBAPI_TRIPLESTORE_FUSEKI_REPOSITORY_NAME`, so we could use this variable instead and have the same behaviour.

Does my thinking make any sense?

mpro7 (approved):

Makefile
BalduinLandolt: as discussed

PR #3272, opened 2024-06-10
feat: Add endpoint to erase projects (DEV-3681)

seakayone:
### Task Description/Number

Issue Number: DEV-

### PR Type

- [ ] Yes


webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/KnoraProjectService.scala
```diff
@@ -75,6 +77,8 @@ final case class KnoraProjectService(knoraProjectRepo: KnoraProjectRepo) {
         DuplicateValueException(s"Project with the shortname: '${shortname.value}' already exists"),
       )
 
+  def erase(project: KnoraProject): Task[Unit] = knoraProjectRepo.delete(project)
```
BalduinLandolt, line 80: comment: here the naming gets a bit weird... I suppose the KnoraProjectRepo follows CRUD terminology and therefore uses delete instead of erase?

webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/KnoraUserService.scala
BalduinLandolt, line 1: praise: I love the changes to this file - so much more readable

BalduinLandolt (approved):
Still one test failing due to a changed error message. Other than that it looks good, thanks!  
Just as a "sanity test", have you compared the number of triples in the DB before and after creating a project and deleting it again? They should be equal, I suppose, right?

seakayone:
> Still one test failing due to a changed error message. Other than that it looks good, thanks! Just as a "sanity test", have you compared the number of triples in the DB before and after creating a project and deleting it again? They should be equal, I suppose, right?

No they will not be equal, because the users which were created are not deleted.

webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/KnoraUserService.scala
seakayone: Thanks.

webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/KnoraProjectService.scala
seakayone: Yes, delete on the repository level really removes the data from the database. We do not expose 'delete' methods on the service level (yet), this is done with update/change requests.

Naming is hard and if you have a better idea let me know. For me it feels awkward if we would have to name the methods on the repo level `erase…`. Also the base interface it would then be a `CrueRepository` … hardly recognizable to me.  ¯\_(ツ)_/¯

BalduinLandolt: no no, in the light of this, it's absolutely fine, this line just reads a bit awkward at first. But it really doesn't hurt, it was just a typical case of "double dip" when reading the code.

BalduinLandolt:
> > Still one test failing due to a changed error message. Other than that it looks good, thanks! Just as a "sanity test", have you compared the number of triples in the DB before and after creating a project and deleting it again? They should be equal, I suppose, right?
> 
> No they will not be equal, because the users which were created are not deleted.

right... I suppose you could create a project without users.  
But I didn't mean to cause you unnecessary work with that remark, so I don't think it's necessary, unless it gives you peace of mind. But to me, the PR looks good.

webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/KnoraUserService.scala
```diff
@@ -245,7 +247,30 @@ object KnoraUserService {
...
-    final case class UserServiceError(message: String)
+    sealed trait UserServiceError {
+      def message: String
+    }
+    private def msg(userIri: UserIri, reason: String, value: StringValue) =
+      s"User ${userIri.value} is $reason ${value.value}."
+
+    final case class NotGroupMember(userIri: UserIri, groupIri: GroupIri) extends UserServiceError {
+      override def message: String = msg(userIri, "is not member of group", groupIri)
+    }
```
mpro7, line 258: IMO `private def msg` is not necessary. I would merge it with below message, but if you want to keep it, remove `is` from the string, because it's doubled by passed `reason`.

```suggestion
    final case class NotGroupMember(userIri: UserIri, groupIri: GroupIri) extends UserServiceError {
      override def message: String = s"User ${userIri.value} is not member of group ${groupIri.value}."
    }
```

mpro7 (approved):
Great job, thanks! Do you plan to add some tests which are testing erasure of the project?

seakayone:
> Great job, thanks! Do you plan to add some tests which are testing erasure of the project?

I am currently writing some integration tests. This and the error message fixup will be provided in a followup PR.

PR #3276, opened 2024-06-12
refactor: Build queries with query builder

BalduinLandolt:
### Task Description/Number

Issue Number: DEV-

### PR Type

- [ ] Yes


siers:
In general, I see that the change is fairly straightforward – add the triples for the structured INSERT query from the .txt files, add tests (I hope you've just saved the printed outputs).

Hence, I don't think there's much to worry about here.

However, purely to make high level navigating of the file possible and to make it a bit more pleasant to the eye, I would suggest to
1.  break off the big `for (.. <- ..) {`s as well as the standalone `match` statements into methods
2. use `.foreach` instead of `for` statements
3. refactor `.foreach(c => ….andHas(…, _)) to `_*` (splat) syntax where possible, see: https://github.com/dasch-swiss/dsp-api/blob/5f0cc71f2a7e2248debfa98ee77dd264042f02e4/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/service/KnoraUserRepoLive.scala#L120-L135
4. use rename the KnoraBaseVocab import to something short (see if .* import is fitting), just to reduce clutter.

BalduinLandolt:
> In general, I see that the change is fairly straightforward – add the triples for the structured INSERT query from the .txt files, add tests (I hope you've just saved the printed outputs).
> 
> Hence, I don't think there's much to worry about here.
> 
> However, purely to make high level navigating of the file possible and to make it a bit more pleasant to the eye, I would suggest to
> 
> 1. break off the big `for (.. <- ..) {`s as well as the standalone `match` statements into methods
> 2. use `.foreach` instead of `for` statements
> 3. refactor `.foreach(c => ….andHas(…, _)) to `_*` (splat) syntax where possible, see: https://github.com/dasch-swiss/dsp-api/blob/5f0cc71f2a7e2248debfa98ee77dd264042f02e4/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/service/KnoraUserRepoLive.scala#L120-L135
> 4. use rename the KnoraBaseVocab import to something short (see if .* import is fitting), just to reduce clutter.

Thanks for your input!

I have previously (cf. https://github.com/dasch-swiss/dsp-api/pull/3276/commits/3e589019c8372eac10e777321fc0d8f613240508) also asserted that the output from the old template-based implementation is isomorphic to the new output, so I think we can be quite confident that it works (apart from the places where I missed test coverage, and so it's broken now... but I'm on it ^^).

That's exactly the kind of suggestion I was hoping for, thanks. I have been staring at this mess for so long that I couldn't see clearly how to make it less ugly...  
Regarding `KnoraBaseVocab`: this will go away anyway, it's currently mostly duplication some other vocabulary we already have. I just added it during development to have everything in one place.

webapi/src/main/scala/org/knora/webapi/slice/resources/repo/model/ResourceCreateModels.scala
```diff
@@ -0,0 +1,134 @@
...
+
+package org.knora.webapi.slice.resources.repo.model
+
+import java.time.Instant
+import java.util.UUID
+
+import org.knora.webapi.IRI
+
+final case class ResourceReadyToCreate(
+  resourceIri: IRI,
```
seakayone, line 14: *thought*: It would be nice to use `InternalIri` instead of the `String` type alias to `IRI` in these case classes.

At least this way it is a bit more clear that the provided iri should be internal. WDYT?

webapi/src/main/scala/org/knora/webapi/responders/v2/resources/CreateResourceV2Handler.scala
```diff
@@ -467,9 +469,9 @@ final case class CreateResourceV2Handler(
                     DateValueInfo(
                       valueHasStartJDN = valueHasStartJDN,
                       valueHasEndJDN = valueHasEndJDN,
-                      valueHasStartPrecision = valueHasStartPrecision,
-                      valueHasEndPrecision = valueHasEndPrecision,
-                      valueHasCalendar = valueHasCalendar,
+                      valueHasStartPrecision = valueHasStartPrecision.toString(),
+                      valueHasEndPrecision = valueHasEndPrecision.toString(),
+                      valueHasCalendar = valueHasCalendar.toString(),
```
seakayone, line 474: *Question*: Is there a reason why the `DateValueInfo` should not accept the typed values?

IMHO `DatePrecisionV2` and `CalendarNameV2` are perfectly good types and should be turned into a `String` when building the query.

seakayone (commented):
Reviewing whether the queries stayed the same is tough and I just sampled it.

In the `ResourcesRepoLiveSpec` you have gone to great length to ensure that this is the case I guess. Would it make sense to test not only the query equality there but also if these queries work as expected? I guess our `integration` tests should cover this - at least partly - already, did you check if they are missing some?

BalduinLandolt:
> Reviewing whether the queries stayed the same is tough and I just sampled it.
> 
> In the `ResourcesRepoLiveSpec` you have gone to great length to ensure that this is the case I guess. Would it make sense to test not only the query equality there but also if these queries work as expected? I guess our `integration` tests should cover this - at least partly - already, did you check if they are missing some?

During development I constantly compared the produced result not only with my expectation, but also with the previous implementation (cf. https://github.com/dasch-swiss/dsp-api/commit/3e589019c8372eac10e777321fc0d8f613240508 ) so my confidence that nothing has changed is pretty high. Admittedly my tests don't cover all permutations/combinations of possible input, so there is no 100% confidence. But my confidence is pretty high. :)  
Similarly, the integration tests test many things, but probably not _everything_. But I'm not actively aware of anything that is not covered. So again, not 100% confidence, but pretty good.

webapi/src/main/scala/org/knora/webapi/slice/resources/repo/service/ResourcesRepoLive.scala
```diff
@@ -63,19 +67,254 @@ object ResourcesRepoLive {
...
+
+  }
 }
+
+extension (tp: TriplePattern)
+  def andHas(p: RdfPredicate, o: Option[RdfObject]): TriplePattern =
+    o.fold(tp)(o => tp.andHas(p, o))
+
+extension (iri: Iri)
+  def has(p: RdfPredicate, o: Option[RdfObject]): Option[TriplePattern] =
```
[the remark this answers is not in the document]
BalduinLandolt: I hope I have improved it in https://github.com/dasch-swiss/dsp-api/pull/3276/commits/b91fc161a4a9a00080a659321a7266c130a68a83 though I'm not sure I can fully eliminate the confusion.

> Also I think this would be helpful in other places as well.
> Why don't we put it in a separate file TriplePatternOps and make it easier available for the rest of the code base?

I was considering it, but then I didn't want to widen the scope of this PR any further. (If I move it to `TriplePatternOps` then I should also use it where applicable, I feel.)  
If you don't mind, I'd rather do that in a follow-up PR.

webapi/src/main/scala/org/knora/webapi/slice/common/repo/rdf/Vocabulary.scala
```diff
@@ -0,0 +1,74 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.knora.webapi.slice.common.repo.rdf
+
+import org.eclipse.rdf4j.model.Namespace
+import org.eclipse.rdf4j.model.impl.SimpleNamespace
+import org.eclipse.rdf4j.sparqlbuilder.rdf.Rdf.iri
+
+object Vocabulary {
```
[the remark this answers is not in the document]
BalduinLandolt: I was thinking of unifying the two vocabularies, but again didn't want to do it in this PR. I suppose I should not have touched the `attachedToProject` at all.  
Again, I'd rather address this in a separate PR, if that's ok?

webapi/src/main/scala/org/knora/webapi/responders/v2/resources/CreateResourceV2Handler.scala
BalduinLandolt: also done in https://github.com/dasch-swiss/dsp-api/pull/3276/commits/550c7b20ad7a0c78231ea9605491bca3fcebdc86 - let me know what you think

webapi/src/main/scala/org/knora/webapi/slice/resources/repo/service/ResourcesRepoLive.scala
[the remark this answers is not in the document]
seakayone: The scaladoc helps for sure. Thanks.

> (If I move it to TriplePatternOps then I should also use it where applicable, I feel.)

Actually no, you just make it available to use. Integration to existing code can then happen later.

webapi/src/main/scala/org/knora/webapi/slice/common/repo/rdf/Vocabulary.scala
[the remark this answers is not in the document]
seakayone: We will have a split as soon as you merge this. I'd prefer the whole `KnoraBase` vocab to be in a single file, preferably in the common package.

You can merge it as is but let me know I can move it as soon as you are done.

webapi/src/main/scala/org/knora/webapi/slice/resources/repo/model/ResourceCreateModels.scala
seakayone: :D Thanks for the effort. If you do not have vi installed maybe you give the online tutorial a chance: https://www.openvim.com/

webapi/src/main/scala/org/knora/webapi/responders/v2/resources/CreateResourceV2Handler.scala
seakayone: > Do you think that is an issue, then I would add a new model representing it; otherwise I will just reuse them.

Not in this case. Theses data structures are simple and do not depend on something weird. In the long run we would of course want to move them into the slice.

BalduinLandolt: makes sense, thanks!

seakayone (approved):

PR #3304, opened 2024-07-02
chore: Add sqlite to localdev setup of ingest

seakayone:


siers (approved):
Ah, it's the ingest db that it's for, cool.

PR #3326, opened 2024-07-17
refactor: Merge Vocabularies

seakayone:


webapi/src/main/scala/org/knora/webapi/slice/common/repo/rdf/Vocabulary.scala
```diff
@@ -7,68 +7,125 @@ package org.knora.webapi.slice.common.repo.rdf
...
+import org.eclipse.rdf4j.sparqlbuilder.rdf.Iri
+import org.eclipse.rdf4j.sparqlbuilder.rdf.Rdf
 import org.eclipse.rdf4j.sparqlbuilder.rdf.Rdf.iri
 
+import org.knora.webapi.messages.OntologyConstants.KnoraAdmin.KnoraAdminPrefixExpansion
+import org.knora.webapi.messages.OntologyConstants.KnoraBase.KnoraBasePrefixExpansion
+import org.knora.webapi.slice.admin.AdminConstants.adminDataNamedGraph
+import org.knora.webapi.slice.admin.AdminConstants.permissionsDataNamedGraph
+
 object Vocabulary {
```
BalduinLandolt, line 19: thought: do we need a top level "vocabulary" object at all? I think `KnoraAdmin`, `KnoraBase` could just as well be top-level

BalduinLandolt (approved):
👍

webapi/src/main/scala/org/knora/webapi/slice/common/repo/rdf/Vocabulary.scala
seakayone: The compiler does not need it...
I do like to have a name for the collection of these iris, it also makes importing simple.
What would be your argument for removing this `Vocabulary` object?

BalduinLandolt: no strong arguments, it just felt a bit redundant to have the the "vocabulary" both in the file name and the object. But you're right, in terms of importing it actually adds value, so never mind my comment, I'll resolve it.

PR #3386, opened 2024-10-10
feat: Allow uploading JSON text files (DEV-4222)

siers:
This should be enough, it seems like not all extensions are tested in unit tests, so this was tested only locally.


mpro7 (commented):
How about to keep the alphabetical order wherever you added JSON format? It would require change TXT position too :)

siers:
There you go.

seakayone (approved):

BalduinLandolt (approved):

PR #3468, opened 2025-01-21
fix/ontology cache

seakayone:
- **inline OntologyTriplestoreHelpers trait**
- **inline OntologyCacheHelpers trait**
- **replace with ZIO.unit**
- **Replace cacheUpdatedOntology(SmartIri,ReadOntologyV2) with refreshing cache from db**
- **replace OntologyCache.deleteOntology with refreshing from DB**

### Task Description/Number

Issue Number: DEV-

### PR Type

- [ ] Yes

### DIFF
--- a/integration/src/test/scala/org/knora/webapi/core/LayersTest.scala
+++ b/integration/src/test/scala/org/knora/webapi/core/LayersTest.scala
@@ -28,7 +28,6 @@ import org.knora.webapi.responders.admin.*
 import org.knora.webapi.responders.v2.*
 import org.knora.webapi.responders.v2.ontology.CardinalityHandler
 import org.knora.webapi.responders.v2.ontology.OntologyCacheHelpers
-import org.knora.webapi.responders.v2.ontology.OntologyCacheHelpersLive
 import org.knora.webapi.responders.v2.ontology.OntologyTriplestoreHelpers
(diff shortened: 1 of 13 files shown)


integration/src/test/scala/org/knora/webapi/slice/ontology/repo/service/OntologyCacheSpec.scala
BalduinLandolt, line 1: praise: Nice to see so much deleted!

question: Do you think it would be worth adding a test to assert that any change to an ontology also triggers a cache reload?

BalduinLandolt (approved):
Nice simplification!

Question: Are you confident that the extra work the API now has to do does not cause any significant performance regressions? Should we dedicate some time to test this specifically?

seakayone:
> does not cause any significant performance regressions

The added overhead is refreshing the ontology cache, which is ~ .4 seconds or less. However, where this is added I  removed the after update checks which my guess is take a similar amount of time. All in all I think the "performance" did not decrease significantly.

> Do you think it would be worth adding a test to assert that any change to an ontology also triggers a cache reload?

No.

BalduinLandolt:
Grat, thanks!

mpro7 (approved):

PR #3527, opened 2025-03-07
refactor: Introduce Core and BaseModule (DEV-4686)

seakayone:
### Description
- **refactor: Introduce CoreModule and BaseModule**
- **refactor: Move components into CoreModule and BaseModule**
- **refactor: Move`IriConverter` to `org.knora.webapi.slice.ontology.domain.service.IriConverter`**

I have been experience [Class too big compile errors again](https://github.com/zio/zio/issues/8855). 
In order to reduce the class size I am continuing to move services into modules.
This is the first step and introduces the CoreModule and BaseModule.

I am currently playing on the idea of merging the values/resources/resourceinfo and ontology slices into a core slice.
Hence the name `CoreModule` rather than `OntologyModule`. 
In this core slice we could modularize even further. 
Using a single core module would make it easier to move the responders into the core module and then refactor it further, wdyt?


siers (approved):
Sounds good.

PR #3546, opened 2025-03-21
feat: Unify resource annotations wording (DEV-4393)

mpro7:
Updates the resource classes labels and comments, which are used on the FE side for displaying properties.


BalduinLandolt (approved):
Thanks!

PR #3547, opened 2025-03-25
feat: Add new route for incoming links (DEV-4722)

mpro7:

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
+++ b/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
@@ -96,14 +96,14 @@ trait SearchResponderV2 {
     query: ConstructQuery,
     schemaAndOptions: SchemaRendering,
     user: User,
-    offset: Option[Int],
+    offset: Option[Int] = None,
   ): Task[ReadResourcesSequenceV2]
 
   def gravsearchV2(
     query: IRI,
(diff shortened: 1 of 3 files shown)


webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -117,6 +113,57 @@ trait SearchResponderV2 {
...
+         |?incomingRes  knora-api:isPartOf <$resourceIri> .
+         |?incomingRes knora-api:seqnum ?seqnum .
+         |}
+         |
+         |} OFFSET $offset
+         |""".stripMargin
+
+    gravsearchV2(query, rendering, user)
+  }
+
```
BalduinLandolt, line 166: suggestion: In order to keep the abstraction levels intact, I would move the implementation of this route to the trait implementation, rather than implementing it directly in the trait.

webapi/src/main/scala/org/knora/webapi/slice/search/api/SearchEndpoints.scala
```diff
@@ -298,6 +316,16 @@ final case class SearchRestService(
     response     <- renderer.render(searchResult, opts)
   } yield response
 
+  def incomingLinksSearch(
```
BalduinLandolt, line 319: the route is called "searchIncomingLinks", so maybe this method should/could be called the same?

PR #3557, opened 2025-04-09
feat: Limit Gravsearch to a single project by query parameter

BalduinLandolt:

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/prequery/InferenceOptimizationService.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/prequery/InferenceOptimizationService.scala
@@ -23,15 +23,16 @@ import org.knora.webapi.messages.util.search.StatementPattern
 import org.knora.webapi.messages.util.search.UnionPattern
 import org.knora.webapi.messages.util.search.ValuesPattern
 import org.knora.webapi.messages.util.search.WhereClause
-import org.knora.webapi.messages.v2.responder.ontologymessages.ReadOntologyV2
 import org.knora.webapi.slice.admin.domain.model.KnoraProject.ProjectIri
 import org.knora.webapi.slice.admin.domain.model.KnoraProject.Shortcode
 import org.knora.webapi.slice.admin.domain.service.ProjectService
+import org.knora.webapi.slice.ontology.domain.service.OntologyRepo
 import org.knora.webapi.slice.ontology.repo.service.OntologyCache
 
 final case class InferenceOptimizationService(
   private val projectService: ProjectService,
   private val ontologyCache: OntologyCache,
+  private val ontologyRepo: OntologyRepo,
 )(implicit private val stringFormatter: StringFormatter) {
 
   /**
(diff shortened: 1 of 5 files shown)


webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/prequery/InferenceOptimizationService.scala
```diff
@@ -120,6 +122,20 @@ final case class InferenceOptimizationService(
...
     } yield relevantOntologiesMaybe
   }
+
+  /**
+   * Returns the set of ontologies of a given project to which the search is limited.
+   *
+   * @param projectIri the IRI of the project.
+   * @return the set of ontology IRIs of the project.
+   */
+  def getProjectOntologies(projectIri: ProjectIri): Task[Option[Set[SmartIri]]] =
```
seakayone, line 132: Rather than manually having to traverse the Ontology data structure I would recommend to directly use the `OntologyRepo` in the `SearchResponder`.

Something like is much easier to read and understand:

```
val ontologyIris: List[SmartIri] = ontologyRepo.findByProject(projectIri).map(_.ontologyMetadata.ontologyIri)
val ontologyIrisAsOptionSet: Option[Set[SmartIri]] = Option.when(ontologyIris.nonEmpty)(ontologyIris.toSet)
```

seakayone: Also the `InferenceOptimizationService` should probably not have intricate knowledge about the cache data structure.

seakayone (changes requested):
Great addition, l like it.

However I have one small problem with your changes:

We are in progress of phasing out traversal of the ontology cache data structure in favour of using the OntologyRepo.

I am happy to approve this change once the `getProjectOntologies` in the `InferenceOptimizationService` is gone.

seakayone:
I would also love to see a test for this feature.

webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/prequery/InferenceOptimizationService.scala
BalduinLandolt: You're right, much nicer: https://github.com/dasch-swiss/dsp-api/pull/3557/commits/5558d8e7da91ce13a0fb935b5e1e2efc24eca2bc

In other places, `InferenceOptimizationService` has way too intricate knowledge about the cache, but I would prefer to keep that out of the scope of this PR.

BalduinLandolt: ah... way, now I actually understand you suggestion, I think - will adjust the code some more.

BalduinLandolt: https://github.com/dasch-swiss/dsp-api/pull/3557/commits/d9506f307f0aa7807eb976b9d8ef4a8e4e0245d9

seakayone (approved):

BalduinLandolt:
> I would also love to see a test for this feature.

Absolutely! Would you be ok with me adding tests in a follow-up PR.  
In my experience, search is a bit nasty to test because it requires a fairly deep understanding of the test data, to come up with a search term that shows exactly what one wants to show.

mpro7 (approved):

PR #3573, opened 2025-04-29
feat: Add dedicated endpoints for getting StillImageRepresentations (DEV-4836)

mpro7:
Resolves backend part of DEV-4836.


BalduinLandolt (approved):

webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -325,6 +360,120 @@ final case class SearchResponderV2Live(
...
+   * @param limitToProject  the option to limit the search to the specific project.
+   * @return a [[ResourceCountV2]] representing the number of StillImageRepresentations that have been found.
+   */
+  def searchStillImageRepresentationsCountV2(
+    resourceIri: IRI,
+    user: User,
+    limitToProject: Option[ProjectIri],
+  ): Task[ResourceCountV2] = {
+    val query: String =
+      s"""
```
seakayone, line 431: Shouldn't this be the exact same query as in the search case?

If so, IMHO, we should not be duplicating it then.

mpro7: True, these queries are almost the same. I thought to inline it but then forgot. Thanks for the reminder. Done in [5e60efe](https://github.com/dasch-swiss/dsp-api/pull/3573/commits/5e60efee4ec418a5b295d7f003ca4b37165964a0).

PR #3604, opened 2025-05-20
feat: Add default CopyrightHolders when creating a new project (DEV-4929)

seakayone:
### Description

Add default `CopyrightHolder`s for AI and unknown cases.
Allow to provide `allowedCopyrightHolders` when creating a project.
Always add the default `CopyrightHolder`s to the `allowedCopyrightHolders` during project creation.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/slice/admin/api/ProjectsEndpoints.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/admin/api/ProjectsEndpoints.scala
@@ -193,7 +193,12 @@ final case class ProjectsEndpoints(
 
     val postAdminProjects = baseEndpoints.securedEndpoint.post
       .in(projectsBase)
-      .in(jsonBody[ProjectCreateRequest])
+      .in(
+        jsonBody[ProjectCreateRequest].description(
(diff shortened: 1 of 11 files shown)


webapi/src/main/scala/org/knora/webapi/slice/admin/api/ProjectsEndpoints.scala
```diff
@@ -193,7 +193,12 @@ final case class ProjectsEndpoints(
 
     val postAdminProjects = baseEndpoints.securedEndpoint.post
       .in(projectsBase)
-      .in(jsonBody[ProjectCreateRequest])
+      .in(
+        jsonBody[ProjectCreateRequest].description(
+          "The property `enabledLicenses` is optional and if not provided the DaSCH recommended licensed will be used. " +
```
seakayone, line 198:
```suggestion
          "The property `enabledLicenses` is optional and if not provided the DaSCH recommended licenses will be used. " +
```

BalduinLandolt (approved):

mpro7 (approved):
What are `allowedCopyrightHolders`? I see they are aggregated with `copyrightHolders` but the name doesn't make it clear for me. Thanks for the explanation.

PR #3605, opened 2025-05-20
feat: Add experimental resource metadata endpoint

seakayone:
### Description

This is meant to be an experimental endpoint which is currently returning the metadata information as json.

It logs out the duration of the query, so that we can understand whether or not we need pagination.

Since this is just explorative I have not bothered to add any tests yet.

The produced query looks like this:

```sparql

SELECT DISTINCT ?classIri ?createdAt ?creator ?deletedAt ?label ?modifiedAt ?resourceIri
WHERE { 
  GRAPH <http://www.knora.org/data/0803/incunabula> { 
    ?resourceIri a ?classIri ;
      knora-base:creationDate ?createdAt ;
      knora-base:attachedToUser ?creator ;
      rdfs:label ?label .
    OPTIONAL { ?resourceIri knora-base:lastModificationDate ?modifiedAt . }
    OPTIONAL { ?resourceIri knora-base:deleteDate ?deletedAt . } 
  }
  ?classIri rdfs:subClassOf* knora-base:Resource .
}
```

The endpoint added is `v2/resources/projects/:shortcode/metadata/resources`. This is also not final.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/slice/resources/api/ResourcesEndpoints.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/resources/api/ResourcesEndpoints.scala
@@ -8,21 +8,43 @@ package org.knora.webapi.slice.resources.api
 import sttp.model.HeaderNames
 import sttp.model.MediaType
 import sttp.tapir.*
+import sttp.tapir.json.zio.jsonBody
 import sttp.tapir.model.UsernamePassword
 import sttp.tapir.server.PartialServerEndpoint
 import zio.ZLayer
+import zio.json.DeriveJsonCodec
+import zio.json.JsonCodec
 
+import java.time.Instant
 import scala.concurrent.Future
 
 import dsp.errors.RequestRejectedException
 import org.knora.webapi.config.GraphRoute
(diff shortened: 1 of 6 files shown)


webapi/src/main/scala/org/knora/webapi/slice/resources/api/ResourcesEndpoints.scala
```diff
@@ -55,6 +77,15 @@ final case class ResourcesEndpoints(
       case _            => None
     }(d => (d, d))
 
+  val getResourcesMetadata = baseEndpoints.securedEndpoint.get
+    .in(base / "projects" / projectShortcode / "metadata" / "resources")
+    .out(jsonBody[List[ResourceMetadataDto]])
+    .description(
+      "<b>This endpoint is currently not stable and may change in the future.</b> " +
+        "Get metadata of all resources in a project. " +
```
mpro7, line 85:
```suggestion
      "<b>This is an experimental endpoint and may change in the future.</b> " +
        "It gets metadata of all resources in a project. " +
```

mpro7 (approved):
If this is connected to any issue, please include it at least in the PR title. Thanks!

PR #3615, opened 2025-05-28
feat: Add environment to OpenTelemetry and improve setup (DEV-4994)

seakayone:
### Description

This PR addresses several significant issues with the current OpenTelemetry setup:

* The environment was not set, causing all traces to be grouped into the same bucket. As a result, we couldn’t determine the origin of spans in Sentry.
* The release was not specified, making it harder to track changes across different releases.
* OpenTelemetry was never properly shut down, which could lead to resource leaks or starvation.
* The layer setup was convoluted and included unused code, which has now been streamlined.

Other changes:
* Clarified the naming of the OpenTelemetry configuration and layer objects.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -153,6 +153,7 @@ val customScalacOptions = Seq(
   "-Wvalue-discard",
   "-Xmax-inlines:64",
   "-Wunused:all",
+  "-Xfatal-warnings",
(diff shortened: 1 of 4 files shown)


BalduinLandolt (approved):

seakayone:
Rebased on main and removed setting as global as this leads to runtime errors. Sorry for the push force.

PR #3624, opened 2025-06-05
refactor: Merge all base spec into E2ESpec

seakayone:


siers (approved):
I suppose since the tests still work, the underlying integrations should be correct, @BalduinLandolt? :)

seakayone:
> LGTM
> 
> Question: I was under the assumption, that the IT specs have a different setup (with regard to SIPI). Is that not/no-longer the case? Or does this have any implications for the tests that were changed from ITSpec to E2ESpec?

All removed tests had the same zio layers `LayersTestLive.layer` for the setup. AFAIU no existing tests use the possibility to provide different layers. I am currently working on maybe even removing the mock setup to simplify the test setup more. This hopefully will help me to get the tests green in the other zio-http PR at some point.

PR #3641, opened 2025-06-17
feat: Add get project license by iri endpoint

seakayone:
### Description
* Add public endpoint `GET /admin/projects/shortcode/:shortcode/licenses/:licenseIri endpoint`

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/slice/admin/api/ProjectsLegalInfoEndpoints.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/admin/api/ProjectsLegalInfoEndpoints.scala
@@ -25,6 +25,9 @@ object ProjectLicenseDto {
   given JsonCodec[ProjectLicenseDto] = DeriveJsonCodec.gen[ProjectLicenseDto]
   given Ordering[ProjectLicenseDto]  = Ordering.by(_.labelEn)
 
+  def from(license: License, project: KnoraProject): ProjectLicenseDto =
+    from(license, project.enabledLicenses.contains(license.id))
+
   def from(license: License, isEnabled: Boolean): ProjectLicenseDto =
     ProjectLicenseDto(
       license.id.value,
@@ -85,6 +88,11 @@ final case class ProjectsLegalInfoEndpoints(baseEndpoints: BaseEndpoints) {
     )
     .description("Get the available licenses for use within this project.")
 
+  val getProjectLicensesIri = baseEndpoints.publicEndpoint.get
+    .in(base / "licenses" / licenseIriPath)
+    .out(jsonBody[ProjectLicenseDto])
+    .description("Get a specific license by its IRI for use within this project.")
+
   val putProjectLicensesEnable = baseEndpoints.securedEndpoint.put
     .in(base / "licenses" / licenseIriPath / "enable")
     .description("Enable a license for use within this project. The user must be project admin or system admin.")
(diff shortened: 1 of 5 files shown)


webapi/src/main/scala/org/knora/webapi/slice/admin/api/ProjectsLegalInfoEndpoints.scala
```diff
@@ -85,6 +88,11 @@ final case class ProjectsLegalInfoEndpoints(baseEndpoints: BaseEndpoints) {
     )
     .description("Get the available licenses for use within this project.")
 
+  val getProjectLicensesIri = baseEndpoints.publicEndpoint.get
```
BalduinLandolt, line 91: remark: to me, the name of this is misleading (it actually first made me misunderstand what this PR is about).  
How would you feel about `getProjectLicenseByIri` - my reason being: 
- `get__Iri` suggests that it returns an IRI, `get__ByIri` indicates that it gets something specified through its IRI.
- the plural in `Licenses` to me sounds like it would return multiple licenses, which it doesn't.

BalduinLandolt (approved):
Overall looks good, thanks!  
See my remark on naming

webapi/src/main/scala/org/knora/webapi/slice/admin/api/ProjectsLegalInfoEndpoints.scala
seakayone: The naming of the endpoint `val`s is consistent with the naming of the path the endpoints define, hence the plural, it is `licenses` in the path _not_ `license`. The respective service method respect the singularity and are named according.

Probably it should be named `getProjectLicensesLicenseIri`, wdyt?

BalduinLandolt: Doesn't really roll off the tongue, does it? Also, I'm still irritated by the `Iri` in the end, which to me sounds as if it returns an IRI. But I don't want to block this because of a naming debate, so I leave the call to you.

BalduinLandolt: It's a question of consistency vs. clarity: Consistent would be `getAdminProjectsByShortcodeLicensesByLicenseIri `, but I think we agree that this is not what we want. To me, `getProjectLicenseByIri` or even `getLicenseByIri` makes it clear "what it does", which I typically value in code.

But again, I do not insist here, I just remarked it. If you feel differently, then by all means, go ahead.

seakayone: I appreciate your input, and clarity is important to me as it is for you. We have a different understand about what is "clear" and useful in the context of the endpoint definitions.

The usual use case for me as a developer is: when I open an `*Endpoint` file I would like to see the endpoints named as close to their paths so that I can find them quickly using the outline tool of my IDE. Mostly I know which path I have to go to. This is the clarity I need when working with these files. For example `getLicenseByIri` would not represent the path of that endpoint clear enough, it would represent the intention of what the endpoint was designed for. When looking for an endpoint then would need to translate this to a path which makes me think and costs time and effort. 

That is the reason why the endpoints naming try to resemble the path and not the intention. In the rest of the code intention is more important and I would totally agree with your naming proposals.

PR #3646, opened 2025-06-20
chore: Add CLAUDE.md

seakayone:

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,168 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,168 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+DSP-API is the Digital Humanities Service Platform API - a Scala-based REST API for managing semantic data and digital assets in the humanities. The project uses ZIO for functional programming, Pekko (Apache Pekko) for actors, and integrates with Apache Jena Fuseki triplestore and Sipi media server.
```
BalduinLandolt, line 7: Having Pekko in there may guide it towards using Pekko, which is not necessarily what you want. I don't mind it there, but if claude suddenly starts proposing Actor-based solutions, it may be worth removing it form here, or specifying that we're replacing the actor-based approach.

BalduinLandolt (approved):
LGTM

CLAUDE.md
seakayone: Agree, it is a valid point. I will keep that in mind next time I am updating it.

Closing this PR as it will be merged with https://github.com/dasch-swiss/dsp-api/pull/3647

PR #3660, opened 2025-06-27
fix: Resolve intermittent logger initialization timeout on startup (DEV-5093)

jnussbaum:
## Summary
- Fix intermittent logger initialization timeout causing ActorSystem startup failures
- Increase Pekko logger startup timeout from 5s to 30s in configuration
- Add exponential backoff retry logic with 3 attempts for ActorSystem initialization

## Problem
The application occasionally fails to start with logger timeout errors:
```
Logger log1-Slf4jLogger did not respond within Timeout(5000 milliseconds)
```
This causes a cascading failure where the ActorSystem shuts down during initialization, leading to application startup failure.

## Solution
1. **Configuration Change**: Added `logger-startup-timeout = 30s` to `application.conf` to give logger more time to initialize under load
2. **Retry Logic**: Implemented exponential backoff retry (1s, 2s, 4s delays) with 3 attempts in `PekkoActorSystem.scala` for resilience

## Test plan
- [x] Code compiles successfully
- [ ] Test startup under various load conditions
- [ ] Verify logs show proper error handling and retries when needed
- [ ] Test with existing integration tests

🤖 Generated with [Claude Code](https://claude.ai/code)


webapi/src/main/resources/application.conf
```diff
@@ -9,6 +9,7 @@ pekko {
     logging-filter = "org.apache.pekko.event.slf4j.Slf4jLoggingFilter"
     log-dead-letters = off
     log-dead-letters-during-shutdown = off
+    logger-startup-timeout = 30s
```
seakayone, line 12: Let's first try if only this configuration change solves your problem.

From the logs I can see that the ` Slf4jLogger started` takes roughly seven seconds, so a timeout of 15 seconds should be more than sufficient.

If the ActorSystem does not startup properly I do not want it to be retried, instead it should fail immediately. Please remove the retry from its `acquire`

seakayone (approved):

webapi/src/main/resources/application.conf
seakayone: Let's try it out.

jnussbaum:
@seakayone It worked 🥳 https://github.com/dasch-swiss/daschland-scripts/actions/runs/15929963689/job/44949719256

PR #3696, opened 2025-07-24
fix: Ensure groups field is not marked as optional in OpenAPI spec

BalduinLandolt:
Fixes OpenAPI schema generation issue where the `groups` field was incorrectly marked as optional.

Adds custom Tapir schema to explicitly mark the `groups` sequence as non-optional.

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/messages/admin/responder/groupsmessages/GroupsMessagesADM.scala
+++ b/webapi/src/main/scala/org/knora/webapi/messages/admin/responder/groupsmessages/GroupsMessagesADM.scala
@@ -4,6 +4,8 @@
  */
 
 package org.knora.webapi.messages.admin.responder.groupsmessages
+import sttp.tapir.Schema
+import sttp.tapir.generic.auto.*
 import zio.json.DeriveJsonCodec
 import zio.json.JsonCodec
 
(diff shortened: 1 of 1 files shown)


webapi/src/main/scala/org/knora/webapi/messages/admin/responder/groupsmessages/GroupsMessagesADM.scala
```diff
@@ -18,6 +20,14 @@ import org.knora.webapi.slice.admin.domain.model.Group
 final case class GroupsGetResponseADM(groups: Seq[Group]) extends AdminKnoraResponseADM
 object GroupsGetResponseADM {
   implicit val codec: JsonCodec[GroupsGetResponseADM] = DeriveJsonCodec.gen[GroupsGetResponseADM]
+
+  // Custom schema to ensure 'groups' field is not marked as optional in OpenAPI
+  // This addresses the Tapir library issue: https://github.com/softwaremill/tapir/issues/993
+  implicit val schema: Schema[GroupsGetResponseADM] = {
+    // Create a custom schema for Seq[Group] that is not optional
```
seakayone, line 27: True, this is necessary because tapir schema auto derivation deems all collections as being optional, thus treating an empty array or a missing a property the same.

In the long run we should probably figure out how to use the proper type in the case class, e.g. `NonEmptyChunk`. However, in a short attempt I have not been successful in making the auto derivation work with this type yet.

Small hint, instead of trying to provide the `Schema.schemaForIterable` you can directly modify the derived schema as I have done when fixing a similar issue for Julien with the `PagedResponse`:

```scala
inline given [A: Schema]: Schema[PagedResponse[A]] = Schema
    .derived[PagedResponse[A]]
    .modify(_.data)(_.copy(isOptional = false))
```

PR #3736, opened 2025-08-18
fix: Erase value from deleted resource

subotic:
# Allow erasing values from deleted resources

This PR refactors the resource retrieval methods in `ResourcesResponderV2` to separate the core functionality from the business rule that replaces deleted resources with `DeletedResource` objects.

Two new methods were introduced:
- `getResourcePreview` - retrieves resources without replacing deleted ones
- `getResources` - retrieves resources without replacing deleted ones

The existing methods now call these new methods and then apply the deletion replacement logic:
- `getResourcePreviewV2` calls `getResourcePreview`
- `getResourcesV2` calls `getResources`

This change allows the `canRemoveValue` method to use `getResources` directly, which enables erasing values from deleted resources. Previously, this wasn't possible because `getResources` would replace deleted resources with `DeletedResource` objects.

A new test case was added to verify that values can be erased from deleted resources.

Additionally, the previously similar problem with `eraseResourceV2`, was now correctly fixed.

### DIFF
--- a/modules/test-it/src/test/scala/org/knora/webapi/responders/v2/ValuesEraseSpec.scala
+++ b/modules/test-it/src/test/scala/org/knora/webapi/responders/v2/ValuesEraseSpec.scala
@@ -19,6 +19,7 @@ import org.knora.webapi.messages.StringFormatter
 import org.knora.webapi.messages.v2.responder.resourcemessages.CreateResourceRequestV2
 import org.knora.webapi.messages.v2.responder.resourcemessages.CreateResourceV2
 import org.knora.webapi.messages.v2.responder.resourcemessages.CreateValueInNewResourceV2
+import org.knora.webapi.messages.v2.responder.resourcemessages.DeleteOrEraseResourceRequestV2
 import org.knora.webapi.messages.v2.responder.valuemessages.*
 import org.knora.webapi.sharedtestdata.SharedTestDataADM.rootUser
 import org.knora.webapi.slice.admin.domain.model.KnoraProject.Shortcode
@@ -184,6 +185,16 @@ object ValuesEraseSpec extends E2EZSpec {
         val0 <- TestHelper.createIntegerValue(res1)
         val1 <- TestHelper.deleteIntegerValue(val0, res1)
         _    <- TestHelper.eraseIntegerValue(val1, res1)
+      } yield assertCompletes
+    },
+    test("erase a value from a deleted resource") {
+      for {
+        res1 <- TestHelper.createResource
+        val0 <- TestHelper.createIntegerValue(res1)
+        // Get fresh resource data after the value was created to have the current lastModificationDate
+        updatedResource <- TestHelper.findActiveResource(res1.iri)
+        _               <- TestHelper.deleteResource(updatedResource)
+        _               <- TestHelper.eraseIntegerValue(val0, updatedResource)
       } yield assertCompletes
     },
   ).provideSomeAuto(TestHelper.layer, ValueRepo.layer)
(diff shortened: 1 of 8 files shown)


modules/test-it/src/test/scala/org/knora/webapi/responders/v2/ValuesEraseSpec.scala
```diff
@@ -572,6 +583,25 @@ final case class TestHelper(
                 .map(_.toMap)
   } yield ResourceWithValues(created, values)
 
+  def getUpdatedResource(resourceIri: ResourceIri): Task[ActiveResource] = for {
```
seakayone, line 586: Following the naming scheme in the TestHelper I would suggest to call this `findActiveResource`.

The method body completely ignores the fact that the resource was "updated" at some point.

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -589,8 +587,65 @@ final case class ResourcesResponderV2(
...
+   * @param valueUuid         if defined, requests only the value with the specified UUID.
+   * @param versionDate       if defined, requests the state of the resources at the specified time in the past.
+   * @param withDeleted       if defined, indicates if the deleted resource and values should be returned or not.
+   * @param showDeletedValues if false, deleted values will be shown as DeletedValue
+   * @param targetSchema      the target API schema.
+   * @param schemaOptions     the schema options submitted with the request.
+   * @param requestingUser    the user making the request.
+   * @return a [[ReadResourcesSequenceV2]].
+   */
+  def getResources(
```
seakayone, line 637: I find it confusing that we have now two `getResources` methods which produce two different results, how about renaming the other `getResourcesWithDeletedResource` to convey its intent?

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -648,42 +703,63 @@ final case class ResourcesResponderV2(
...
+  /**
+   * Get the preview of a resource as it is in the triplestore.
+   *
+   * @param resourceIris   the resource to query for.
+   * @param withDeleted    indicates if the deleted resource should be returned or not.
+   *
+   * @param requestingUser the user making the request.
+   * @return a [[ReadResourcesSequenceV2]].
+   */
+  private def getResourcePreview(
```
seakayone, line 755: As with my [comment above ](https://github.com/dasch-swiss/dsp-api/pull/3736/files#r2284955424) I find the nearly same name for different methods confusing. It would help me to rename the above to `getResourcePreviewWithDeletedResource`.

subotic:
### Merge activity

* **Aug 19, 2:30 PM UTC**: @subotic merged this pull request with [Graphite](https://app.graphite.dev/github/pr/dasch-swiss/dsp-api/3736).

webapi/src/main/scala/org/knora/webapi/responders/v2/ResourcesResponderV2.scala
```diff
@@ -579,7 +576,54 @@ final case class ResourcesResponderV2(
...
+      propertyIri,
+      valueUuid,
+      versionDate,
+      withDeleted,
+      showDeletedValues,
+      targetSchema,
+      schemaOptions,
+      requestingUser,
+    ).map { apiResponse =>
+      // BR: Deleted resources are to be replaced with DeletedResource.
```
siers, line 601: What does "BR" mean? I can't find any attested meanings.

PR #3739, opened 2025-08-19
refactor: Split ResourcesEndpointsE2ESpec, extract GET /v2/resources tests

seakayone:
### Description
Move all tests for `GET /v2/resources/:resourceIri` to a dedicated file.

### DIFF
--- a/modules/test-e2e/src/test/scala/org/knora/webapi/e2e/v2/ResourcesEndpointsGetResourcesE2ESpec.scala
+++ b/modules/test-e2e/src/test/scala/org/knora/webapi/e2e/v2/ResourcesEndpointsGetResourcesE2ESpec.scala
@@ -10,6 +10,7 @@ import sttp.client4.UriContext
 import sttp.model.MediaType
 import zio.ZIO
 import zio.test.assertTrue
+
 import org.knora.webapi.E2EZSpec
 import org.knora.webapi.e2e.InstanceChecker
 import org.knora.webapi.e2e.v2.ResourcesRouteV2E2ESpec.aThingWithHistoryIri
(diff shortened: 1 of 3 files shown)


siers:
@BalduinLandolt if a wrapping method `test2` (or something) would be created that would fix the title, would that be satisfactory? it wouldn't feel like a brand new test framework then

seakayone:
I have used zio's `Gen` for property based testing in the past and when property based testing is the approach this is the way to go.

However my approach in this is parameterized testing. There is a difference.

Property-based testing checks whether a program holds for a wide range of automatically generated inputs by asserting general properties, while parameterized testing runs the same test logic against a fixed set of explicitly provided input values. 

While you can use `check(Gen.fromIterable(testCases)` for a parameterized approach, it does not create different tests with different descriptions. I wanted to retain the test descriptions as they contained vital information about why this case was tested, which could not easily be derived from the parameter values.

IMHO, the thirty lines of common test setup in the TestCase are far from inventing an own test framework. I am simply using zio and zio-test to generate standard zio tests.

seakayone:
As for your argument, that you cannot run a single test by itself: That is true.

I have chosen this approach nonetheless for another advantage. That it now is easy do understand the differences between each and every test because only the parameters differ and not the code.

BalduinLandolt:
> I do not have a strong opinion. It saddens me that the advantages I was seeing were not reflected in you arguments. However, I value you feedback and have inlined the test by [replacing generating the test with duplicating the code](https://github.com/dasch-swiss/dsp-api/pull/3739/commits/93c90bc2fba19e4b43c1c29a2720425c13c601fe).

That was because I had not seen it - GitHub does not seem to load in new messages while one is writing a comment, and I had started my message while there was only your first reply there. It was not my intention to ignore your argument.

Thank you for being accommodating of my feedback!

BalduinLandolt (approved):

BalduinLandolt:
@seakayone thanks for pointing this out. I have to say, though, that this does not change my opinion in the matter: To me, this pattern seems more like an obfuscation than a simplification. But maybe at the end of the day, this is a matter of taste.  
If you feel that our code base would really profit from dynamic test generation, feel free to write an RFC so we can have a proper discussion on the matter; and with an ADR spawned from it, it would at least not be undocumented and not such an outlier in regard to all the other tests.

PR #3753, opened 2025-08-28
fix: Fix double encoding of comments when creating a resource (DEV-5356)

seakayone:
### Description

Adds tests to ensure create resource and update value with comments with line breaks are working correctly.

Removes double encoding of comment.

Related to https://github.com/dasch-swiss/dsp-api/pull/3737


siers (approved):
LGTM. Does this interfere with Balduin's efforts or is it another issue entirely?

seakayone:
> LGTM. Does this interfere with Balduin's efforts or is it another issue entirely?

It is related for sure and stems from the same cause - the create query was migrated to rdf4j's sparql builder.

PR #3774, opened 2025-09-15
chore: Bump tapir (include zio-http) (DEV-4405)

siers:


seakayone (commented):
Compile error because of deprecation:
> ```in trait ZIO is deprecated since 2.1.21: Use `catchAll```

PR #3797, opened 2025-09-23
fix: Gravsearch limiting to project wrongly excludes knora-api entities

BalduinLandolt:
### Description

Previously, when limiting a Gravsearch  (or Gravsearch count) query to a project, it would only consider resource classes defined in the ontologies of the project.  
This wrongly excluded any classes of a type defined in `knora-base`, such as `LinkObj` or `Region` etc.

The fix adds the `knora-base` ontology to the ontologies considered for search when `limitToProject` is set.


BalduinLandolt:
@seakayone tests have been added.

I had to jump more hoops than expected, also it did not behave fully correctly, so that needed to be fixed.  
If anything is unclear, I'm happy to clarify.

PR #4068, opened 2026-04-09
refactor: Migrate to rdf4j gravsearch getResourceWithSpecificProperties (DEV-6230)

seakayone:
- **refactor: Migrate to rdf4j getResourceWithSpecifiedProperties**


seakayone:
@claude review, check that both the rdf4j query and the removed twirl template query are sematically identical, also check if the resulting query is valid SPARQL.

siers (approved):

PR #4072, opened 2026-04-13
refactor: Migrate ResourceIri and ValueIri to standalone types (DEV-6231)

seakayone:
Extracts `ResourceIri` and `ValueIri` from the `KnoraIris` hierarchy into standalone `StringValue` case classes with built-in validation, removing the dependency on `SmartIri` for construction and usage. This eliminates unnecessary indirection through `IriConverter` and `StringFormatter` at call sites, simplifying both production code and tests.

Unlike ontology, class, and property IRIs, `ResourceIri` and `ValueIri` do not have different ontology schema representations (complex vs. simple), so they never needed the `SmartIri` machinery. This makes them safe to extract into plain validated string types.

Also consolidates the `ResourceId` opaque type into the `ResourceIri` companion and removes dead code from `StringFormatter` and `IriConverter`.

Slightly [related is the documentation update PR](https://github.com/dasch-swiss/dsp-api/pull/4073).


siers (approved):

siers:
> Unlike ontology, class, and property IRIs, ResourceIri and ValueIri do not have different ontology schema representations (complex vs. simple), so they never needed the SmartIri machinery. This makes them safe to extract into plain validated string types.

👍

PR #4074, opened 2026-04-14
fix(tracing): Fix traceparent propagation from dsp-app through API to Fuseki (DEV-5672)

seakayone:
The OpenTelemetry Java agent's thread-local context is incompatible with ZIO's fiber-based execution model. This caused the agent to overwrite SDK-injected `traceparent` headers with unlinked trace contexts, and the Tapir `OpenTelemetryTracing` interceptor's server spans were invisible to ZIO's `FiberRef`-based `ContextStorage`.

Replace the Tapir interceptor with a ZIO-native middleware that extracts incoming `traceparent` and creates server spans via ZIO `Tracing`. Disable the Java agent's HTTP client and Netty auto-instrumentation on the API container so the SDK-based propagation is not overwritten. Add a local dsp-app config with Faro tracing enabled and CORS URLs configured for local development.

Related config change: https://github.com/dasch-swiss/ops-deploy/pull/1279


siers:
So the automatic instrumentation is now only for ingest?
What's the app-config? Is it duplicating any of the conf files in `resources/`?

siers (approved):

seakayone:
> So the automatic instrumentation is now only for ingest? What's the app-config?

good point, I haven't looked into ingest, but this is likely affected by the same problem thread-local vs. zio-fibers. I will tackle this in a follow up issue.

> Is it duplicating any of the conf files in `resources/`?

This file is referenced by the docker-compose and configures the app in the localdev environment to send traces to the local dev environment otel stack.

PR #4086, opened 2026-04-20
fix(project-migration): Include system admin project members as normal users in export (DEV-6281)

seakayone:
Project members or project admins who were also members of the SystemAdmin group on the source instance were silently dropped from project exports, so they disappeared from the project after re-import.

The export SPARQL `CONSTRUCT` no longer filters out system-admin users; instead the `isInSystemAdminGroup` flag is rewritten to `false` in the exported admin data. The source instance's system-admin membership therefore does not leak into the target instance, but the user keeps their project/group memberships.

Fixes [DEV-6281](https://linear.app/dasch/issue/DEV-6281/projectadminprojectmember-wird-falschlich-aus-dem-projekt-entfernt).

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/slice/export/domain/AdminModelScoping.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/export/domain/AdminModelScoping.scala
@@ -23,7 +23,7 @@ import org.knora.webapi.messages.OntologyConstants.KnoraAdmin
 object AdminModelScoping {
 
   // Shared Jena property/resource constants for admin model operations.
-  // Package-private so rewriteExistingUserTriples / rewriteNewUserTriples can reuse them.
+  // Package-private so the import service (ProjectMigrationImportService) can reuse them.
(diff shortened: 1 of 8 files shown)


webapi/src/main/scala/org/knora/webapi/slice/export/domain/AdminModelScoping.scala
```diff
@@ -94,4 +94,37 @@ object AdminModelScoping {
...
+      if (existing.isEmpty) None
+      else {
+        val wasTrue =
+          existing.exists(stmt => stmt.getObject.isLiteral && stmt.getLiteral.getLexicalForm.equalsIgnoreCase("true"))
+        existing.foreach(model.remove)
+        val _ = model.add(user, isInSystemAdminGroupProp, falseLiteral)
+        Option.when(wasTrue)(user.getURI)
+      }
+    }
+  }
```
siers, line 129: super duper wordy this function

siers (approved):
the change is so deep in the context, it is hard to follow

PR #4087, opened 2026-04-21
feat: Move Fuseki Docker image publishing to dsp-api (DEV-6277)

mpro7:
## Summary

Moves the Fuseki Docker image build and publish pipeline from `docker-apache-jena-fuseki` into `dsp-api`. Migration lands at the current production version `5.5.0-3`, fixing the pre-existing version drift between `Dependencies.scala` (was `5.5.0-2`) and `docker-compose.yml` (`5.5.0-3`).

- `fuseki/`: Dockerfile with `ARG FUSEKI_VERSION=5.5.0-3` (image tag, read by Makefile) + `ARG JENA_VERSION=5.5.0` (Apache binary, used for download/verify), plus four verbatim files from the old repo
- `Makefile`: `docker-build-fuseki-image` and `docker-publish-fuseki-image` targets; `docker-build-fuseki-image` added as a dependency of `docker-build`
- `docker-compose.yml`: `build: context: ./fuseki` added to the `db` service
- `project/Dependencies.scala`: `fusekiImage` bumped from `5.5.0-2` to `5.5.0-3`
- `build-and-test.yml`: conditional Fuseki image bootstrap in `test-it` and `test-e2e` jobs; new `check-fuseki-version-consistency` job asserting Dockerfile, Dependencies.scala, and docker-compose.yml all carry the same version tag
- `docker-publish-fuseki.yml`: new workflow (push to main, `fuseki/**` path filter) with QEMU, Docker Hub publish, and Jenkins DEV trigger (`Service=db&Version=<version>`)
- `publish-release.yml` + `publish-from-branch.yml`: add QEMU, update `setup-buildx-action` to `@v3`, add explicit `make docker-publish-fuseki-image` step

## Test plan

- [x] `make docker-build-fuseki-image` builds the image locally
- [x] Trigger `publish-from-branch.yml` manually from this branch to verify end-to-end build/push
- [x] Confirm `test-it` and `test-e2e` jobs pass
- [ ] After merging: verify Jenkins `trigger-dev-deployment` step succeeds and `RELEASE-DEV.json` updates
- [ ] After merging: archive `docker-apache-jena-fuseki` (confirm no undocumented consumers with Lukas Stoeckli first)

## Related

- Spec: dasch-specs PR #88
- Linear: DEV-6277

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -223,7 +223,7 @@ jobs:
       - name: Assert fuseki/Dockerfile, Dependencies.scala, and docker-compose.yml agree
         run: |
           set -euo pipefail
-          DOCKERFILE_VER=$(grep '^ARG FUSEKI_VERSION=' fuseki/Dockerfile | cut -d= -f2 | tr -d '"[:space:]')
+          DOCKERFILE_VER=$(grep '^ARG IMAGE_VERSION=' fuseki/Dockerfile | cut -d= -f2 | tr -d '"[:space:]')
           DEPS_VER=$(grep 'val fusekiImage' project/Dependencies.scala | sed 's/.*:\([^"]*\)".*/\1/' | tr -d '[:space:]')
           COMPOSE_VER=$(grep 'image: daschswiss/apache-jena-fuseki:' docker-compose.yml | sed 's/.*apache-jena-fuseki://' | tr -d '[:space:]')
           echo "fuseki/Dockerfile:          $DOCKERFILE_VER"
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -215,6 +215,24 @@ jobs:
       - name: Build docs
         run: just docs-build
 
+  check-fuseki-version-consistency:
+    name: Check Fuseki version consistency
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Assert fuseki/Dockerfile, Dependencies.scala, and docker-compose.yml agree
+        run: |
+          set -euo pipefail
+          DOCKERFILE_VER=$(grep '^ARG FUSEKI_VERSION=' fuseki/Dockerfile | cut -d= -f2 | tr -d '"[:space:]')
+          DEPS_VER=$(grep 'val fusekiImage' project/Dependencies.scala | sed 's/.*:\([^"]*\)".*/\1/' | tr -d '[:space:]')
+          COMPOSE_VER=$(grep 'image: daschswiss/apache-jena-fuseki:' docker-compose.yml | sed 's/.*apache-jena-fuseki://' | tr -d '[:space:]')
+          echo "fuseki/Dockerfile:          $DOCKERFILE_VER"
+          echo "project/Dependencies.scala: $DEPS_VER"
+          echo "docker-compose.yml:         $COMPOSE_VER"
+          [ "$DOCKERFILE_VER" = "$DEPS_VER" ] || { echo "ERROR: Dockerfile ($DOCKERFILE_VER) != Dependencies.scala ($DEPS_VER)"; exit 1; }
+          [ "$DOCKERFILE_VER" = "$COMPOSE_VER" ] || { echo "ERROR: Dockerfile ($DOCKERFILE_VER) != docker-compose.yml ($COMPOSE_VER)"; exit 1; }
+          echo "All Fuseki versions match: $DOCKERFILE_VER"
+
   check-formatting:
     name: Check formatting
     runs-on: ubuntu-latest
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -106,6 +106,14 @@ jobs:
         uses: dasch-swiss/dsp-api/.github/actions/preparation@main
         with:
           java-version: ${{ env.JAVA_VERSION }}
+      - name: Build Fuseki image locally (only when fuseki/ changed)
+        shell: bash
+        run: |
+          set -euo pipefail
+          CHANGED=$(git diff --name-only origin/${{ github.base_ref || 'main' }}...HEAD)
+          if echo "$CHANGED" | grep -q "^fuseki/"; then
+            make docker-build-fuseki-image
+          fi
       - name: Run integration tests
         run: make test-it
       - name: WebApi IT Test Report
@@ -136,6 +144,14 @@ jobs:
         uses: dasch-swiss/dsp-api/.github/actions/preparation@main
         with:
           java-version: ${{ env.JAVA_VERSION }}
+      - name: Build Fuseki image locally (only when fuseki/ changed)
+        shell: bash
+        run: |
+          set -euo pipefail
+          CHANGED=$(git diff --name-only origin/${{ github.base_ref || 'main' }}...HEAD)
+          if echo "$CHANGED" | grep -q "^fuseki/"; then
+            make docker-build-fuseki-image
+          fi
       - name: Run end-to-end tests
         run: ./sbtx "test-e2e/test"
       - name: WebApi E2E Test Report
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -106,7 +106,7 @@ jobs:
         uses: dasch-swiss/dsp-api/.github/actions/preparation@main
         with:
           java-version: ${{ env.JAVA_VERSION }}
-      - name: Build Fuseki image locally (only when fuseki/ changed)
+      - name: Build Fuseki image if fuseki/ changed
         shell: bash
         run: |
           set -euo pipefail
@@ -144,7 +144,7 @@ jobs:
         uses: dasch-swiss/dsp-api/.github/actions/preparation@main
         with:
           java-version: ${{ env.JAVA_VERSION }}
-      - name: Build Fuseki image locally (only when fuseki/ changed)
+      - name: Build Fuseki image if fuseki/ changed
         shell: bash
         run: |
           set -euo pipefail
--- a/.github/workflows/docker-publish-fuseki.yml
+++ b/.github/workflows/docker-publish-fuseki.yml
@@ -1,0 +1,61 @@
+name: Docker publish Fuseki image
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'fuseki/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false  # never cancel a publish mid-push
+
+permissions:
+  contents: read
+
+jobs:
+  publish-fuseki:
+    name: Publish Fuseki image
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    outputs:
+      version: ${{ steps.fuseki_version.outputs.version }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Get Fuseki version
+        id: fuseki_version
+        run: |
+          set -euo pipefail
+          V=$(grep '^ARG FUSEKI_VERSION=' fuseki/Dockerfile | cut -d= -f2 | tr -d '"[:space:]')
+          [[ "$V" =~ ^[A-Za-z0-9._-]+$ ]] || { echo "FUSEKI_VERSION invalid: $V"; exit 1; }
+          echo "version=$V" >> "$GITHUB_OUTPUT"
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+      - name: Build and publish Fuseki image to Docker Hub
+        run: make docker-publish-fuseki-image
+
+  trigger-dev-deployment:
+    name: Trigger deployment to DEV
+    runs-on: ubuntu-latest
+    needs: publish-fuseki
+    permissions: {}
+    steps:
+      - name: Trigger deployment to DEV
+        env:
+          FUSEKI_VERSION: ${{ needs.publish-fuseki.outputs.version }}
+          JENKINS_BASIC_AUTH_USER: ${{ secrets.JENKINS_BASIC_AUTH_USER }}
+          JENKINS_BASIC_AUTH_PASS: ${{ secrets.JENKINS_BASIC_AUTH_PASS }}
+          JENKINS_DEV_WEBHOOK: ${{ secrets.JENKINS_DEV_WEBHOOK }}
+        run: |
+          curl -f -u "${JENKINS_BASIC_AUTH_USER}:${JENKINS_BASIC_AUTH_PASS}" \
+               -X POST "${JENKINS_DEV_WEBHOOK}" \
+               --data-urlencode "Service=db" \
+               --data-urlencode "Version=${FUSEKI_VERSION}"
--- a/.github/workflows/docker-publish-fuseki.yml
+++ b/.github/workflows/docker-publish-fuseki.yml
@@ -27,8 +27,8 @@ jobs:
         id: fuseki_version
         run: |
           set -euo pipefail
-          V=$(grep '^ARG FUSEKI_VERSION=' fuseki/Dockerfile | cut -d= -f2 | tr -d '"[:space:]')
-          [[ "$V" =~ ^[A-Za-z0-9._-]+$ ]] || { echo "FUSEKI_VERSION invalid: $V"; exit 1; }
+          V=$(grep '^ARG IMAGE_VERSION=' fuseki/Dockerfile | cut -d= -f2 | tr -d '"[:space:]')
+          [[ "$V" =~ ^[A-Za-z0-9._-]+$ ]] || { echo "IMAGE_VERSION invalid: $V"; exit 1; }
           echo "version=$V" >> "$GITHUB_OUTPUT"
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
--- a/.github/workflows/publish-from-branch.yml
+++ b/.github/workflows/publish-from-branch.yml
@@ -19,4 +19,3 @@ jobs:
         run: |
           echo ${{ secrets.DOCKER_HUB_TOKEN }} | docker login -u ${{ secrets.DOCKER_USER }} --password-stdin
           make docker-publish
-          make docker-publish-fuseki-image
--- a/.github/workflows/publish-from-branch.yml
+++ b/.github/workflows/publish-from-branch.yml
@@ -10,8 +10,6 @@ jobs:
     steps:
       - name: Run preparatory steps
         uses: dasch-swiss/dsp-api/.github/actions/preparation@main
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v3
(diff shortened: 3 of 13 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -136,6 +144,14 @@ jobs:
...
         with:
           java-version: ${{ env.JAVA_VERSION }}
+      - name: Build Fuseki image if fuseki/ changed
+        shell: bash
+        run: |
+          set -euo pipefail
+          CHANGED=$(git diff --name-only origin/${{ github.base_ref || 'main' }}...HEAD)
+          if echo "$CHANGED" | grep -q "^fuseki/"; then
+            make docker-build-fuseki-image
+          fi
```
seakayone, line 154: (requested change) IT and E2E test should only use published Docker images.

I think we should not build the fuseki image for our dsp-api tests. In general I would like the tests to depend on a defined, stable and published version. This way we can be sure it is testing against the exact same Docker image we will be using in production.

IMHO, when we need to update the fuseki version this should be done in a separate pull request and our new Fuseki Docker image will be published giving us the stable version. The integration of the new version into dsp-api should then be done in a follow up task. 

IMHO, conditionally building the docker image twice for our tests runs does not provide value, adds complexity and makes the tests potentially slower.

.github/workflows/build-and-test.yml
```diff
@@ -199,6 +215,24 @@ jobs:
       - name: Build docs
         run: just docs-build
 
+  check-fuseki-version-consistency:
+    name: Check Fuseki version consistency
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Assert fuseki/Dockerfile, Dependencies.scala, and docker-compose.yml agree
```
seakayone, line 223: As described [in my comment above](https://github.com/dasch-swiss/dsp-api/pull/4087/changes#r3122275138) I think that updating the Fuseki docker container and integration into dsp-api can and may be separate tasks making it possible for `fuseki/Dockerfile` and `docker-compose.yml` having version drift.

In `Dependencies.scala` we have a `fuskeiImage` version that should match the `docker-compose.yml` container tag. 

I like your thinking of replacing the current comment with a ci test. Though to be honest, I feel like the `grep`ping and `sed`ding through a bunch of files really feels brittle and might become more effort than it helps us. We can try it out, for now.

Makefile
```diff
@@ -67,8 +67,29 @@ docker-publish-ingest-image: # publish ingest image to Dockerhub
...
+	  -t $(FUSEKI_DOCKER_IMAGE) \
+	  --push \
+	  fuseki/
+
 .PHONY: docker-build
-docker-build: docker-build-dsp-api-image docker-build-sipi-image docker-build-ingest-image ## build and publish all Docker images locally
+docker-build: docker-build-dsp-api-image docker-build-sipi-image docker-build-ingest-image docker-build-fuseki-image ## build and publish all Docker images locally
 
 .PHONY: docker-publish
 docker-publish: docker-publish-dsp-api-image docker-publish-sipi-image docker-publish-ingest-image ## publish all Docker images to Dockerhub
```
seakayone: `docker-build` now also builds Fuseki. 
It might be a good idea to have a comment on `docker-publish` to explain why you deliberately left out the `docker-publish-fuseki-image`.

seakayone (changes requested):

.github/workflows/build-and-test.yml
mpro7: TLDR:  Keeping Option A. The conditional build is the architecturally consistent choice given Fuseki now lives in dsp-api — the whole point is to validate Dockerfile and app changes together. Option B re-introduces the two-repo coordination overhead we just eliminated. The main downside (testing against an unpublished image) is mitigated by IT/E2E passing before the merge triggers the publish.

Option A: Keep conditional build                                                                                    
                                                                                                                      
Gains:
  - Atomic PR — Dockerfile, version bump, and any app-side config changes land together                               
  - IT/E2E validate the new image against dsp-api before it reaches Docker Hub                                        
  - Faster feedback — you discover Fuseki/app incompatibilities before publishing                                     
  - Aligns with the monorepo principle: infra and the code that uses it testable together (the whole point of moving Fuseki into dsp-api)                                                                                                
                                                                                                                      
Losses:                                                                                                             
  - CI tests against an unpublished, amd64-only local image — not the exact multi-arch artifact that goes to          
  production                                                                                                          
  - Extra conditional logic in CI (git diff, grep) — brittle if base ref changes or the grep mismatches
  - Two additional steps in test-it and test-e2e, slightly slower when fuseki/ changed                                
  - If the conditional logic has a bug, the build silently skips — hard to notice                                     
                                                                                                                      
Option B: Two-PR flow                                                                                               

Gains:                                                                                                              
  - Tests always run against a known, published, immutable multi-arch image — identical to what production uses
  - Simpler CI — no conditional logic, no git diff                                                                    
  - Clean separation: "publish new Fuseki" vs "adopt new Fuseki in dsp-api"
                                                                                                                      
Losses:         
  - Every Fuseki change requires two sequenced PRs with coordination overhead
  - No pre-publish validation that the new Dockerfile actually works with dsp-api — you discover breakage only after the image is already on Docker Hub                                                                                
  - Logically coupled changes (Dockerfile + app config) artificially split across two PRs                             
  - Partially defeats the purpose of moving Fuseki into dsp-api in the first place
                                                                                                                      
Best practice:  
The right answer depends on change frequency and coupling. The industry split is:                                   
  - Monorepo/tight coupling: build locally in CI, publish on merge — Option A. Standard in monorepos (Nx, Turborepo, Google) where infra and app code coevolve.                                                                          
  - Separate repos/loose coupling: always test against published images — Option B. Correct when the Dockerfile and the app are truly independent.                                                                                      
                                                                                                                      
Since we just moved Fuseki into dsp-api specifically to enable atomic coevolution, Option A is the architecturally consistent choice. Option B is the right pattern for the old world (separate docker-apache-jena-fuseki repo). Adopting it now would give us the worst of both worlds: Fuseki in dsp-api but still requiring two PRs per change.
                                                                                                                   
The one concrete risk in Option A — publishing a broken image — is mitigated by the fact that IT/E2E run against the locally-built image first. If they pass, the publish is safe.

.github/workflows/build-and-test.yml
mpro7: Fair point — keeping it for now as it's already caught real drift. Happy to revisit if it causes friction.

Makefile
mpro7: Done in [73f78f6](https://github.com/dasch-swiss/dsp-api/pull/4087/commits/73f78f6862d91b8b24defdb0441cc03e2b217db4).

seakayone:
(request changes) Is this publishing the Fuseki image twice? — Confirmed, worse than twice.

  The same tag daschswiss/apache-jena-fuseki:<FUSEKI_VERSION> is pushed from three independent call sites with no cross-workflow coordination:

  ```
  ┌──────────────────────────────────────────┬──────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────┐
  │                 Trigger                  │                            Caller                            │                       Line                        │
  ├──────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────┤
  │ push: branches: [main], paths: fuseki/** │ docker-publish-fuseki.yml → make docker-publish-fuseki-image │ .github/workflows/docker-publish-fuseki.yml:42-43 │
  ├──────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────┤
  │ release: [published]                     │ publish-release.yml → make docker-publish-fuseki-image       │ .github/workflows/publish-release.yml:24          │
  ├──────────────────────────────────────────┼──────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────┤
  │ workflow_dispatch (any branch)           │ publish-from-branch.yml → make docker-publish-fuseki-image   │ .github/workflows/publish-from-branch.yml:22      │
  └──────────────────────────────────────────┴──────────────────────────────────────────────────────────────┴───────────────────────────────────────────────────┘
```
IMHO we only need a single publishing for stable Fuseki version.

 Concrete race: A release-please merge that touches fuseki/** fires docker-publish-fuseki.yml (on the push) and publish-release.yml (on the release) concurrently — both docker buildx build --push to the same tag with no shared concurrency group. Whichever finishes last wins; trigger-dev-deployment can be signalled against
  whichever manifest happens to be on Docker Hub at webhook time.

NOTE: the Makefile comment at Makefile:94-95 — touched by the most recent chore commit 73f78f686 ("clarify docker-publish excludes Fuseki image") — says docker-publish excludes Fuseki, but the two release workflows then unconditionally add a second make docker-publish-fuseki-image step. The comment documents the Makefile target
   in isolation but misrepresents the combined CI behavior.
   
   
Additionally make sure the publishing of the Fuseki Docker image is depending on the it/e2e tests to succeed first.

Once you fix the two following points I am happy to approve:
1. Multiple Fuseki Docker image publishing replaced by single stable Fuseki image
2. Make Fuseki Docker image publishing dependent on the success of the it/e2e test.

File: .github/workflows/build-and-test.yml (hunk 1 above)
Comment (seakayone) [reply .github/workflows/build-and-test.yml:154]: The additional value I have not seen yet is "IT/E2E validate the new image against dsp-api before it reaches Docker Hub". My main concern of "test against what goes to production" is addressed by the fact that we have a post-merge run in our release PRs which test against the published docker image. 

So as far as I understand what your are proposing is:
Build and test the Fuseki Docker image locally within the same PR as the api code changes, then publish it only after merge, enabling atomic validation of tightly coupled code and infrastructure.

I can go with this approach. I do not advocate for a two stage PR workflow anymore.

However, the current implementation does not exactly do this .
Currently the Fuseki Docker Image is [published multiple times](https://github.com/dasch-swiss/dsp-api/pull/4087#issuecomment-4295071832) _and_ the publishing of the Fuseki Docker image is unconditional as it does not depend on the it/e2e tests to succeed first.

File: .github/workflows/docker-publish-fuseki.yml [hunk 4]
```diff
@@ -0,0 +1,61 @@
...
+        env:
+          FUSEKI_VERSION: ${{ needs.publish-fuseki.outputs.version }}
+          JENKINS_BASIC_AUTH_USER: ${{ secrets.JENKINS_BASIC_AUTH_USER }}
+          JENKINS_BASIC_AUTH_PASS: ${{ secrets.JENKINS_BASIC_AUTH_PASS }}
+          JENKINS_DEV_WEBHOOK: ${{ secrets.JENKINS_DEV_WEBHOOK }}
+        run: |
+          curl -f -u "${JENKINS_BASIC_AUTH_USER}:${JENKINS_BASIC_AUTH_PASS}" \
+               -X POST "${JENKINS_DEV_WEBHOOK}" \
+               --data-urlencode "Service=db" \
+               --data-urlencode "Version=${FUSEKI_VERSION}"
```
SamuelBoerlin, line 61: ~~Since we are now using only the `FUSEKI_VERSION` and dropped the `-N` suffix (e.g. `5.5.0-3`), we may be forced to re-publish with the same tag even if we have to make adjustments to our `Dockerfile` for whatever reason.
This may be fine depending on the changes to the Dockerfile, but since it is now a possibility that we could or have to re-publish under the same tag, we should pin `Version` here to the image index digest, i.e. `6.0.0@sha256:0e3ff6c760fc735bd2c1ef4d4a376ee66535c7c19f1b270073e518de7b21a824` instead of just `6.0.0`, so that we can ensure consistent deployments.
However, this will also require a small change in ops-deploy for which I'll create a PR momentarily.~~
Actually never mind, the `FUSEKI_VERSION` format is still the same as before :)

.github/workflows/docker-publish-fuseki.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: docker/setup-buildx-action@v3
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+      - name: Build and publish Fuseki image to Docker Hub
+        run: make docker-publish-fuseki-image
+
+  trigger-dev-deployment:
```
seakayone, line 45: Do we need a second dev deployment trigger?

For now any Fuseki Version update in ops-deploy is meant to stay manual, isn't it?
If so then this here should not exist. In fact, then this trigger causes a deployment for the db without a version change. 

We have another trigger which takes care of automatically deploying the api:
```bash
workflows/docker-publish.yml
30:  trigger-dev-deployment:
```

The new trigger here might even interfere with the api deployment trigger, both triggered workflows may be overlapping, I am not sure if that is supported by the Jenkins jobs or if this will cause a problem. I'd recommend to remove the db deployment trigger here.

seakayone: After clarifying with @SamuelBoerlin this is a non issue.

Either way in which order those deployment triggers are fired they will not interfere with each other.
We have two subsequent deployments on dev and eventually will have the desired versions deployed.
The additional downtime and possibly failing first deployments (old api does not like new db version or other way around) will be rectified with the second deployment, this is an acceptable side effect for now - especially given this only happens a couple of times a year in the dev environment.

fuseki/Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+       tini bash curl ca-certificates findutils coreutils pwgen procps \
+    ; \
+    rm -rf /var/lib/apt/lists/*
+
+
+# Update below according to https://jena.apache.org/download/
+# and checksum for apache-jena-fuseki-<version>.tar.gz.sha512
+ARG FUSEKI_VERSION=5.5.0-3
```
SamuelBoerlin: Nit: I'd prefer we call this something like `IMAGE_VERSION` (our version we publish, including the `-3` suffix) and keep `FUSEKI_VERSION` as the actual Fuseki version.

Makefile
```diff
@@ -67,11 +67,32 @@ docker-publish-ingest-image: # publish ingest image to Dockerhub
...
+	  --push \
+	  fuseki/
+
 .PHONY: docker-build
-docker-build: docker-build-dsp-api-image docker-build-sipi-image docker-build-ingest-image ## build and publish all Docker images locally
+docker-build: docker-build-dsp-api-image docker-build-sipi-image docker-build-ingest-image docker-build-fuseki-image ## build and publish all Docker images locally
 
 .PHONY: docker-publish
-docker-publish: docker-publish-dsp-api-image docker-publish-sipi-image docker-publish-ingest-image ## publish all Docker images to Dockerhub
+docker-publish: docker-publish-dsp-api-image docker-publish-sipi-image docker-publish-ingest-image ## publish all Docker images to Dockerhub (Fuseki excluded: published separately via docker-publish-fuseki-image)
```
SamuelBoerlin: Nit: this is getting a bit confusing IMO as it's inconsistent with `docker-build`. Maybe something like `docker-publish-dsp-images`?

SamuelBoerlin (changes requested):
One requests, but otherwise looks okay to me besides the points @seakayone already raised:
I think we should continue to treat publishing of Fuseki independently of dsp-api releases for now, as it has been this way so far, so we should only publish on merge in `docker-publish-fuseki.yml‎` and not in the other workflows. We could revisit this in the future though.  
I had assumed that the pre-merge IT/E2E test would be sufficient, so rerunning them after merge would not be necessary, as it wouldn't allow you to merge if they failed. But I can also see how it would be safer to always re-run them again post-merge/pre-publish anyways just in case.  

Request: since we plan on archiving `dasch-swiss/docker-apache-jena-fuseki` after this, can we please have a brief README.md in `fuseki/` with some key infos:
- what this image is
- that the DB comes pre-configured with the `dsp-repo` dataset as defined in `dsp-repo.ttl`
- configuration (`ADMIN_PASSWORD` env var)
- how to we update Jena/Fuseki (-> update FUSEKI_VERSION, JENA_VERSION, FUSEKI_SHA512, docker-compose etc.)
- how to publish (-> new pr, publish on merge)

mpro7:
> Additionally make sure the publishing of the Fuseki Docker image is depending on the it/e2e tests to succeed first.
> 
> Once you fix the two following points I am happy to approve:
> 
> 1. Multiple Fuseki Docker image publishing replaced by single stable Fuseki image
> 2. Make Fuseki Docker image publishing dependent on the success of the it/e2e test.

1. Fixed in [63da517](https://github.com/dasch-swiss/dsp-api/pull/4087/commits/63da517fbee1babf3c77202ecb06fa10e770de53). ~~The `publish-from-branch.yml` is manually triggered workflow for deployment from the branch before merging. It should stay there. No automatic race is possible.~~ This is not implemented as it should be for testing purposes, so I removed it from `publish-from-branch.yml` too in [7f9b256](https://github.com/dasch-swiss/dsp-api/pull/4087/commits/7f9b25644861874ddc1f668365d7b34cfde4c76d).
2. Fuseki publish already implicitly depends on IT/E2E passing — `docker-publish-fuseki.yml` only fires on push to main, which requires branch protection checks (including IT/E2E) to pass before the merge is allowed. As from my understanding. no additional workflow-level dependency is needed.

Makefile
mpro7: This pointed actually the real problem, which IMO is the addition `docker-build-fuseki-image` as a dependency of `docker-build` in this PR. If Fuseki publish is intentionally separate, Fuseki build should be separate too. The cleaner fix would be probably:
- `remove docker-build-fuseki-image` from `docker-build` dependencies
- developers who need a local Fuseki build run `make docker-build-fuseki-image` explicitly
- for the stack, Docker pulls the published image automatically if not locally present
- both `docker-build` and `docker-publish` then consistently exclude Fuseki
Does it make sense?

fuseki/Dockerfile
mpro7: Good point, added in [3b5975b](https://github.com/dasch-swiss/dsp-api/pull/4087/commits/3b5975ba6f2cad35a15f73655962d2ccbf0c0aa7) as well as the README file. Let me know if it looks good or there are missing/redundant things.

Makefile
SamuelBoerlin: Yeah, that's also a reasonable option 👍

fuseki/Dockerfile
SamuelBoerlin: Thanks! Looks good.

.github/workflows/publish-release.yml
```diff
@@ -12,9 +12,11 @@ jobs:
     steps:
       - name: Run preparatory steps
         uses: dasch-swiss/dsp-api/.github/actions/preparation@main
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
```
SamuelBoerlin: This is redundant now that the fuseki/build publish was removed from this workflow, no?

.github/workflows/publish-from-branch.yml
```diff
@@ -10,9 +10,11 @@ jobs:
     steps:
       - name: Run preparatory steps
         uses: dasch-swiss/dsp-api/.github/actions/preparation@main
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
```
SamuelBoerlin: This is redundant now that the fuseki/build publish was removed from this workflow, no?

Makefile
mpro7: Great! Fixed here: [92333a3](https://github.com/dasch-swiss/dsp-api/pull/4087/commits/92333a3e316fb18b87f764f09d1a1cefd993211f).

SamuelBoerlin (approved):
Probably should remove those two now redundant "Set up QEMU" steps, otherwise LGTM!

.github/workflows/publish-from-branch.yml
mpro7: Good catch! Removed here: [80fd02f](https://github.com/dasch-swiss/dsp-api/pull/4087/commits/80fd02f368275924de95dfa334ac9fd3df57c7a0).

.github/workflows/publish-release.yml
mpro7: Good catch! Removed here: [80fd02f](https://github.com/dasch-swiss/dsp-api/pull/4087/commits/80fd02f368275924de95dfa334ac9fd3df57c7a0).

seakayone (approved):

PR #4088, opened 2026-04-23
feat(tracing): Use endpoint path template for low-cardinality span names and add trace<->log correlation (DEV-6292)

seakayone:
Span names included the fully-resolved request path (e.g. `HTTP GET /admin/lists/http%3A%2F%2Frdfh.ch%2Flists%2F...`), which exploded the cardinality of `traces_spanmetrics_latency_bucket` in Grafana. OTel HTTP semconv prescribes low-cardinality span names based on the route template, not the realized URL.

The outer zio-http middleware now names the span `HTTP ${method}` only — the semconv-compliant fallback when no route is known. Once Tapir matches an endpoint, a new `EndpointInterceptor` renames the active span to `HTTP ${method} ${endpoint.showPathTemplate}` (e.g. `HTTP GET /admin/lists/{listIri}`), pulling the current span from the fiber-backed `ContextStorage`. `tapir-opentelemetry-tracing` is not an option here because it relies on thread-local OTel context, incompatible with ZIO's fiber-based propagation.

Also records OTel HTTP semconv attributes on the server span so traces can be searched by route template or realized path without inflating span-metric cardinality:

- `http.request.method` — e.g. `GET`
- `url.path` — raw request path, e.g. `/admin/lists/http%3A%2F%2F...`
- `http.route` — matched template, e.g. `/admin/lists/{listIri}` (only set when an endpoint matches)
- `http.response.status_code` — e.g. `200`
- `user_agent.original` — verbatim `User-Agent` header, when present

Logs emitted while handling a request are now annotated with `trace_id` and `span_id` (sourced from the extracted OTel context), so log lines can be pivoted from a trace in Grafana and vice versa without manual correlation.

Applied to both dsp-api (`DspApiServer`) and ingest (`IngestApiServer`).


siers (approved):

ingest/src/main/scala/swiss/dasch/infrastructure/IngestApiServer.scala
```diff
@@ -38,8 +54,47 @@ final class IngestApiServer(
...
+  // http.route. Keeps span names low-cardinality per OTel HTTP semantic conventions:
+  // https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
+  private def spanNameInterceptor: EndpointInterceptor[Task] = new EndpointInterceptor[Task] {
+    private def updateSpanMetadata(ep: AnyEndpoint): UIO[Unit] =
+      ctxStore.get.flatMap { ctx =>
+        ZIO.succeed {
+          val method = ep.method.map(_.method).getOrElse("HTTP")
+          val route  = ep.showPathTemplate(showQueryParam = None, showQueryParamsAs = None)
+          val span   = Span.fromContext(ctx)
+          val _      = span.updateName(s"HTTP $method $route")
```
SamuelBoerlin:
```suggestion
          val _      = span.updateName(s"$method $route")
```
https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name recommends `{method} {target}` as span name without the `HTTP` prefix.

SamuelBoerlin (changes requested):
Great, thanks! I can't comment much on the code. I just have one suggestion regarding the span name.

ingest/src/main/scala/swiss/dasch/infrastructure/IngestApiServer.scala
seakayone: https://github.com/dasch-swiss/dsp-api/pull/4088/commits/a1141e88ebb12298cd74e4f12ac1db1f29389dc2 fixed it

SamuelBoerlin (approved):

PR #4091, opened 2026-04-27
test: Pin TDB2 transactions to a single thread in TriplestoreServiceInMemory

seakayone:
Apache Jena TDB2 stores transaction state in a `ThreadLocal`, so `begin` / `commit` / `abort` / `end` must all run on the same OS thread. The previous implementation in `TriplestoreServiceInMemory` wrapped the lifecycle and dependent resources (`QueryExecution`, `ResultSet`, `Model`) in chained `ZIO.acquireRelease` blocks, which ZIO is free to schedule across different threads. When release landed on a thread that never saw `begin`, TDB2 threw `TransactionException: Not in an active transaction`.

This is the root cause of the intermittent `CardinalityServiceLiveSpec` failure and explains why the same exception surfaced sporadically across unrelated PRs.

Collapse the entire txn lifecycle and any dependent resource consumption into a single `ZIO.attemptBlocking` block via a `withTx` helper, pinning everything to one blocking-pool thread for the duration. Test infrastructure only — no production code changes.

### DIFF
--- a/webapi/src/test/scala/org/knora/webapi/store/triplestore/api/TriplestoreServiceInMemory.scala
+++ b/webapi/src/test/scala/org/knora/webapi/store/triplestore/api/TriplestoreServiceInMemory.scala
@@ -7,7 +7,6 @@ package org.knora.webapi.store.triplestore.api
 
 import org.apache.jena.query.*
 import org.apache.jena.rdf.model
-import org.apache.jena.rdf.model.Model
 import org.apache.jena.rdf.model.ModelFactory
 import org.apache.jena.riot.RDFDataMgr
 import org.apache.jena.tdb.TDB
@@ -21,11 +20,11 @@ import zio.Scope
 import zio.Task
 import zio.UIO
 import zio.ULayer
-import zio.URIO
 import zio.ZIO
 import zio.ZLayer
 import zio.stream.ZStream
 
+import java.io.ByteArrayOutputStream
 import java.io.InputStream
 import java.nio.charset.StandardCharsets
 import java.nio.file.Path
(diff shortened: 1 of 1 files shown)


webapi/src/test/scala/org/knora/webapi/store/triplestore/api/TriplestoreServiceInMemory.scala
```diff
@@ -83,28 +82,38 @@ object TestTripleStore {
...
-    def acquire(query: String, ds: Dataset)                     = ZIO.attempt(QueryExecutionFactory.create(query, ds))
-    def release(qExec: QueryExecution): ZIO[Any, Nothing, Unit] = ZIO.succeed(qExec.close())
-    getDataSetWithTransaction(ReadWrite.READ).flatMap(ds => ZIO.acquireRelease(acquire(query, ds))(release))
-  }
+  // TDB2 stores transaction state in a ThreadLocal: begin / commit / abort / end
+  // must all run on the same OS thread. ZIO.attemptBlocking pins the synchronous
+  // body to a single blocking-pool thread for its full duration, so collapsing
+  // the whole txn lifecycle (and any dependent ResultSet/Model/QueryExecution
+  // consumption) into one block is the safe pattern. f must be plain Scala/Java
+  // and must fully materialise its result — no lazy iterators may escape.
```
siers:
```
  // ... f must be plain Scala/Java
  // and must fully materialise its result — no lazy iterators may escape.
```
slop

seakayone: How would you better describe it?

siers (approved):
Cool

webapi/src/test/scala/org/knora/webapi/store/triplestore/api/TriplestoreServiceInMemory.scala
seakayone: I have simplified the very wordy explanation, thanks for your input.

PR #4092, opened 2026-04-27
chore: Verify API healthcheck in CI (DEV-6247)

siers:
a quick one-two

proof it works: https://github.com/dasch-swiss/dsp-api/actions/runs/24993169094/job/73183477997?pr=4092

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,3 +1,5 @@
+version: '3'
+
 services:
   app:
     image: daschswiss/dsp-app:latest
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3'
-
 services:
   app:
     image: daschswiss/dsp-app:latest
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,3 +1,5 @@
+version: '3'
```
seakayone: AFAIU, the version field in a docker-compose became obsolete.

> The top-level version property is defined by the Compose Specification for backward compatibility. It is only informative and you'll receive a warning message that it is obsolete if used.

https://docs.docker.com/reference/compose-file/version-and-name/

seakayone: When using docker compose I now get this warning:
```bash
docker compose up
WARN[0000] /Users/christian/git/dasch/dsp-api/chore-test-docker-healthcheck/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```
Please remove it.

seakayone (changes requested):

seakayone:
Other than the `version` comment this change LGTM.

docker-compose.yml
siers: A certain someone suggested adding it, because it noticed I'm using `docker-compose` instead of just `docker compose` 😅 

Removed

seakayone (approved):

PR #4097, opened 2026-04-29
fix(export): fix CSV export timeout for large resource classes (DEV-6288)

mpro7:
[DEV-6288](https://linear.app/dasch/issue/DEV-6288/timeout-on-data-export-csv-download)

## Summary

`POST /v3/export/resources` timed out for classes with 30k+ rich-text resources. Root cause: PR #4018 introduced `queryStandoff = true` in `ExportService` unconditionally, loading all standoff markup per batch. `textValueColumn` only ever reads `StandoffFootnoteTag` nodes — all other markup was loaded and discarded. For 31k resources this caused a ~10x per-batch payload increase. Additionally, two pre-existing timeout mismatches in the triplestore layer were exposed during testing.

## Changes

**1. Narrow standoff loading to footnote tags only (`ExportService` / `GetResourcePropertiesAndValuesQuery`)**
- Add `standoffTagFilter: Option[SmartIri] = None` to `GetResourcePropertiesAndValuesQuery.build` — when `Some(tagIri)`: constrains `?standoffNode rdf:type <tagIri>` in CONSTRUCT and WHERE, drops `targetHasOriginalXMLID` join
- Thread through `ReadResourcesServiceLive.readResourcesSequencePar` (trait + impl); all existing callers keep `None` (full standoff, unchanged)
- `ExportService.exportResources` passes `standoffTagFilter = Some(footnoteTagIri)`

**2. Use maintenance timeout (120s) for standoff CONSTRUCT queries**
- `readResourcesSequence_` now passes `SparqlTimeout.Maintenance` when `queryStandoff = true`
- Only affects `ExportService.exportResources` — the only caller passing `queryStandoff = true`

**3. Align HTTP read timeout with SPARQL query timeout**
- `executeSparqlQuery` now sets `readTimeout = sparqlTimeout + 10s` on the sttp request
- sttp4 defaults to a 1-minute read timeout; Maintenance queries (120s SPARQL) were being dropped by the HTTP client at ~60s before Fuseki could respond

## Blast radius

- `standoffTagFilter` defaults to `None` — zero impact on existing callers
- Maintenance timeout: only fires for `queryStandoff = true`, only called from export
- HTTP read timeout: Standard queries go from 60s → 30s (safe: Fuseki responds within ms of its 20s kill); Maintenance/Gravsearch go from 60s → 130s (fixes latent issue)
- UPDATE queries bypass `executeSparqlQuery` — unaffected

## Tests

- `sbt "testOnly *GetResourcePropertiesAndValuesQuerySpec*"` — 14 query shape tests pass (U-2: type constraint present when filter set, no `targetHasOriginalXMLID`; U-3: full standoff and `targetHasOriginalXMLID` present when no filter)
- `sbt "testOnly *ExportServiceSpec*"` — 14 export tests pass
- `sbt check` — format + lint clean
- Manual: 31.6k resources exported successfully in ~8.8 min against dev Fuseki

## Known limitation

The full export response is buffered before being sent to the client — connections exceeding the ingress 60s idle timeout will still fail on prod for very large classes. Tracked in [DEV-6336](https://linear.app/dasch/issue/DEV-6336/improve-csv-export-performance-for-large-resource-classes).

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/slice/resources/repo/GetResourcePropertiesAndValuesQuery.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/resources/repo/GetResourcePropertiesAndValuesQuery.scala
@@ -20,6 +20,8 @@ import java.util.UUID
 import dsp.valueobjects.UuidUtil
 import org.knora.webapi.IRI
 import org.knora.webapi.messages.SmartIri
+import scala.util.chaining.scalaUtilChainingOps
+
 import org.knora.webapi.slice.common.QueryBuilderHelper
 import org.knora.webapi.slice.common.repo.rdf.Vocabulary.KnoraBase
 
@@ -150,20 +152,11 @@ object GetResourcePropertiesAndValuesQuery extends QueryBuilderHelper {
 
     val standoffPatterns =
       if (queryStandoff)
-        standoffTagFilter match {
-          case Some(tagIri) =>
-            Seq(
-              valueObject.has(KnoraBase.valueHasStandoff, standoffNode),
-              standoffNode.isA(Rdf.iri(tagIri.toIri)).andHas(standoffProperty, standoffValue),
-            )
-          case None =>
-            Seq(
-              valueObject.has(KnoraBase.valueHasStandoff, standoffNode),
-              standoffNode
-                .has(standoffProperty, standoffValue)
-                .andHas(KnoraBase.targetHasOriginalXMLID, targetOriginalXMLID),
-            )
-        }
+        Seq(valueObject.has(KnoraBase.valueHasStandoff, standoffNode))
+          .pipe(_ :+ (standoffTagFilter match {
+            case Some(tagIri) => standoffNode.isA(Rdf.iri(tagIri.toIri)).andHas(standoffProperty, standoffValue)
+            case None         => standoffNode.has(standoffProperty, standoffValue).andHas(KnoraBase.targetHasOriginalXMLID, targetOriginalXMLID)
+          }))
       else Seq.empty
 
     val linkPatterns =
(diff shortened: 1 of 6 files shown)


webapi/src/main/scala/org/knora/webapi/slice/resources/service/ReadResourcesServiceLive.scala
```diff
@@ -119,7 +122,9 @@ final case class ReadResourcesServiceLive(
                 maybeVersionDate = versionDate.map(_.value),
                 queryAllNonStandoff = true,
                 queryStandoff = queryStandoff,
+                standoffTagFilter = standoffTagFilter,
               ),
+              if (queryStandoff) SparqlTimeout.Maintenance else SparqlTimeout.Standard,
```
siers, line 127: What does this mean to use a "maintenance" timeout?

webapi/src/main/scala/org/knora/webapi/slice/resources/repo/GetResourcePropertiesAndValuesQuery.scala
```diff
@@ -267,20 +281,33 @@ object GetResourcePropertiesAndValuesQuery extends QueryBuilderHelper {
...
+                  .andHas(KnoraBase.standoffTagHasStartIndex, startIndex),
+              )
+              .and(
+                standoffTag
+                  .has(KnoraBase.standoffTagHasInternalReference, targetStandoffTag)
+                  .and(targetStandoffTag.has(KnoraBase.standoffTagHasOriginalXMLID, targetOriginalXMLID))
+                  .optional(),
+              )
+              .filter(Expressions.gte(startIndex, Rdf.literalOf(0)))
+        }
```
siers: this repetition is a bit heavy. there is a pipe operator in scala.util.chaining that adds a `.pipe(x=>x)` which would be a way to solve this nicely

https://www.scala-lang.org/api/2.13.5/scala/util/ChainingOps.html

webapi/src/main/scala/org/knora/webapi/slice/resources/repo/GetResourcePropertiesAndValuesQuery.scala
```diff
@@ -147,12 +150,20 @@ object GetResourcePropertiesAndValuesQuery extends QueryBuilderHelper {
...
+              standoffNode.isA(Rdf.iri(tagIri.toIri)).andHas(standoffProperty, standoffValue),
+            )
+          case None =>
+            Seq(
+              valueObject.has(KnoraBase.valueHasStandoff, standoffNode),
+              standoffNode
+                .has(standoffProperty, standoffValue)
+                .andHas(KnoraBase.targetHasOriginalXMLID, targetOriginalXMLID),
+            )
+        }
```
siers: this is less repetitive as the other one, but still

webapi/src/main/scala/org/knora/webapi/store/triplestore/impl/TriplestoreServiceLive.scala
```diff
@@ -381,9 +381,10 @@ case class TriplestoreServiceLive(
       case SparqlTimeout.Gravsearch  => triplestoreConfig.gravsearchTimeout
     }
 
-    val params  = Map(("query", query.sparql), ("timeout", timeout.toSeconds.toString))
-    val uri     = targetHostUri.addPath(paths.query)
-    val request = authenticatedRequest.post(uri).header("Accept", acceptMimeType).body(params)
+    val params      = Map(("query", query.sparql), ("timeout", timeout.toSeconds.toString))
+    val uri         = targetHostUri.addPath(paths.query)
+    val readTimeout = FiniteDuration(timeout.toSeconds + 10, TimeUnit.SECONDS)
+    val request     = authenticatedRequest.post(uri).header("Accept", acceptMimeType).body(params).readTimeout(readTimeout)
```
siers, line 387: what's the significance of this? why is this feature changing timeouts in TriplestoreServiceLive.scala?

webapi/src/main/scala/org/knora/webapi/slice/resources/service/ReadResourcesServiceLive.scala
mpro7: Maintenance is the 120s timeout tier, already used for repo dumps and other long-running operations (vs Standard = 20s). Standoff CONSTRUCT queries over 500-resource batches can easily exceed 20s on large datasets — reusing this tier fits the intent. We could introduce a dedicated Export tier but there's no reason to diverge from what already exists.

webapi/src/main/scala/org/knora/webapi/store/triplestore/impl/TriplestoreServiceLive.scala
mpro7: Pre-existing bug exposed by this PR: sttp4 defaults to a 1-minute HTTP read timeout for all requests. Maintenance queries (120s SPARQL) were being dropped by the HTTP client at ~60s before Fuseki could respond. The fix aligns the HTTP read timeout with the SPARQL timeout + 10s. It touches `TriplestoreServiceLive` because that's where the HTTP request is constructed.

webapi/src/main/scala/org/knora/webapi/slice/resources/repo/GetResourcePropertiesAndValuesQuery.scala
mpro7: Good catch! Fixed in [67b4c35](https://github.com/dasch-swiss/dsp-api/pull/4097/commits/67b4c350d36230860354fa7baad2ae8aab67e932).

webapi/src/main/scala/org/knora/webapi/slice/resources/repo/GetResourcePropertiesAndValuesQuery.scala
mpro7: Fixed in [67b4c35](https://github.com/dasch-swiss/dsp-api/pull/4097/commits/67b4c350d36230860354fa7baad2ae8aab67e932).

siers (approved):

PR #4099, opened 2026-04-29
chore: auto-symlink .env in worktrees for run-with-dev-db

mpro7:
## Summary

- Auto-creates a symlink to the main worktree's `.env` when `just run-with-dev-db` is run from a git worktree where `.env` is absent
- Adds VPN requirement note to the recipe comment

## Test plan

- [ ] Run `just run-with-dev-db` from a fresh worktree (no `.env`) — symlink should be created automatically
- [x] Run again — symlink already exists, should be skipped without error
- [ ] Run from the main repo — `.env` exists, no symlink logic triggered


BalduinLandolt (approved):
LGTM

Note, the following doesn't work if you manage worktrees manually, with loom or any other such tool.  <br>But Claude Code also has the capability of managing worktrees now, and then they allow for a `.worktreeinclude` file ([https://code.claude.com/docs/en/worktrees#copy-gitignored-files-into-worktrees](<https://code.claude.com/docs/en/worktrees#copy-gitignored-files-into-worktrees>)). I just learned about this today. ^^  <br>Again, this only works for claude-managed worktrees, but might be worth making use of.

mpro7:
> LGTM
> 
> Note, the following doesn't work if you manage worktrees manually, with loom or any other such tool. But Claude Code also has the capability of managing worktrees now, and then they allow for a `.worktreeinclude` file (https://code.claude.com/docs/en/worktrees#copy-gitignored-files-into-worktrees). I just learned about this today. ^^ Again, this only works for claude-managed worktrees, but might be worth making use of.

AFAIK it suppose to work with loom as well. Thanks for the `.worktreeinclude`, it might make this condition redundant, but worth to keep it for now. I will also try to use the other way.

PR #4104, opened 2026-05-05
refactor: Remove IIIFRequestMessageHandler indirection

seakayone:
The `MessageRelay` is a transitional shim from the Akka actor → ZIO migration: a `Ref[List[MessageHandler]]` plus an untyped `ask[R]` that routed `RelayedMessage` instances to whichever handler claimed responsibility. The long-term goal is direct method calls; the relay only still exists because of circular dependencies between responders.

`IIIFRequestMessageHandler` was not one of those real cycles — it only routed `SipiGetTextFileRequest` to `SipiService.getTextFileRequest`, a single call site dressed up as a relay handler.

This PR removes that indirection:

- `SipiService.getTextFileRequest` now takes `(fileUrl, senderName)` and returns `Task[String]` directly. The `requestingUser` parameter was unused; the response wrapper held a single `String`.
- The handler, its layer, the `IIIFRequest`/`SipiRequest` traits, and the request/response wrappers are deleted.
- Callers (`ResourcesResponderV2`, `StandoffResponderV2`) now call `sipiService` directly, dropping several already-dead constructor params along the way.


siers (approved):
Nice!

PR #4107, opened 2026-05-05
refactor: Remove MessageRelay from OntologyResponderV2

seakayone:
Remove `MessageRelay` from `OntologyResponderV2` and migrate former callers to now call `OntologyCacheHelpers` or `OntologyResponderV2` directly .

`OntologyResponderV2` no longer extends `MessageHandler`, no longer subscribes to `MessageRelay`, and the layer no longer requires it. Unused request case classes that existed only as relay envelopes are deleted, along with the dispatch methods in the responder.


siers (approved):
Nice!

PR #4108, opened 2026-05-05
feat(project-migration): Add optional error message to failed data tasks

seakayone:
Failed data tasks (project migration export/import) now carry an optional human-readable error message that is persisted and surfaced to API callers. This lets clients show the actual failure reason instead of a generic "Failed" status.


siers (approved):

Nora-Olivia-Ammann:
Thank you!

PR #4110, opened 2026-05-06
feat: RegionPreviewValue support (DEV-6300)

siers:
**`RegionPreviewValue` support (DEV-6300)**

New value type: a resource points at a `Region` via `hasRegionPreview`. Only the link (`isRegionPreviewOf`) is stored. The IIIF preview URL is computed at read time, never persisted.

**Pipeline**

```
Write: JSON-LD isRegionPreviewOf  →  persist  <value> kb:isRegionPreviewOf <region>
Read:  RegionPreviewValue (iiifUrl = None)  →  fetch region (geometry) + still image (filename)
       →  iiifUrl = …/pct:x,y,w,h/full/0/default.jpg     (fetched resources not returned)
```

**Change summary**

- `knora-base.ttl` & `OntologyConstants.scala` — `RegionPreviewValue` / `hasRegionPreview` / `isRegionPreviewOf` / `iiifUrl`; v51 → v52
- `ValueMessagesV2.scala` — `RegionPreviewValueContentV2` (value type, JSON-LD, parser); `GeomValueContentV2.parsePoints`
- `ApiComplexV2JsonLdRequestParser.scala`, `ResourceCreateModels.scala`, `ResourcesRepoLive.scala` — write path: parse + persist `isRegionPreviewOf`
- `ConstructResponseUtilV2.scala` — read path: dispatch `RegionPreviewValue`; `iiifUrl = None` (deferred to the service)
- `ReadResourcesServiceLive.scala` — `augmentRegionPreviewValues`: region geometry → `pct:x,y,w,h`, still-image filename → IIIF URL
- `ReadResourcesServiceLiveSpec.scala` – finally has a standalone test! previously implicitly tested
- `GravsearchTypeInspectionUtil.scala`, `ResourceMessagesV2.scala`, `Vocabulary.scala` — supporting constants/accessors
- acceptance test fixture updates


siers:
Ran claude as the reviewer, there are two things

webapi/src/main/resources/knora-ontologies/knora-base.ttl
```diff
@@ -333,6 +333,19 @@
     :subjectClassConstraint :Region .
 
 
+###  http://www.knora.org/ontology/knora-base#hasRegionPreview
+
+:hasRegionPreview
+    rdf:type                owl:ObjectProperty ;
+    rdfs:subPropertyOf      :hasValue ;
+    rdfs:label              "has Region preview"@en ;
```
BalduinLandolt: please add translations to all the labels - these are used by DSP-APP for internationalization, and when they're missing it looks really ugly to users who use the app in a non-english language. (Claude-generated translations are good enough, but ask it to make them consistent with the vocabulary that's already in use.)

webapi/src/main/resources/knora-ontologies/knora-base.ttl
```diff
@@ -2792,3 +2805,39 @@
...
+
+# This property is generated on the fly and not present in the database
+
+:iiifUrl
+    rdf:type                owl:DatatypeProperty ;
+    rdfs:label              "is an IIIF URL"@en ;
+    rdfs:comment            "IIIF URL for the region preview"@en ;
+    rdfs:subPropertyOf      :valueHas ;
+    :subjectClassConstraint :RegionPreviewValue ;
+    :objectDatatypeConstraint  xsd:string .
```
BalduinLandolt, line 2855: I don't think we need this in the ontology, if it's ever only materialised in externally. I think we have other cases that are the same, which get added in the course of the internal-to-external schema translation

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
```diff
@@ -2518,6 +2529,57 @@ object LinkValueContentV2 {
...
+ * @param regionIri        the IRI of the Region resource this value points to.
+ * @param iiifUrl          the IIIF preview URL for the region, computed at read time in `ReadResourcesServiceLive`.
+ * @param comment          a comment on this [[RegionPreviewValueContentV2]], if any.
+ */
+case class RegionPreviewValueContentV2(
+  ontologySchema: OntologySchema,
+  regionIri: ResourceIri,
+  iiifUrl: Option[String] = None,
+  comment: Option[String] = None,
+) extends ValueContentV2 {
```
BalduinLandolt, line 2545: Nitpick (can stay like that, just jumped at me): Technically, this not very cleanly modelled: `iiifUrl` will always be `None` when reading from the DB, ands will always be `Some()` after it's resolved downstream. So, in principal, these are two separate models.

webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
```diff
@@ -2518,6 +2529,57 @@ object LinkValueContentV2 {
...
+  override def valueType: SmartIri = {
+    implicit val stringFormatter: StringFormatter = StringFormatter.getGeneralInstance
+    OntologyConstants.KnoraBase.RegionPreviewValue.toSmartIri.toOntologySchema(ontologySchema)
+  }
+
+  override def valueHasString: String = regionIri.value
+
+  override def toOntologySchema(targetSchema: OntologySchema): RegionPreviewValueContentV2 =
+    copy(ontologySchema = targetSchema)
```
BalduinLandolt, line 2554: does that mean that this value type is identical for simple and complex schema?

webapi/src/main/scala/org/knora/webapi/slice/common/repo/rdf/Vocabulary.scala
```diff
@@ -157,7 +157,9 @@ object Vocabulary {
     val standoffTagHasStart: Iri             = iri(kb + "standoffTagHasStart")
     val standoffTagHasEnd: Iri               = iri(kb + "standoffTagHasEnd")
 
-    val hasTextValueType: Iri = iri(kb + "hasTextValueType")
+    val hasTextValueType: Iri  = iri(kb + "hasTextValueType")
+    val isRegionPreviewOf: Iri = iri(kb + "isRegionPreviewOf")
+    val isRegionOf: Iri        = iri(kb + "isRegionOf")
```
BalduinLandolt: why is that added here?

webapi/src/main/scala/org/knora/webapi/slice/resources/service/ReadResourcesServiceLive.scala
```diff
@@ -204,6 +234,94 @@ final case class ReadResourcesServiceLive(
...
+    regions.resources.flatMap { region =>
+      for {
+        regionSelector <- geometryToIiifRegion(region)
+        imageIri       <- region.isRegionOfValueReferredIri
+        image          <- imagesByIri.get(imageIri)
+        stillImage     <- image.values.values.flatten.map(_.valueContent).collectFirst {
+                        case si: StillImageFileValueContentV2 => si
+                      }
+      } yield region.resourceIri ->
+        s"${appConfig.sipi.externalBaseUrl}/${image.projectADM.shortcode}/${stillImage.fileValue.internalFilename}/$regionSelector/full/0/default.jpg"
```
BalduinLandolt:
```suggestion
        s"${appConfig.sipi.externalBaseUrl}/${image.projectADM.shortcode}/${stillImage.fileValue.internalFilename}/$regionSelector/max/0/default.jpg"
```

This produces an invalid IIIF URL, should be "max", not "full". Compare:
https://iiif.stage.dasch.swiss/0854/4iI0VeNIZe4-DhXMWwDizxh.jpx/pct:39.796687,24.423475,9.266166,17.576948/full/0/default.jpg
https://iiif.stage.dasch.swiss/0854/4iI0VeNIZe4-DhXMWwDizxh.jpx/pct:39.796687,24.423475,9.266166,17.576948/max/0/default.jpg

Also, this suggests that you haven't manually verified that the change actually works.  
And Claude spotted this bug extremely quickly, so presumably you didn't run reviewing on the PR either.  
I consider both a prerequisite, before calling a PR ready.

BalduinLandolt (changes requested):
Added some nitpicks, and there is one small but important bug, the IIIF URLs are invalid.  
Also, if that's reasonably doable, maybe an integration test that verifies that it produces IIIF URLs which SIPI can handle, would prevent issues like this. But if it's too much of a pain, then no need to bother.

siers:
I ran through the reviewer once more, none of those concerned me.

BalduinLandolt (approved):

PR #4112, opened 2026-05-06
refactor: Extract StandoffMappingService and remove MessageRelay usage

seakayone:
Extract a dedicated `StandoffMappingService` for XML→standoff mapping and XSL transformation lookups so callers can invoke these operations directly instead of routing through `MessageRelay`. The service owns mapping caching, triplestore queries, ontology validation, and Sipi file access.

Replace raw mapping IRI strings with `StandoffMappingIri` and `StandoffMappingElementIri` value types for type safety and validation at API boundaries. A separate `GetXslTransformationMetadataQuery` avoids the layer cycle that previously kept the lookup inside the responder.


siers (approved):
I can't follow this behemoth, but I see one file deleted, another created and a bunch of oneliners. 👍

PR #4113, opened 2026-05-08
refactor: Remove MessageRelay and MessageHandler abstractions

seakayone:
The `MessageRelay`/`MessageHandler` indirection was a temporary scaffold for the actor-to-ZIO migration. With the last responders ported, it no longer pays for itself: every call still went through a single subscriber, so the relay added a stringly-typed `ask[A]` and runtime dispatch in front of what is otherwise a normal method call.

Replacing it with direct service calls makes call sites type-checked end-to-end, removes a layer of subscription wiring, and lets `ResourcesResponderV2` and `ValuesResponderV2` depend on the services they actually need (`ReadResourcesService`) rather than on a generic relay.

Also removes the orphaned scaffolding the relay leaves behind: the `ResponderRequest` / `KnoraRequestV2` / `KnoraRequestADM` / `StoreRequest` / `ListsResponderRequestV2` tagging traits, the `Responder.handleUnexpectedMessage` fallback, and the `UnexpectedMessageException` it raised — all of which only existed to feed the relay's runtime dispatch. ADR-0007, which described the now-defunct architecture, is marked superseded.


siers (approved):
grandios

PR #4115, opened 2026-05-08
refactor: Replace ZioHelper utilities with native ZIO operations

seakayone:
Drops a thin internal utility layer in favour of ZIO's standard API so the call sites read with no indirection.

- Inline `ZioHelper.sequence` at its two callers (`ConstructResponseUtilV2`, `CreateResourceV2Handler`) using `ZIO.foreach` directly over the map
- Delete `ZioHelper`, `LogAspect`, and `ZioHelperSpec` — no remaining callers
- Remove unused exception and utility classes


webapi/src/main/scala/org/knora/webapi/messages/util/ConstructResponseUtilV2.scala
```diff
@@ -1132,108 +1131,114 @@ final class ConstructResponseUtilV2(
...
+                                  valueCreationDate = valueCreationDate,
+                                  valueHasUUID = valueHasUUID,
+                                  valueContent = otherValueContentV2,
+                                  previousValueIri = previousValueIri,
+                                  deletionInfo = valueDeletionInfo,
+                                )
+                            }): ReadValueV2
+                          }
+                          .map(property -> _)
+                      }
```
siers, line 1241: isn't this a little bulky? 😆

siers (approved):

PR #4117, opened 2026-05-12
feat(project-migration): Reject built-in user references in attachedToUser on project migration import

seakayone:
Reject project-migration imports whose `data.nq` attaches resources or values to the built-in `SystemUser` or `AnonymousUser` via `knora-base:attachedToUser`. These IRIs exist on every target instance; leaving them in imported data would silently bind project content to system-managed identities, so the import must surface the conflict and require pre-migration cleanup.

Enforced as a SHACL `NodeShape` (`AttachedToUserNotBuiltInShape` in `data-shapes.ttl`) that targets `sh:targetSubjectsOf knora-base:attachedToUser` with `sh:not [ sh:in ( knora-admin:SystemUser knora-admin:AnonymousUser ) ]`. Runs as part of the existing SHACL pass — no extra parse of `data.nq`, no extra Jena `Model` lifecycle.

Scope note: covers `attachedToUser` only (the actual RDF property behind both `creator` and `valueCreator` in the codebase). References to the built-ins via other paths are not enforced by this shape.

Fixes this issue reported by @Nora-Olivia-Ammann


Nora-Olivia-Ammann (approved):

BalduinLandolt (approved):

Nora-Olivia-Ammann:
@BalduinLandolt FYI we have the SystemUser in data graphs on both prod and ls-prod. I am working on a more detailed analysis of all the clean-up.

PR #4121, opened 2026-05-20
chore: Upgrade Apache Jena Fuseki to 6.1.0 (DEV-5935)

seakayone:
Bump the DaSCH Fuseki image to Apache Jena Fuseki 6.1.0 (`daschswiss/apache-jena-fuseki:6.1.0-0`) to stay on the supported upstream release line.

Jena 6.1 changed how invalid Lucene input in `text:query` surfaces: when embedded in a sub-select or OPTIONAL it is now raised at evaluation time as `TextIndexParseException` and returned as HTTP 500 instead of the previous parse-time HTTP 400. Both still indicate bad user input, so the triplestore client maps either status (when the body contains `Text search parse error`) to `BadRequestException`, preserving the public API contract.

Also cleans up the long-standing Dockerfile `LegacyKeyValueFormat` warnings while the file is being touched.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
   db:
     # should be the same version as in Dependencies.scala,
     # make sure to use the same version in ops-deploy repository when deploying new DSP releases!
-    image: daschswiss/apache-jena-fuseki:5.5.0-3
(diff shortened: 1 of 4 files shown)


siers (approved):
Oops, forgot to click submit, but I'd opened the PR

PR #4124, opened 2026-05-26
feat: stream CSV export response chunked from dsp-api

mpro7:
[DEV-6336](https://linear.app/dasch/issue/DEV-6336)

## Summary

- Converts `POST /v3/export/resources` from a buffered response (JVM accumulates full CSV, browser waits 8+ min) to `Transfer-Encoding: chunked` streaming — time-to-first-byte drops from minutes to seconds
- Header bytes are emitted before any batch fetch starts; each 500-resource batch is encoded and flushed as it resolves with parallelism 5
- Row order (by `rdfs:label`) is preserved and made deterministic

## Changes

**Configuration:**
- `batchSize` and `parallelism` are operational knobs in `application.conf` under `app.export` (`batch-size = 500`, `parallelism = 5`), each with an env override (`KNORA_WEBAPI_EXPORT_BATCH_SIZE` / `KNORA_WEBAPI_EXPORT_PARALLELISM`). Read via `AppConfig` in `ExportService` so values can change per deployment without a code change/release/deploy.

**Core streaming:**
- `ExportService.exportResources` returns `Task[ZStream[Any, Throwable, Byte]]`; the one-time setup (ordered IRI+label fetch, link-label map, vocabulary, header bytes) runs as an effect, then per-batch rows stream via `ZStream.mapZIOPar(5)`
- `ExportRestService` runs setup eagerly (`.orDie`) before the response status is committed, so a setup failure is a 5xx rather than a `200 OK` with a truncated/empty body
- `FindResourcesService.findResourceIrisOrderedByLabel` returns `(ResourceIri, label)` pairs; the cross-batch link-label map is built from them (no second query). Shared `valuesClauseFor` keeps the subclass-expansion invariant in one place
- `CsvService`: `writeRowToString` / `writeHeaderToString` for incremental per-row encoding
- `ReadResourcesServiceLive.readResourcesSequence`: extended with `queryStandoff`, `skipRetrievalChecks`, `standoffTagFilter` (additive, defaults preserve behaviour)

**Endpoint wiring:**
- `ExportEndpoints`: output is `statusCode(Ok) ++ header("Content-Disposition") ++ streamBinaryBody(ZioStreams)(Csv())`
- `ExportRestService`: returns `(contentDisposition, stream)`; `Clock.instant` reflects request-entry (filename timestamp shifts ≤1 s)

**Ordering:**
- `findResourceIrisOrderedByLabel` selects `?label` and sorts in the JVM by `(label, resourceIri)`. Case-sensitive label order matches the pre-streaming `.sortBy(_.label)`; resource IRI is the deterministic tie-break (the old code relied on Fuseki's non-reproducible scan order). The three `ExportEndpointsE2ESpec` golden CSVs were reordered for the two tied-label pairs.

## Tests

- 5 golden CSV tests materialise the `ZStream` via `runCollect`
- Ordering regression test (scrambled labels → alphabetical) + tie-break fix-confirmation test (two "Mike" resources, asserted sorted by resource IRI)
- **Setup-failure test**: a failing setup surfaces on the effect without consuming the stream (guards the 5xx-not-200 path)
- **Cross-batch link-label test**: `batchSize = 1` forces a link target into a different batch; asserts the label still resolves
- Streaming-shape test (header byte before any batch completes) + mid-stream failure propagation test
- **E2E**: asserts the response is chunked (no buffered `Content-Length`)
- OAI golden updated for the two new "Mike" resources
- **Config fix-confirmation test**: with `app.export.batch-size = 100`, 600 IRIs produce six batches of 100 — proves the configured value drives batching (not the old hard-coded literal)
- **`ExportStreamingIT`** (`modules/test-it/`): sweeps a `(parallelism, batchSize)` matrix against incunabula `page` (4024 resources) on real Fuseki — wall-time regression guard plus a re-runnable vehicle for choosing the config defaults. Excluded from default CI (multi-minute); run via `sbt "test-it/testOnly *ExportStreamingIT*"`.

## Default tuning values

The config defaults (5, 500) were chosen from a matrix run on a Fuseki testcontainer (incunabula `page`, 4024 resources, ms post-warmup):

```
parallelism | bs=100  bs=250  bs=500  bs=1000 bs=2000
        1   | 6626    6697    7163    8082    10871
        2   | 3472    3470    3822    4467    6552
        5   | 1999    2140    2326    2690    7001    ← default
       10   | 2209    2135    2137    2852    6837
       20   | 1894    1787    2074    2689    6775
```

Parallelism past 5 plateaus (no measurable gain, adds Fuseki contention under multi-tenant load); batchSize above 1000 hits a Fuseki query-planning cliff (non-linear cost growth past ~1000 IRI bindings). These are starting points — ops can override per deployment, e.g. once prod moves to our own hardware.

## Review

Multi-perspective review (Scala/ZIO, consistency, performance, security, simplicity, pattern) run on the branch. Addressed: the truncated-200 setup-failure regression (above), a redundant second full-class label query, the missing cross-batch + chunked-header tests, and removal of dead `toCsv`/`ExportedCsv`. Security found no injection — the IRI validators close the `VALUES` string-interpolation surface. Post-review fixes: revert literal 0x1E byte to `^^` escape, `Map[IRI, String]` → `Map[ResourceIri, String]`, `ZIO.attempt(throw)` → `ZIO.fromOption(...).orElseFail(...)`.

## Test plan

- `sbt "webapi/testOnly *ExportServiceSpec*"` — 21 tests pass; `sbt fmt` / `sbt check` clean
- E2E (`make test-e2e`, Docker) — verified in CI
- `sbt "test-it/testOnly *ExportStreamingIT*"` — 1 test passes; matrix above


webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
```diff
@@ -139,35 +144,86 @@ final case class ExportService(
...
-                         queryStandoff = true,
-                         skipRetrievalChecks = true,
-                         standoffTagFilter = Some(footnoteTagIri),
-                       )
-
-      propertyIriInfos <- propertyIriInfos(selectedProperties)
-      labelSmartIri    <- iriConverter.asSmartIri(OntologyConstants.Rdfs.Label)
-      propsWithInfos    = selectedProperties.map(p => (p, propertyIriInfos.get(p)))
-      headers           = rowHeaders(propsWithInfos, labelSmartIri, language, includeIris, includeArkUrls)
+    batchSize: Int = 500, // production value; overridable only so tests can force resources across batch boundaries
```
BalduinLandolt: having such magic values in code is a bit of a smell. Perhaps this should be a configurable value, rather than hard-coding it?

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
```diff
@@ -139,35 +144,86 @@ final case class ExportService(
...
+        headerBytes       = Chunk.fromArray(csvService.writeHeaderToString(rowBuilder).getBytes(StandardCharsets.UTF_8))
+      } yield StreamingExportContext(orderedIris, linkLabels, propsWithInfos, vocabularies, rowBuilder, headerBytes)
+
+    // `setup` runs eagerly as part of this Task so that failures during setup (ordered-IRI fetch, link-label
+    // fetch, vocabulary load, header encoding) surface to the caller *before* the response status is committed
+    // — i.e. a 5xx, not a `200 OK` with an empty/truncated body. Only per-batch failures after the first bytes
+    // have been flushed remain an inherent (unavoidable) mid-stream truncation.
+    setup.map { ctx =>
+      val iriPosition: Map[ResourceIri, Int] = ctx.orderedIris.zipWithIndex.toMap
+      val parallelism                        = 5
```
BalduinLandolt: this, too, could be a config value, so we could tune it without code changes

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
```diff
@@ -291,7 +347,7 @@ final case class ExportService(
   }
 
   private def stringFormat(s: String): String =
-    s.replaceAll("<[^>]+>", "").replaceAll("\u001e", " ")
+    s.replaceAll("<[^>]+>", "").replaceAll("", " ")
```
BalduinLandolt: what's the deal with this change?

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
```diff
@@ -209,7 +265,7 @@ final case class ExportService(
     propsWithInfo: List[(PropertyIri, Option[ReadPropertyInfoV2])],
     includeIris: Boolean,
     includeArkUrls: Boolean,
-    resources: Map[IRI, ReadResourceV2],
+    linkLabels: Map[IRI, String],
```
BalduinLandolt: this seems like weaker typing, right?

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/FindResourcesService.scala
```diff
@@ -64,26 +69,70 @@ final case class FindResourcesServiceLive(
...
+      (projectGraph.value, valuesClause)
+    }
+
+  private def buildClassQuery(project: KnoraProject, classIri: ResourceClassIri): Task[Select] =
+    valuesClauseFor(project, classIri).map { (projectGraph, valuesClause) =>
       Select.gravsearch(
         s"""PREFIX knora-base: <${KB.NS.getName}>
            |SELECT DISTINCT ?$resourceIriVar WHERE {
-           |  GRAPH <${projectGraph.value}> { ?$resourceIriVar a ?$classIriVar }
+           |  GRAPH <$projectGraph> { ?$resourceIriVar a ?$classIriVar }
```
BalduinLandolt, line 94: is this change sound?

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/FindResourcesService.scala
```diff
@@ -64,26 +69,70 @@ final case class FindResourcesServiceLive(
...
            |  VALUES ?$classIriVar { $valuesClause }
            |}""".stripMargin,
       )
     }
 
+  private def buildClassQueryOrderedByLabel(project: KnoraProject, classIri: ResourceClassIri): Task[Select] =
+    valuesClauseFor(project, classIri).map { (projectGraph, valuesClause) =>
+      // ?label is selected so findResourceIrisOrderedByLabel can sort in the JVM; Fuseki's ORDER BY tie-break
+      // for equal labels is not reproducible, so the authoritative sort happens in Scala.
+      Select.gravsearch(
```
BalduinLandolt, line 104: I hope this is not really gravsearch here, and the method is just called that. Otherwise this will be slow, I suspect.

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/FindResourcesService.scala
```diff
@@ -64,26 +69,70 @@ final case class FindResourcesServiceLive(
...
+    classIri: ResourceClassIri,
+  ): Task[Seq[(ResourceIri, String)]] =
+    for {
+      sparql <- buildClassQueryOrderedByLabel(project, classIri)
+      rows   <- triplestore.query(sparql).map(_.results.bindings)
+      pairs  <- ZIO.foreach(rows) { row =>
+                 for {
+                   value <- ZIO.attempt(
+                              row.rowMap.getOrElse(resourceIriVar, throw new InconsistentDataException("")),
+                            )
```
BalduinLandolt: I think `ZIO.fromOption` would be the better call here, than throwing the exception, just to catch it again and return it in a ZIO.

webapi/src/main/scala/org/knora/webapi/slice/resources/service/ReadResourcesServiceLive.scala
BalduinLandolt, line 1: why has this file changed?

webapi/src/test/scala/org/knora/webapi/slice/resources/service/ReadResourcesServiceFake.scala
BalduinLandolt, line 1: same as above

BalduinLandolt (commented):
I generally don't know enough about this to give much meaningful feedback. Added some questions and nitpicks, though.

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
mpro7: Now validated by `ExportStreamingIT` (added in this PR, `modules/test-it/.../ExportStreamingIT.scala`). The IT sweeps a 5×5 matrix against incunabula `page` (4024 resources) on real Fuseki — milliseconds, post-warmup:

```
parallelism | bs=100  bs=250  bs=500  bs=1000 bs=2000
        1   | 6626    6697    7163    8082    10871
        2   | 3472    3470    3822    4467    6552
        5   | 1999    2140    2326    2690    7001    ← current default
       10   | 2209    2135    2137    2852    6837
       20   | 1894    1787    2074    2689    6775
```

Reads: parallelism past 5 plateaus (no real gain, adds Fuseki contention risk under multi-tenant load); batchSize cliff above 1000 (Fuseki query-planning cost grows non-linearly past ~1000 IRI bindings); current (5, 500) sits within 16% of the matrix optimum at this dataset size. Not plumbing through `AppConfig` — the defaults are validated, and adding a knob no one needs to turn is just operational noise. If real-world usage at much larger scale ever diverges, the same IT can re-sweep and motivate a change then.

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
mpro7: see above https://github.com/dasch-swiss/dsp-api/pull/4124#discussion_r3318214346

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
mpro7: Functional no-op — Scala parses both `""` (escape) and the literal 0x1E byte to the same character. Editor/formatter artifact; reverting to the escape form for readability and greppability.

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
mpro7: Fair point — the key is always a `ResourceIri` (built from `iri.value -> label` and read with `linkLabels.get(lvc.referredResourceIri.value)`). Switching to `Map[ResourceIri, String]` so the `.value` falls away at the call site.

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/FindResourcesService.scala
mpro7: Pure extraction — the query is identical, I just lifted the subclass-expansion + project-graph resolution into `valuesClauseFor` so `buildClassQueryOrderedByLabel` shares it (the comment on lines 76–78 explains why keeping the subclass invariant in one place matters). No behavioural change.

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/FindResourcesService.scala
mpro7: Misleading name on the triplestore side — `Select.gravsearch(sparql)` is just `Select(sparql, SparqlTimeout.Gravsearch)`. It doesn't run a Gravsearch search; it picks the generous timeout. The actual query is a plain SPARQL SELECT with a VALUES clause and `OPTIONAL { ?r rdfs:label ?l }`. Same shape as the sibling `buildClassQuery`. Renaming the helper is worth a follow-up but out of scope here.

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/FindResourcesService.scala
mpro7: Good catch. Switching both row extractions (this one and the existing one in `findResources`) to `row.rowMap.get(resourceIriVar).someOrFail(InconsistentDataException(""))` — same failure type, no exception roundtrip.

webapi/src/main/scala/org/knora/webapi/slice/resources/service/ReadResourcesServiceLive.scala
mpro7: Three additive params (`queryStandoff`, `skipRetrievalChecks`, `standoffTagFilter`) added to the sync `readResourcesSequence`. They already exist on the `…Par` variant; streaming runs one batch at a time via the sync method, so it needs the same surface. All default to safe values (`false` / `None`) so existing callers are untouched. `standoffTagFilter` is what lets the export ask the triplestore for only footnote tags instead of fetching all standoff and filtering in the JVM.

webapi/src/test/scala/org/knora/webapi/slice/resources/service/ReadResourcesServiceFake.scala
mpro7: Same — fake must implement the trait, so it picks up the three new params with the same defaults.

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
BalduinLandolt: 1. I wasn't questioning the number, I was pointing out that it should be a configuration item rather than a hard-coded number.
2. These numbers don't tell us much as long as they are not run against prod; and even then, they might no longer hold, once we move prod to our own hardware. So again, it's not a question of "what is the best number", but a question of "should we really have to change code, release and deploy, to be able to change this number".

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
BalduinLandolt: Hm... makes sense. if we only use the label, not the resource, then it makes sense not to pass the entire thing; and then String is of course the right typing.

modules/test-it/src/test/scala/org/knora/webapi/slice/api/v3/export_/ExportStreamingIT.scala
BalduinLandolt, line 1: Not sure this test is really needed. But I don't mind it.

BalduinLandolt (approved):
See my latest comments:

- not sure if that parllelism comparison test is needed
- maybe I was to fast complaining about this change with the string type
- I still think we should put values that we may want to tune in config, not in code (batch size, parallelism)

modules/test-it/src/test/scala/org/knora/webapi/slice/api/v3/export_/ExportStreamingIT.scala
mpro7: Cost is low: excluded from default CI, lives in `test-it/`, runs only on demand and might be useful for further calibration. But I am also fine to remove it.

BalduinLandolt: nah, if cost is lo, then there's no reason to remove it

webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
mpro7: Thanks for pushing back, added in [767ce3a](https://github.com/dasch-swiss/dsp-api/pull/4124/commits/767ce3a429943eb6682f294cb327ec8e6c357d8c).

BalduinLandolt: Great, thanks!

PR #4125, opened 2026-05-28
docs: Add CONVENTIONS.md and REVIEW.md

BalduinLandolt:
## Summary

- Add top-level `CONVENTIONS.md` (work-phase agent reference card) and `REVIEW.md` (review-phase checklist), each linking out to `docs/development/` for depth — same pattern dsp-repository and sipi already use.
- Add `docs/development/dsp-api-error-handling.md` covering the fail-vs-die distinction, typed `IO[E, A]` vs `Task[A]`, the V3 error-variant pattern, and the narrow legacy `throw` carve-out.
- Trim the redundant Error Handling subsection in `dsp-api-conventions.md` to a pointer, so the two docs can't drift.
- Wire both new top-level files into `CLAUDE.md`.

## Motivation

- The dasch-claude-plugins `scala-zio-reviewer` agent looks for `CONVENTIONS.md` at repo root; today it falls back to a generic skill whose patterns contradict our actual conventions (it recommends trait + `*Live` splits, separate interface files, etc.).
- The on-PR rules (no `unsafeFrom` in RestServices, three-tier endpoint split, fail vs die discipline, V3 typed errors) were scattered across CLAUDE.md and several `docs/development/*.md` files. A single review checklist makes them enforceable.
- The fail-vs-die distinction — including the rule to always carry information with `ZIO.dieMessage` / `.orDieWith` — was not documented anywhere before.

## Test plan

- [ ] `npx markdownlint-cli -c .markdownlint.yml --disable MD013 MD040 -- CONVENTIONS.md REVIEW.md docs/development/dsp-api-error-handling.md docs/development/dsp-api-conventions.md` passes (verified locally)
- [ ] Links between CONVENTIONS.md / REVIEW.md / CLAUDE.md / `docs/development/*.md` resolve


REVIEW.md
```diff
@@ -0,0 +1,66 @@
...
+
+- [ ] `sbt check` passes (formatting + linting)
+- [ ] `sbt test` passes; integration / E2E tests cover changed behaviour
+- [ ] Commits follow [Conventional Commits](https://www.conventionalcommits.org/) — see `CONVENTIONS.md` for the prefix → changelog mapping enforced by `release-please`
+- [ ] "Knora" not used in commit messages, PR titles, docs, comments, or human-readable text — keep it to legacy packages / class names only
+- [ ] Every new source file has the Apache-2.0 + SPDX header (`sbt headerCheckAll`)
+
+### Services & API layering
+
+- [ ] New services are `final case class … private (…)` with `ZLayer.derive` in the companion — **no** new trait + `*Live` split for plain domain services
```
seakayone:
```suggestion
- [ ] New services are `final class (…)` with `ZLayer.derive` in the companion — **no** new trait + `*Live` split for plain domain services
```

Services usually have a public constructor, I am not sure if a private construtor would work with ZIO layer derive.

Also, I prefer now to use a `class` for the reason that the members are not public, for `case class`es the members are public and my IDE for example cannot determine then if a member is unused. Using `class` the unused member become apparent in the IDE.

docs/development/dsp-api-error-handling.md
```diff
@@ -0,0 +1,165 @@
...
+| Recoverable failure   | ZIO error channel — `ZIO.fail`, typed `IO[E, A]`  | `Result::Err`       |
+| Invariant violation   | `die` family — `ZIO.dieMessage`, `.orDieWith`     | `panic!` / `expect` |
+
+A *recoverable failure* is one a caller could reasonably handle: a 4xx response, a business rule violation, a missing resource. An *invariant violation* is unreachable state — reaching it means our model is wrong. If you can describe under what conditions the caller would handle it, it's recoverable.
+
+## Recoverable failures
+
+### Exception hierarchy
+
+dsp-api uses a sealed hierarchy of `Exception` subtypes as the failure values. These are not thrown — they're carried in the ZIO error channel.
```
seakayone: The `KnoraException` hierachy is only applicable for the responders and v2 rest services, newer code often has service specific error models which are mapped by the v3 rest services to the respective response errors.

In a later paragraph this is described properly but it might be confusing to see the `Exception hierachy` being applicable to the whole code base.

docs/development/dsp-api-error-handling.md
```diff
@@ -0,0 +1,165 @@
...
+
+Where the error set is small and obvious, narrow the channel (Style B / C). Where admin / v2 code already uses `Task[A]` + `BadRequestException`, keep it — we're not retrofitting.
+
+## Invariant violations
+
+When a branch represents state we believe is unreachable — a `None` after we've just checked `nonEmpty`, an unknown variant we've already exhaustively matched, an IRI we just successfully constructed — use the `die` family. **Always carry information about the invariant.**
+
+### Use `dieMessage` and `orDieWith`, not bare `die` / `orDie`
+
+`ZIO.dieMessage(String)` and `.orDieWith(e => Throwable)` exist precisely to attach an explanation. The message states the invariant in code — like Rust's `expect("…")` vs `unwrap()`. No comment needed.
```
seakayone:
```suggestion
`ZIO.dieMessage(String)` and `.orDieWith(e => Throwable)` exist precisely to attach an explanation. The message states the invariant in code. No comment needed.
```
The sentence is entirely clear without the Rust example, not sure if the Rust analogy is really useful here.

docs/development/dsp-api-error-handling.md
```diff
@@ -0,0 +1,165 @@
...
+- Do **not** introduce *new* non-effectful chunks just to keep throwing.
+- New effectful code uses `ZIO.fail` (recoverable) or `ZIO.dieMessage` (invariant).
+- When you touch one of the legacy `throw` sites, replacing it with the effectful equivalent is a welcome incidental cleanup — but not required as a precondition for unrelated changes.
+
+## Quick reference
+
+| You want to…                                             | Use                                                 |
+|----------------------------------------------------------|-----------------------------------------------------|
+| Reject a malformed request                               | `ZIO.fail(BadRequestException(...))`                |
+| Signal a missing resource                                | `ZIO.fail(NotFoundException(...))` or `.someOrFail` |
```
seakayone: As described above theses two examples here failing with a `RequestRejectedException` are only applicable to the v2 endpoints. v3 must be a `V3ErrorInfo`.

CONVENTIONS.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Stack
+
+Scala 3, ZIO 2, Tapir, zio-json, sbt. Package root: `org.knora.webapi`. Triplestore: Apache Jena Fuseki. Media server: Sipi.
+
+## Code Conventions
+
+### Services
+
+- `final case class ServiceName(private val …)` with `val layer = ZLayer.derive[ServiceName]` in the companion. No trait + `*Live` split for new domain services.
```
seakayone:
```suggestion
- `final class ServiceName(val …)` with `val layer = ZLayer.derive[ServiceName]` in the companion. No trait + `*Live` split for new domain services.
```
See above. Since this is redundant information maybe provide a reference here to the more detailed explanation.

CONVENTIONS.md
```diff
@@ -0,0 +1,93 @@
...
+- **Carve-out**: trait + `*Live` is allowed for repos and other seams that need in-memory test doubles (`KnoraGroupRepo` / `KnoraGroupRepoLive`, `OntologyRepo`, `JwtService`, `Authenticator`). New repo traits ship an in-memory companion at `webapi/src/test/.../service/<Name>InMemory.scala`.
+
+### API layer — three-tier split (mandatory)
+
+- `*Endpoints.scala` — Tapir endpoint definitions only.
+- `*ServerEndpoints.scala` — wiring of endpoint to RestService method. `.zServerLogic(_ => …)` for public, `.serverLogic(restService.x)` for secured.
+- `*RestService.scala` — handler logic. Every secured method has the curried shape `def x(user: User)(args…): Task[Resp]`. The body follows: **auth check → business logic → `format.toExternal(...)`**.
```
seakayone: Ideally `RestService`s contain a minimal amount of business logic and only handle auth and presentation logic. Plain `Service`s should contain the business logic.

CONVENTIONS.md
```diff
@@ -0,0 +1,93 @@
...
+- Register new endpoints in `CompleteApiServerEndpoints`.
+- Request / response DTOs live in a sibling object inside the `*Endpoints.scala` file (or a `*RequestsAndResponses.scala` next to it), not in a separate `model/` package.
+
+### Value objects
+
+- `final case class X private (value: …)` extending the appropriate `*Value` base, with `object X extends StringValueCompanion[X]` (or `WithFrom[I, X]`).
```
seakayone:
```suggestion
- `final case class X private (value: I)` extending the appropriate `*Value` base, with `object X extends  WithFrom[I, X]` or a specialized companion e.g. StringValueCompanion[X]` in case `I` is `String`.
```

CONVENTIONS.md
```diff
@@ -0,0 +1,93 @@
...
+- Smart constructor: `def from(value): Either[String, X]`. `unsafeFrom` is for known-good inputs only. Codecs declared with `given`. Full pattern in `docs/development/dsp-api-value-types.md`.
```
seakayone:
```suggestion
- Smart constructor: `def from(value): Either[String, X]`. `unsafeFrom` is for known-good inputs only, checks invariants and throws if invalid. Codecs declared with `given`. Full pattern in `docs/development/dsp-api-value-types.md`.
-  Factory functions should be named `makeNew`
```

CONVENTIONS.md
```diff
@@ -0,0 +1,93 @@
...
+
+### IRI handling
+
+- At the API boundary: `ZIO.fromEither(X.from(iri)).mapError(BadRequestException.apply)`.
```
seakayone: See above, the `BadRequestException` is v2 specific.

CONVENTIONS.md
```diff
@@ -0,0 +1,93 @@
...
+- Full rules — fail vs die, typed `IO[E, A]` vs `Task[A]`, V3 typed errors, the legacy `throw` carve-out — in `docs/development/dsp-api-error-handling.md`.
+
+### SPARQL
+
+Never concatenate query strings. Use rdf4j SparqlBuilder via the helpers in `slice/common/repo`. See `docs/development/dsp-api-sparql-queries.md`.
+
+### Imports & formatting
+
+- No fully-qualified class names in code bodies — import at the top of the file.
+- Order: stdlib / ZIO → third-party → internal `org.knora.*` (alphabetical within each group), single blank line between groups.
```
seakayone: AFAIU the order is rewritten anyway with `sbt fmt`, no need to define it here.

CONVENTIONS.md
```diff
@@ -0,0 +1,93 @@
...
+- Scalafmt via `sbt fmt`; CI runs `sbt check`.
+- Every source file carries the Apache-2.0 + SPDX header (managed by `sbt headerCreateAll` / `headerCheckAll`).
+
+### Naming in human-readable text
+
+The legacy name "Knora" lives on in packages (`org.knora.webapi`) and class names — leave them. Do **not** use "Knora" in commit messages, PR titles, docs, comments, spec files, or learning documents. Use "dsp-api".
```
seakayone: As much as the wish for `Knora` being legacy exists, the ontologies were never renamed and are in active use. I am currently writing code where a method `toKnoraBaseOntology` would get a very awkward name when following the rule strictly.

CONVENTIONS.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Testing Conventions
+
+- `object XSpec extends ZIOSpecDefault`; `suite("X should")(test("...") { … })`; layers via `.provide(...)`. Use `TestAspect.withLiveClock` for time-dependent tests.
+- New repo traits ship an in-memory companion under `webapi/src/test/.../service/<Name>InMemory.scala`.
+- Test data: prefer self-contained fixtures next to the component. Only fall back to shared `test_data/` sets when unavoidable. When adding to a shared set, verify an actual *instance* of the scenario exists — not just that the schema supports it.
+- Test locations: unit → `webapi/src/test/scala/`; integration → `modules/test-it/`; ingest integration → `modules/test-ingest-integration/`; E2E → `modules/test-e2e/`; shared utilities → `modules/testkit/`.
+
```
seakayone, line 66: Would it make sense to shortly mention or describe `GoldenTest`?

CONVENTIONS.md
```diff
@@ -0,0 +1,93 @@
...
+| `fix:`       | Bug Fixes                  |
+| `perf:`      | Performance Improvements   |
+| `docs:`      | Documentation              |
+| `test:`      | Tests                      |
+| `deprecated:`| Deprecated                 |
+| `refactor:`  | Maintenances               |
+| `build:`     | Maintenances               |
+| `chore:`     | Maintenances               |
+
+Breaking changes use `!` (e.g. `feat!:`). Do not use "Knora" in commit messages; use "dsp-api".
```
seakayone, line 83: I also like to use the _optional scope_ for commit in order to mark the change affecting a specific release artifact. 

For example `feat(dsp-api): Change A`, `bug(sipi): Fix B` ....

REVIEW.md
```diff
@@ -0,0 +1,66 @@
...
+- [ ] Trait + `*Live` is acceptable only for repos and explicit test seams; any such new repo ships an in-memory companion at `webapi/src/test/.../service/<Name>InMemory.scala`
+- [ ] Endpoints split across the three tiers: `*Endpoints` (definition) → `*ServerEndpoints` (wiring) → `*RestService` (logic) — and registered in `CompleteApiServerEndpoints`
+- [ ] Secured RestService methods are curried `def x(user: User)(args…): Task[Resp]` (Tapir wiring depends on this)
```
seakayone: Nit pick" the method definition is not in fact currying, it enables currying. Currying is not used at all here.

Using multiple parameter lists enables wiring in the `RestService` method in the `ServerEndpoint` very concise without having to enumerate all params. This makes adding additional params to the endpoints easier, you simply have to add the params to the endpoint definition and the method, no change in the `ServerEndpoint` necessary.

REVIEW.md
```diff
@@ -0,0 +1,66 @@
...
+- [ ] Secured RestService bodies follow **auth check → business logic → `format.toExternal(...)`** — `toExternal` is not skipped
+- [ ] `AuthorizationRestService.ensure*` results are re-used (don't reload the project / group after the auth call returned it)
+- [ ] Request / response DTOs live next to the `*Endpoints` (sibling object or `*RequestsAndResponses.scala`), not in a separate `model/` package
+
+### IRI handling
+
+- [ ] IRI conversion happens at the API boundary: `ZIO.fromEither(X.from(iri)).mapError(BadRequestException.apply)`
```
seakayone: IRI conversion is only necessary for ontology, resource class and value IRIs which have different ApiSchema representations. Other IRIs, like `ValueIri` or `ResourceIri` can in fact be used in the endpoint definitions.

seakayone:
Until now I have treated `CLAUDE.md` as what you have now have introduced as `CONVENTIONS.md`. 

In `CLAUDE.md` I have added short examples with links to the deep dive in `doc/development/...`. Would it make sense to migrate the existing examples to the conventions file, so the scala-zio-reviewer would be able to pick these up too? 

What do you think belongs into the conventions and what belongs into the claude file?

BalduinLandolt:
## Comment Resolution Summary

Addressed the open (unresolved) review threads from @seakayone. All changes are documentation-only.

### Changes Made

- `docs/development/dsp-api-error-handling.md:20` — Scoped the `KnoraException` hierarchy to the responders / v2 rest services and noted that new v3 code uses service-specific error models (with a link to Style C), so the hierarchy is not read as applying codebase-wide.
- `docs/development/dsp-api-error-handling.md` (quick reference) — Marked the `BadRequestException` / `NotFoundException` rows as admin / v2 and broadened the v3 row to "Reject / fail a new v3 endpoint", so v2-specific guidance is not applied to v3.
- `CONVENTIONS.md:20` + `REVIEW.md` — RestServices now described as auth / presentation only with minimal business logic; business logic lives in plain `*Service`s ("auth check → delegate to service → `toExternal`").
- `CONVENTIONS.md:32` + `REVIEW.md` — IRI boundary mapping marked admin / v2; v3 maps the failure to the endpoint's typed `V3ErrorInfo` variant instead of `BadRequestException`.
- `CONVENTIONS.md:49` + `REVIEW.md` — Dropped the manual import-ordering rule; noted that `sbt fmt` orders imports automatically.
- `CONVENTIONS.md:55` + `REVIEW.md` — Carved out `knora-base` / `knora-admin` ontology identifiers (e.g. `toKnoraBaseOntology`) from the "no Knora in human-readable text" rule, since those ontologies were never renamed and are in active use. The rule now targets free-form prose only.
- `CONVENTIONS.md:63` — Added a short `GoldenTest` description (mix in the trait, `assertGolden(actual, "suffix")`, expected value stored under `src/test/resources/`, regenerate with `rewrite = true`) with a pointer to `webapi/src/test/scala/org/knora/webapi/GoldenTest.scala`.
- `CONVENTIONS.md:80` — Documented the optional Conventional Commits scope, e.g. `feat(dsp-api): …`, `fix(sipi): …`.
- `REVIEW.md:20` + `CONVENTIONS.md` — Corrected "curried" to "multiple parameter lists" and explained the wiring benefit (adding a param touches only the endpoint definition and the method).
- `REVIEW.md:27` + `CONVENTIONS.md` — Clarified that only ontology / resource-class / property IRIs (schema-dependent representations) are taken as `IriDto` and converted via `IriConverter`; simple IRIs like `ProjectIri`, `ResourceIri`, `ValueIri` are used directly in the endpoint definition.

### Notes

- The general thread about migrating `CLAUDE.md` examples into `CONVENTIONS.md` (what belongs in conventions vs the claude file) is a design discussion rather than a concrete change request, so it was left for direct reply rather than edited blindly.
- Line lengths in these reference files already exceed the 120-char MD013 limit throughout (one bullet per line by design); the edits follow the same established style.

Commit: 5a78e58e0

BalduinLandolt:
@seakayone thanks for the review!

I applied all your suggestions and had Claude work through all your comments after verifying the with everything I either agree or I don't know and take your word for it. :)

> Until now I have treated `CLAUDE.md` as what you have now have introduced as `CONVENTIONS.md`.
> 
> In `CLAUDE.md` I have added short examples with links to the deep dive in `doc/development/...`. Would it make sense to migrate the existing examples to the conventions file, so the scala-zio-reviewer would be able to pick these up too?
> 
> What do you think belongs into the conventions and what belongs into the claude file?

Regarding that question: Perhaps these things could/should be moved to the conventions file, but I don't see it as urgent, because in the claude.md file they won't be missed either, and they shouldn't bloat the context either.  
Generally, my take is to keep claude.md as short as possible, because too long agent files have been proven to be counter-productive. To my knowledge, the best thing one can do is use "progressive disclosure", providing the agent with the information where it can find what, rather than putting the thing right into the context.  
But for this case, I have no way of judging if it would be better or worse to move these to the context file, or not.

seakayone (approved):

PR #4126, opened 2026-05-29
feat: optional explicit value order on create (DEV-6573)

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
@BalduinLandolt @siers 
Claude still flags some critical issues, but I already did about 7 cycles of review and fix...

BalduinLandolt:
> @BalduinLandolt @siers Claude still flags some critical issues, but I already did about 7 cycles of review and fix...

Could you paste a sample of some of the critical issues Claude flags? That would help to judge if they are worth following up on, or not

BalduinLandolt (commented):
To me, everything looks reasonable.

Would still be curious to see what issues Claude still flags.

Also, maybe I missed something… but I only saw tests for the Values responder, not for the Resources responder. Is this intentional?

Nora-Olivia-Ammann:
> To me, everything looks reasonable.
> 
> Would still be curious to see what issues Claude still flags.
> 
> Also, maybe I missed something… but I only saw tests for the Values responder, not for the Resources responder. Is this intentional?

I don't know. I said it must be implemented when creating new values and resources. And it said it did... is that not correct then?

Nora-Olivia-Ammann:
@BalduinLandolt , there are more, but these are the first two:

7 Critical | 9 Warning | 5 Suggestion**

---

### `webapi/.../slice/resources/repo/service/ValueRepo.scala`

- **Critical** — `checkDuplicateOrder` uses `.orDie` on `triplestore.query(...)`, which turns any triplestore error (database unreachable, malformed response) into a fatal fiber defect rather than a proper 5xx. The method signature `IO[BadRequestException, Unit]` hides this: callers believe the only failure mode is a 400, but the JVM can also crash the fiber silently. Use `Task[Unit]` and let the caller decide what an infra failure means. *(scala-zio-reviewer)*

- **Critical** — `checkDuplicateOrder` is the repo-layer enforcement for ordinary values, but the same check for **link values** was added at the **responder layer** (`ValuesResponderV2.createLinkValueV2AfterChecks`). This splits enforcement across two layers asymmetrically: a future reader adding a new value type or refactoring must know about both call sites. The ordinary-value pattern (check inside repo) should be the model; the link-value check belongs in `ValueRepo` or a `createLinkValue` equivalent. *(pattern-recognition-specialist)

BalduinLandolt:
Ok, that's at the level of _should fix_, not _must fix_, so this shouldn't block us here. But letting it clean up for some more rounds wouldn't hurt either, I suppose.

Nora-Olivia-Ammann:
> Ok, that's at the level of _should fix_, not _must fix_, so this shouldn't block us here. But letting it clean up for some more rounds wouldn't hurt either, I suppose.

So you think I should just run it some more?

BalduinLandolt:
If claude can resolve these issues, I see no reason not to do it

BalduinLandolt (approved):
Looks good, thanks!

PR #4132, opened 2026-06-01
fix(export): lower CSV export batch-size default to 100 (DEV-6336)

mpro7:
[DEV-6336](https://linear.app/dasch/issue/DEV-6336)

## Problem

Streaming CSV export (`POST /v3/export/resources`) stalls mid-download and fails with an **empty body and no error** for large rich-text classes (e.g. `ugofoscolo` `Image`) — reproducibly around a fixed point (~14%). Lighter classes (~10k light-text resources) work fine.

## Root cause

The stream flushes **one chunk per fully-materialized batch**, so `batch-size` also bounds the gap between bytes on the wire — a batch's standoff `CONSTRUCT` + row encoding must finish before its rows are sent. The default `500` was tuned on light-text data (incunabula `page`). For rich-text/standoff-heavy classes a 500-resource batch takes far longer; combined with order-preserving `mapZIOPar` (a slow in-order batch blocks all downstream flushes while later completed batches buffer), the response goes silent longer than the **ingress 60s idle timeout** → the connection is dropped → Angular's blob download discards the partial body. Because resources are label-ordered, the heavy cluster sits at a fixed position, so it fails at the same ~14% each time.

This is the **same 60s ingress idle timeout** that #4097 (DEV-6288) documented as its known limitation, explicitly deferred to DEV-6336:

> *The full export response is buffered before being sent to the client — connections exceeding the ingress 60s idle timeout will still fail on prod for very large classes. Tracked in DEV-6336.*

Streaming (the original DEV-6336 work) fixed time-to-first-byte but not the sub-60s inter-byte cadence, so the timeout still bites — just mid-transfer instead of at the start.

## Change

- **`app.export.batch-size` default `500 → 100`** so even heavy batches complete well under the idle window, keeping bytes flowing. Overridable per deployment via `KNORA_WEBAPI_EXPORT_BATCH_SIZE` (raise it for light-resource projects).
- `ExportStreamingIT`: repoint the regression-guard combo `(5, 500) → (5, 100)` and document that the IT measures total throughput, **not** the inter-flush gap that drives the production failure.

The per-batch query load is unchanged; only the on-wire flush cadence. The #4124 tuning matrix shows `bs=100` is at worst neutral vs `bs=500` on light data (and avoids Fuseki's large-`VALUES` planning cliff), so no throughput regression is expected.

## Follow-up (infra, not in this PR)

Raise `nginx.ingress.kubernetes.io/proxy-read-timeout` (and confirm `proxy-buffering` off) for the export route in ops-deploy. No batch size can guarantee <60s for a single pathologically large resource; the ingress bump is the robust backstop.

## Immediate mitigation

`KNORA_WEBAPI_EXPORT_BATCH_SIZE=100` can be set on dev/prod now — no deploy required.

## Tests

- `sbt "webapi/testOnly *ExportServiceSpec*"` — 21 pass (incl. the config fix-confirmation test asserting `batch-size` drives batching, and the streaming-shape/ordering/cross-batch tests).
- No new test added: a `batchSize == 100` assertion would be a config-mirror, and a timing test can't reproduce the ingress idle-timeout condition (no ingress in tests; `runCollect` drains the whole stream). The existing batching + streaming-shape guards cover the behavioral contract.

### DIFF
--- a/modules/test-it/src/test/scala/org/knora/webapi/slice/api/v3/export_/ExportStreamingIT.scala
+++ b/modules/test-it/src/test/scala/org/knora/webapi/slice/api/v3/export_/ExportStreamingIT.scala
@@ -23,8 +23,14 @@ import org.knora.webapi.slice.ontology.repo.service.OntologyCache
  *
  * Two purposes:
  *   1. Tuning vehicle -- emits per-combo timings so we can pick production defaults from data, not guesses.
- *   2. Regression guard -- the matrix asserts that the production combo (5, 500) completes within a generous wall-time
+ *   2. Regression guard -- the matrix asserts that the production combo (5, 100) completes within a generous wall-time
  *      budget and produces the expected number of CSV rows.
+ *
+ * Caveat: this IT `runCollect`s the whole stream and measures *total throughput*. It does NOT observe the
+ * inter-flush gap that drives the real production failure mode (a single slow in-order batch going silent past the
+ * ingress 60s idle timeout). incunabula `page` is also light-text data; rich-text/standoff classes have far higher
+ * per-batch latency. Treat these timings as a throughput sanity check, not as validation of the idle-timeout fix --
+ * that lives in the conservative `app.export.batch-size` default.
  *
  * Excluded from default CI because it is multi-minute (Fuseki testcontainer startup + incunabula bulk-load + matrix).
  * Run on demand with: `sbt "test-it/testOnly *ExportStreamingIT"`.
@@ -75,7 +81,7 @@ object ExportStreamingIT extends E2EZSpec {
                     } yield (parallelism, batchSize, java.time.Duration.ofNanos(t1 - t0), lines)
 
         // Warmup: prime Fuseki query plan cache + JVM JIT for the production combo. Discarded.
-        _ <- runOnce(5, 500)
+        _ <- runOnce(5, 100)
 
         results <- ZIO.foreach(matrix) { case (p, b) => runOnce(p, b) }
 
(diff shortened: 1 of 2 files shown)


siers (approved):

mpro7:
## Empirical validation against dev data

Tested the local API (this branch) against **dev Fuseki**, real problem class `084B ugofoscolo:Image` (**31,613 resources**), with a local nginx standing in for the prod ingress (`proxy_read_timeout`, `proxy_buffering off`).

| Run | Idle timeout | Result |
|---|---|---|
| Direct, bs=100 | none | ✅ 31,614 rows · TTFB 1.9s · total 196s · **max inter-chunk gap 43s** |
| Via nginx, bs=100 | **60s** (prod ingress) | ✅ full 31,614 rows — fix survives the real timeout |
| Via nginx, bs=100 | 30s (< 43s gap) | ❌ drops at **4,701 rows = 14.9%** · `http 200` truncated · `curl exit 18` · nginx `upstream timed out while reading upstream` |

The 30s run reproduces the reported symptom exactly: **stall at ~14%, no error surfaced, empty/partial body** (status already committed to 200, so the mid-stream drop can't become a 5xx). It also confirms the mechanism — a single in-order batch whose fetch+encode gap exceeds the ingress idle timeout drops the connection. `bs=500` reproduces this against the real 60s ingress, because its per-batch gap is ~5× the bs=100 gap (≈200s ≫ 60s) — and may also hit the 120s `SparqlTimeout.Maintenance` cap.

### Caveat → sharpened recommendation

bs=100 survives 60s here, but the worst gap was **43s — only ~17s of headroom**. On a busier prod Fuseki or a heavier class that margin can vanish. So:

- **bs=100 is necessary but not comfortably sufficient.** Worth considering a lower default (e.g. 50).
- **The ingress `proxy-read-timeout` bump (ops-deploy) should be treated as required, not optional** — no batch size can guarantee <60s for a single pathologically large resource.

Repro recipe (local API + dockerized nginx with `proxy_read_timeout` 60s vs 30s, POST to `/v3/export/resources` for the class above) available on request.

PR #4138, opened 2026-06-02
feat(ontology): Reify LinkValues in transformer

seakayone:
### Description

A `knora-base` `LinkValue` is an `rdf:Statement` reification of a direct link between two resources, and the public Knora-API v2 representation does not carry that reified form. This stage-2 step rebuilds it.

For each `LinkValue`, the transformer derives the direct property from the link property (dropping the `Value` suffix), emits the direct-link triple between the resource and the target, and reifies the value with `rdf:subject`/`rdf:predicate`/`rdf:object`, a `valueHasRefCount` of 1, and a `valueHasString` of the target IRI — mirroring how the API itself writes link values. The transient `linkValueHasTargetIri` is dropped.

The reification is keyed on `(resource, directProperty, target)` so the upcoming standoff conversion can find and bump the reference count of an existing link rather than create a duplicate.


Nora-Olivia-Ammann (approved):
LGTM

BalduinLandolt (approved):

PR #4140, opened 2026-06-03
feat: Add create-only project data-graph import endpoints (v3)

seakayone:
### Description

Adds a v3 API to create a new project's data graph from a knora-api (v2 external schema) JSON-LD payload, so that data prepared outside DSP can be imported through the REST API without a full BagIt migration package. Until now the ontology transformer existed only as internal code with no API surface.

New endpoints (system-admin only, gated behind the new `allow-project-data-import` feature flag, disabled by default):

- `POST /v3/projects/{projectIri}/data-imports` — streams JSON-LD, returns `202` with a task ID (async, same pattern as the migration import)
- `GET /v3/projects/{projectIri}/data-imports/{importId}` — status polling
- `DELETE /v3/projects/{projectIri}/data-imports/{importId}` — task cleanup

Key design decisions:

- **Create-only**: importing fails with `409 data_graph_exists` if the project already has a data graph — checked synchronously at request time and re-verified before the upload, since data-graph imports are not mutually exclusive with migration imports (each task kind has its own `DataTaskState`).
- The transformer now emits N-Quads with every quad assigned to the project's data named graph, derived from the project — `@graph` declarations in the payload are ignored, making cross-graph contamination structurally impossible. Synthesised creation dates use `xsd:dateTime` to conform to knora-base and the SHACL shapes.
- The permissions literal is resolved once from the project's DOAPs for the importing system admin (ProjectAdmin-group DOAP has first precedence and applies uniformly); per-class/per-property DOAPs are not resolved per entity.
- SHACL validation reuses the migration import's shapes; the project's ontology graphs and the admin data graph are fetched from the triplestore since the payload carries instance data only.

See `docs/03-endpoints/api-v3/project-data-import.md` for the endpoint reference.

### DIFF
--- a/docs/03-endpoints/api-v3/project-data-import.md
+++ b/docs/03-endpoints/api-v3/project-data-import.md
@@ -1,0 +1,109 @@
+# Project Data Import
+
+The project data import API creates a new project's data graph from a [knora-api (v2 external schema)](../api-v2/introduction.md)
+JSON-LD payload. The server transforms the payload into the internal knora-base representation, validates it, and
+streams it into the triplestore.
+
+Unlike the [Project Migration Export/Import](project-migration.md) — which moves a complete project (admin data,
+ontologies, permissions, assets) between instances as a BagIt package — the data import handles **instance data
+only**. The project and its ontologies must already exist on the instance, and the project's data graph must not
+exist yet (create-only).
+
+## Endpoints
+
+For request/response schemas, error codes, and interactive testing, see the
+[OpenAPI documentation](https://api.dasch.swiss/api/docs/#/API%20v3).
+
+| Route                                                | Method   | Description                             |
+| ---------------------------------------------------- | -------- | --------------------------------------- |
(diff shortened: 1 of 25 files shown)


BalduinLandolt (approved):
👍

Nora-Olivia-Ammann:
The description looks good to me. There is one blocker we must not forget @BalduinLandolt .
For avisblatt and other projects we must be able to resolve the DOAP on a per resource and property level.

BalduinLandolt:
Yes, yesterday @christian.kleinboelting  and I talked about this, and there are some limitations we need to resolve before this can serve to bring Avisblatt to Prod - such as permissions, or being able to create the resources in the name of a user that is not SysAdmin. But these things can be resolved moving forward, and this PR should enable you to test the new bulk import before it is 100% finished.

Nora-Olivia-Ammann:
@BalduinLandolt Absolutely!
Do we have a project where we have the checklist written down?

BalduinLandolt:
@christian.kleinboelting has a plan document with open points noted; currently only locally, but he said he'd push it to the specs repo. We can turn that document into a project.

PR #4154, opened 2026-06-18
feat(maintenance): increase the time-out for the `TriplestoreService` in the maintenance endpoint (DEV-6655)

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
@BalduinLandolt It turned out that this was sufficient, no memory increase was needed. This now works locally with the default API docker config (the compose file in the root). That should work for the servers too, right?

BalduinLandolt (changes requested):
This change is problematic in two ways:

1. we should not apply a specific timeout (`triplestoreConfig.maintenanceTimeout`) to a general-purpose method (this changes all update queries). Instead we should follow the established pattern of allowing to override the timeout on the query object (missing on `Update`, but present on `Select` and `Construct`, using the `SparqlTimeout` enum.
2. padding a config-value with `+ 20` doesn't make sense. If that is needed, then the value should be changed in the config.

Nora-Olivia-Ammann:
I tried it out, it also works with this version.

BalduinLandolt (approved):
👍

PR #4173, opened 2026-07-02
fix: don't inject order index into JSON-LD value objects on resource create

BalduinLandolt:
### Description

**Bug fix.**

Creating a resource that carries a bare datatype property serialised as JSON-LD *value objects* (e.g. `knora-api:hasResourceAuthorship` sent as `[{"@value": "...", "@type": "xsd:string"}, ...]`, which is how dsp-tools serialises it) failed with:

```
400  JsonLdError: A value object with disallowed entries has been detected [INVALID_VALUE_OBJECT]
```

**Root cause:** `ApiComplexV2JsonLdRequestParser.injectOrderIndices` appends an internal `orderIndex` entry to *every* element of *every* top-level array property. For value objects (literals) this produces `{"@value": ..., "@type": ..., "orderIndex": 0}` — an invalid value object — so JSON-LD expansion rejects the whole request.

This never surfaced before because:
- **file** authorship (`hasAuthorship`) is nested inside the file-value object, not a top-level array key, so it was never touched;
- authorship sent as **plain strings** (`["a", "b"]`) has non-object array elements, so it was never touched — which is the only form the existing tests used.

**Fix:** only inject the order index into node objects; skip JSON-LD value objects (those with an `@value` member). Order indices are only meaningful for Knora value nodes, and bare datatype properties are unordered.

Found while running dsp-tools' e2e round-trip against a dev API image for [dsp-tools#2350](https://github.com/dasch-swiss/dsp-tools/pull/2350) (resource-side legal metadata). Without this, any `xmlupload` of a resource with resource-side authorship 400s.

**Tests:** added a parser test covering the typed-value-object serialisation; the existing plain-string test is retained, so both formats are now covered. `webapi/testOnly *ApiComplexV2JsonLdRequestParserSpec` passes (65 tests); scalafmt clean.

**Breaking change?** No.

https://claude.ai/code/session_013tvarQfKWHQiNi1zKEMuZt

### DIFF
--- a/webapi/src/main/scala/org/knora/webapi/slice/common/ApiComplexV2JsonLdRequestParser.scala
+++ b/webapi/src/main/scala/org/knora/webapi/slice/common/ApiComplexV2JsonLdRequestParser.scala
@@ -74,7 +74,10 @@ final case class ApiComplexV2JsonLdRequestParser(
         val modified = obj.fields.map {
           case (key, Json.Arr(values)) if !key.startsWith("@") =>
             val indexed = values.zipWithIndex.map {
-              case (Json.Obj(fields), idx) =>
+              // only node objects carry an order; JSON-LD value objects (those with an @value entry,
+              // e.g. a bare datatype property such as hasResourceAuthorship) must be left untouched,
+              // as an extra entry would make them an invalid value object during expansion
+              case (Json.Obj(fields), idx) if !fields.exists(_._1 == "@value") =>
                 Json.Obj(fields :+ (OrderIndexProperty -> Json.Num(idx)))
               case (other, _) => other
             }
--- a/webapi/src/test/scala/org/knora/webapi/slice/common/ApiComplexV2JsonLdRequestParserSpec.scala
+++ b/webapi/src/test/scala/org/knora/webapi/slice/common/ApiComplexV2JsonLdRequestParserSpec.scala
@@ -2074,6 +2074,38 @@ object ApiComplexV2JsonLdRequestParserSpec extends ZIOSpecDefault {
           Set(Authorship.unsafeFrom("Lotte Reiniger"), Authorship.unsafeFrom("Hilma af Klint")),
       )
     },
+    test("createResourceRequestV2 parses hasResourceAuthorship sent as typed value objects (dsp-tools serialisation)") {
+      for {
+        _      <- ZIO.serviceWithZIO[KnoraProjectRepo](_.save(TestDataFactory.someProject))
+        uuid   <- Random.nextUUID
+        result <- service(
+                    _.createResourceRequestV2(
+                      """{
+                        |  "@type": "ex:Thing",
+                        |  "rdfs:label": "test resource",
+                        |  "ka:attachedToProject": { "@id": "http://rdfh.ch/projects/0001" },
+                        |  "ka:hasResourceAuthorship": [
+                        |    { "@value": "Lotte Reiniger", "@type": "xsd:string" },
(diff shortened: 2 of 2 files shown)


webapi/src/main/scala/org/knora/webapi/slice/common/ApiComplexV2JsonLdRequestParser.scala
```diff
@@ -74,7 +74,10 @@ final case class ApiComplexV2JsonLdRequestParser(
         val modified = obj.fields.map {
           case (key, Json.Arr(values)) if !key.startsWith("@") =>
             val indexed = values.zipWithIndex.map {
-              case (Json.Obj(fields), idx) =>
+              // only node objects carry an order; JSON-LD value objects (those with an @value entry,
+              // e.g. a bare datatype property such as hasResourceAuthorship) must be left untouched,
+              // as an extra entry would make them an invalid value object during expansion
```
siers, line 79: 👀

siers (approved):

webapi/src/main/scala/org/knora/webapi/slice/common/ApiComplexV2JsonLdRequestParser.scala
BalduinLandolt: Could you elaborate? ^^

This comment I think is actually helpful, otherwise I would have no idea why this `if !fields.exists(_._1 == "@value")` guard is there, just from looking at it.

siers: I am looking and printing some values at ValueOrderingSpec to understand how these indexes are inserted, but I'm still not sure I follow, so I suppose I'm out of depth here.

Interesting that in ApiComplexV2JsonLdRequestParserSpec, there is no test that shows a OrderIndexProperty being created/tested, if I'm reading it right

siers (approved):

webapi/src/main/scala/org/knora/webapi/slice/common/ApiComplexV2JsonLdRequestParser.scala
BalduinLandolt: yea, this is a bit of an ugly hack - it used to be that values were parsed from a JSON list, so they preserved order. Eventually this was changed to reading the payload as an RDF object and extracting data from there; but because RDF properties are order agnostic, this introduced a regression, where the order became random (or inversed, or something like that). We fixed this by adding this order injection (which in and by itself feels a bit hacky, but solved the problem). And later we exposed the option to pass value order explicitly, but for compatibility reasons we had to keep the order injection.

This value order injection is only supposed to happen to values (i.e. objects of the knora-base:Value types), not to literals. And it checks simply for "is it a JSON-LD object?" which approximates "is it a Value Object?".

Now it turned out that the following JSON-LD triple:

```jsonc
{
  // resource
  "hasAuthorship": "Foo"
}
```

can also be expressed as:

```jsonc
{
  // resource
  "hasAuthorship": {
    "@type": "xsd:string",
    "@value": "Foo"
  }
}
```

And that is what dsp-tools is sending.  
And obviously, this is now also a JSON object, so it looked like a value object to the order injector.

BalduinLandolt: And there is a good chance we're missing tests for many of these things. But that would be out of scope for this PR.

PR #4175, opened 2026-07-02
test: dedupe e2e console logger and harden flaky replace-user-iri auth tests

BalduinLandolt:
## Problem

CI run [28602440910](https://github.com/dasch-swiss/dsp-api/actions/runs/28602440910/job/84813904024) was unreadable: the *expected* "project already exists" import failures from `ProjectMigrationImportE2ESpec` produced **282 identical ERROR stack-trace blocks**, burying the actual failure.

## Root cause (commit 1)

zio-test's sbt runner installs every spec's `bootstrap` into a shared runtime. `E2EZSpec` installed a fresh `consoleLogger(...)` layer per spec class, so with **47** `E2EZSpec` subclasses in the e2e run, 47 distinct logger instances accumulated in `FiberRef.currentLoggers` — every server-side ERROR line was printed 47 times (2 log calls × 3 failed imports × 47 loggers = 282 blocks).

**Fix:** create a single shared `ZLogger` instance in `object E2EZSpec` and install it via `Runtime.addLogger`. `FiberRef.currentLoggers` is a `Set`, so installing the same instance once per spec bootstrap is idempotent — one logger for the whole run.

## Flaky auth test (commit 2)

The run's actual failure was `MaintenanceReplaceUserIriE2ESpec` — "returns 403 when the authenticated user is not SystemAdmin" — a client-side timeout, the same runner-contention failure mode analyzed in #4169. That PR hardened only the sibling `MaintenanceReplaceUserIriInProjectE2ESpec`; this commit applies the identical `@@ TestAspect.timeout(30.seconds) @@ TestAspect.flaky` mitigation to the users-endpoint spec's 401/403 tests. (The re-run of 28602440910 passed, confirming the flake.)

## Verification

- A/B test with two specs in one sbt invocation (`MaintenanceReplaceUserIriE2ESpec` + `ProjectMigrationImportE2ESpec`): baseline prints each ERROR line **2×**, fixed prints it exactly **1×**; all 14 tests pass in both runs.
- `test-e2e/Test/compile`, `sbt fmt` ✅

https://claude.ai/code/session_01A33MeZAQDyeaL8tWiRyXpN

### DIFF
--- a/modules/testkit/src/main/scala/org/knora/webapi/E2EZSpec.scala
+++ b/modules/testkit/src/main/scala/org/knora/webapi/E2EZSpec.scala
@@ -36,13 +36,6 @@ abstract class E2EZSpec extends ZIOSpec[E2EZSpec.Environment] {
   implicit val sf: StringFormatter = StringFormatter.getInitializedTestInstance
   val faker: Faker                 = new Faker()
 
-  private val testLogger: ULayer[Unit] = Runtime.removeDefaultLoggers >>> consoleLogger(
-    config = {
-      val default = ConsoleLoggerConfig.default
-      default.copy(filter = default.filter.withRootLevel(LogLevel.Error))
-    },
-  )
-
   /**
    * The OpenTelemetry layer the application under test runs with. Defaults to the stdout-exporter setup;
    * instrumentation specs override it (e.g. with an in-memory span exporter) to assert on emitted spans.
@@ -51,7 +44,7 @@ abstract class E2EZSpec extends ZIOSpec[E2EZSpec.Environment] {
   protected def otelLayer: ULayer[api.OpenTelemetry & Tracing & ContextStorage] = OtelSetup.layer
 
   override lazy val bootstrap: ULayer[E2EZSpec.Environment] =
-    testLogger >>>
+    E2EZSpec.testLogger >>>
       TestContainerLayers.all >+>
       LayersLive.remainingLayer(otelLayer) >+>
       ApiModule.layer >+>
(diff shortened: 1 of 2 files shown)


modules/testkit/src/main/scala/org/knora/webapi/E2EZSpec.scala
```diff
@@ -90,6 +83,20 @@ abstract class E2EZSpec extends ZIOSpec[E2EZSpec.Environment] {
 
 object E2EZSpec {
 
+  // zio-test's sbt runner installs every spec's `bootstrap` into a shared runtime, so a per-spec
+  // `consoleLogger(...)` layer adds a fresh logger instance for each E2EZSpec subclass in the run,
+  // duplicating every log line once per spec class. `FiberRef.currentLoggers` is a Set, so
+  // installing this single shared logger instance instead is idempotent.
+  private val sharedConsoleLogger: ZLogger[String, Any] = {
```
siers, line 90: another one of these AI comments that's missing the mark. would you agree?

siers (approved):

modules/testkit/src/main/scala/org/knora/webapi/E2EZSpec.scala
BalduinLandolt: I can remove it, if you prefer. But personally, I don't mind it, because I'm not reading that code. And if it prevents future AI runs from being confused about this, that's a plus.  
But as you prefer!

BalduinLandolt: merged it as is, because the log noise was really bothersome in another CI run I had to look at. But I'm happy to change it any time

PR #4176, opened 2026-07-03
chore: add checked-in Metals MCP server for agent workflows

BalduinLandolt:
## What

Adds a checked-in, project-scoped **`metals` MCP server** so Claude Code / agent workflows get real Scala language intelligence in this repo — compiler diagnostics, type-aware usage search, symbol docs and lookup — instead of relying on grep/read.

## Changes

- **`.mcp.json`** (new, repo root): defines the `metals` server, launched as `metals-mcp --workspace ${CLAUDE_PROJECT_DIR:-.} --transport stdio`. stdio means each session spawns/manages its own process — no long-lived HTTP server or stale hardcoded port. `${CLAUDE_PROJECT_DIR:-.}` (not a bare `.`) is used because a spawned stdio server's cwd is not guaranteed to be the repo root.
- **`.claude/settings.json`**: `enabledMcpjsonServers: ["metals"]` to pre-approve the server for the team.
- **`docs/development/dsp-api-metals-mcp.md`** (new): full guidance — prerequisites (`cs install metals-mcp`, Metals ≥ 1.6.6), the high-value tools, the `edit → compile-file → query` usage pattern (with `compile-file` as source of truth), cold-start expectations on a large multi-module build, and the concurrency caveats (one server per checkout; separate worktrees OK with resource cost; **launch from the dsp-api directory under LOOM** since `.mcp.json` discovery is launch-cwd-only).
- **`CLAUDE.md`**: short pointer section to the new doc.

## Notes for reviewers

- Requires `metals-mcp` on each dev's PATH. If absent, the server just fails to start and agents fall back to grep/read — nothing else breaks.
- On recent Claude Code, checked-in approvals are ignored in untrusted folders, so a one-time folder-trust prompt may still appear on first use (documented).
- Pre-existing markdownlint issues elsewhere in `CLAUDE.md` were intentionally left untouched to keep the diff scoped; the new doc and the added section lint clean.

https://claude.ai/code/session_01LRmMML6ZcLJscjhfmfVrRy

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -29,5 +29,5 @@
   },
   "hooks": {},
   "includeCoAuthoredBy": true,
-  "enabledMcpjsonServers": []
+  "enabledMcpjsonServers": ["metals"]
 }
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,8 @@
+{
+  "mcpServers": {
+    "metals": {
+      "command": "metals-mcp",
+      "args": ["--workspace", "${CLAUDE_PROJECT_DIR:-.}", "--transport", "stdio"]
+    }
+  }
+}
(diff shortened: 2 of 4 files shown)


siers (approved):
👍

PR #4177, opened 2026-07-03
test: remove ExportStreamingIT benchmark sweep from CI

BalduinLandolt:
## Problem

A CI run flagged an IT test taking >4 minutes: [run 28651211149](https://github.com/dasch-swiss/dsp-api/actions/runs/28651211149/job/84969431571?pr=4174). `ExportStreamingIT` runs a `(parallelism × batchSize)` matrix sweep (25 combos + a warmup) against a Fuseki testcontainer — 26 full exports of 4024 `incunabula:page` resources — and prints per-combo timings. That's a benchmark, not a test.

```
[ExportStreamingIT] parallelism= 1  batchSize= 100   10996 ms  (4025 lines)
[ExportStreamingIT] parallelism= 1  batchSize= 250   12215 ms  (4025 lines)
...  (25 rows)
```

Its docstring claimed it was "Excluded from default CI," but nothing enforced that (no `TestAspect.ignore`, no tag filter), so it ran on every PR.

## Change

- Delete `ExportStreamingIT`.
- Drop the now-stale `ExportService` comment that referenced it.

## Why it's safe to remove

- **Tuning purpose is done.** The sweep existed to pick production defaults; those were locked in by #4132 (batch-size → 100).
- **Its only assertion was a 5-minute wall-time "regression guard"** — low-signal and flake-prone on shared CI runners (the docstring itself notes the timings are "a throughput sanity check, not validation of the idle-timeout fix").
- **Correctness is already covered:**
  - `ExportServiceSpec` (unit, against triplestore): golden CSV output, alphabetical ordering, IRI tie-breaking, cross-batch link-label resolution (`batchSize=1`), and batch-size defaulting from `AppConfig`.
  - `ExportEndpointsE2ESpec`: chunked streaming response plus the include-options.

Saves ~4 minutes per CI run.

https://claude.ai/code/session_01TftfAcyT4QJGb9eXHc6bTE


siers (approved):
😁 

I could see the value in some test that is checking that the running time is approximately correct, but yeah, 4 minutes is not it 😁

PR #4158, opened 2026-06-21
feat: add post-create resource authorship edit endpoint (DEV-6669)

danielasubotic:
Fixes DEV-6669

> **Stacked on [#4157](https://github.com/dasch-swiss/dsp-api/pull/4157) (api-2, DEV-6662)** — base branch is `feature/dev-6662-per-resource-authorship`, not `main`. Review/merge #4157 first, then rebase this onto main.

## Motivation

api-2 (#4157) added the `knora-base:hasResourceAuthorship` property with read + write-on-create. This PR (**api-2b**) adds the **post-create edit** path so a resource's authorship can be changed after creation — needed by dsp-app app-3's inline authorship edit (DEV-6667). It was split out of api-2 because it has no FileValue precedent and warranted a focused, design-first PR (spec §4.B).

## Summary

A new `Modify`-gated endpoint `PUT /v2/resources/authorship` that replaces a resource's authorship (an empty list clears it). Modeled end-to-end on the existing `updateResourceMetadataV2` chain.

## Key Changes

- **Messages** (`ResourceMessagesV2`): `UpdateResourceAuthorshipRequestV2` + `UpdateResourceAuthorshipResponseV2` (with `toJsonLDDocument` emitting `knora-api:hasResourceAuthorship` + the new `lastModificationDate`).
- **Parser** (`ApiComplexV2JsonLdRequestParser`): `updateResourceAuthorshipRequestV2` (reuses the `resourceAuthorship` accessor added in api-2).
- **SPARQL** (`ChangeResourceAuthorshipQuery`, new): mirrors `ChangeResourceMetadataQuery` — DELETE the old `hasResourceAuthorship` triples + last-mod-date, INSERT the new authorship (one triple per author) + new last-mod-date, with the same optimistic-lock `lastModificationDate` WHERE handling.
- **Responder** (`ResourcesResponderV2.updateResourceAuthorshipV2`): `IriLocker`, class check, `ensureNoConflictingChange`, `checkResourcePermission(_, Permission.ObjectAccess.Modify, _)`, run the update, verify the new last-mod-date.
- **Endpoint + wiring**: `putResourcesAuthorship` in `ResourcesEndpoints`, wired through `ResourcesServerEndpoints` and `ResourcesRestService`.

## Challenges and Decisions

[...]

PR #4157, opened 2026-06-21
feat: add per-resource authorship property with read and write-on-create (DEV-6662)

danielasubotic:
Fixes DEV-6662

## Motivation

The resource-side legal-metadata feature needs per-resource **authorship** so a data record can credit its authors independently of the project default and of file-level authorship. This is **api-2** of the roadmap (spec: dasch-specs `03-feat-legal-metadata-resource-side-plan.md` §4.B). It follows api-1 (project-level fields, dsp-api#4156) and is deployed alongside it.

## Summary

Adds a new built-in property `knora-base:hasResourceAuthorship` (a flat datatype property on the resource node) and the **read** and **write-on-create** paths for it. The heavier, no-precedent **post-create edit endpoint is split into DEV-6669 (api-2b)** so it gets a focused, design-first PR.

This split aligns with the frontend: app-3's create-form (set authorship at creation) needs only this PR; only its *inline edit* needs api-2b.

## Key Changes

- **Ontology:** `knora-base:hasResourceAuthorship` (subjectClassConstraint `:Resource`, minCardinality 0); `KnoraBaseVersion` 51→52 with a `MigrateOnlyBuiltInGraphs` plugin registered for v52 (built-in-graph reload, no data migration — also reloads api-1's knora-admin additions). Mirrored in `OntologyConstants` (KnoraBase + KnoraApiV2) and `KnoraBaseToApiV2ComplexTransformationRules` (datatype property, subjectType `knora-api:Resource`).
- **Read:** `resourceAuthorship` on `ReadResourceV2` + `toJsonLD` (emitted only when set); extracted in `ConstructResponseUtilV2`; and `GetResourcePropertiesAndValuesQuery` extended (CONSTRUCT + OPTIONAL WHERE) because it enumerates resource predicates rather than returning them all.
- **Write-on-create:** parsed from the `POST /v2/resources` JSON-LD (`ApiComplexV2JsonLdRequestParser`), carried on `CreateResourceV2`, threaded to `ResourceReadyToCreate`, and written as one triple per author in `ResourcesRepoLive` (no triples when empty).
- **Docs:** `knora-base.md` Resources section.

## Challenges and Decisions

- **Version-bump convention.** `KnoraBaseVersion` was already 51; the established pattern (see `RepositoryUpdatePlan` v51) is to bump and register a `MigrateOnlyBuiltInGraphs` plugin — not a custom data migration. Done for v52.
[...]

PR #4156, opened 2026-06-20
feat: add project-level resource-side legal metadata fields and endpoint (DEV-6661)

danielasubotic:
Fixes DEV-6661

## Motivation

DSP stores **data** (the RDF resource record) and **assets** (files) side by side. Today dsp-app only lets researchers set license / copyright holder / authorship on assets; the data record has no machine-readable legal field, so projects hand-type license prose into the project description — failing FAIR R1.1 and SNSF reporting. This is the first PR (**api-1**) of the resource-side legal-metadata feature (spec: dasch-specs `03-feat-legal-metadata-resource-side-plan.md`; roadmap DEV-6475).

## Summary

Adds three project-level, read-only-on-GET legal fields and a dedicated admin endpoint to set them. Additive and migration-free (the new admin-ontology properties are optional). Per-resource authorship — the heavier workstream — follows in **api-2** (DEV-6662).

## Key Changes

- `KnoraProject` gains `dataLicense: Option[LicenseIri]`, `dataCopyrightHolder: Option[CopyrightHolder]`, `dataAuthorship: List[Authorship]` (defaulted, so existing constructors and built-in projects are unaffected).
- knora-admin ontology: three optional properties (`hasDataLicense`, `hasDataCopyrightHolder`, `hasDataAuthorship`) on `:knoraProject`, mirrored in `OntologyConstants` and `Vocabulary`.
- Repo serialization (`KnoraProjectRepoLive`) reads/writes the new triples; the admin project GET DTO exposes them read-only with zio-json defaults (backward-compatible for older payloads — dsp-app client, dsp-tools, repository.dasch.swiss).
- New endpoint `PUT /admin/projects/{shortcode}/legal-info/resource-side` (extends the existing `ProjectsLegalInfoEndpoints` / `ProjectsLegalInfoRestService`): admin-gated via `ensureSystemAdminOrProjectAdminByShortcode`, validates the license against the six CC-BY licenses only, returns the saved values.

## Challenges and Decisions

- **No version bump / no migration.** Adding optional admin-ontology properties needs no `KnoraBaseVersion` bump — precedent: commit `4199c2aed` added `hasAllowedCopyrightHolder` the same way without touching `package.scala`. The new properties reload into existing triplestores when api-2's v52 built-in-graph reload fires; until then the feature works on the data triples directly (the repo reads/writes them without depending on the ontology graph).
[...]

PR #4182, opened 2026-07-07
docs: capture ontology, SPARQL and lenient-read conventions from the legal-metadata reviews (DEV-6719)

danielasubotic:
Part of DEV-6719 — https://linear.app/dasch/issue/DEV-6719

## Motivation

The resource-side legal metadata reviews (#4156, #4157, #4158) surfaced conventions that reviewers had to correct by hand because they were written down nowhere: the `hasDefault*` naming rule (whose absence cost a breaking rename across four repos, #4174), sort-on-read for repeated RDF literals, the WITH-graph scoping rationale (the in-code comment was initially wrong and had to be corrected in review), lenient reads of persisted data, and the built-in-ontology version-bump rules. This PR writes them down where the next contributor (human or agent) will find them.

## Summary

Docs-only. Extends the existing convention docs — no new files, no mkdocs nav changes.

## Key Changes

- `docs/development/dsp-api-conventions.md` — new **Ontology Conventions** section: `hasDefault*` naming for overridable project-wide defaults, RDF-name vs wire-key divergence, pointer to the version-bump rules.
- `docs/development/dsp-api-sparql-queries.md` — why `WITH` matters for updates (ungraphed WHERE silently no-ops without a union default graph) + the data-graph invariant, sourced from the corrected comments in `ChangeResourceAuthorshipQuery`; new **Repeated Properties Have No Order** section (sort-on-read, precedent `keywords`/`defaultDataAuthorship`).
- `docs/development/dsp-api-error-handling.md` — new **Reading persisted data leniently** section (skip + `ZIO.logWarning`, never 500 the whole read, never drop silently), with the real pattern from `ConstructResponseUtilV2`.
- `docs/05-internals/development/updating-repositories.md` — new **Changing the Built-in Ontologies** subsection: the three cases (transform plugin / `MigrateOnlyBuiltInGraphs` reload / no bump for ride-along additive props) and the rule that `knoraApiOntologyWithValueObjects.jsonld` is regenerated by `OntologyFormatsE2ESpec`, never hand-edited.
- `CONVENTIONS.md` / `REVIEW.md` — matching reference-card bullets and checklist lines (new "Ontology & RDF" blocks).

## Challenges and Decisions

- The WITH-graph text paraphrases the corrected review commits rather than re-deriving the semantics — re-derivation is how the wrong comment happened in the first place.
- Deliberately **not** re-added: the testing conventions from #4158 (already in CONVENTIONS.md/REVIEW.md) and the value-ordering learnings (already in dasch-specs).

## Gotchas

None — docs only.

## Test Plan

- `markdownlint` clean on all six files (run with the repo config).
- Claims verified against main: `RepositoryUpdatePlan.scala:38-39` (v51/v52 plugin registrations), `KnoraProjectRepoLive.scala:119/127` (sort-on-read), `ConstructResponseUtilV2` (lenient-read pattern), `OntologyFormatsE2ESpec` (fixture regeneration).


docs/development/dsp-api-conventions.md
```diff
@@ -258,6 +258,30 @@ object AuthServiceLiveSpec extends ZIOSpecDefault {
...
+### Naming overridable project-wide defaults
+
+A property that holds a **project-wide default that can be overridden per resource** is named
+`hasDefault*` in the ontology, with `default*` as the JSON payload key — following the
+`default_permissions` precedent (e.g. `knora-admin:hasDefaultDataAuthorship` / `defaultDataAuthorship`).
+A property that applies directly (not overridable) does not take the prefix (e.g. `hasDataLicense`).
+
+Get this right before merging: payload keys are cross-repo API. The initial `hasDataAuthorship`
+had to be renamed after merge, which took a breaking change across four repositories
+(dsp-api#4174, dsp-app, dsp-tools, dasch-specs).
```
BalduinLandolt: this part is not really relevant and should be removed

docs/development/dsp-api-sparql-queries.md
```diff
@@ -505,6 +518,24 @@ See the [Gravsearch API documentation](../03-endpoints/api-v2/query-language.md)
...
+Repeated datatype triples (`<s> <p> "a", "b", "c"`) are a **set** — SPARQL result order is not
+guaranteed, and an independent read may return the values in any order, even if the store happens
+to preserve insertion order most of the time.
+
+Consequences:
+
+- **Sort on read.** When a repeated property is exposed as a list, sort it by value at the
+  read boundary so round-trips are deterministic and the PUT response agrees with a later GET.
+  Precedent: `keywords` and `defaultDataAuthorship` in `KnoraProjectRepoLive.toEntity`
+  (both `.sortBy(_.value)`).
```
BalduinLandolt: this is only true for testing - in production code, if the order matters, sorting on the result set normally doesn't make sense (due to paging), and should be done on the query

CONVENTIONS.md
```diff
@@ -43,6 +43,14 @@ Scala 3, ZIO 2, Tapir, zio-json, sbt. Package root: `org.knora.webapi`. Triplest
 
 Never concatenate query strings. Use rdf4j SparqlBuilder via the helpers in `slice/common/repo`. See `docs/development/dsp-api-sparql-queries.md`.
 
+### Ontology & RDF
+
+- Overridable project-wide defaults: `hasDefault*` in the ontology, `default*` as payload key (precedent `default_permissions`, `hasDefaultDataAuthorship`). Payload keys are cross-repo API — a post-merge rename is a breaking change across four repos. Details in `docs/development/dsp-api-conventions.md` § Ontology Conventions.
```
BalduinLandolt: this bullet point is conflating two things: naming conventions and breaking changes upon renaming.  
The first can stay here, even though naming conventions should probably be their own file/paragraph eventually. The latter isn't really relevant here.

CONVENTIONS.md
```diff
@@ -43,6 +43,14 @@ Scala 3, ZIO 2, Tapir, zio-json, sbt. Package root: `org.knora.webapi`. Triplest
...
+- Repeated datatype triples are **unordered**: sort on read for deterministic round-trips (see `keywords` / `defaultDataAuthorship` in `KnoraProjectRepoLive`); never assert insertion order in tests. See `docs/development/dsp-api-sparql-queries.md` § Repeated Properties Have No Order.
```
BalduinLandolt: this is only really relevant for testing, so maybe it should be in a different section?

BalduinLandolt (approved):
Please address my comments below, then it looks good

docs/development/dsp-api-conventions.md
danielasubotic: Removed (41cda29) — the background lives in the dasch-specs cross-repo doc (dasch-swiss/dasch-specs#142).

docs/development/dsp-api-sparql-queries.md
danielasubotic: Right — reframed (41cda29): order-relevant result sets use ORDER BY in the query (post-hoc sorting doesn't compose with paging); the read-boundary sort stays only for the small entity literal lists (keywords/defaultDataAuthorship) and deterministic round-trips, with an explicit note that it's not a substitute for ORDER BY.

CONVENTIONS.md
danielasubotic: Trimmed to the naming rule + pointer (41cda29).

CONVENTIONS.md
danielasubotic: Moved to Testing Conventions and reworded per your other comment (41cda29); same move in REVIEW.md.

BalduinLandolt (approved):
👍

PR #4184, opened 2026-07-08
feat(gravsearch): add matchFulltext function

BalduinLandolt:
## What

Adds a new Gravsearch filter function `knora-api:matchFulltext(?resource, "terms")` that searches everywhere the full-text index covers for a resource — its `rdfs:label`, any text value, any value comment, or a list node (incl. sub-nodes) referenced by one of its list values. It gives the same result set as `GET /v2/search/{term}`, but expressed as a Gravsearch function so it can be combined with structured criteria in a single query.

Implements [DEV-6715](https://linear.app/dasch/issue/DEV-6715).

## How

- New AST nodes: an opaque `GroupPattern` (survives all optimizer/inference passes unchanged) and a `CoalesceFunction`.
- The `matchFulltext` handler expands the FILTER into an index-anchored block: a Lucene hit anchors the match, two `OPTIONAL`s resolve it to the containing resource (value path / list path), and a `COALESCE` + closing `BIND` bind the user's resource variable.
- Type-inference support so the resource argument is recognised as a `knora-api:Resource`.
- Constraints mirroring the semantics: must be the top-level FILTER expression, at most once per query, search-string validation mirroring the fulltext endpoint, and SPARQL-literal escaping of the search term (`"`, `\`, LF, CR).
- Deterministic inferencer variable suffixes in `SelectTransformer` so the generated SPARQL is golden-snapshot-testable.

## Testing

- Golden-snapshot prequery + count-prequery tests for the classless, class-restricted, and in-UNION expansions.
- E2E coverage asserting parity with `GET /v2/search`, plus deletion edge cases and embedded-newline escaping.
- Parser and type-inference unit tests.

## Notes

- A pre-existing SPARQL-literal escaping gap in the sibling `matchText`/`matchLabel` functions was surfaced during this work and is tracked separately in [DEV-6725](https://linear.app/dasch/issue/DEV-6725); `matchFulltext` itself is not affected.
- The last four commits are code-review follow-ups (shared `notDeletedFilter` helper, deferred counter mutation, test dedup, doc corrections); all golden snapshots are unchanged, confirming the refactors are behavior-preserving.

https://claude.ai/code/session_01NwwhWDuecVQitC11zWnosQ

### DIFF
--- a/docs/03-endpoints/api-v2/query-language.md
+++ b/docs/03-endpoints/api-v2/query-language.md
@@ -698,6 +698,35 @@ except that the first argument is a variable representing a resource:
 FILTER knora-api:matchLabel(?book, "Zeitglöcklein")
 ```
 
+### Filtering on Any Indexed Text of a Resource
+
+The function `knora-api:matchFulltext` searches for matching words anywhere the
+full-text index covers for a resource: its `rdfs:label`, any of its text values, any
+of its value comments, or a list node (including sub-nodes) referenced by one of its
+list values. It gives the same result set as the `GET /v2/search/{term}` endpoint,
+expressed as a Gravsearch function so it can be combined with structured criteria in
+one query.
+
+The first argument is a variable representing a resource (main or linked). The
+second argument is a string literal containing the search terms, with the same
+[Lucene Query Parser syntax](../../07-lucene/lucene-query-parser-syntax.md) support
+as `knora-api:matchText` and the fulltext endpoint (Lucene's default operator is a
+logical OR when submitting several search terms).
+
+This function can only be used as the top-level expression in a `FILTER`, and at
+most once per query — a second call is rejected, since evaluating it twice is
+significantly slower and no known use case needs it.
+
+```sparql
+FILTER knora-api:matchFulltext(?book, "Zeitglöcklein")
+```
+
+Deleted resources and deleted values never match. A search term containing `"` or
+`\` is safely escaped before being passed to the full-text index. As with
+`knora-api:matchText`, the search term is otherwise passed through to Lucene as-is,
+including its default result-count limit for very high-hit terms — the same
+limitation the fulltext endpoint has.
+
 ### Filtering on Resource IRIs
 
 A `FILTER` can compare a variable with another variable or IRI
--- a/docs/03-endpoints/api-v2/query-language.md
+++ b/docs/03-endpoints/api-v2/query-language.md
@@ -724,8 +724,9 @@ FILTER knora-api:matchFulltext(?book, "Zeitglöcklein")
 Deleted resources, deleted text values, and deleted list nodes never match. One
 subtlety on the list path: a resource can still match via a *deleted* list value, as
 long as the list node that value points to is itself not deleted — for the list path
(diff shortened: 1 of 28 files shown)


modules/webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/transformers/SelectTransformer.scala
```diff
@@ -27,16 +27,29 @@ class SelectTransformer(
 
   override def transformFilter(filterPattern: FilterPattern): Task[Seq[QueryPattern]] = ZIO.succeed(Seq(filterPattern))
 
+  // Gives each statement a unique suffix for the VALUES variables the inference pass may introduce
+  // (e.g. `?resTypes0`, `?resTypes1`, ...), instead of the random one `OntologyInferencer` falls back
+  // to. Determinism lets prequery golden-snapshot tests assert on the rendered SPARQL; a *constant*
+  // suffix would be wrong here, since it would give unrelated statements (e.g. `?mainRes a Resource`
+  // and `?val a Value`) the same VALUES variable, intersecting their blocks into empty results.
+  private var statementCounter = 0
```
siers, line 35: `var` for an integer and this ZIO.succeed increment business? odd

docs/03-endpoints/api-v2/query-language.md
```diff
@@ -698,6 +698,38 @@ except that the first argument is a variable representing a resource:
...
+```
+
+Deleted resources, deleted text values, and deleted list nodes never match. One
+subtlety on the list path: a resource can still match via a *deleted* list value, as
+long as the list node that value points to is itself not deleted — for the list path
+the deletion check is on the matched list node, not on the intervening list value. A
+search term containing `"`, `\`, or a newline is safely escaped before being passed to
+the full-text index. As with `knora-api:matchText`, the search term is otherwise passed
+through to Lucene as-is, including its default result-count limit for very high-hit
+terms — the same limitation the fulltext endpoint has.
```
siers, line 732: These are two paragraphs written into one.

siers (approved):

modules/webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/transformers/SelectTransformer.scala
BalduinLandolt: absolutely nasty... but copying 1:1 the patterns of other gravsearch query transformers.  
A refactor is long over due, here - but not in this PR.

docs/03-endpoints/api-v2/query-language.md
BalduinLandolt: 👍

modules/webapi/src/main/scala/org/knora/webapi/messages/util/search/gravsearch/transformers/SelectTransformer.scala
siers: Haha, gotcha

PR #4186, opened 2026-07-09
feat: include resource-side legal info in the CSV download when legal metadata is set (DEV-6721)

danielasubotic:
Fixes [DEV-6721](https://linear.app/dasch/issue/DEV-6721/include-resource-legal-info-in-the-download-when-legal-metadata-is-set)

## Motivation

When a project's resources are downloaded via the Data Browser "Download" (the v3 CSV resource export), the project's resource-side legal metadata was not included — even when set. A researcher downloading the data therefore had no record of how they may legally reuse it, which defeats the purpose of capturing the legal metadata (R5-SC-04).

## Summary

The v3 CSV export now includes the resource-side (data-side) legal metadata as columns, gated per field: as soon as any one field is set, its column appears; when none is set, the CSV is byte-for-byte unchanged. Backend-only — the export contract is unchanged, so no frontend work is needed.

## Key Changes

- `ExportService`: adds up to three columns after `Label`, each gated independently:
  - **Data License** — project-wide (`KnoraProject.dataLicense`), resolved from the built-in license catalog to its label (e.g. "CC BY 4.0"), same on every row.
  - **Copyright Holder** — project-wide (`KnoraProject.dataCopyrightHolder`), same on every row.
  - **Authorship** — per-resource (`ReadResourceV2.resourceAuthorship`), multi-value joined with ` :: `.
- `FindResourcesService.anyResourceHasAuthorship`: class-scoped `LIMIT 1` existence probe deciding whether the Authorship column is included (needed before streaming, since headers precede rows).
- Tests: positive case (all three columns + differing per-resource authorship) and a column-level "unchanged when unset" case; existing golden tests act as the byte-level none-set regression guard.

## Challenges and Decisions

- **Authorship is per-resource, not project-level.** It is written on resources with no dependency on the project's license/copyright-holder (verified in `CreateResourceV2Handler`), and a project can clear its default after resources inherited it. So the Authorship column is decided from the actual resource data (existence probe), not from the project object.
- **License label without new layer wiring.** `LicenseRepo` is consumed inside `AdminModule` and not exposed at the top-level layer, so injecting it would ripple across modules. Licenses are a fixed built-in catalog (the admin API only enables/disables built-ins), so the label is resolved from the static `License.BUILT_IN` — no new dependency.
- **Per-field, not all-or-nothing.** Columns are added independently (matching the export's existing conditional columns like ARK URL / link IRIs), so setting only a license shows only the Data License column.

## Gotchas

- Column headers are hardcoded English, consistent with the existing fixed columns (`Resource IRI`, `Label`, `ARK URL`).
- The OAI/JSON export and the v2 project metadata export are intentionally out of scope.

## Test Plan

- `sbt "webapi/testOnly *ExportServiceSpec"` — 23 pass (incl. golden none-set guard + the two new tests).
- `sbt fmt` + `sbt "scalafixAll --check"` clean.
- Manual: on a local stack, set resource-side legal via `PUT /admin/projects/shortcode/{shortcode}/legal-info/resource`, then `POST /v3/export/resources` (`Accept: text/csv`) and confirm the expected column(s) appear; unset → CSV unchanged.


modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
```diff
@@ -78,6 +79,36 @@ final case class ExportService(
   private val footnoteTagIri: SmartIri         = OntologyConstants.Standoff.StandoffFootnoteTag.toSmartIri
   private val footnoteContentPropIri: SmartIri = OntologyConstants.Standoff.StandoffFootnoteTagHasContent.toSmartIri
 
+  // Resource-side (data-side) legal columns for the CSV export. Each field is included independently — as soon
+  // as one is set for the project/resources, its column appears; when none is set, nothing is added and the CSV
+  // is byte-for-byte unchanged. License + copyright holder are project-wide (same on every row); authorship is
+  // per-resource. Header/cell order is kept identical here so the streamed header and rows stay in lockstep.
```
BalduinLandolt: This comment doesn't add value.

And I would argue that this should always be included in the CSV, even if there aren't any values.

modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
```diff
@@ -78,6 +79,36 @@ final case class ExportService(
...
+        (if includeAuthorship then List(AuthorshipHeader) else Nil)
+
+    def cells(resource: ReadResourceV2): ListMap[String, String] =
+      (if includeLicense then ListMap(DataLicenseHeader -> dataLicenseLabel.getOrElse(""))
+       else ListMap.empty[String, String]) ++
+        (if includeCopyright then ListMap(CopyrightHolderHeader -> dataCopyrightHolder.getOrElse(""))
+         else ListMap.empty[String, String]) ++
+        (if includeAuthorship then
+           ListMap(AuthorshipHeader -> resource.resourceAuthorship.map(_.value).mkString(ValueSep))
+         else ListMap.empty[String, String])
```
BalduinLandolt: this could be massively simplified if we always include these columns

modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/FindResourcesService.scala
```diff
@@ -135,6 +144,30 @@ final case class FindResourcesServiceLive(
...
+        s"""PREFIX knora-base: <${KB.NS.getName}>
+           |SELECT ?$resourceIriVar WHERE {
+           |  GRAPH <$projectGraph> {
+           |    ?$resourceIriVar a ?$classIriVar ;
+           |                     knora-base:hasResourceAuthorship ?authorship .
+           |  }
+           |  VALUES ?$classIriVar { $valuesClause }
+           |} LIMIT 1""".stripMargin,
+      )
+    }
```
BalduinLandolt: having to query the DB just to decide which headers to include seems like a bad idea

BalduinLandolt (changes requested):
I don't think it's a good idea to leave out columns that are empty. that's inconsistent, potentially confusing to users, hides the fact that there is important data missing, and it adds a lot of complexity to the implementation (and makes it slower, because a DB query is needed to determine which headers to include).

I would propose we change that design, which will make things much more straight forward.  
When that's done, I'm happy to review again.

danielasubotic:
Reworked per your feedback: the three columns (Data License, Copyright Holder, Authorship) are now always included, blank when there are no values. This drops the per-field gating and removes the `anyResourceHasAuthorship` DB probe entirely — `FindResourcesService` is back to its previous shape. Export goldens (unit + e2e) regenerated to include the columns (blank for the test projects, which have no resource-side legal). Ready for another look.

modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
```diff
@@ -78,6 +79,12 @@ final case class ExportService(
   private val footnoteTagIri: SmartIri         = OntologyConstants.Standoff.StandoffFootnoteTag.toSmartIri
   private val footnoteContentPropIri: SmartIri = OntologyConstants.Standoff.StandoffFootnoteTagHasContent.toSmartIri
 
+  // Resource-side (data-side) legal columns, always present in the export: project-wide data license and
+  // copyright holder (same on every row, blank when the project has none) plus per-resource authorship.
+  private val DataLicenseHeader     = "Data License"
+  private val CopyrightHolderHeader = "Copyright Holder"
+  private val AuthorshipHeader      = "Authorship"
```
BalduinLandolt, line 90: nitpick: the fact that these new ones are defines as constants here, but the pre-existing headers are not here, indicates that this is solved inconsistently.

modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
```diff
@@ -171,10 +178,21 @@ final case class ExportService(
...
     )
 
+    // The data license and copyright holder are project-wide, so they are resolved once and repeated on every
+    // row (blank when the project has none). The license IRI is resolved to its human label from the built-in
+    // catalog. Per-resource authorship is read off each resource in `exportSingleRow`.
+    val dataLicenseLabel: String =
+      project.dataLicense
+        .map(iri => License.BUILT_IN.find(_.id == iri).map(_.labelEn).getOrElse(iri.value))
+        .getOrElse("")
+    val dataCopyrightHolder: String = project.dataCopyrightHolder.map(_.value).getOrElse("")
```
BalduinLandolt: `opt.map(f).getOrElse(a)` could be written as `opt.fold(a)(f)` - this could be added to the repo style conventions as preferable

BalduinLandolt (approved):
much better! just two nitpicks remarked

modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
danielasubotic: Addressed in the rework: the legal columns are now always present (blank when the project has none), and that comment is gone.

modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
danielasubotic: Addressed in the rework: the columns are always included now, so the per-field gating is gone.

modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/FindResourcesService.scala
danielasubotic: Addressed: the `anyResourceHasAuthorship` DB probe was removed; `FindResourcesService` is back to its previous shape.

modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
danielasubotic: Aligned upward in d40e0f8: `ARK URL` / `Resource IRI` / `Label` are constants now too, so every header name is single-sourced.

modules/webapi/src/main/scala/org/knora/webapi/slice/export/api/service/ExportService.scala
danielasubotic: Switched to `fold` in d40e0f8 (both the license label and the copyright holder).

PR #4187, opened 2026-07-09
build: bazelify the build with rules_scala + rules_oci

subotic:
## What & why

Migrate the whole build to Bazel. All Scala compiles with `rules_scala`, tests run under `bazel test`, and all four container images (`daschswiss/knora-sipi`, `daschswiss/knora-api`, `daschswiss/dsp-ingest`, `daschswiss/apache-jena-fuseki`) build with `rules_oci`. CI runs entirely through `just` (which wraps Bazel) against a shared **NativeLink** remote build backend for a remote cache + executor. `just` is the single build/test/publish entry point; the Makefile has been removed (all targets ported to the justfile).

**sbt stays fully functional in parallel** for a validation period — it remains the Scala build of record and is not retired here. Removing the sbt-parallel paths (and `./sbtx`) will be a separate PR a few weeks after this lands.

## What's included

- **Scala 3.3.7 → 3.8.4** migration (both sbt and Bazel).
- **Bazel foundation:** `rules_scala` 7.2.6 (from BCR), a single pinned `@maven` repo (`maven.install` + committed `maven_install.json`), a custom Scala toolchain, a hermetic LLVM C toolchain + JDK 25, workspace-status stamping, and a `buildinfo` rule replacing sbt-buildinfo.
- **All modules compile and test under Bazel**, byte-matching sbt's pass counts. A single custom `DspZTestJUnitRunner` backs both sbt and Bazel test runs (avoids zio-test's double-`bootstrap` and keeps container startups to ~1 per JVM).
- **All four images build with `rules_oci`** (sipi/api/ingest/fuseki), each verified live: the stack comes up healthy, distributed traces flow, and Fuseki's TDB2 store persists across a container restart.
- **Driver layer on Bazel:** the `justfile`, `docker-compose.yml` healthchecks, and CI all call Bazel.
- **Remote build execution:** CI points Bazel at the shared NativeLink backend for a remote cache + executor. See `docs/development/dsp-api-rbe.md`.

## Deliberately retained during the validation window

- sbt for the formatting check (`just check`) and a temporary tag-drift gate (`just check-docker-image-tag`), both removable once validation closes.
- Fuseki's *publish* path stays on the Dockerfile / `docker buildx`; its Bazel image is still exercised every CI run as a test preload and guarded by `//tools/oci:image_versions_match_sbt`.


SESSION-HANDOFF.md
BalduinLandolt, line 1: Is it necessary to check in this file here? Could it live somewhere else?

BalduinLandolt (approved):
I could't really "review" this PR, but looked over if - seems sane at least.

If possible, I'd prefer not to have the session handoff document checked in when merging.

Also, the PR description suggests that this is work in progress; since the PR description will become the commit message on main, I would frame that differently.

SESSION-HANDOFF.md
subotic: no, will remove before merge.

PR #4190, opened 2026-07-11
feat: eliminate redundant asset compression in project-migration export (DEV-6592)

BalduinLandolt:
[DEV-6592](https://linear.app/dasch/issue/DEV-6592)

## Summary

- The shared BagIt writer (`modules/bagit`) always DEFLATE-compressed every entry, so the project-migration export compressed incompressible media **twice**: once when ingest bags assets into `assets.zip`, and again when webapi wraps that already-compressed zip in the outer bag (the "double-deflate").
- This adds a per-entry `Compression { Deflate, Store }` knob and uses it: assets are stored in **both** passes, while the compressible `rdf/` N-Quads stay DEFLATEd.
- `Store` is implemented as **DEFLATE level 0** (`Deflater.NO_COMPRESSION`), **not** true `STORED` — it drops into the existing single-pass streaming writer with no CRC pre-pass. The ZIP entry method stays `DEFLATED` and the level is not recorded in the archive, so old and new bags remain interchangeable in **both** directions and the read/import/validation path is untouched.

## Changes

- **`modules/bagit`**: new `Compression` enum (with mandatory scaladoc on `Store`); required, no-default `compression` field on `PayloadEntry.File`/`.Directory`; `BagCreator` threads it through `collectFiles`/`writePayloadEntries`, sets the deflate level per entry before `putNextEntry`, and resets to the default level before the tag files.
- **ingest**: `ProjectService.exportProjectAsBagIt` stores assets (`Compression.Store`).
- **webapi**: `ProjectMigrationExportService.createBagItZip` stores `assets/assets.zip` and keeps `rdf/` on `Deflate` — killing the double-deflate.
- All 21 `PayloadEntry` construction sites updated to pass `Compression` explicitly (compile-enforced by the no-default field).
- **Docs**: note the compression policy in the migration export-format section.

## Test Plan

Built test-first (red -> green per behavior):

- `BagItSpec` — a `Store` entry is written uncompressed while a `Deflate` entry shrinks (1 MB-of-zeros probe, entry sizes read via `java.util.zip.ZipFile`); a nested `Store`-in-`Store` bag round-trips byte-for-byte and validates; a backward-compat all-`Deflate` bag validates and extracts.
- `ProjectMigrationExportServiceSpec` — the outer migration bag stores `data/assets/assets.zip` and compresses `data/rdf/admin.nq` (enlarged the stubbed `assets.zip` and injected a sizeable compressible RDF payload so the assertions are not vacuous).
- `ImportServiceSpec` (ingest) — the real export -> import SHA-512 round-trip now exercises `Store`.
- `sbt check` clean (scalafmt + scalafix + headers); `bagit` 128 / `ingest` 28 / `webapi` export+import 48 tests green.
- Adversarial review (scala-zio, consistency, simplicity, performance): no critical/warning findings.

**Pending (separate step):** the `main`-vs-branch before/after A/B on the Alice in DaSCHland fixture (REQ-3.1/3.2 — packing throughput / CPU / bag size), to be run end-to-end on a local stack and recorded here.

https://claude.ai/code/session_018ZqH2kNsewQK9L4q6UvLrY

### DIFF
--- a/docs/03-endpoints/api-v3/project-migration.md
+++ b/docs/03-endpoints/api-v3/project-migration.md
@@ -55,6 +55,10 @@ data/
 
 All RDF data is serialized as N-Quads and uses the internal knora-base schema.
 
+Within the archive, the `data/rdf/` N-Quads are DEFLATE-compressed, while `data/assets/assets.zip` is stored
+without effective compression: its contents are already-compressed media, so re-compressing them only wastes
+CPU. This is why `assets.zip` does not shrink relative to its source.
+
 The `bag-info.txt` file includes metadata fields:
 
 | Field                  | Description                                     |
--- a/modules/bagit/src/main/scala/org/knora/bagit/domain/PayloadEntry.scala
+++ b/modules/bagit/src/main/scala/org/knora/bagit/domain/PayloadEntry.scala
@@ -7,7 +7,21 @@ package org.knora.bagit.domain
 
 import zio.nio.file.Path
 
+enum Compression {
+
+  /** DEFLATE at the default level (~6). */
+  case Deflate
+
+  /**
+   * No effective compression: written as DEFLATE level 0 (`Deflater.NO_COMPRESSION`), NOT the ZIP `STORED`
+   * method — the entry method stays `DEFLATED`. Chosen over true STORED to keep the single-pass streaming
+   * writer: STORED requires the CRC-32 and size up front, which would force a second read pass over every
+   * file. The observable effect is compressed size ~= uncompressed size, not a `STORED` method flag.
+   */
+  case Store
+}
+
 enum PayloadEntry {
-  case File(relativePath: String, sourcePath: Path)
-  case Directory(prefix: String, sourcePath: Path)
+  case File(relativePath: String, sourcePath: Path, compression: Compression)
+  case Directory(prefix: String, sourcePath: Path, compression: Compression)
 }
--- a/modules/bagit/src/main/scala/org/knora/bagit/internal/BagCreator.scala
+++ b/modules/bagit/src/main/scala/org/knora/bagit/internal/BagCreator.scala
@@ -13,6 +13,7 @@ import java.io.File
 import java.io.IOException
 import java.security.MessageDigest
 import java.time.LocalDate
+import java.util.zip.Deflater
 import java.util.zip.ZipEntry
 import java.util.zip.ZipOutputStream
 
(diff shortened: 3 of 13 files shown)


BalduinLandolt:
## Verification — before/after A/B (Alice in DaSCHland), REQ-3.1 / REQ-3.2

Real end-to-end migration export, `main` vs this branch, against the **same** locally-ingested Alice fixture (project `0854`, **971 MB / 474 files**: 158 originals, 103 JP2, 24 MP3, 1 MP4, 9 PDF, 18 CSV, 158 `.info` sidecars). Both the `dsp-api` **and** `dsp-ingest` images were swapped between runs (Pass #1 lives in ingest, Pass #2 in webapi); Fuseki, the ingested assets and Sipi were held constant. One run per side, local machine (fast NVMe). CPU is the cumulative container cgroup `usage_usec` delta bracketing each export.

| Metric | `main` (baseline) | this branch | Δ |
|---|--:|--:|--:|
| Wall-clock | 195.4 s | 160.8 s | **−34.6 s (−17.7%)** |
| ingest CPU (Pass #1) | 56.7 s | 36.8 s | **−19.9 s (−35.1%)** |
| dsp-api CPU (Pass #2, the double-deflate) | 58.3 s | 44.7 s | **−13.6 s (−23.3%)** |
| **total CPU** | 114.9 s | 81.5 s | **−33.4 s (−29.1%)** |
| Bag size | 992,693,859 B | 1,018,085,053 B | +25.4 MB (**+2.56%**, within the ≤3% REQ-3.2 bound) |
| Asset identity | — | — | **478/478 inner files byte-for-byte identical** |

Per-pass wall-clock (from `BagProgress` logs): Pass #1 ingest bagging ~88 s → ~68 s; Pass #2 outer bag ~82 s → ~67 s.

### Policy verified in the produced bags

- **Inner `assets.zip` (Pass #1):** on `main` the compressible originals shrink (`.info` 0.54, `.csv` 0.39, `.xml` 0.18, `.pdf` 0.93, `.orig` 0.90); on the branch **everything is stored** (all ratios ≈ 1.00, the tiny > 1.0 being the DEFLATE-0 wrapper overhead). Incompressible media (`.jpx` 0.996, `.mp3`/`.mp4` ≈ 0.98) was never compressible either way.
- **Outer bag (Pass #2):** `data/assets/assets.zip` compresses to ratio **1.000 on both** — i.e. `main` burns CPU DEFLATE-ing an already-compressed ~991 MB blob for **zero** size gain (the double-deflate), while the branch stores it. `data/rdf/*.nq` stays DEFLATE-compressed on both (0.05–0.12), so REQ-1.3 holds.
- **Identity:** all 478 inner asset files are byte-for-byte identical between `main` and the branch, and the RDF payloads match.

### Interpretation

- **Compression CPU waste is eliminated** (total CPU −29%, both passes), leaving only the unavoidable SHA-512 hashing of ~1 GB.
- **Wall-clock improves less than CPU (−18%)** because the export is partly I/O-bound — the two-pass 1 GB byte-movement + hashing is the floor. This is the anticipated *"byte movement is the floor → prioritise the I/O-reduction phase next"* finding, now measured rather than assumed. A flat-ish wall-clock is a finding, not a failure.
- **Size (+2.56%)** comes entirely from storing Alice's compressible bits (`.info` sidecars, PDFs, CSVs). Alice is a near-worst case for this bound; for the media-dominated projects this targets (e.g. 0810 = 14.3 GB video), the compressible fraction is negligible and the delta trends toward the DEFLATE-0 wrapper overhead (~0.03%).
- **Scaling:** the CPU win scales ~linearly with asset volume (the double-deflate in particular), so asset-heavy projects save substantially more CPU; how much surfaces as wall-clock vs freed CPU depends on whether that environment is CPU- or I/O-bound.

subotic (approved):

PR #4193, opened 2026-07-17
feat(dsp-api): implement Region Preview values (DEV-6758)

Nora-Olivia-Ammann:
[DEV-6758](https://linear.app/dasch/issue/DEV-6758)

Implements the **dsp-api + dsp-docs** side of Region Preview Links — the deploy-gate repo that must reach dev / a release before the App (`06`) and dsp-tools (`07`) work can go green. Executes the implementation plan `05-feat-region-preview-links-api-implementation-plan.md` (dasch-specs), completing the merged v1 `RegionPreviewValue` shape as a distinct leaf value type.

## Motivation

`RegionPreviewValue` lets a resource reference a `Region` directly (without link-value reification) so the App can render a preview of a region on an image. The v1 shape was incomplete: a dangling `salsah-gui:RegionPreview` reference, an unenforced target contract, no delete-protection, and a single computed `iiifUrl`. This PR finishes it.

## Changes

- **Ontology & delivery** — define `salsah-gui:RegionPreview` (ttl + Scala allowlist), drop the v1 `:iiifUrl` term, bump `KnoraBaseVersion` 52→53 with a `MigrateOnlyBuiltInGraphs` migration.
- **Value model** — reshape `RegionPreviewValueContentV2` from `iiifUrl`-only to the computed read surface: crop URL + highlight box (rectangle geometry only), full-page thumbnail (`full/,512`), bounded full-image identity (`{@id,@type,rdfs:label}`), and the still image's legal metadata. crop/thumbnail render as `xsd:anyURI` typed literals. Fix the `from`-builder comment mis-binding.
- **Referential integrity** — target must exist, be a `Region`, and be same-project, on both write paths (the load-bearing `ValuesResponderV2.checkPropertyObjectClassConstraint` fix plus the create-resource path).
- **Delete-protection** — a `Region` referenced by a live preview cannot be deleted/erased (`isResourceInUse` UNION branch keyed on the value node's `isRegionPreviewOf`, with a load-bearing `isDeleted false` filter that releases the region once the preview is deleted).
- **Read surface** — computed post-read via a bounded `SystemUser` read; image resolution is independent of geometry (thumbnail/identity/legal for any geometry, crop/highlight for rectangles only).
- **Served ontology** — 7 `hasPreview*` runtime properties + class cardinalities declared in the transformation rules; generated ontology fixtures regenerated.
- **v3 CSV export** — crop cell + region IRI (via a new `isIriBearingProp`).
- **dsp-docs** — `RegionPreviewValue`, the gui element, the create/update payload, and the read surface (incl. the two-tier permission model).

## Challenges and Decisions

- **Referential integrity is on two paths.** The primary add-value/update-value path routes through `checkPropertyObjectClassConstraint`, which previously only checked the value's own type. A new branch there enforces existence + is-a-Region (keyed on `knora-base:Region`, not the property's `RegionPreviewValue` constraint) **and** the value-type-vs-property constraint.
- **Bounded elevated read (deliberate permission departure).** Once the value's own permission gate passes, identity + legal metadata are returned via a `SystemUser` read regardless of the user's permissions on the region/image; image pixels stay enforced by Sipi. Only a fixed field set is projected — never the full still-image resource.
- **`iiifUrl` re-homes to knora-api runtime properties** rather than living in `knora-base`, matching `fileValueAsUrl`/`stillImageFileValueHasDimX`.
- **Delete-protection query grouping.** The two `isResourceInUse` branches are grouped as `{ {..} UNION {..} }` inside one `GRAPH` block (a first attempt emitted `GRAPH{} UNION GRAPH{}`, which Fuseki rejects with 400 — caught by the integration run and fixed).

## Test Plan

- `sbt webapi/compile`, `sbt check` — clean.
- Unit (`sbt webapi/test ...`): `ReadResourcesServiceLiveSpec` (read surface vs real data), `ApiComplexV2JsonLdRequestParserSpec` (comment-bearing parse), `ValueContentV2Spec` (rectangle-gate `parseShape`).
- Integration (`sbt test-it/test ...`, Docker): `ValuesResponderV2Spec` (referential integrity, add + update paths) and `ResourcesResponderV2Spec` (delete-protection + release) — 169 passing.
- e2e (`sbt test-e2e/test ...`, Docker): `OntologyFormatsE2ESpec` passing; served-ontology fixtures regenerated and committed.

## Gotchas / follow-ups (tracked on DEV-6758)

- **Legal-property `subjectType` (resolved — deliberate reuse):** the reused `hasCopyrightHolder`/`hasAuthorship`/`hasLicense` are declared `subjectType = FileValue` but added as cardinalities on `RegionPreviewValue`. This is intentional: `InstanceChecker` validates instances by class **cardinalities** + property **`objectType`**, never `subjectType` (see `InstanceChecker.checkRec`), so a region preview whose image carries legal metadata validates fine — `RegionPreviewValue` declares the cardinalities and the objectTypes match what the read path emits. Keeping `subjectType = FileValue` is the accurate domain (a region preview *projects* the image's legal metadata; it doesn't own copyright), and reusing the exact FileValue predicates is what `06`/`07` rely on. Broadening to `:Value` was considered and rejected as a globally-looser, less-truthful change that fixes only a cosmetic inconsistency. See the analysis comment below.
- Deferred tests (remaining): restricted-image read, `ExportService` region-preview case, `InsertValueQueryBuilder` golden. Kept out to avoid destabilizing the shared `anything-data.ttl`. *(The non-rectangle-geometry read path and the create-resource-inline referential integrity are now covered — see below.)*

## Cross-repo

First of three (the deploy gate). This branch is feature-complete for local verification by `06`/`07` against a local build; downstream CI stays red-by-design until this deploys to dev / releases. R12 (`objectType = knora-api:RegionPreviewValue`) is confirmed by the regenerated `anythingOntologyWithValueObjects` fixture. Thumbnail size is baked as `full/,512` (R13) — verify in the App during the pre-merge local run.

## Screenshots

[Attach manually if UI changes] — N/A (API + docs only).

---

## Added beyond the PRD — region colour (`hasPreviewColor`)

> **Note:** This was implemented although it is **not part of the original PRD / plan `05`.** The API now also serves the region's colour (`knora-base:hasColor`) as **`knora-api:hasPreviewColor`** on each region-preview value, so the App can tint the highlight box to match the region's chosen colour. It is computed on read like the sibling preview fields — a bare string, geometry-independent (present for any geometry), declared `ZeroOrOne` on `RegionPreviewValue`. The generated ontology fixture was regenerated and the docs updated. Covered by `ReadResourcesServiceLiveSpec` at the content level (rectangle + non-rectangle) and in the JSON-LD body (precise `hasPreviewColor` → value assertion). Commit `fbd50b131`.


Nora-Olivia-Ammann:
### Note: legal-property `subjectType` on `RegionPreviewValue` is intentional

A review flagged that the reused legal predicates `hasCopyrightHolder` / `hasAuthorship` / `hasLicense` are declared with `subjectType = knora-api:FileValue` but are added as cardinalities on `knora-api:RegionPreviewValue` (which is `⊑ :Value`, not a `FileValue`). Recording why this is fine and left as-is:

- **`InstanceChecker` never checks `subjectType`.** `InstanceChecker.checkRec` validates each instance property against (1) the class's **cardinalities**, (2) the cardinality **count**, and (3) the property's **`objectType`**. There is no `subjectType` check. Since `RegionPreviewValue` declares the cardinalities for these three properties, and their `objectType` (string / string-array / `License` IRI) matches exactly what the read path emits, a region preview whose still image carries legal metadata validates against the served ontology.
- **`subjectType = FileValue` is the accurate domain.** Copyright / authorship / license belong to the file asset; a `RegionPreviewValue` merely *projects* the underlying still image's legal metadata onto the read response. Reusing the exact FileValue predicates (same keys) is deliberate so the App and dsp-tools handle legal metadata uniformly across file values and region previews.
- **Broadening `subjectType` to `:Value` was considered and rejected.** It's a one-line change but it would loosen the domain of the *shared* FileValue properties globally (implying any `Value` — `IntValue`, `TextValue`, … — can have a copyright holder), trading an accurate domain for a vague one to fix a purely cosmetic inconsistency that has no runtime effect.

Net: no code change; the served ontology (`OntologyFormatsE2ESpec`) and the read rendering are both verified. The earlier "instance-level `InstanceChecker`" follow-up is therefore closed as settled.

Nora-Olivia-Ammann:
Added the two highest-impact follow-up tests (commit 021ff6bbe):

- **Create-resource-inline referential integrity** (integration): creating a resource with an inline region preview whose target is not a Region is rejected; a valid Region target succeeds. This covers the create-resource path (4.1/4.2) — previously only the add/update-value path was tested.
- **Non-rectangle read gate** (unit): a preview whose region has polygon geometry emits thumbnail + full-image identity + legal but no crop URL / highlight box. Uses a self-contained fixture loaded only by `ReadResourcesServiceLiveSpec` (reuses `a-thing-picture`; the shared `anything-data.ttl` is untouched).

Remaining deferred tests: restricted-image read, `ExportService` region-preview case, `InsertValueQueryBuilder` golden — all lower-risk (`isIriBearingProp` equals `isLinkValueProp` for non-region properties, already exercised by the export golden tests).

Nora-Olivia-Ammann:
I tried it out with the APP and TOOLS drafts and that is where I noticed that the colour should also be consistent with the Region display, it felt really odd. So far everything seems to work.

docs/03-endpoints/api-v2/reading-and-searching-resources.md
```diff
@@ -44,6 +44,28 @@ See the interfaces `Resource` and `ResourcesSequence` in module
...
+`knora-api:isRegionPreviewOf` target. On read, the API augments it with fields computed from the referenced region's
+geometry and the still image it is part of:
+
+- `knora-api:hasPreviewCropUrl` — a IIIF URL of the cropped region, and
+  `knora-api:hasPreviewHighlightBoxX/Y/W/H` — the region's bounding box as percentages (0–100). These are present
+  **only when the region's geometry is a rectangle**; for other geometries they are omitted.
+- `knora-api:hasPreviewThumbnailUrl` — a IIIF URL of the full page at a fixed height, present for any geometry.
+- `knora-api:hasPreviewColor` — the region's colour (its `knora-base:hasColor`), present for any geometry.
+- `knora-api:hasPreviewFullImage` — a bounded reference (`@id`, `@type`, `rdfs:label`) to the still image resource.
+- `knora-api:hasCopyrightHolder`, `knora-api:hasAuthorship`, `knora-api:hasLicense` — the still image's legal
```
BalduinLandolt: I'm not convinced by the naming of these properties:
- we're in the context of a preview value, so I'm not sure all of those need the "preview" prefix (also, not all of them have it, so it's inconsistent).
- not sure we need the "crop" part, as this is essentially the heart of the preview value; but here I would keep the "preview"
- re-using `hasCopyrightHolder` etc. seems problematic to me, because normally these properties make statement about the copyright of the subject of the triple. Here, in contrast, they make a statement about the copyright of a linked resource's file value.

Here's my naming suggestion:

```suggestion
- `knora-api:hasPreviewUrl` — a IIIF URL of the cropped region, and
  `knora-api:hasHighlightBoxX/Y/W/H` — the region's bounding box as percentages (0–100). These are present
  **only when the region's geometry is a rectangle**; for other geometries they are omitted.
- `knora-api:hasThumbnailUrl` — a IIIF URL of the full page at a fixed height, present for any geometry.
- `knora-api:hasColor` — the region's colour (its `knora-base:hasColor`), present for any geometry.
- `knora-api:hasFullImage` — a bounded reference (`@id`, `@type`, `rdfs:label`) to the still image resource.
- `knora-api:hasFullImageCopyrightHolder`, `knora-api:hasFullImageAuthorship`, `knora-api:hasFullImageLicense` — the still image's legal
```

But that's of course up for debate.

modules/test-it/src/test/scala/org/knora/webapi/responders/v2/ValuesResponderV2Spec.scala
```diff
@@ -1210,25 +1210,44 @@ object ValuesResponderV2Spec extends E2EZSpec { self =>
...
+        // aThing is a Thing, not a Region.
+        valuesResponder(_.createValueV2(createPreviewParams(aThingIri), anythingUser1, randomUUID)).exit
+          .map(err => assert(err)(failsWithA[OntologyConstraintException]))
+      },
+      test("reject creating a region preview whose target is in another project (REQ-1.3)") {
+        // zeitgloecklein is in the incunabula project (0803).
+        valuesResponder(_.createValueV2(createPreviewParams(zeitgloeckleinIri), anythingUser1, randomUUID)).exit
+          .map(err => assert(err)(failsWithA[BadRequestException]))
+      },
+      test("create a region preview value (REQ-1.1)") {
```
BalduinLandolt: I'm not sure we should add references to acceptance criteria to test names (but admittedly it doesn't hurt)

modules/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/ontologymessages/KnoraBaseToApiV2ComplexTransformationRules.scala
```diff
@@ -418,6 +418,72 @@ object KnoraBaseToApiV2ComplexTransformationRules extends OntologyTransformation
       .withRdfLabelEn("External Url to the image")
       .withRdfCommentEn("External Url to the image")
 
+  // Runtime (computed-on-read) fields of a region preview value. They are declared here as knora-api-only
+  // properties (never stored in knora-base), so a read instance validates against the served ontology.
+  private val HasPreviewCropUrl = makeOwlDatatypeProperty(KA.HasPreviewCropUrl, XSD.ANYURI)
+    .withSubPropertyOf(KA.ValueHas)
+    .withSubjectType(KA.RegionPreviewValue)
+    .withRdfLabelEn("has region preview crop URL")
+    .withRdfCommentEn("IIIF URL of the cropped region image (computed on read; rectangle geometry only).")
```
BalduinLandolt: On that note, please see https://linear.app/dasch/issue/DEV-6754/region-geometry-decide-how-to-handle-non-rectangular-shapes-circles#comment-523f148f

modules/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
```diff
@@ -1460,13 +1462,21 @@ object GeomValueContentV2 {
   } yield GeomValueContentV2(ApiV2Complex, geom, comment)
 
   final case class Point(x: Double, y: Double)
-  private final case class GeomShape(points: List[Point])
+  // Public so the read-surface augmentation can gate crop/highlight on `geomType`. The JSON key `type` is
+  // reserved, hence the `@jsonField` rename; zio-json ignores the other geometry keys we do not model here.
+  final case class GeomShape(
+    @jsonField("type") geomType: Option[String],
+    points: List[Point],
+  )
```
BalduinLandolt: I don't understand this

modules/webapi/src/main/scala/org/knora/webapi/responders/v2/resources/CheckObjectClassConstraints.scala
```diff
@@ -79,6 +79,29 @@ object CheckObjectClassConstraints {
...
+                  propertyRequiresValue(resourceIdForErrorMsg, propertyIri, objectConstraints)
+                }
+              regionClass     = Constants.Region.toSmartIri
+              targetClass     = linkTargetClasses(rpvc.regionIri.value)
+              targetClassInfo = entityInfo.classInfoMap(targetClass)
+              _              <-
+                ZIO.when(!targetClassInfo.allBaseClasses.contains(regionClass)) {
+                  ZIO.fail(
+                    OntologyConstraintException(
+                      s"${resourceIdForErrorMsg}Resource <${rpvc.regionIri.value}> cannot be referenced by property <${propertyIri.toComplexSchema}>, because it is not a member of class <${regionClass.toComplexSchema}>",
```
BalduinLandolt: Not sure if that error message is optimal

BalduinLandolt (approved):
All in all looks good. See my comments; and the things we discussed.

docs/03-endpoints/api-v2/reading-and-searching-resources.md
Nora-Olivia-Ammann: They look good to me, claude change it.

modules/test-it/src/test/scala/org/knora/webapi/responders/v2/ValuesResponderV2Spec.scala
Nora-Olivia-Ammann: Good point, claude change it.

modules/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/ontologymessages/KnoraBaseToApiV2ComplexTransformationRules.scala
Nora-Olivia-Ammann: Good, claude will investigate it separately.

modules/webapi/src/main/scala/org/knora/webapi/messages/v2/responder/valuemessages/ValueMessagesV2.scala
Nora-Olivia-Ammann: Me neither, claude should explain it.

modules/webapi/src/main/scala/org/knora/webapi/responders/v2/resources/CheckObjectClassConstraints.scala
Nora-Olivia-Ammann: Thanks, claude should improve it.

mpro7:
@Nora-Olivia-Ammann just noticed browsing the pull requests, the scope in the PR title doesn't make sense `feat(dsp-api)` - subject to change or remove.

Nora-Olivia-Ammann:
## Summary of changes since the last review

Three commits on top of the reviewed state: the review-feedback fixes, plus two follow-ups (region label, and a bounding-box crop for non-rectangular regions). Overview for reviewers below — 11 files, +526/−236.

### ⚠️ API contract change — please note

On a **complex-schema read**, a `RegionPreviewValue` now **always** carries `hasPreviewUrl`, `hasThumbnailUrl`, the four `hasHighlightBox{X,Y,W,H}`, `hasPreviewColor`, and `hasFullImage`. Their ontology cardinality was tightened from `0-1` to **`owl:cardinality 1`**, so consumers can rely on them being present. This is justified because `knora-base:Region` requires the underlying properties (`hasGeometry` min 1, `isRegionOf`/`isRegionOfValue` 1, `hasColor` 1). Only the still-image legal metadata stays optional. This affects the **read** contract only — create/update is unchanged (these are read-only computed projections).

---

### 1. Review-feedback fixes (`18952383e`)

**Property renaming** (code + docs + regenerated ontology):

| Before | After |
|---|---|
| `hasPreviewCropUrl` | `hasPreviewUrl` |
| `hasPreviewThumbnailUrl` | `hasThumbnailUrl` |
| `hasPreviewHighlightBox{X,Y,W,H}` | `hasHighlightBox{X,Y,W,H}` |
| `hasPreviewFullImage` | `hasFullImage` |
| `hasPreviewColor` | *(kept — `hasColor` already denotes a Region's colour)* |
| reused `hasCopyrightHolder` / `hasAuthorship` / `hasLicense` | dedicated `hasFullImageCopyrightHolder` / `hasFullImageAuthorship` / `hasFullImageLicense` |

- The dedicated legal properties (subject `RegionPreviewValue`) resolve the concern that reusing `FileValue`'s predicates made a statement about the wrong subject — they now describe the *linked still image*, not the value.
- Clearer `OntologyConstraintException` message when a region-preview target is not a `Region`.
- Expanded the `GeomShape` comment (the "I don't understand this" thread) to explain the `valueHasGeometry` JSON model and the `@jsonField("type")` rename.
- Dropped the `(REQ-1.x)` acceptance-criteria tags from test names.

### 2. Serve the region label (`f5c839804`)

- `isRegionPreviewOf` is now expanded on read into a bounded `{ @id, @type, rdfs:label }` reference (mirroring `hasFullImage`), instead of only `{ @id }`. The label/type come from the region resource already fetched during augmentation — **no extra query**. The write path (only reads `@id`) and the simple schema are unchanged.

### 3. Bounding-box crop for any geometry + tightened cardinalities (`1305bcc7f`)

- The crop URL and highlight box are now computed for **any geometry** (previously rectangle-only), at natural aspect ratio — nothing is forced square:
  - rectangle / polygon → tight bounding box of the vertices;
  - circle / ellipse → bounding box from centre ± `radius` (`GeomShape` gained a `radius` field).
- Corners are clamped to `[0,1]` so a shape at the image edge can't produce an out-of-range IIIF selector. **Rectangle output is byte-identical** to before.
- Cardinalities tightened to `ExactlyOne` (see the contract note above).
- Added a circle region to the read-surface fixture; new polygon + circle crop tests.

---

### Verification

- `ReadResourcesServiceLiveSpec` (rectangle / polygon / circle / augment / JSON-LD body) — green
- `ValuesResponderV2Spec` incl. `create a region preview value` — green (confirms tightened read cardinalities do not affect create/update)
- `OntologyFormatsE2ESpec` — green; the `knoraApiOntologyWithValueObjects.jsonld` approval file was regenerated
- `sbt fmt` + markdownlint — clean

PR #4201, opened 2026-07-21
docs: capture SPARQL pattern-order conventions from the tile-permission regression (DEV-6800)

BalduinLandolt:
## Summary

Captures the conventions from the DEV-6796 tile-permission regression (fixed in #4199/#4200) so this class of regression is caught at review time instead of on prod. Per DEV-6800, this goes into conventions/review docs rather than a standalone learning.

**`docs/development/dsp-api-sparql-queries.md`** — new section *Pattern Order and Query Performance*:

- Selective patterns first, always before `OPTIONAL` blocks (left-joins evaluate in document order; prod runs TDB2 on the fixed heuristic without `stats.opt`, so emitted order is load-bearing).
- The `pattern.and(group)` nesting trap — it emits `{ pattern . { … } }` instead of splicing into a flat group.
- Every `OPTIONAL` on a multi-valued property multiplies intermediate rows; prefer whole-subject fetch when a mapper consumes the model (DEV-6798).
- Shape-changing builder inputs (e.g. `AbstractEntityRepo.entityProperties`) are query changes — review the pinned/golden query diff.

**`CONVENTIONS.md`** — the two rules as work-phase bullets under § SPARQL.

**`REVIEW.md`** — two new checklist items under § SPARQL (pattern order / flat grouping; builder-shape changes must show up in a pinned/golden spec diff).

Also normalizes pre-existing underscore emphasis to asterisks (markdownlint MD049) in untouched sections of the SPARQL guide.

## Test plan

- Documentation only; `markdownlint` passes on the three files.

Resolves DEV-6800

https://claude.ai/code/session_014cRYN8pdJTmWkPti7xTxbY


siers (approved):
Technically I find the writing dense, but I assume this is for machine consumption.

PR #4207, opened 2026-07-23
fix: pass explicit Lucene hit limit in search queries to prevent silent 10k truncation (DEV-6823)

BalduinLandolt:
Resolves [DEV-6823](https://linear.app/dasch/issue/DEV-6823) (parent: [DEV-6822](https://linear.app/dasch/issue/DEV-6822))

## Problem

Every `text:query` sent to Fuseki's Lucene text index omitted the optional limit argument, so Jena silently capped the lookup at **10,000 hits** (`TextIndexLucene.MAX_N`). The cap is applied **before** the SPARQL filters (`limitToProject`, `limitToResourceClass`, `isDeleted`), and since wildcard/prefix queries are constant-score, the surviving 10k are an arbitrary subset of all matching text in the entire database, across all projects.

Result: resources were randomly unfindable in search-by-label (link-value autocomplete, advanced search) whenever the typed prefix matched more than 10k text snippets dataset-wide. This is the root cause behind the long-standing "inconsistent search" bug family (DEV-4589, DEV-5481, DEV-4904, likely DEV-3695), previously closed as not fixable. Full analysis, stage measurements, and reproduction recipes: [DEV-6822](https://linear.app/dasch/issue/DEV-6822).

Verified on stage (prod mirror): the search-by-label query for "Anton" (Webern project, class Person) returns 0 without a limit and the correct 8 persons with one; raw hit counts confirm the cap at exactly 10,000 (true count for `Anton*`: 11,782).

## Changes

- New shared constant `OntologyConstants.Fuseki.luceneHitLimit = 1000000` (~10× headroom over the worst prefix measured on prod-scale data, `uni*` = 94,187 hits; deliberately not unbounded since Lucene pre-allocates the hit-queue).
- `SearchQueries.selectCountByLabel` / `constructSearchByLabel`: `(rdfs:label "…")` → `(rdfs:label "…" 1000000)`.
- `SearchFulltextQuery` (fulltext route + count): `text:query "…"` → `text:query ("…" 1000000)`. Same bug, same one-line fix — folded in here; the Gravsearch call sites need AST work and are tracked separately in [DEV-6824](https://linear.app/dasch/issue/DEV-6824).

## Tests

- New `SearchQueriesSpec` (golden files) pins the generated search-by-label SPARQL including the limit.
- `SearchFulltextQuerySpec` expected queries updated.
- Local test data cannot reproduce the 10k cap, so correctness on prod-scale data was verified manually against the stage DB (measurements in DEV-6822/DEV-6823).

## Latency note

With the limit, worst-case terms (`uni*`) take ~1.6 s on the search route and ~9 s on the count route on stage (join-dominated, not Lucene). Typical terms ~1.5 s. Correct-and-slower beats fast-and-empty; count-route optimization can follow up if needed.

https://claude.ai/code/session_01SVPsnzF7SCjCutrnQ6ioCb

### DIFF
--- a/modules/webapi/src/main/scala/org/knora/webapi/messages/OntologyConstants.scala
+++ b/modules/webapi/src/main/scala/org/knora/webapi/messages/OntologyConstants.scala
@@ -1082,5 +1082,11 @@ object OntologyConstants {
 
   object Fuseki {
     val luceneQueryPredicate = "http://jena.apache.org/text#query"
+
+    // Without an explicit limit, Jena caps text index lookups at 10'000 hits (TextIndexLucene.MAX_N) and
+    // silently discards an arbitrary subset of matches *before* the SPARQL filters apply, making resources
+    // unfindable (DEV-6822). The worst realistic prefix measured on prod-scale data yields ~95k hits, so 1M
+    // leaves ample headroom while still bounding the Lucene collector allocation.
+    val luceneHitLimit: Int = 1000000
   }
 }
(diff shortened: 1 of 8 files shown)


derschnee68:
@claude review

siers (approved):

PR #4208, opened 2026-07-23
feat(dsp-api): cache file permission checks on the image-serving path (DEV-6806)

BalduinLandolt:
[DEV-6806](https://linear.app/dasch/issue/DEV-6806)

## Summary

- A single tiled image fires many identical `GET /admin/files/{shortcode}/{filename}` permission checks in a short burst. This puts a short-lived, in-memory `zio.cache.Cache` (`AssetPermissionsCache`) in front of the unchanged `AssetPermissionsResponder`, keyed by `(UserIri, Shortcode, InternalFilename)`.
- On a hit the previously computed decision (`permissionCode` + optional `restrictedViewSettings`) is returned without a triplestore round-trip, and zio-cache's single-flight collapses the simultaneous first-tile salvo into one resolution.
- Only successful decisions are retained (failures get `Duration.Zero`), expiry is lazy after a configurable TTL (default 2 min), capacity is bounded (default 10 000), and `hits`/`misses`/`size` are exposed as unlabeled Prometheus gauges.

## Changes

**Cache & config**
- New `AssetPermissionsCache` (wrapping service, `ZLayer.scoped`) in `responders.admin`; the responder stays a pure resolver. On a miss the lookup re-derives the `User` from its IRI via `UserService.findUserByIri` (the same path the auth layer uses), so anonymous requests resolve through the built-in fallback.
- `file-permission-cache { ttl, capacity }` HOCON section with env overrides; `AppConfig.FilePermissionCacheConfig` with load-time validation (`0 < ttl <= 10 min` — a staleness guard added from security review — and `capacity >= 1`).

**Wiring**
- `FilesServerEndpoints`, `AdminApiModule.Dependencies`, and `LayersLive` now route through the cache; the responder is still produced (the cache depends on it and `AssetPermissionsResponderSpec` still uses it).

**Build**
- `dev.zio:zio-cache:0.2.4` added to sbt (`Dependencies.scala`) **and** the Bazel build (`MODULE.bazel`, `modules/webapi/BUILD.bazel`, re-pinned `maven_install.json`).

## Test Plan

- **Unit** (`AssetPermissionsCacheSpec`, pure JVM): hit/miss stats, single-flight dedup, failure-not-cached, TTL expiry (`TestClock`), capacity bound, per-user key independence, deny (`code 0`) round-trip. ✅ 7/7
- **Config** (`AppConfigSpec`): asserts `ttl`/`capacity` load from HOCON and that `ttl = 0` / `capacity = 0` fail validation. ✅
- **Integration** (`AssetPermissionsCacheE2ESpec`, real triplestore via testcontainers): authenticated code 6, anonymous code 1 + settings, cross-project not-found, and cached-vs-direct-responder equivalence for the anonymous (constant-vs-re-derived) and authenticated cases. ✅ 5/5
- `sbt check` (scalafmt + scalafix) clean; `bazel build //modules/webapi:webapi //modules/webapi:test` and `//modules/test-it:test` compile.

## Notes for the reviewer

- **Metrics (REQ-3.1) — verified live.** Ran the built image in the full local stack: after 6 anonymous requests to one file, `/metrics` (:3339) reported `file_permission_cache_hits 5.0`, `file_permission_cache_misses 1.0`, `file_permission_cache_size 1.0` — the three unlabeled gauges are exposed via the Prometheus connector and track cache activity as designed.
- **Pre-existing Bazel/sbt version drift (not from this PR).** `//tools/deps:maven_versions_match_sbt` is red on `main` for ~21 unrelated deps (tapir, HikariCP, rdf4j, opentelemetry, flyway, …) that drifted between `Dependencies.scala` and `MODULE.bazel`. This PR's `zio-cache` addition is consistent on both sides; the drift is out of scope for DEV-6806 and flagged for a separate maintenance PR.

https://claude.ai/code/session_01MHdLgHyuoxFPv8bcpaPN55

### DIFF
--- a/modules/webapi/src/main/scala/org/knora/webapi/responders/admin/AssetPermissionsCache.scala
+++ b/modules/webapi/src/main/scala/org/knora/webapi/responders/admin/AssetPermissionsCache.scala
@@ -1,0 +1,119 @@
+/*
+ * Copyright © 2021 - 2026 Swiss National Data and Service Center for the Humanities and/or DaSCH Service Platform contributors.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.knora.webapi.responders.admin
+
+import zio.*
+import zio.cache.Cache
+import zio.cache.Lookup
+import zio.metrics.Metric
+import zio.metrics.PollingMetric
+
+import dsp.errors.BadRequestException
+import dsp.errors.NotFoundException
+import org.knora.webapi.config.AppConfig
+import org.knora.webapi.slice.admin.domain.model.InternalFilename
+import org.knora.webapi.slice.admin.domain.model.KnoraProject.Shortcode
+import org.knora.webapi.slice.admin.domain.model.User
+import org.knora.webapi.slice.admin.domain.model.UserIri
+import org.knora.webapi.slice.admin.domain.service.UserService
+import org.knora.webapi.slice.api.admin.model.PermissionCodeAndProjectRestrictedViewSettings
+
+/**
+ * A short-lived, in-memory cache in front of [[AssetPermissionsResponder]] for the IIIF tile-serving path (DEV-6806).
+ *
+ * A single tiled image fires many identical `GET /admin/files/{shortcode}/{filename}` checks in a short burst; caching
+ * the decision keyed by `(UserIri, Shortcode, InternalFilename)` sheds the repeated `FileValuePermissionsQuery`
+ * round-trip and permission computation within that burst, and zio-cache's single-flight behaviour collapses the
+ * simultaneous first-tile salvo into one resolution.
+ *
+ * The responder stays a pure resolver: this is a wrapping service, so the cache is self-contained and independently
+ * testable. Only successful decisions are retained (failures re-resolve on the next request), expiry is lazy after a
+ * configured TTL, and capacity is bounded — see [[AssetPermissionsCache.makeCache]].
+ */
+final case class AssetPermissionsCache(
+  private val cache: Cache[
+    AssetPermissionsCache.CacheKey,
+    Throwable,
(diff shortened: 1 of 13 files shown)


modules/webapi/src/main/scala/org/knora/webapi/responders/admin/AssetPermissionsCache.scala
```diff
@@ -0,0 +1,119 @@
...
+                   // anonymous IRI, so no special-casing is needed. This introduces a benign TOCTOU divergence from
+                   // the uncached path: the uncached path receives an already-hydrated `User` and never re-checks
+                   // existence, whereas here a user deleted between requests would surface as a `NotFoundException`.
+                   users
+                     .findUserByIri(key.userIri)
+                     .someOrFail(NotFoundException(s"No user found for ${key.userIri.value}"))
+                     .flatMap(
+                       responder.getPermissionCodeAndProjectRestrictedViewSettings(_)(key.shortcode, key.filename),
+                     )
+                 }
```
siers: I can't immediately tell whether this should be the case that half of the implementation is inside the Layer, not the instance itself. 

Otherwise the PR is 👍

siers (approved):

modules/webapi/src/main/scala/org/knora/webapi/responders/admin/AssetPermissionsCache.scala
BalduinLandolt: Yea, this is a bit weird about zio-cache: Because the cache is defined as a function `key -> value`, the lookup is not where the call happens, but where the cache is defined. Took me some getting used to.  
But admittedly, extracting this into a lookup function would be nicer. I'll do that.

PR #4194, opened 2026-07-21
chore: Dependency updates (patch + minor + all, combined)

siers:
Combines the three open dependency-update chore PRs into one branch, integrated sequentially and verified to compile green locally.

Supersedes:
- #4170 (patch)
- #4171 (minor)
- #4192 (all)

## Version bumps

**Patch (#4170)**
- Scala 3.3.7 → 3.3.8
- rdf4j 5.3.1 → 5.3.2
- spring-security-core 7.0.5 → 7.0.6
- tapir 1.13.19 → 1.13.25
- zio-opentelemetry 3.1.17 → 3.1.18
- sbt 1.12.11 → 1.12.13
- CI coverage-report path scala-3.3.7 → scala-3.3.8

**Minor (#4171)**
- refined 0.11.3 → 0.11.4
- datafaker 2.5.4 → 2.7.0
- opentelemetry (otel) 1.62.0 → 1.63.0
- opentelemetry-semconv 1.41.1 → 1.42.0
- flyway 12.7.0 → 12.10.0
- HikariCP 7.0.2 → 7.1.0
- sbt-scalafix 0.14.6 → 0.14.7

**All (#4192)**
- bouncycastle bcprov-jdk15to18 1.84 → 1.85

## Source change required by a bump

- `OntologyInferencer.scala`: Scala 3.3.8 requires nullary methods be called with `()`. Changed two `Random.nextInt.abs` call sites to `Random.nextInt().abs`. This is a genuine fatal warning under CI's `-Xfatal-warnings` that a plain local compile misses (local dev strips the flag via `project/LocalSettings.scala`).

## Verification

- `clean compile Test/compile` passes green with `-Xfatal-warnings` forced on after each merge.
- Full test suite not run locally — relying on CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4210, opened 2026-07-23
build(dsp-api): sync Bazel/sbt maven versions and add drift-sync tooling

BalduinLandolt:
## Summary

`//tools/deps:maven_versions_match_sbt` is red on `main`: the scala-steward bump (#4194) advanced `project/Dependencies.scala` (the sbt build) but not the Bazel side, so `MODULE.bazel`'s `maven.install` drifted behind for 21 shared coordinates. Since CI builds with Bazel, this affects every PR. This PR does three things:

1. **Fixes the current drift** — bumps `MODULE.bazel` to match `Dependencies.scala` and re-pins `maven_install.json`.
2. **Fixes the regression that the sync surfaced** — `SipiIT`, broken by the transitive `zio-http` upgrade the bump pulls in.
3. **Adds a CI workflow so this can't silently recur** — auto-syncs the Bazel manifests on any dependency-update PR and commits the fix back.

## Changes

**1. Version sync** — bumped in `MODULE.bazel`, re-pinned in `maven_install.json` (`bazel run @unpinned_maven//:pin`):

| Dependency | Bazel was | now (matches sbt) |
|---|---|---|
| `com.softwaremill.sttp.tapir:*` (6 artifacts) | 1.13.19 | 1.13.25 |
| `io.opentelemetry:*` sdk/exporters/sdk-testing | 1.62.0 | 1.63.0 |
| `io.opentelemetry.semconv:opentelemetry-semconv` | 1.41.1 | 1.42.0 |
| `dev.zio:zio-opentelemetry` | 3.1.17 | 3.1.18 |
| `org.eclipse.rdf4j:*` client/shacl/sparqlbuilder | 5.3.1 | 5.3.2 |
| `eu.timepit:refined` | 0.11.3 | 0.11.4 |
| `org.bouncycastle:bcprov-jdk15to18` | 1.84 | 1.85 |
| `org.flywaydb:flyway-core` | 12.7.0 | 12.10.0 |
| `com.zaxxer:HikariCP` | 7.0.2 | 7.1.0 |
| `net.datafaker:datafaker` | 2.5.4 | 2.7.0 |
| `org.springframework.security:spring-security-core` | 7.0.5 | 7.0.6 |

**2. `SipiIT` redirect fix** — aligning the versions upgrades `tapir` 1.13.19 → 1.13.25, transitively pulling **`zio-http` 3.8.0 → 3.11.2**. zio-http 3.11.2's `Client` no longer follows redirects, so `SipiIT` (which relies on Sipi's `303` from `/{prefix}/{id}/file` to the canonical IIIF URL) saw `SeeOther` instead of the permission-derived `Ok`/`Unauthorized`/`NotFound`. `SipiIT.requestGet` now follows redirects explicitly, re-resolving the target path through the test container so it always hits the mapped port.

**3. Auto-sync in CI** (so the drift can't recur unnoticed — scala-steward opens these PRs unattended, so a manual recipe alone wouldn't help):
- `.github/workflows/sync-bazel-deps.yml` runs on any PR that touches `project/Dependencies.scala`, runs `just sync-bazel-maven-versions`, and commits the resulting `MODULE.bazel` / `maven_install.json` changes back to the PR branch — pushed with the `GH_TOKEN` PAT so `build-and-test` re-runs against the synced state. Path-filtered so the sync commit (Bazel files only) doesn't re-trigger it (no loop); same-repo PRs only; release PRs skipped; `workflow_dispatch` for manual/test runs.
- `tools/deps/check_maven_versions.py --fix` (the workflow's engine) rewrites `MODULE.bazel` to match `Dependencies.scala` — both the `maven.install` artifacts list and the `maven.artifact(...)` override blocks; default (check) behaviour is unchanged. `just sync-bazel-maven-versions` (`--fix` + re-pin) is the one-command local/manual fallback.
- `docs/05-internals/development/third-party.md` documents the sbt+Bazel duality and both paths (was stale: "managed by SBT").

## Root cause (this is why the tooling is here)

scala-steward is sbt-native and updates only `project/Dependencies.scala`, never the Bazel manifests. **CI is Bazel-only, and on `main` the Bazel side still used the old versions — so the drift was masking a behaviour-changing bump (#4194's tapir → zio-http upgrade) that CI never exercised.** Had Bazel been kept in step, #4194's own CI would have caught the `SipiIT` break at bump time. `//tools/deps:maven_versions_match_sbt` *catches* drift but couldn't *fix* it; `just sync-bazel-maven-versions` now does.

## Test Plan

- `bazel test //tools/deps:maven_versions_match_sbt` → **PASS** (68 shared coordinates agree; was failing on 21).
- `bazel build //modules/webapi:webapi //modules/ingest:ingest` → builds clean with the bumped versions.
- `--fix` round-trip: introduced drift in both `MODULE.bazel` forms, `--fix` restored the file **byte-for-byte** to the synced state; `markdownlint` clean on the edited doc.
- `SipiIT` fix verified via CI (its WireMock uses a fixed port that's occupied locally).

## Validating the auto-sync workflow

A workflow added in a PR only takes effect once it's on the default branch, and it's path-filtered to `project/Dependencies.scala` (which this PR doesn't touch) — so `sync-bazel-deps.yml` does **not** run on this PR. It validates on the next dependency-update PR (scala-steward, 1st/15th) or via `workflow_dispatch` from a branch with drift. Worth a reviewer's eye on the token/permissions (`GH_TOKEN` push-back, `contents: write`, fork guard) since it can't be exercised here.

https://claude.ai/code/session_01MHdLgHyuoxFPv8bcpaPN55

### DIFF
--- a/.github/workflows/sync-bazel-deps.yml
+++ b/.github/workflows/sync-bazel-deps.yml
@@ -1,0 +1,68 @@
+name: Sync Bazel deps with sbt
+
+# scala-steward (and any human dependency bump) updates only project/Dependencies.scala; it has no
+# awareness of the Bazel manifests. If MODULE.bazel / maven_install.json aren't kept in step, the Bazel
+# build CI runs fails (//tools/deps:maven_versions_match_sbt, and the actual build). Since those PRs are
+# opened unattended, this workflow does the sync automatically: on any PR that touches Dependencies.scala
+# it runs `just sync-bazel-maven-versions` and commits the resulting MODULE.bazel / maven_install.json
+# changes back to the PR branch (which then re-runs the normal build-and-test checks against the synced
+# state). Manual `just sync-bazel-maven-versions` remains available as a fallback.
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+    paths:
+      - project/Dependencies.scala
+  # Manual trigger for testing / one-off syncs: dispatch it from the branch that has drifted.
+  workflow_dispatch: {}
+
+# Needs write to push the sync commit back to the PR branch.
+permissions:
+  contents: write
+
+concurrency:
+  group: sync-bazel-deps-${{ github.head_ref || github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  sync:
+    name: Sync MODULE.bazel to Dependencies.scala
+    runs-on: ubuntu-latest
+    # Same-repo PRs only (we cannot push to a fork's branch); skip release PRs; always allow manual dispatch.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      (github.event.pull_request.head.repo.full_name == github.repository &&
+       !startsWith(github.head_ref, 'release-please'))
+    steps:
+      - name: Checkout the PR branch
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ github.head_ref || github.ref_name }}
+          # PAT (not the default GITHUB_TOKEN) so the pushed sync commit re-triggers the
+          # build-and-test workflow; a push made with GITHUB_TOKEN would not.
+          token: ${{ secrets.GH_TOKEN }}
+          fetch-depth: 0
+      - name: Install Nix (provides Bazel + just for the re-pin)
+        uses: DeterminateSystems/determinate-nix-action@v3
+      - name: Enable Nix build cache
+        uses: DeterminateSystems/magic-nix-cache-action@v14
+      - name: Restore Bazel repository cache
+        uses: actions/cache@v5
+        with:
+          path: ~/.cache/bazel-repo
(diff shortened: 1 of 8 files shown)


subotic (approved):
Isn't this a bit too much? Why not simply turn of renovate until Marcin or someone else figures out how to update Bazel dependencies and at that time, we can simply remove SBT. This will need to happen in the next 1-2 weeks anyway.

BalduinLandolt:
> Isn't this a bit too much? Why not simply turn of renovate until Marcin or someone else figures out how to update Bazel dependencies and at that time, we can simply remove SBT. This will need to happen in the next 1-2 weeks anyway.

You're right, this is over-doing it.  
I'll merge it anyways, because we already have it, and because Claude mixed in some important stuff (deps upgrading and the necessary fixes along with it).  
But we should simplify this for sure.

PR #4214, opened 2026-07-24
chore: fix misleading object-class-constraint error message

Nora-Olivia-Ammann:
### Description

**What kind of change does this PR introduce?** Bug fix to a user-facing error message (no change to validation behaviour).

**What is the current behavior?**

When a resource is created/updated with a link (or value) whose target does not satisfy the property's `objectClassConstraint`, the API returns an `OntologyConstraintException` whose message lists the acceptable classes in a confusing way:

```
Resource <...> cannot be the object of property <...#isPartOfChapterValue>,
because it does not belong to class <...#Chapter,...#Chapter>
```

Two problems combine:

1. **Malformed formatting** — the list of acceptable classes is joined with a bare comma and wrapped in a single pair of angle brackets, so it reads like one broken IRI `<A,B>` instead of "one of `<A>`, `<B>`". (`mkString(",")` inside `<...>`.)
2. **Duplication** — the constraint class appears twice (`<Chapter,Chapter>`). The acceptable-class list is built as `objectConstraint +: findDirectSubclassesBy(objectConstraint)`, but `findDirectSubclassesBy` already includes the class itself (a class is registered as a subclass of itself in the ontology cache — see the `subclassesWithSelf` naming in `OntologyRepoLive.findAllSubclassesBy`). So the constraint class is contributed once by the prepend and once by the self-inclusive lookup.

**What is the new behavior?**

Same message for both affected paths (`propertyInvalid`, `propertyRequiresValue`), now readable:

```
Resource <...> cannot be the object of property <...#isPartOfChapterValue>,
because it does not belong to any of the following classes: <...#Chapter>
```

- Each acceptable class IRI is wrapped in its own angle brackets and separated by `, ` (extracted into a small `formatObjectClassConstraints` helper).
- The constraint list is de-duplicated at the source in `propertysObjectConstraint`.

The validation logic itself is unchanged — only the human-readable error text.

**Does this PR introduce a breaking change?** No. Only the wording of an error message changes; there is no stable contract on error-message text.

### Tests

- New unit spec `CheckObjectClassConstraintsSpec` covering the formatting helper (each IRI individually bracketed, comma-space separated, empty list, and a regression guard that duplicates never produce a `<A,B>`-style token).
- Strengthened the existing integration test in `ResourcesResponderV2Spec` ("not create a resource with a link to a resource of the wrong class for the link property") to assert the message end-to-end: the new phrasing is present, no bare-comma-joined IRIs (`,http`), and the constraint class is listed exactly once. Kept order-independent because `findDirectSubclassesBy` iterates a map (unstable subclass order).

### DIFF
--- a/modules/test-it/src/test/scala/org/knora/webapi/responders/v2/ResourcesResponderV2Spec.scala
+++ b/modules/test-it/src/test/scala/org/knora/webapi/responders/v2/ResourcesResponderV2Spec.scala
@@ -1459,9 +1459,21 @@ class ResourcesResponderV2Spec extends E2EZSpec { self =>
           projectADM = anythingProject,
         )
 
-        resourceResponder(_.createResource(CreateResourceRequestV2(inputResource, anythingUser2, randomUUID))).exit.map(
-          actual => assert(actual)(failsWithA[OntologyConstraintException]),
-        )
+        // objectClassConstraint of hasOtherThing is anything:Thing, which has subclasses. The error must
+        // list the acceptable classes with each IRI in its own angle brackets (no malformed `<A,B>` token)
+        // and the constraint class must appear only once (it is a subclass of itself in the ontology cache).
+        val thingIri = "http://0.0.0.0:3333/ontology/0001/anything/v2#Thing"
+        resourceResponder(
+          _.createResource(CreateResourceRequestV2(inputResource, anythingUser2, randomUUID)),
+        ).exit.map { actual =>
+          val message = actual.causeOption.flatMap(_.failureOption).map(_.getMessage).getOrElse("")
+          assert(actual)(failsWithA[OntologyConstraintException]) &&
+          assertTrue(
+            message.contains("does not belong to any of the following classes:"),
+            !message.contains(",http"),
+            message.split(java.util.regex.Pattern.quote(s"<$thingIri>"), -1).length - 1 == 1,
+          )
+        }
       },
       test("not create a resource with invalid custom permissions") {
         val resourceIri   = ResourceIri.makeNew(anythingProject.shortcode)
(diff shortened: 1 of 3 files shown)


modules/test-it/src/test/scala/org/knora/webapi/responders/v2/ResourcesResponderV2Spec.scala
```diff
@@ -1459,9 +1459,21 @@ class ResourcesResponderV2Spec extends E2EZSpec { self =>
...
+        val thingIri = "http://0.0.0.0:3333/ontology/0001/anything/v2#Thing"
+        resourceResponder(
+          _.createResource(CreateResourceRequestV2(inputResource, anythingUser2, randomUUID)),
+        ).exit.map { actual =>
+          val message = actual.causeOption.flatMap(_.failureOption).map(_.getMessage).getOrElse("")
+          assert(actual)(failsWithA[OntologyConstraintException]) &&
+          assertTrue(
+            message.contains("does not belong to any of the following classes:"),
+            !message.contains(",http"),
+            message.split(java.util.regex.Pattern.quote(s"<$thingIri>"), -1).length - 1 == 1,
```
BalduinLandolt, line 1474: Nitpick: Scala has it's own Regex machinery. I don't think we need to fall back to Java here. And I would assume, a Regex could be written, that makes this assertion nicer than `.length -1 == 1`. But it doesn't really matter.

BalduinLandolt (approved):

modules/test-it/src/test/scala/org/knora/webapi/responders/v2/ResourcesResponderV2Spec.scala
Nora-Olivia-Ammann: sorry, automerge ignored your comment. If you want i can open a new PR?

BalduinLandolt: the `java.` is the giveaway. 😉 <br>No, seriously, it doesn't matter. If you want to open a new PR, the more valuable thing would be to add a statement to the conventions/review file, that scala idioms are to be preferred over java; and let it sweep the codebase for java-uses, to reduce the pattern, so it doesn't reproduce it from what's already around. But frankly, I would guess you have more valuable things to do with your time.

PR #4219, opened 2026-07-24
feat(dsp-api): admin SPARQL query passthrough, read-only (DEV-6734)

BalduinLandolt:
[DEV-6734](https://linear.app/dasch/issue/DEV-6734)

Milestone 1 of the admin SPARQL passthrough surface: the cross-cutting foundations plus a **read-only** SPARQL 1.1 query endpoint. M2 (update) is a separate plan and reuses everything here.

## Motivation

Staff lost the sustainable path to triplestore inspection when the DB password was encrypted and the DB port closed, and the planned Fuseki to in-process Jena move will remove the standalone store entirely. This restores the most frequent, lowest-risk capability — inspecting data the curated API does not surface — through dsp-api, with per-user identity, no shared secret, a kill-switch, and attributable logging.

## What this adds

`POST /admin/sparql/query`, registered only where a deployment sets `allow-sparql-passthrough` (default off, read at startup). It forwards a `SystemAdmin`'s query untouched and relays the store's status, `Content-Type` and body verbatim.

- **Config**: the flag plus four guardrails under `app.triplestore.sparql-passthrough` (store timeout 120s, request cap 1 MiB, response ceiling 64 MiB, concurrency backstop 8), each env-overridable and validated at load. A startup WARNING echoes the resolved flag so an accidentally-enabled passthrough is alertable.
- **Seam**: `TriplestoreService.rawQuery`, the one method taking an arbitrary query string. This is what an in-process Jena impl or a remote client (M3) reimplements.
- **Endpoint**: the three-tier split, on a new bearer-only `BaseEndpoints.bearerSecuredEndpoint`.
- **Store side**: `arq:httpServiceAllowed=false` in `modules/fuseki/dsp-repo.ttl`, plus docs and an entrypoint warning.

## Challenges and decisions

**The channel split is the design.** A store error — malformed SPARQL, a 406, the store's own timeout response — is a *successful* `RawSparqlResponse` carrying the store's status, so it can be relayed. `rawQuery` therefore bypasses `doHttpRequest`/`checkResponse`, which would turn a user's bad query into a dsp-api 500 instead of the store's 400, and uses `asStreamAlways` so a 4xx is not folded into a failure. dsp-api's own failures ride the error channel.

**Error variants are per-endpoint, never shared.** A typed variant in `BaseEndpoints.errorOutputs` serializes the exception message verbatim, so putting these there would make every endpoint — including public ones — start returning them. All six outcomes are enumerated via `errorOutVariantsPrepend`; an outcome without a variant falls through to the default and is reported as a generic 500.

**Every error type is new and storeless.** `TriplestoreConnectionException` is deliberately not reused: its message embeds the store URI. No `cause` is attached either, since an sttp transport exception's own message carries the target URI and the OTel agent records exception messages and stacktraces. The store URI goes to the server-side log instead.

**Bearer-only.** HTTP basic runs a bcrypt verification on every call, and because security logic runs before server logic that cost would sit upstream of every bound this surface establishes — the one unbounded part of a route built to bound itself. Omitting basic also keeps the `WWW-Authenticate: Basic` challenge off a route the API docs advertise. No cookie input is what keeps it CSRF-safe under reflected-origin CORS.

## Gotchas

- **Tapir's `NotAcceptableInterceptor` blocks arbitrary `Accept` headers.** It answers an empty 406 before the server logic runs when no declared output media type matches, so `Accept: text/csv` and `text/turtle` never reached the store, and the store's own 406 was unreachable. Declaring the body `*/*` does *not* help — the matching treats a wildcard as meaningful only on the requested-range side. The interceptor is replaced with one exempting this single route by path template; every other endpoint behaves exactly as before. The template is a constant with a test asserting it equals the endpoint's own rendered value.
- **The plan's JVM-arg SSRF mitigation does not exist.** Measured: `-Darq:httpServiceAllowed=false` is ignored (the outbound call was still attempted), and Fuseki 6.1.0 has no `--set`. The assembler `ja:context` is the only mechanism, and it must sit on the dataset the service references. Because the entrypoint copies that file only when absent and `/fuseki` is a persistent volume, existing environments keep their old config across redeploys — so the docs make a **live per-host SERVICE probe** the precondition, and the entrypoint warns.
- **Two different timeout units**, both measured: Fuseki's per-request `timeout` parameter is in **seconds** (dsp-api's existing convention is correct), while `arq:queryTimeout` and `--timeout` are in **milliseconds**.
- **The request cap is a policy check, not a memory bound**, and the code says so. The body is already materialised when it runs; what makes it safe is that the body is always fully consumed, which keeps the connection usable. Failing the stream at the cap instead would abort it without draining — the `autoRead=false` keep-alive hazard.
- **`user.userIri` can throw**: it re-validates the IRI. A log line must never be able to fail the request it describes, so the raw `user.id` is logged.
- The concurrency backstop uses an atomic `Ref#modify`, not a semaphore: `withPermit` would queue, accumulating fibers and connections rather than rejecting.

## Test Plan

77 new/changed tests, all passing:

- `webapi` unit (49): config defaults and validation; `rawQuery` against the in-memory double (SELECT/ASK/CONSTRUCT/DESCRIBE, JSON/XML/CSV, 400 on parse failure, 406); RestService authz, flag gating, query-string rejection, request cap; backstop reject-not-queue with a latch; route registration and the path-template drift guard.
- `test-it` (12): `rawQuery` against real Fuseki — TSV, DESCRIBE, store 4xx relayed, store-side timeout cancellation, unreachable store with the message compared for equality, response ceiling; plus the span-status description guard.
- `test-e2e` (16): 401/403/200, basic rejected while working elsewhere, cookie rejected, no Basic challenge, both request forms, 415, `?query=` 400, Content-Type relay both ways, malformed-query relay, and a >1 MiB body giving 413 with the next request on the same connection succeeding.

`sbt check` and `sbt "scalafixAll --check"` pass. `sbt test` has 5 failures (`DateUtilSpec`, `UpgradePluginPR3112Spec`, `ValuesEndpointsE2ESpec`, `SearchEndpointsPostGravsearchWithTypeInferenceComplexSchemaE2ESpec`) — **all reproduced identically on clean `origin/main`** in a throwaway worktree, and none touch code in this diff.

## Not included — needed before enabling this anywhere

- **ops-deploy**: an ansible `false` default for the flag (the templates render it unconditionally, so prod's state is decided there, not by dsp-api's default); the per-host SERVICE verification; the optional `arq:queryTimeout` floor; an alert on the prod flag being on.
- **The `SERVICE`-refused integration test** cannot be written yet: `FusekiTestContainer` uses the published `daschswiss/apache-jena-fuseki` tag, so the TTL change only reaches tests after a new image is published and `Dependencies.scala` bumped. The behaviour was verified manually against 6.1.0 (422, `SERVICE execution disabled`).
- No E2E test for the backstop 503: whether concurrent requests genuinely overlap depends on store latency, so it would be flaky. It is covered deterministically at unit level with a latch.
- The `union-default-graph-differs-across-environments` learning doc in `dasch-specs` is still wrong and needs correcting or retiring.

https://claude.ai/code/session_01Jg8eNFcfAydSpTx7QoqQp7


mpro7:
@claude review

mpro7 (approved):

PR #4228, opened 2026-07-28
build: remove sbt; Bazel now owns compile, test, formatting, linting, and deps

subotic:
### Description

Closes out the Bazel validation window by **removing sbt entirely**. Bazel (rules_scala / rules_oci via the Nix dev shell) is now the sole build tool — compile, test, images, formatting, linting, and dependency management all run through it.

**What kind of change:** build/tooling (maintenance). No runtime/API behavior change — `/version` reports the same sipi/fuseki values as before.

**What changed**

- **Formatting / linting → Bazel.** rules_scala scalafmt toolchain (`tools/scalafmt/`) driving `.scalafmt.conf`. Import grouping is configured **third-party-first** (a leading `[".*"]` group), matching the previous scalafix `OrganizeImports` convention, so the reformat is a small formatting-only delta (16 files, isolated in the `style:` commit and listed in `.git-blame-ignore-revs`) rather than a wholesale import relocation. scalafix is dropped: unused imports already fail the compiler (`-Wunused:all -Werror`).
- **No relative imports → `//tools/lint`.** A new `//tools/lint:no_relative_imports` check (py_test, mirroring `//tools/license`, wired into `just check`) forbids **top-level** relative imports. scalafmt reorders the top-level import block syntactically and would otherwise hoist a relative import above the wildcard that brings its prefix into scope and break the build — this replaces the safety the removed semantic scalafix rule provided. Local/member imports (e.g. `import Codecs.*` inside a def) are unaffected. The 8 existing top-level relative imports are fully-qualified in a preceding `refactor:` commit.
- **License headers → `//tools/license`** (SPDX check + `just header-fix`), replacing sbt-header.
- **Dependencies → Renovate.** `MODULE.bazel`'s `maven.install` is the sole version source; `.github/renovate.json` opens update PRs and `renovate-pin.yml` re-pins `maven_install.json`. scala-steward, the sync workflow, `dependency-graph.yml`, and both `//tools` drift checks are removed.
- **Image versions single-sourced.** sipi is read from the pinned base image's `org.opencontainers.image.version` OCI label (`//tools/buildinfo:oci_config_label`; the MODULE.bazel digest is its sole source). fuseki's Jena dist version lives once in `MODULE.bazel` as `FUSEKI_DIST_VERSION`, consumed by the `image_versions.fuseki` tag and the `fuseki_dist` http_archive (strip_prefix + url); the image tag stays literal for Renovate. `FusekiTestContainer` gets the image from the test target's `env`, not `BuildInfo`. `tools/workspace_status.sh` is back to pure git.
- **`just` + CI.** `fmt` / `check` / `header-fix` / `run-with-dev-db` and the CI formatting job now run on Bazel; `check` also runs the no-relative-imports lint. The temporary tag-drift gate is removed; `docker-publish-fuseki-image` reads the tag from its single source in `MODULE.bazel`; `metals-bootstrap` uses bazel-bsp; `flake.nix` ships `metals-mcp`. The build-and-test jobs bootstrap via the **local** `./.github/actions/preparation` (not `@main`), so the sbt-less tree runs its own preparation step (main's copy still has an sbt cache that keys off deleted `build.sbt`/`project` files).
- **Deleted:** `build.sbt`, `project/`, `sbtx`, `.sbtopts`, `.scalafix.conf`, `.scala-steward.conf`.

**Verification (local, this branch):** full Scala compile of `//modules/...` under `-Werror`; `just check` green (scalafmt + license + no-relative-imports, 915 files); webapi + affected test modules compile; `//modules/fuseki:load` analyzes. The unit/integration/e2e test suites and the formatting gate are being validated in **CI**, which this PR unblocked (it could not run at all until the rebase resolved the merge conflict).

**Metals works on Bazel 9 (local bazel-bsp patch).** bazel-bsp 4.0.x (the server Metals 1.6.7 downloads) predates Bazel 9; its sync aspect breaks two ways on Bazel 9.1.1, both fixed locally so `list-modules` / `compile-file` / `get-usages` work: (1) a **macOS-scoped** `--incompatible_autoload_externally` flag in `.bazelrc` restores the `JavaInfo`/`CcInfo` globals Bazel 9 removed (macOS-only — local dev is macOS, CI is Linux and never runs bazel-bsp; a no-op for our own build); (2) `tools/metals/patch_bazelbsp_core.py` rewrites bazel-bsp’s struct-returning aspect to a real provider, re-applied on every invocation by a `bazel_binary` wrapper (`.bazelproject` → `tools/metals/bazel-bsp-wrapper.sh`) since bazel-bsp re-copies its aspects on each import. Vendored patch, **not** a rules_scala 7.x issue (rules_scala 7.2.6 + Scala 3.8.4 build fine); remove once bazel-bsp supports Bazel 9 upstream (scalameta/metals#8268). Fresh worktree: `just metals-bootstrap` then the metals `import-build` tool. Documented in `CLAUDE.md`, `CONVENTIONS.md`, `docs/development/dsp-api-metals-mcp.md`.

**Other information:** the version single-sourcing rule is written up in `CONVENTIONS.md` → "Versions single-sourced".


tools/license/check_spdx_header.py
```diff
@@ -0,0 +1,51 @@
...
+
+py_test (not an sh_test around a python3 shebang) so it resolves the hermetic
+interpreter registered in MODULE.bazel — same rationale as //tools/deps.
+"""
+import re
+import sys
+
+# The `©` char is intentionally omitted from the pattern for encoding
+# robustness; this substring is specific enough to identify the header.
+COPYRIGHT_RE = re.compile(r"2021 - \d{4} Swiss National Data and Service Center for the Humanities")
```
BalduinLandolt, line 19: If I understand correctly, this would allow any 4-digit number, and not check for the current year, right?

BalduinLandolt (approved):
I haven't reviewed everything in detail, but looks reasonable.

@raitis.veinbahs FYI

PR #4234, opened 2026-07-31
fix(testkit): build the E2E server layer interruptibly so request handlers match production (DEV-6864)

BalduinLandolt:
[DEV-6864](https://linear.app/dasch/issue/DEV-6864)

## Problem

`E2EZSpec` builds `DspApiServer.layer` inside `TestAspect.beforeAll`, whose `ZIO.acquireRelease` acquire runs **uninterruptibly**. zio-http captures the runtime flags of the fiber that builds the server layer (`AppRef.empty` → `ZIO.runtime`, preserved by `NettyDriver.addApp`) into the `Runtime` it forks every request-handler fiber from. Net effect: **every request handler in the E2E stack runs with the Interruption flag disabled** — unlike production, where `Main` builds the layers on the interruptible main fiber.

This silently breaks anything inside a request handler that relies on interrupting a fiber:

- `race`/`raceFirst` with a non-terminating loser deadlocks (`raceFibersWith` forks contenders with the parent's flags; the finisher `loser.interrupt *> exit` then waits forever) — this is exactly the DEV-6864 raced breadth-guard hang, see the corrected learning in dasch-swiss/dasch-specs#166
- `timeout` inside handlers cannot interrupt
- zio-http's own connection-close listener (`fiber.interruptAsFork`) is a no-op, so client-disconnect behavior is untested
- `fork`ed fibers inside handlers inherit interruption-off, so even explicit `Fiber#interrupt` degrades to awaiting natural completion

## Fix

Build the server layer under `ZIO.interruptible` in `E2EZSpec.prepare` (one line + comment). This restores production parity; production layer builds are interruptible too, so no safety regression.

## Verification

- `SearchEndpointsGetSearchE2ESpec` passes on this branch (21.6s).
- Regression demo: the raced-guard commit `f23528c2f` — which hung deterministically on every probed fulltext search (510s timeout cascade) — **passes in 19.5s with only this change applied**, confirming the E2E setup, not the guard's real async effects, was the missing variable.
- Upstream context: [zio/zio#11044](https://github.com/zio/zio/issues/11044); the flag-capture question is filed as [zio/zio-http#4240](https://github.com/zio/zio-http/issues/4240).

Not audited here: `MockDspApiServer` (SipiIT) is built via `provideSomeLayerShared`, a different code path; its canned-response handlers don't rely on interruption.

https://claude.ai/code/session_01Gwpb3zGjZRQ7bPoNQYXsmS


derschnee68:
@claude review

derschnee68 (approved):

PR #4236, opened 2026-08-03
fix(admin): don't key the asset-permission cache on the decorative shortcode

BalduinLandolt:
## What

`GET /admin/files/{shortcode}/{filename}` identifies the file by its internal `filename` alone (a globally-unique asset ID). The `{shortcode}` path segment only ever affected the permission-code-1 (restricted view) branch, yet the permission cache introduced in DEV-6806 keyed on `(userIri, shortcode, filename)`. An anonymous caller varying only the shortcode over one known filename could therefore fill the bounded 10k-entry cache with duplicate entries that all resolve to the same value, evicting the genuinely-hot tile entries the cache exists to serve — with no authentication and no invalid response to signal it.

## Change (Option B — shortcode non-authoritative)

- The shortcode no longer affects the answer for **any** permission code, and is **dropped from the cache key** (now `(userIri, filename)`).
- Code 1 still resolves the project by IRI (EntityCache-served `findById`) for its restricted-view settings, but no longer compares the shortcode — so the divergent `"... in project {shortcode}"` 404 message is gone; only `"No file value was found for filename ..."` remains.
- The endpoint keeps the `{shortcode}` segment for URL compatibility and its description now documents it as non-authoritative.

This closes the amplification vector by construction, adds no per-request triplestore query on the tile hot path, and reconciles the error messages.

### Behaviour change
A permission-code-1 request with a *wrong* project shortcode previously returned `404`; it now returns `200 {"permissionCode":1, ...}` like every other code. The shortcode is documented as decorative.

## Acceptance criteria (DEV-6867)
- [x] Filling the cache by varying only the shortcode no longer evicts other `(user, filename)` entries — shortcode is out of the key.
- [x] Shortcode treatment is consistent across permission codes (never authoritative) and documented as such.
- [x] "file not found" / "not in this project" error messages reconciled (the latter is removed).
- [x] No per-request triplestore query added on the tile hot path.

## Verification
- `bazel build //modules/webapi` + `//modules/test-it:test` — compile clean under `-Wunused:all -Werror`.
- `//modules/webapi:test --test_filter=.*AssetPermissionsCacheSpec.*` — passes.
- `just check` (scalafmt, SPDX headers, no-relative-imports) — passes.
- Integration specs (`AssetPermissionsResponderSpec`, `AssetPermissionsCacheE2ESpec`) compile; not run locally (need Docker + triplestore) — relying on CI. `AssetPermissionsCacheE2ESpec` gains a regression test asserting two different shortcodes for the same `(user, filename)` yield the same decision.

## Follow-up
DEV-6892 — deprecate this route in favour of a shortcode-independent one, since the segment is now decorative.

Fixes DEV-6867.

### DIFF
--- a/modules/test-e2e/src/test/scala/org/knora/webapi/e2e/admin/AdminFilesE2ESpec.scala
+++ b/modules/test-e2e/src/test/scala/org/knora/webapi/e2e/admin/AdminFilesE2ESpec.scala
@@ -14,6 +14,7 @@ import org.knora.testrunner.DspZTestJUnitRunner
 import org.knora.webapi.E2EZSpec
 import org.knora.webapi.messages.store.triplestoremessages.RdfDataObject
 import org.knora.webapi.sharedtestdata.SharedTestDataADM.*
+import org.knora.webapi.slice.admin.domain.model.KnoraProject.Shortcode
 import org.knora.webapi.slice.api.admin.model.PermissionCodeAndProjectRestrictedViewSettings
 import org.knora.webapi.slice.api.admin.model.ProjectRestrictedViewSettingsADM
 import org.knora.webapi.testservices.ResponseOps.*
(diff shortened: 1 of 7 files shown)


mpro7:
@claude review

mpro7 (approved):

PR #4239, opened 2026-08-03
feat(admin): view-restrictions reporting API (DEV-6778)

derschnee68:
## What

Adds a read-only **view-restrictions reporting API** for the admin "view restrictions" screen (iteration 1h). For a project, it reports how many items are hidden from each of the three built-in audiences (anonymous, authenticated, project member) and lets an admin drill into the affected resources and their restricted parts.

Two endpoints under `admin/projects/iri/{projectIri}/view-restrictions`:

- **`GET .../summary`** — per-audience counts of hidden items, grouped by resource class or by the property that carries the restriction (`groupBy`), optionally filtered to one item type (`itemType` = `all`/`resource`/`file`/`value`/`comment`).
- **`GET .../items`** — paginated list of the resources affected under a given class/property, each with its per-audience visibility and its nested restricted file values / values / comments.

Both require the requesting user to be **project admin on the target project or system admin**.

## How

- Slice-architecture wiring: `ViewRestrictionsEndpoints` → `ViewRestrictionsServerEndpoints` → `ViewRestrictionsRestService` → `ViewRestrictionsService` → `ViewRestrictionsRepo`, registered through `AdminDomainModule` / `AdminApiModule` / `AdminApiServerEndpoints`.
- **Visibility is resolved with the real permission model** (`PermissionUtilADM.getUserPermissionADM`) against a synthetic non-admin user per audience. Permission code 0 → `Hidden`; code 1 → `RestrictedView` only for file values (renderable, watermarked/reduced), else collapses to `Hidden`; code ≥ 2 → `Visible`. Summary counts hidden items only; restricted-view is surfaced per item on the drill-down.
- **Comment semantics:** a comment is a `valueHasComment` literal on a value, sharing that value's permissions — so under `itemType=all` a commented value is counted once, not double-counted.
- SPARQL uses the rdf4j query builder (no string concat), with a conservative regex pre-filter that drops provably-open rows so the scan stays proportional to the number of restrictions.
- **Large-project guardrail (v1):** each query is capped at `ViewRestrictionsRepo.ScanCap` rows; hitting the cap flags the summary `approximate = true` (counts become a lower bound). A configurable cap / cache is a documented follow-up (PRD §7/§9).
- Adds a committed OpenAPI spec (`docs/openapi/openapi.yaml`) and a `just dump-openapi` recipe so the frontend can codegen a typed client. Includes the PRD.

## Testing

- `ViewRestrictionsServiceSpec` — pure visibility mapping, summary counts (cumulative, deleted-excluded), drill-down, and the values/files/comments + property-grouping paths over seeded Turtle fixtures.
- `AdminViewRestrictionsE2ESpec` — 200 for sysadmin and project admin, 401 anon, 403 non-admin and cross-project admin, plus the drill-down paging shape.

## Notes / follow-ups

- Label/ontology names are derived from IRI local-names pending proper ontology-label resolution.
- `restrictedView` state, property-first grouping refinements, and the large-project guardrail hardening are left for the DEV-6777 feasibility spike.


derschnee68:
@claude review

docs/prd/2026-07-28-view-restrictions-api.md
BalduinLandolt, line 1: PRDs should not live in the API repo but in [dasch-specs](https://github.com/dasch-swiss/dasch-specs).

docs/prd/2026-07-28-view-restrictions-api.md
```diff
@@ -0,0 +1,414 @@
...
+| Code | Level | Screen eye-state |
+| --- | --- | --- |
+| `0` | no access | **hidden** (`visibility_off`) |
+| `1` | RestrictedView (RV) | **restricted view** (`blur_on`) — image file values only |
+| `≥ 2` | View / Modify / … | **visible** (`visibility`) |
+
+**Count semantics (decided):** the summary *number* counts **hidden only** (code `0`).
+Restricted-view (code `1`) is surfaced as its own eye-state on drill-down rows but is
+**not** added to any summary count. Because RV only applies to image file values, the
+RV state can only appear on file-value rows.
```
BalduinLandolt: technically, RV can be added to resource/values other than files, it's just not meaningful. I will check, if we have that in data or not.

docs/prd/2026-07-28-view-restrictions-api.md
```diff
@@ -0,0 +1,414 @@
...
+Fills the expanded rows under one group. Returns the affected resources and, nested,
+the restricted file values / values / comments, each with its per-audience visibility.
+
+```
+GET /admin/projects/{projectIri}/view-restrictions/items
+      ?groupBy=resourceClass | property
+      &group={class-or-property-IRI}          (required — the expanded row)
+      &itemType=all | resource | file | value | comment
+      &page=1&page-size=25                     (PageAndSize; max size 100)
+```
```
BalduinLandolt: I'm not sure, how much it would change - but since we have two "modes" of how these endpoints work ("resource mode" and "property mode"), perhaps the endpoints would be cleaner if these two modes were modelled as different endpoints (maybe `.../view-restrictions/propety|class` or something like that). Have you considered this option?

docs/prd/2026-07-28-view-restrictions-api.md
```diff
@@ -0,0 +1,414 @@
...
+## 9. Open Questions
+
+1. **Large-project guardrail.** ~~With no cache in v1, what protects the summary request…~~
+   **Resolved (see §0):** each query is capped at `ScanCap` rows and the summary reports
+   `approximate = true` when the cap is hit. A configurable cap and a cached summary remain the
+   performance follow-up.
+2. **Definition of "affected resource" in the drill-down** when only a nested part is
+   restricted but the resource itself is fully visible — confirmed included (resource is
+   listed with `resourceVisibility` all-visible and the restricted parts nested). Confirm
+   this is the desired reading for property mode too.
```
BalduinLandolt: I don't fully get this - if you want me to confirm this, maybe you could explain it some more, so I understand?

docs/prd/2026-07-28-view-restrictions-api.md
```diff
@@ -0,0 +1,414 @@
...
+3. **Property-mode counting of comments.** A comment is attached to a value of some
+   property; in property mode, should a restricted comment be grouped under that value's
+   property, or under a synthetic "comment" grouping? (Assumed: under the value's
+   property.)
```
BalduinLandolt: sounds reasonable to me

docs/prd/2026-07-28-view-restrictions-api.md
```diff
@@ -0,0 +1,414 @@
...
+4. **Ontology/label localisation.** Which language are `label` / `propertyLabel`
+   returned in — the requesting user's preferred language, or the project default?
```
BalduinLandolt: I would take the user's preferred language, and only if that's not available, fall back

docs/prd/2026-07-28-view-restrictions-api.md
```diff
@@ -0,0 +1,414 @@
...
+5. **Stable group ordering.** Should groups be ordered by descending anonymous-hidden
+   count (as 1h shows), and is that ordering part of the contract or a UI concern?
```
BalduinLandolt: I don't think that matters very much, right?

docs/prd/2026-07-28-view-restrictions-api.md
```diff
@@ -0,0 +1,414 @@
...
+6. **Exact project-admin authorization predicate** to reuse (the same check as other
+   project-scoped admin write endpoints) — confirm against `BaseEndpoints` /
+   `ProjectRestService`.
```
BalduinLandolt: I think there is  a `ensureProjectOrSystemAdmin` function in the auth code, that should do exactly what you want.

docs/prd/2026-07-28-view-restrictions-api.md
```diff
@@ -0,0 +1,414 @@
...
+
+---
+
+## 10. Next Steps
+
+1. Review & resolve the Open Questions (esp. #1 guardrail and #6 auth predicate).
+2. Run a **perf spike**: benchmark a single per-class count query on the largest
+   available fixture/dev project to size the guardrail decision.
```
BalduinLandolt: did you do that? do we have numbers?

docs/prd/2026-07-28-view-restrictions-api.md
```diff
@@ -0,0 +1,414 @@
...
+3. Turn this PRD into an implementation plan (`/plan`) — scaffold the 5-file admin slice
```
BalduinLandolt: I guess the plan thing didn't happen, right?

modules/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/ViewRestrictionsRepo.scala
```diff
@@ -0,0 +1,312 @@
...
+    val comment     = variable("comment")   // bound iff the value carries a valueHasComment literal
+    val creator     = variable("creator")
+    val permissions = variable("permissions")
+
+    var wherePattern = resource
+      .isA(resClass)
+      .andHas(KnoraBase.attachedToProject, Rdf.iri(projectIri.value))
+      .andHas(KnoraBase.isDeleted, Rdf.literalOf(false))
+      .and(resClass.has(zeroOrMore(RDFS.SUBCLASSOF), KnoraBase.Resource))
+      .and(resource.has(RDFS.LABEL, label).optional())
```
BalduinLandolt: I think there's some room for query performance improvement here, but that would need some time to figure out, so let's go with this for now, and look into query improvements in a later round

justfile
```diff
@@ -249,6 +249,17 @@ docs-build: docs-build-dependent docs-ingest-build
...
+# Dump the live OpenAPI spec (served by the running API's docs server) to docs/openapi/openapi.yaml.
+# The frontend (dsp-app / dsp-js) can codegen a typed client from this committed file.
+# Requires a running stack (`just stack-start`); override DOCS_URL for a remote/other-port server.
+dump-openapi DOCS_URL="http://localhost:3339/docs/docs.yaml":
+    #!/usr/bin/env bash
+    set -euo pipefail
+    mkdir -p docs/openapi
+    echo "Fetching OpenAPI spec from {{DOCS_URL}} ..."
+    curl -fsS "{{DOCS_URL}}" -o docs/openapi/openapi.yaml
+    echo "Wrote docs/openapi/openapi.yaml ($(wc -l < docs/openapi/openapi.yaml | tr -d ' ') lines)"
```
BalduinLandolt: having this target here makes perfect sense to me; but could we git-ignore the generated file rather than checking it in? I assume this is only needed for local development, right?

modules/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/ViewRestrictionsRepo.scala
```diff
@@ -0,0 +1,312 @@
...
+
+object ViewRestrictionsRepo extends QueryBuilderHelper {
+
+  val layer = ZLayer.derive[ViewRestrictionsRepo]
+
+  /**
+   * Max rows scanned per query — the large-project guardrail (PRD §7/§9). A blunt cap in v1; a
+   * configurable value / cache is the follow-up. Kept generous so real projects are not truncated.
+   */
+  val ScanCap: Int = 10000
```
BalduinLandolt: I think it would be reasonable to make this configurable from the start, and just set the default to 10k

BalduinLandolt (changes requested):
Three changes I would like to request:

- If the openAPI spec file is only needed for local development, I'd prefer this to be git ignored, rather than checked in. We already have a checked-in copy in the app repo, and I don't see why we should have another copy here.
- PRDs should not live in the code repo, but in dasch-specs. (this used to be done differently - I think this happened due to the old version of the plugin you used.)
- Some tests for the query generator would be nice.

Other than that, things look very reasonable to me.  
(due to the large OpenAPI file, Claude review seems to fail on this PR; I'm running it locally right now.)

I added some remarks to the PRD, maybe some of it is relevant, probably not all of it, though.

One more thing I realized only now: If comments on values don't have their own permissions, does it actually make sense to have the comments thing separate? I'd leave it as is for now, because we signed off the design like this. But I'd like to discuss this for a second iteration. Maybe comments should be removed from this entirely.

BalduinLandolt:
My local Claude review (which hat problems with the big OpenAPI spec file too) found the following things:

---

# Code Review — PR #4239: view-restrictions reporting API (DEV-6778)

**Scope:** `feature/dev-6778…` vs `main`, Scala/ZIO admin slice.
**Result:** 0 Critical · 4 Warning · 3 Suggestion. No merge-blockers; several spec-compliance gaps.
Findings below survived independent adversarial verification.

---

## Warnings

### W1 · `groupBy=Property` + `itemType=Resource` returns an empty report (violates AC7)

- **File:** `modules/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/ViewRestrictionsRepo.scala:61-68`
- **What:** In `findRestrictedObjects`, `wantResources` is false in property mode (requires
  `groupBy == ResourceClass`) **and** `wantValues` is false for `itemType == Resource` (not in
  `{All, File, Value, Comment}`). Both queries are skipped.
- **Effect:** `GET …/view-restrictions/summary?groupBy=Property&itemType=Resource` (and the `items`
  route) returns zero groups / an empty page even when the project has restricted items. PRD AC7
  requires `itemType=resource` to behave as `all` in property mode.
- **Fix:** When `groupBy == Property`, coerce `Resource → All` (set `wantValues` true).

### W2 · `restrictedView` reported for any file value, not only images (violates AC10)

- **File:** `modules/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsService.scala:99`
- **What:** The repo binds `?fileClass` for any `rdfs:subClassOf* knora-base:FileValue`, so
  audio/PDF/video/archive values are also `ItemType.File`. `rvEligible = (row.itemType == ItemType.File)`
  then maps permission code 1 to `Visibility.RestrictedView` for them.
- **Effect:** (1) Drill-down emits `visibility=restrictedView` on non-image files — AC10 forbids this
  (RV is renderable for images only). (2) In the summary (which counts Hidden only), such a file
  contributes 0 instead of collapsing to Hidden, undercounting per AC3.
- **Fix:** Gate `rvEligible` on the image file type specifically (e.g. bind
  `knora-base:StillImageFileValue` in the query rather than the generic `FileValue`).

### W3 · Unvalidated `group` query param → HTTP 500 (should be 400) + SPARQL-injection surface

- **File:** `modules/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/ViewRestrictionsRepo.scala:237,302`
  (param declared at `…/api/admin/ViewRestrictionsEndpoints.scala:40`)
- **What:** `group` is `query[String]("group")` with no validator, passed untouched to `Rdf.iri(g)`.
  rdf4j does not validate/escape the string.
- **Effect:** A value containing a space/`>` (e.g. `?group=foo bar`) yields a malformed query that
  Fuseki rejects; the failure is not a typed endpoint error, so the client gets **HTTP 500 instead of
  400**. Because the string is unescaped it is also a read-only, admin-gated SPARQL-injection surface.
  *(Note: the reviewer's cross-project data-exfiltration claim was refuted — the query structure
  prevents it — but the 500 + missing validation is real.)*
- **Fix:** Accept as a typed/validated IRI (value object or `IriDto` + `IriConverter`) and map
  conversion failure to `BadRequestException` at the API boundary, as the sibling `projectIri` does.

### W4 · `items()` discards `result.capped` → silent truncation, wrong total (violates AC12)

- **File:** `modules/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsService.scala:158,187`
- **What:** `items()` reads `result.rows` but ignores `result.capped`. When a single group exceeds
  `ScanCap` (10000) the query truncates.
- **Effect:** `total = ordered.size` under-reports, resources past row 10000 disappear from every page,
  and there is no `approximate` signal to the client — whereas `summary()` does propagate it for the
  same data. Inconsistent with AC12.
- **Also:** The class-doc comment at `ViewRestrictionsService.scala:33-37` is **stale** — it describes
  the large-project guardrail as a pending follow-up, but PRD §0 marks it resolved and the code
  implements it.
- **Fix:** Surface `capped` on the paged response (or otherwise signal truncation), and refresh the comment.

---

## Suggestions

### S1 · Unknown project returns 403, not 404 (AC5)

- **File:** `modules/webapi/src/main/scala/org/knora/webapi/slice/api/admin/service/ViewRestrictionsRestService.scala:35,47`
- **What:** `ensureSystemAdminOrProjectAdminById` does `findById(...).someOrFail(ForbiddenException)`,
  so a well-formed but non-existent `projectIri` yields 403 even for a system admin. AC5 requires 404
  for an unknown project.
- **Note:** This matches the shared admin-API convention, so it may be an intentional deviation —
  worth a deliberate decision rather than a silent one.

### S2 · OpenAPI example uses the external namespace for `propertyIri`

- **File:** `modules/webapi/src/main/scala/org/knora/webapi/slice/api/admin/ViewRestrictionsEndpoints.scala:225`
- **What:** The `RestrictedItem` example sets
  `propertyIri = "http://api.knora.org/ontology/knora-api/v2#hasStillImageFileValue"` (external), but
  the endpoint actually emits the **internal** `knora-base` IRI. The same example's `resourceClassIri`
  correctly uses the internal namespace.
- **Fix:** Use the internal IRI the endpoint really returns, so client codegen consumers aren't misled.

### S3 · Missing test coverage for AC5, AC7, AC12

- **Files:** `modules/webapi/src/test/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsServiceSpec.scala:317`;
  `modules/test-e2e/src/test/scala/org/knora/webapi/slice/api/admin/AdminViewRestrictionsE2ESpec.scala`
- **What:** The property-mode suite only covers `itemType=All`. No test for `groupBy=Property` +
  `itemType=Resource` (AC7 — this gap is why W1 went unnoticed), no summary/items consistency test
  (AC12), and no E2E coverage of malformed-IRI 400 / unknown-project 404 (AC5).
- **Fix:** Add these cases; the AC7 case should be added alongside the W1 fix.

---

## Refuted (not reported as findings)

- **Perf:** `ScanCap` LIMIT doesn't bound triplestore work; in-memory pagination is O(group size)
  per request; `parsePermissions` runs 3×/row — all produce correct output and are largely documented
  v1 follow-ups.
- **Security:** cross-project data exfiltration via the `group` injection — not achievable given the
  query structure (the injectable surface itself is captured in W3).
- **Consistency:** `Vocabulary` `linkValue`/`LinkValue` duplicate — both resolve to the correct IRI.
  *(Aside: the pre-existing lowercase `linkValue` points at the `LinkValue` class, not a `linkValue`
  property — a latent oddity, but pre-existing and unused by this PR.)*

modules/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsService.scala
```diff
@@ -0,0 +1,198 @@
...
+      PermissionUtilADM.getUserPermissionADM(
+        entityCreator = row.creator,
+        entityProject = projectIri.value,
+        entityPermissionLiteral = row.permissions,
+        requestingUser = audienceUser(audience, projectIri),
+      ),
+      // Only still-image file values can be served in restricted view; RV on anything else — including
+      // non-image file values (audio/PDF/video), resources, ordinary values and comments — means
+      // "not fully visible" and collapses to Hidden. See AC10.
+      rvEligible = row.isImageFile,
```
BalduinLandolt: This here is the problematic line: this applies `rvElegible` only, if the thing at hand is an image file value.

modules/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsService.scala
```diff
@@ -0,0 +1,198 @@
...
+   * partial-view rendering, so code 1 collapses to [[Visibility.Hidden]] (the object is simply not fully
+   * visible to that audience) rather than showing a misleading "restricted view" state.
+   */
+  def visibilityOf(permission: Option[Permission.ObjectAccess], rvEligible: Boolean): Visibility =
+    permission.map(_.code).getOrElse(0) match {
+      case 0                                                                   => Visibility.Hidden
+      case c if c == Permission.ObjectAccess.RestrictedView.code && rvEligible => Visibility.RestrictedView
+      case c if c == Permission.ObjectAccess.RestrictedView.code               => Visibility.Hidden
+      case _                                                                   => Visibility.Visible
+    }
```
BalduinLandolt, line 127: Here it happens (especially line 86): because `rvElegible` can only be true for images, all other types (Resources, regular Values, file Values that are not images) will be displayed as "hidden", even though they have actually RV set.

This results in a dashboard that displays other permissions than we actually have in the DB, which isn't the point of having the dashboard.

BalduinLandolt (changes requested):

BalduinLandolt:
@claude review

BalduinLandolt (approved):

PR #4247, opened 2026-08-06
feat(dashboards): bring the Backend dashboard under Git Sync, fixing two dead panels

BalduinLandolt:
Stacked on #4246 — review that one first.

## Motivation

Second of the dashboards moving into `grafana-dashboards/` under the ownership split (dev-team-owned dashboards live here; host/fleet dashboards stay with ops). "Backend" covers dsp-api, dsp-ingest and the Fuseki triplestore, all three of which are built from this repo.

It also demonstrates why the move is worth doing. Two of its eleven panels have been rendering nothing, and nothing outside Grafana could show that:

- **Tapir requests** plotted 2xx with `service="DSP_svc_api"` but non-2xx with `service="api"`. `DSP_svc_api` is the *only* value of that label on `tapir_request_duration_seconds_count`, so the error series was always empty — and an empty series draws a flat zero line that reads as "no errors". Fixed; it now reports 0.0125 req/s on prod.
- **API JVM Memory** queried `jvm_memory_bytes_used` / `jvm_memory_bytes_max`, which do not exist in Prometheus at all. Renamed to the live `jvm_memory_used_bytes` / `jvm_memory_max_bytes` that the INGEST and DB panels on the same dashboard already use. Now reports 362 MB used against a 2.8 GB max on prod.

`fuseki_request_duration` and its `isGravsearch` / `isMaintenance` tags come from this repo — `TriplestoreServiceLive.scala` even carries a comment reading *"existing dashboards keep isGravsearch/isMaintenance"*, i.e. the code is already constrained by a dashboard it cannot see. This PR makes that link visible.

## Summary

- `dsp-api/dsp-backend.json`, uid preserved as `bdpr1ptxz49hcd`, converted to the v2 schema.
- The **Health Status library panel is inlined** as a regular panel. Git Sync cannot version a library element, so a provisioned dashboard referencing one carries a dependency this repo can neither protect nor review. It reported `connectedDashboards: 1`, so nothing else loses a shared panel.
- Panel descriptions written for all eleven panels — every one was empty.

## Validation

- JSON parses; all 11 elements have exactly one layout reference, none dangling, none unplaced, no duplicate panel ids.
- All 15 expressions executed against `grafanacloud-prom` on `dasch-vre-prod-01`; every one returns data, including both fixed panels and the inlined health panel (39 domain/stack rows).

## Deploy note

`bdpr1ptxz49hcd` must be **deleted in Grafana before this merges** — Git Sync cannot take over a uid held by an unmanaged dashboard, and it fails the sync pass rather than skipping the file. The dashboard will be briefly absent between the delete and the merge.

## Not included

**Route usage metrics** was going to be in this batch and is being dropped. It is obsolete three ways over: it reads `service="traefik_traefik"`, which now only exists on three non-DSP boxes (DSP hosts moved to `service="traefik"`); Traefik now emits **JSON**, not the CLF the `pattern` stage parses; and routers are named `vre-prod-01-api@docker`, not the `dsp-*-api@docker` the query matches. Every panel returns nothing. That is a rewrite, not a migration, and route usage is probably better sourced from `tapir_request_total` — which has per-route counts and far longer retention than Loki.

### DIFF
--- a/grafana-dashboards/dsp-api/dsp-backend.json
+++ b/grafana-dashboards/dsp-api/dsp-backend.json
@@ -1,0 +1,1212 @@
+{
+  "apiVersion": "dashboard.grafana.app/v2",
+  "kind": "Dashboard",
+  "metadata": { "name": "bdpr1ptxz49hcd" },
+  "spec": {
+    "title": "Backend",
+    "description": "Health and resource overview for the three JVM services of a DSP stack: dsp-api, dsp-ingest and the Fuseki triplestore, selected by the Instance variable. Fuseki request durations come from dsp-api's own instrumentation (fuseki_request_duration, tagged type/isGravsearch/isMaintenance/isSearch in TriplestoreServiceLive), not from Fuseki itself.",
+    "tags": ["dsp-api", "dsp-ingest", "fuseki"],
+    "cursorSync": "Crosshair",
+    "editable": true,
+    "preload": false,
+    "links": [],
+    "annotations": [
+      {
+        "kind": "AnnotationQuery",
+        "spec": {
+          "builtIn": true,
+          "enable": true,
+          "hide": true,
+          "iconColor": "rgba(0, 211, 255, 1)",
+          "legacyOptions": { "type": "dashboard" },
+          "name": "Annotations & Alerts",
+          "query": {
+            "datasource": { "name": "-- Grafana --" },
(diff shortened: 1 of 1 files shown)


BalduinLandolt:
Folding into #4246 — keeping the dashboard migration as a single PR.

PR #4251, opened 2026-08-12
fix(admin): make the view-restrictions endpoints usable on real projects (DEV-6778)

derschnee68:
## What

The view-restrictions endpoints added in #4239 **time out on real projects**. Reported from the dev
stack as:

```
ERROR org.knora.webapi.store.triplestore.impl.TriplestoreServiceLive
Triplestore timed out while sending a response, after sending statuscode 200.
```

This makes them usable. Counts are unchanged — the endpoint returns exactly what it did before, faster.

## Why it was slow

Two `rdfs:subClassOf` patterns lived in the shared query skeletons (`resourceCore` / `valueCore`) and
were therefore evaluated **per result row** — and rows scale as resources × properties × values:

- `?resClass rdfs:subClassOf* knora-base:Resource` — restricting the grouping key to resource classes.
- `FILTER NOT EXISTS { ?resource a ?sub . ?sub rdfs:subClassOf+ ?resClass }` — keeping only the most
  specific asserted class.

Measured in isolation on the value-count query against `incunabula` (a *small* project, 4231 resources):

| Query variant | Time |
| --- | --- |
| Neither pattern | 0.78s |
| `+ subClassOf* Resource` | 2.10s |
| `+ most-specific-class filter` | 2.46s |
| **Both (what shipped)** | **4.09s** |

A summary request issues several such queries, sequentially — roughly 17 round-trips.

## What changed

**Resolve the class hierarchy once per request** (`ProjectClasses`) instead of per result row:

- The project's asserted resource classes are discovered once and bound with a `VALUES ?resClass`
  clause, replacing the `subClassOf*` guard.
- The most-specific-class filter is emitted **only** when the project actually asserts a class
  together with a strict subclass of it — the only case where it can change the answer, since
  `?resource a ?class` matches asserted types only.

The class list is deliberately the classes **present in the project's data** (6 for `incunabula`, 19
for `anything`), *not* the `Resource` closure from the ontology. `docs/development/dsp-api-sparql-queries.md`
records that inlining a large closure as `VALUES` is catastrophic (DEV-6803: 2.1s → >60s); a project's
asserted classes are a small set that anchors the scan, which is the shape that doc endorses.

**Parallelise independent reads** — the six per-(audience, state) counts and the class population, and
the drill-down's total and page. Bounded at 4 concurrent so one request cannot saturate the store.

## Results

Measured over all 6 local projects × 2 `groupBy` × 5 `itemType` = 60 requests:

| | Before | After |
| --- | --- | --- |
| Total | 39.59s | **22.89s** |
| Worst case (`incunabula`, ResourceClass/All) | 5.53s | **2.71s** |
| Responses identical to baseline | — | **60/60 byte-identical** |

## Review

A local review pass at high effort found a correctness bug in the first commit and nine smaller
issues; the second commit fixes them. The one that mattered:

> The multi-typed probe was **not equivalent** to the filter it gates. The filter's `?subClass` is
> unconstrained, but the probe compared only classes from the discovered list (which is filtered to
> resource classes), and a `sizeIs < 2` short-circuit skipped it entirely for single-class projects.
> A resource typed with a subclass outside that list would have been **counted twice** — on an
> endpoint contracted to return exact counts.

The probe now mirrors the filter exactly, including the `isDeleted false` guard the real queries use.
Also fixed: an empty class list no longer drops the resource-class guard; `withValues` requires a
`WHERE` clause rather than splicing into an arbitrary brace; the parallel fan-out is bounded.

## Testing

- `ViewRestrictionsQuerySpec` — the class-resolution suite is **golden snapshots** (per `REVIEW.md`
  and the SPARQL doc) rather than `contains`/`!contains` assertions, covering the single-typed,
  multi-typed, empty-class-list, discovery, probe, value-count and drill-down query shapes. The
  golden files make clause placement reviewable in the diff — and would have caught the bug above.
- `ViewRestrictionsServiceSpec` — the repo-level tests now resolve `ProjectClasses` from the real
  fixture data, so the discovery step is exercised end to end.
- Verified against the local stack: all 60 summary responses byte-identical to the pre-change
  baseline, `AdminViewRestrictionsE2ESpec` (15 tests) green.

## Notes / follow-ups

- **Known gap, deliberately left:** the class list is discovered in a separate read from the counts,
  so a class first asserted mid-request is missed. Closing it needs a single snapshot across a
  request, which is the cost this change removes. The window is milliseconds on an admin reporting
  dashboard, and the pre-existing total/page split already has the same property. Worth a decision,
  not a blocker.
- `subPropertyOf*` in `valueCore` is deliberately untouched — removing it measured **168s** vs 0.71s.
  It is load-bearing, not a missed optimisation.
- Numbers here are from local test data. The original timeout came from a larger project, so the
  real-world win should be bigger but is unmeasured.

### DIFF
--- a/modules/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsService.scala
+++ b/modules/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsService.scala
@@ -181,19 +181,31 @@ final case class ViewRestrictionsService(
 
   def summary(projectIri: ProjectIri, groupBy: GroupBy, itemType: ItemType): Task[ViewRestrictionsSummary] =
     for {
-      literals <- repo.distinctPermissions(projectIri, itemType, groupBy)
+      // Resolved once and reused by every query below: the project's asserted resource classes, which
+      // replace two per-row `rdfs:subClassOf` traversals in the queries themselves.
+      classes  <- repo.projectClasses(projectIri)
+      literals <- repo.distinctPermissions(projectIri, itemType, groupBy, classes)
       // One count per (audience, state). Both are tiny fixed sets — 3 x 2 — and each classification pass
       // runs over the small distinct-literal set, so this stays cheap regardless of project size.
-      counted <- ZIO.foreach(for (a <- Audience.ordered; s <- countedStates) yield (a, s)) { case (audience, state) =>
-                   val hits = literalsResolvingTo(literals, state, audience, projectIri)
-                   repo.countByGroup(projectIri, groupBy, itemType, hits).map(rows => (audience, state, rows))
-                 }
-      // Every resource class the project has, with its population — independent of any restriction and of
-      // the itemType filter. In class mode this is the row set itself (see below), so a class is reported
-      // with its true size whether or not anything in it is restricted. A property has no resource
-      // population of its own, so property mode has no equivalent and reports none.
-      classTotals <- if (groupBy == GroupBy.ResourceClass) repo.totalResourcesByClass(projectIri)
-                     else ZIO.succeed(Seq.empty)
+      //
+      // Issued in parallel, together with the class population below: the six counts are independent
+      // read-only aggregations over disjoint permission literals, so serialising them just added up their
+      // latencies (the summary was ~17 sequential round-trips).
+      countedAndTotals <-
+        ZIO
+          .foreachPar(for (a <- Audience.ordered; s <- countedStates) yield (a, s)) { case (audience, state) =>
+            val hits = literalsResolvingTo(literals, state, audience, projectIri)
+            repo.countByGroup(projectIri, groupBy, itemType, hits, classes).map(rows => (audience, state, rows))
+          }
+          // Every resource class the project has, with its population — independent of any restriction and
+          // of the itemType filter. In class mode this is the row set itself (see below), so a class is
+          // reported with its true size whether or not anything in it is restricted. A property has no
+          // resource population of its own, so property mode has no equivalent and reports none.
+          .zipPar(
+            if (groupBy == GroupBy.ResourceClass) repo.totalResourcesByClass(projectIri, classes)
+            else ZIO.succeed(Seq.empty),
+          )
+      (counted, classTotals) = countedAndTotals
       // groupId -> per-audience counts. Contributions accumulate per unit, so a group's resource count and
       // its value count stay separate all the way to the response.
       byGroup = counted.foldLeft(Map.empty[String, AudienceCounts]) { case (acc, (audience, state, rows)) =>
@@ -301,15 +313,23 @@ final case class ViewRestrictionsService(
     for {
       // Paging happens in SPARQL: `rows` are already exactly the resources of the requested page, and
       // `total` is an exact COUNT over the whole result set — no scan cap, no Scala-side slicing.
-      total <- repo.countRestrictedResources(projectIri, groupBy, itemType, group)
-      rows  <- repo.findRestrictedObjects(
-                projectIri,
-                groupBy,
-                itemType,
-                group,
-                offset = pageAndSize.size * (pageAndSize.page - 1),
-                limit = pageAndSize.size,
-              )
+      classes <- repo.projectClasses(projectIri)
+      // The page total and the page itself are independent reads over the same snapshot, so they go out
+      // together rather than one after the other.
+      totalAndRows <- repo
+                        .countRestrictedResources(projectIri, groupBy, itemType, group, classes)
+                        .zipPar(
+                          repo.findRestrictedObjects(
+                            projectIri,
+                            groupBy,
+                            itemType,
+                            group,
+                            offset = pageAndSize.size * (pageAndSize.page - 1),
+                            limit = pageAndSize.size,
+                            classes,
+                          ),
+                        )
+      (total, rows) = totalAndRows
       // Assemble one RestrictedResource per affected resource, with its restricted parts nested under it.
       byResource = rows.groupBy(_.resourceIri).toSeq.map { case (resourceIri, resourceRows) =>
                      val head            = resourceRows.head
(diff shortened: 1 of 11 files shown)


derschnee68:
@claude review

modules/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/ViewRestrictionsRepo.scala
```diff
@@ -193,20 +236,27 @@ final case class ViewRestrictionsRepo(
...
 
   private def runResourceQuery(
     projectIri: ProjectIri,
     group: String,
     resourceIris: Seq[String],
+    classes: ProjectClasses,
   ): Task[Seq[RestrictedObjectRow]] =
     triplestore
-      .query(Select(ViewRestrictionsRepo.resourceQuery(projectIri, Some(group), resourceIris)))
+      .query(select(ViewRestrictionsRepo.resourceQuery(projectIri, Some(group), resourceIris, classes), classes))
```
domsteinbach, line 259: this is imho potentially annoying:
"Select()" -> class
vs. new
"select()" -> helper

modules/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/ViewRestrictionsRepo.scala
```diff
@@ -234,9 +284,12 @@ final case class ViewRestrictionsRepo(
     groupBy: GroupBy,
     itemType: ItemType,
     resourceIris: Seq[String],
+    classes: ProjectClasses,
   ): Task[Seq[RestrictedObjectRow]] =
     triplestore
-      .query(Select(ViewRestrictionsRepo.valueQuery(projectIri, Some(group), groupBy, resourceIris)))
+      .query(
+        select(ViewRestrictionsRepo.valueQuery(projectIri, Some(group), groupBy, resourceIris, classes), classes),
```
domsteinbach, line 291: I would rename the helper into selectWithClasses
or toSelect()
To me having an uppercase class
Select() and a lowercase "select()" is not optimal

domsteinbach (approved):

mpro7 (commented):
Not blocking — the measurements carry this and byte-identical responses are the right bar. One structural note for a follow-up, plus one optional nit.

**The two request-scoped constants have no home.** `ProjectClasses` now rides on 23 signatures in `ViewRestrictionsRepo.scala` — and `projectIri` rides on the same 23. This PR didn't create that; it added a second constant to a namespace of static builders that already threaded one. Worth fixing as a constructor, not with `using`/`FiberRef`.

Three things the current shape leaves to convention, which a per-request builder closes structurally:

1. **The `VALUES` splice can be forgotten.** Correctness depends on pairing each builder with `select(query, classes)` (line 113) at every call site. When `iris.nonEmpty`, `resClassPatterns` deliberately emits no class guard (line 400) — so a future method calling `triplestore.query(Select(someNewQuery(iri, classes)))` directly leaves `?resClass` completely unconstrained and silently counts value classes into the totals. Five call sites currently have to remember; nothing in the types enforces it.
2. **`variable("resClass")` is coupled by string.** `select` re-creates the variable (line 114) separately from the builders that consume it. Rename it inside one builder and the `VALUES` clause binds a phantom variable — same silent inflation, no failing test.
3. **The goldens pin a reconstruction, not what's sent.** `ViewRestrictionsQuerySpec` repeats the two-step dance itself (`withValues(resourceCountQuery(...), singleTyped.valuesClause(resClassVar))`), manufacturing its own `resClassVar`. If the builders rendered their own query, the goldens would pin the exact string the repo sends and that boilerplate would go.

Weaker, and latent rather than live: nothing ties a `ProjectClasses` to the project it was resolved from, so across 23 signatures the pairing is convention. Not reachable today — both entry points have a single `projectIri`.

Shape I'd suggest:

```scala
final case class ViewRestrictionsQueries(projectIri: ProjectIri, classes: ProjectClasses)
    extends QueryBuilderHelper {
  private val resClass = variable("resClass")            // owned here, not re-created by string elsewhere
  private def render(q: SelectQuery): Select =           // every builder exits through this
    Select(withValues(q, classes.valuesClause(resClass)))

  def resourceCount(perms: Set[String]): Select = render { ... }
  def valueCount(groupBy: GroupBy, itemType: ItemType, perms: Set[String]): Select = render { ... }
}

// repo.forProject(iri): Task[ProjectReader] — resolves ProjectClasses once, then:
reader.countByGroup(groupBy, itemType, hits)             // no projectIri, no classes
```

`ProjectClasses` then leaves the repo's public API, `ViewRestrictionsService`, and `ViewRestrictionsServiceSpec` (which currently resolves it from fixtures only to pass it back in). The existing golden files are the proof the move is behaviour-preserving — they should come out byte-identical.

Not `(using ProjectClasses)`: the codebase uses `using`/`given` only for typeclass instances (codecs, `Ordering`), it doesn't reduce the signature count — only its visual weight — and it hides exactly the data flow that hazard 1 depends on. Not the ZIO environment or a `FiberRef` either: neither is readable from the pure builders, where the parameter is densest.

**Optional nit for this PR** (~3 lines, kills hazard 2 on its own): hoist the `resClass` variable into one `private val` shared by `select` and the builders, instead of re-creating it by name in both places.

**Separate follow-up, likely a bigger win than this PR:** `docs/development/dsp-api-sparql-queries.md:386` says aggregation inputs and class extents are what needs `GRAPH` scoping (55x measured on a count query), and that `GRAPH <projectDataGraph>` *replaces* `?s knora-base:attachedToProject <p>` (5.4x, DEV-6827). These are aggregations running unscoped with that join. Caveat that makes it non-trivial: here `attachedToProject` is also what keeps values out of `resourceCore` — values carry `attachedToUser`/`hasPermissions`/`isDeleted` but not `attachedToProject` — so the class guard has to stay if the join goes. Also worth considering: caching `ProjectClasses` per project with a short TTL removes the two discovery round-trips this adds, and the PR already accepts a millisecond staleness window for that exact value.

modules/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsService.scala
```diff
@@ -301,15 +318,24 @@ final case class ViewRestrictionsService(
...
-      rows  <- repo.findRestrictedObjects(
-                projectIri,
-                groupBy,
-                itemType,
-                group,
-                offset = pageAndSize.size * (pageAndSize.page - 1),
-                limit = pageAndSize.size,
-              )
+      classes <- repo.projectClasses(projectIri)
+      // The page total and the page itself are independent reads, so they go out together rather than one
```
siers, line 322: This whole comment is unnecessary. 

The `par` prefix is standard lexicon, obviously they are executed in parallel. 

At most you could write that there could be a race condition here, but that was the case before, so what does matter?

Half of the other comments in this PR are as useless.

PR #4257, opened 2026-08-18
build: fix the release-publish scalac worker crash (8 MB stack + run publish over RBE)

mpro7:
## Motivation

The last two release publishes failed: **v37.7.0** (run 32034161094) and **v37.6.0** (run 31381168141). Both died in the same step with the same signature — the Bazel scalac worker crashing while compiling `//modules/webapi:webapi`:

```
ERROR: modules/webapi/BUILD.bazel:56:14: scala @@//modules/webapi:webapi failed:
       Worker process did not return a WorkResponse
error: recipe `docker-publish-dsp-api-image` failed with exit code 1
```

Neither release got its images from the release workflow: v37.6.0 was covered by the concurrent `Docker publish images` run on the same commit, and v37.7.0 needed a manual `Docker Publish from branch` dispatch ~18 h later. It is not release-specific either — the same crash hit 4 of the last 8 `Docker publish images` runs on main (first occurrence 2026-07-28).

## Summary

This PR fixes both the symptom and the exposure:

- **The fix (symptom):** adds `-Xss8m` to `scalac_jvm_flags` in `bazel/toolchains/BUILD.bazel`, restoring (with headroom) the compiler stack size sbt used to set.
- **The fix (exposure):** runs the three Docker publish workflows over the NativeLink remote executor, so the compile and OCI image assembly run remotely — where `:webapi` has never crashed and where publish runs reuse PR CI's cached compile — instead of on a local scalac worker on every publish.

### bazel/toolchains/BUILD.bazel
- `scalac_jvm_flags` becomes a two-element list: the existing OTel system property plus `-Xss8m`, with a comment recording the failure mode so the flag is not pruned as unexplained.

### .github/actions/bazel-rbe + the publish workflows
- The three publish workflows previously ran the `bazel-rbe` action with `stage: cache` (remote cache only, local execution); `build-and-test.yml` already ran `stage: full` (remote executor). With every call site now on the executor, the `stage` input is removed entirely — the action always emits the remote cache **and** executor flags. No `justfile`, `BUILD.bazel`, or `.bazelrc` change is needed: `--remote_download_toplevel` is already set so `bazel run //modules/*:push` gets its image bytes locally.

### docs/development/dsp-api-rbe.md
- Rewritten to current state (no staged-rollout history): every RBE-using CI job runs the remote cache + executor, including the publish workflows, with `-Xss8m` as the `--remote_local_fallback` safety net.

### The worker crash is a compiler StackOverflowError, not a compile error

**Problem:** the printed worker log is ~1000 frames of one repeating cycle in dotty's `staging` phase (`CrossStageSafety.transform` → `TreeMapWithStages.transform` → `TreeMapWithPreciseStatContexts.transformStats/loop`). Bazel truncates the exception header, so the throwable class is not in the log — but ~1024 repeated frames is exactly the JVM's default `MaxJavaStackTraceDepth`, which makes `StackOverflowError` the only fit. `staging` recurses once per statement while walking blocks, so the stack depth it needs scales with module size, and `webapi` has grown past what a default (~1 MB) thread stack supports.

**Tried:** ruling out a code defect and a resource limit. The same commit compiles both ways: for v37.7.0, `Docker publish images` (started 13:15:36) compiled `:webapi` in a local worker for 220 s and passed, while `Publish-release` (started 13:16:02, identical tree) crashed at 130 s. Identical inputs, opposite outcomes — JIT inlining shifts frame sizes run to run, which is how a stack that sits right at the limit behaves. No `OutOfMemoryError` or GC-overhead output anywhere in the logs.

**Solution:** `git log` shows where the setting went. `sbtx` (removed in #4228) launched with `default_jvm_opts_common="-Xms512m -Xss2m -XX:MaxInlineLevel=18"`; the Bazel toolchain only ever carried the OTel system property, so the migration silently halved the compiler's stack. `-Xss8m` restores it with 4× the margin sbt had. rules_scala 7.2.6 passes `scalac_jvm_flags` to the scalac launcher as `--jvm_flag=…` (`scala/private/rule_impls.bzl`), which lands on the worker JVM's command line in both worker and non-worker mode.

### Why PR CI never saw it — and why `stage: full` removes the exposure

**Problem:** `CI-build-and-test` compiles the same target on every PR and has never hit this.

**Cause:** it uses the `bazel-rbe` action with `stage: full`, so `:webapi` is executed on the NativeLink executor (`remote, remote-cache` in its logs). `docker-publish.yml`, `publish-release.yml` and `publish-from-branch.yml` used `stage: cache` — remote cache only, local execution (`remote-cache, worker`). Only the local worker path was exposed. Because exec platform properties are part of the action key, the publish runs also could not reuse CI's remotely-executed result, so they recompiled `:webapi` locally on *every* publish and rolled the dice each time.

**Solution:** run the publish workflows over the remote executor. The compile and image assembly then run remotely (arch-neutral JVM bytecode — no cross-compile need), and a publish on a main commit reuses the action-cache entries PR CI already produced for that commit. Only `bazel run //modules/*:push` runs locally to reach the registry; `--remote_download_toplevel` (already in `.bazelrc`, `build --remote_download_toplevel`) materializes the image tarball on the runner. The `docker-healthcheck` job already runs the analogous `bazel run //modules/*:load` on the executor on every PR, so the "build remote → materialize bytes locally → run locally" path is already proven for OCI targets. `-Xss8m` stays load-bearing: `--remote_local_fallback` degrades to a local compile on a backend outage, which is exactly the crashing path.

## Gotchas

- **Keep an `-Xss` on the scalac toolchain.** The default JVM thread stack is not enough for `webapi` and the requirement grows with the module. If a future change moves compiler JVM flags around, carry `-Xss` with them. It also guards the `--remote_local_fallback` path where the publish compile still runs locally.
- **Keep the publish workflows on the remote executor.** Making them build locally again re-exposes the local-worker compile on every publish. If the `bazel run //modules/*:push` bytes-materialization ever regresses, prefer fixing that over pushing the compile back onto the runner.
- **A green PR proves nothing about the local-worker path.** PR CI compiles remotely; the exposed path was the publish workflows compiling locally.
- **Bazel does not retry a dead worker.** One crash fails the build, so a flake here is a hard release failure, not a slow build.

## Test Plan

- [ ] CI green on this PR (validates the flag is accepted by the launcher, the toolchain still builds, and the workflow YAML is valid; note that PR CI compiles remotely and only healthchecks `:load`, so it does not exercise the publish push path).
- [ ] Merge → `Docker publish images` green on the merge commit. Merging to `main` triggers `docker-publish.yml`, which now compiles on the executor — the `:webapi` compile runs remotely (or is a cache hit), all four `bazel run //modules/*:push` steps push `latest` + the git tag, and `trigger-dev-deployment` fires. This is the first run that actually validates the switch.
- [ ] Next release publish (`Publish-release`) completes without a worker crash.

No automated regression test: the failure is a compiler stack-depth threshold in a Bazel worker, which cannot be asserted from a unit test. The publish path staying green across subsequent runs is the confirming evidence.

## Rollback

If a publish fails because `bazel run //modules/*:push` cannot find the image bytes, make the affected publish workflow(s) build locally again (drop the executor flags for that step) — `-Xss8m` keeps that local compile path safe — then investigate the multi-arch `:index` runfiles-materialization case separately.

### DIFF
--- a/.github/actions/bazel-rbe/action.yml
+++ b/.github/actions/bazel-rbe/action.yml
@@ -4,16 +4,13 @@ description: >
   string (the `flags` output) that each CI step passes to a `just` recipe. Bazel does
   NOT expand env vars in .bazelrc, so the connection + mTLS flags live here (per CI step), not in the
   rc file; the backend-agnostic tuning flags (--remote_download_toplevel, --remote_local_fallback,
-  --repository_cache, …) are in .bazelrc. stage=cache emits remote-cache-only flags; stage=full also
-  adds the remote executor. Flags are empty when mTLS material is absent (fork PR → cold local build).
+  --repository_cache, …) are in .bazelrc. The flags enable the remote cache (AC+CAS) and the remote
+  executor. Flags are empty when mTLS material is absent (fork PR → cold local build).
   dsp-api is single-arch (all runners are x86_64 = the worker arch), so there is no cross-compile /
   target-platform machinery. External archives are cached via --repository_cache + the actions/cache
   step, not a remote downloader.
 
 inputs:
-  stage:
-    description: "cache = remote cache only (local execution); full = also remote executor."
-    default: "cache"
   runner-endpoint:
     description: RE-API endpoint, e.g. grpcs://<host>:50051 — serves the remote cache (AC+CAS) and the remote executor.
     required: true
@@ -38,7 +35,6 @@ runs:
     - id: build
       shell: bash
       env:
-        RBE_STAGE: ${{ inputs.stage }}
         RBE_RUNNER_ENDPOINT: ${{ inputs.runner-endpoint }}
         RBE_CA_CERT: ${{ inputs.ca-cert }}
         RBE_CLIENT_CERT: ${{ inputs.client-cert }}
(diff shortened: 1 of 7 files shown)


subotic (approved):

subotic (approved):
I did some changes. Should be good now.

PR #4258, opened 2026-08-18
feat(ingest): backfill sidecar sizes for existing assets (DEV-6957)

siers:
[DEV-6957](https://linear.app/dasch/issue/DEV-6957/sidecar-file-migration-precompute-file-size)

## Summary

Follow-up to #4255, which added optional `sizeOriginal` / `sizeDerivative` to the asset sidecar (`<assetId>.info`) at ingest time. Those fields are only populated for assets ingested *since* that change — every sidecar already on disk still has them absent. This PR adds the one-off migration that backfills them, other fields (incl. checksums) untouched.

## Changes

**`AssetInfoService.updateSizes(infoFile, original, derivative)`** (new) — reads the sidecar, stamps the two on-disk sizes, writes it back. Every other field, checksums included, is preserved. A file the sidecar references but that is not on disk is logged at WARN and its size left absent, rather than failing the asset.

**`AssetSizeMigrationService`** (new) — walks the asset store, one `AssetFileGroup` per `.info` file, and refreshes each. Original/derivative paths come out of the sidecar rather than being guessed from the filename. Traversal and refresh both run at `StorageService.maxParallelism()`. A sidecar that fails to read or parse is logged and counted, not fatal, so one bad asset cannot strand the rest of the store. Returns an `AssetSizeMigrationReport(found, updated, failed)`.

**`MigrateSizes`** (new) — `ZIOAppDefault` entrypoint, wired as the `//modules/ingest:migrate-sizes` `scala_binary`. Not part of service startup; it reads `STORAGE_ASSET_DIR` the same way the service does, so it must be pointed at the same volume. Exits non-zero when `failed > 0` so partial failures are visible to the caller.

## Running it

Locally, via Bazel:

```bash
bazel run //modules/ingest:migrate-sizes            # whole asset store
bazel run //modules/ingest:migrate-sizes -- 0801    # a single project
```

Against a deployed image, swapping only the main class out of the image's own entrypoint (`java -cp '/sipi/lib/*' swiss.dasch.Main`):

```bash
docker compose run --rm --no-deps --entrypoint java ingest -cp '/sipi/lib/*' swiss.dasch.MigrateSizes
```

`--no-deps` is safe because the migration touches the asset volume and its own config only; it needs no other service. The compose `ingest` service already sets `STORAGE_ASSET_DIR=/opt/images` and mounts the asset volume, so no extra wiring is needed. Append a shortcode to scope it to one project.

Idempotent — a second run recomputes the same sizes. An unknown shortcode fails fast.

## Test Plan

- `bazel test //modules/ingest:test` — full module suite passes, not just the new spec.
- `just fmt` / license headers — clean.
- The `docker compose run` command above, executed against the built image on the local asset store: `Found 6 asset groups` / `Size migration finished: 6 updated, 0 failed of 6`, with a WARN for the one asset whose `.orig` is absent and `sizeDerivative` correctly written into its sidecar.
- New `AssetSizeMigrationServiceSpec`: whole-project refresh with per-group outcomes; unknown project fails; one group discovered per info file with the sidecar's own paths; backfill onto a sidecar lacking both sizes; derivative absent when only the original is on disk; both absent (and each warned) when neither is.

Note the spec's storage layer is provided at suite level, so the asset store is shared across its tests. The project-wide assertions are written as invariants (`updated == found`, `failed == 0`) rather than an exact asset count, so they don't depend on suite ordering.

### DIFF
--- a/modules/ingest/BUILD.bazel
+++ b/modules/ingest/BUILD.bazel
@@ -98,6 +98,14 @@ scala_binary(
     runtime_deps = [":ingest"],
 )
 
+# One-off sidecar size backfill; not part of the service startup.
+scala_binary(
+    name = "migrate-sizes",
+    main_class = "swiss.dasch.MigrateSizes",
+    visibility = ["//visibility:public"],
+    runtime_deps = [":ingest"],
+)
+
 scala_junit_test(
     name = "test",
     srcs = glob(["src/test/scala/**/*.scala"]),
--- a/modules/ingest/src/main/scala/swiss/dasch/MigrateSizes.scala
+++ b/modules/ingest/src/main/scala/swiss/dasch/MigrateSizes.scala
@@ -18,11 +18,16 @@ import zio.*
  * One-off migration backfilling `sizeOriginal`/`sizeDerivative` into every `.info` sidecar in the store.
  * Reads `STORAGE_ASSET_DIR` like the service does, so point it at the same volume.
  *
+ * Locally, via Bazel:
+ *
  * {{{
  *   bazel run //modules/ingest:migrate-sizes            # whole asset store
  *   bazel run //modules/ingest:migrate-sizes -- 0801    # a single project
  * }}}
+ *
+ * Against a deployed image, swapping only the main class out of the image's own entrypoint:
  */
+// docker compose run --rm --no-deps --entrypoint java ingest -cp '/sipi/lib/*' swiss.dasch.MigrateSizes
 object MigrateSizes extends ZIOAppDefault {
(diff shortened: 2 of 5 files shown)


subotic (approved):
Looks good, but how do you want to execute this on the server where you don't have physical access to the files?

siers:
I know a guy in the OPS team. :)

siers:
We tested the running times of size computations and directory traversal. I believe they were quick enough and the script is idempotent enough not to require a migration check.

PR #4260, opened 2026-08-19
perf(admin): trim the view-restrictions permission probes and bound the grouped counts (DEV-6778)

derschnee68:
Follow-up to #4251, which made the view-restrictions endpoints usable on real projects. That PR removed two `rdfs:subClassOf` traversals the triplestore was re-evaluating per result row; this one removes two more patterns that were doing no work, bounds the per-query cost, and pins the invariants the dsp-app page depends on.

No API surface change: endpoints, DTOs, the service layer and `docs/openapi/openapi.yaml` are untouched, so the generated dsp-das client regenerates identically and needs no update.

## Changes

**`perf`: drop the most-specific-class filter from the permission probes.** `distinctResourcePermissionsQuery`/`distinctValuePermissionsQuery` inherited `FILTER NOT EXISTS { … subClassOf+ … }` from the shared cores, but they `SELECT DISTINCT ?permissions` — the filter can only remove rows, and removing rows cannot change which literals survive. `resClassPatterns` takes a `dedupeRows` flag, defaulting to true so the counting and paging queries are unchanged. Measured on a synthetic 46k-resource / 265k-value project with 40% of resources multi-typed (so the filter was actually emitted): **121.2s → 15.8s** on the value probe, result set byte-identical (16 literals, verified by diff).

**`perf`: drop the unused creator join from the permission probes.** The same two queries bound `attachedToUser ?creator` without ever projecting or constraining it — an unconstrained join whose only effect was to multiply intermediate rows before `DISTINCT` collapsed them. `distinctPermissions`' own scaladoc already said the creator is irrelevant here; the query didn't match. **875ms → 720ms (−18%)** on the 5k project. `bindCreator` joins `dedupeRows` as a flag; in both cores the creator keeps its position in the pattern chain, so toggling it cannot reorder the rendered SPARQL of unrelated queries.

**`perf`: chunk the grouped counts by resource class.** Each grouped count was one query over the project's whole class list, and Fuseki applies `query-timeout` (20s) per request — so past some size that single query is cancelled mid-stream regardless of how efficient the per-row work is. `runCountByGroup` now fans out over chunks of 3 classes and `mergeCounts` sums per `(groupId, unit)`.

Please read that commit message before reviewing this one: it is a correctness-preserving restructuring that bounds per-query cost, **not** a throughput win, and it does **not** rescue the large case. At 46k resources / 265k values the summary still takes 40-49s through the API and returns 500 for `itemType=File`/`Value` — one 3-class chunk is ~4.4s idle and 13-17s under the request's own concurrency. Fuseki's log confirms the fan-out is active (133 grouped-count queries where the unchunked form issues ~12); the volume is simply past what a 20s budget absorbs, and the unchunkable `distinctPermissions` probes run first. An earlier draft of that message claimed "8.1s → 1.4s slowest chunk" from hand-run queries against an idle store; that figure does not survive real concurrency and has been retracted in the message.

**`test`: pin the summary invariants dsp-app relies on.** The dsp-app page shows `totals` above a column of per-group `counts` and cross-checks neither, and sums `group.totalResources` client-side for the footer. Two tests assert `totals` equals the fold of the group counts in both grouping modes, with the fold spelled out independently rather than reusing the service's `plus`.

More importantly, a `manyClassesTurtle` fixture with five classes sharing one property. Every pre-existing fixture has at most three classes and `ClassChunkSize` is 3, so **the merge was a no-op in every test**: mutating `mergeCounts` from `_ + _` to `math.max` passed the entire suite. With the new fixture that mutation fails on the property-mode case where one group is populated from every chunk.

## Verification

- `just check` and `bazel test //modules/webapi:test` green.
- Correctness of the chunked merge verified against the triplestore at the chunk size the code uses: three chunks of three classes, summed, reproduce the single-query result exactly for `groupBy=Property`, all 10 property counts identical.
- All 10 summary variants (2 `groupBy` × 5 `itemType`) on `incunabula` return **0.9-2.7s**; drill-down pages in 1.4s.
- Exercised through the dsp-app View restrictions page against a local API built from this branch.

## Known ceiling, not a regression

A 46k-resource / 265k-value project still exceeds `query-timeout`. Closing that needs a longer timeout tier for admin reporting, more Fuseki heap than the local `-Xmx3G`, or a different shape for the `distinctPermissions` probes, which have no group key to chunk on. Worth a separate ticket rather than stretching this PR.

Also worth recording: no project in the local test data asserts a class together with one of its ancestors — all four probe 0 — so the most-specific-class filter never fires on that data and the first commit's 121s → 16s is insurance for a shape that does not currently occur, not a win visible today.

The synthetic-data generator used for these measurements is deliberately **not** included here.

### DIFF
--- a/modules/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/ViewRestrictionsRepo.scala
+++ b/modules/webapi/src/main/scala/org/knora/webapi/slice/admin/repo/ViewRestrictionsRepo.scala
@@ -140,7 +140,7 @@ final case class ViewRestrictionsRepo(
         resourceCounts <-
           if (wantResources)
             runCountByGroup(
-              ViewRestrictionsRepo.resourceCountQuery(projectIri, permissions, classes),
+              ViewRestrictionsRepo.resourceCountQuery(projectIri, permissions, _),
               CountUnit.Resources,
               classes,
             )
@@ -154,7 +154,7 @@ final case class ViewRestrictionsRepo(
                   groupBy,
                   ViewRestrictionsRepo.effectiveItemType(groupBy, itemType),
                   permissions,
-                  classes,
+                  _,
                 ),
               CountUnit.Items,
               classes,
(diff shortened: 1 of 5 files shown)


derschnee68:
@claude review

domsteinbach (approved):
Can you address the reviews issues A and B? 
For the other issues pointed out we need numbers - not claims. The total time needed for a response is now higher for smaller projects, for larger projects it is way smaller. There is no more timeout and the total work which is done is not really higher (except for really small projects). Having 10+ tiny queries against fuseki instead of one huge one is effectively not more work.

siers (approved):
All changes are contained within the ViewRestrictionsRepo that you've been working on. There are some things that I can't follow, but go ahead and try it out

PR #4262, opened 2026-08-20
perf(admin): chunk the grouped value counts by property as well as class

derschnee68:
Follow-up to #4260, which trimmed the permission probes and bounded the grouped counts by chunking them over the project's resource classes. This adds the second chunking axis that work was missing.

No behaviour change: the counts are a pure fan-out and merge of a query the triplestore already answered the same way. The endpoint contract and response shape are untouched.

## Why class chunking alone is not enough

Chunking by resource class bounds each query, but it has a floor: once a chunk is a single class it cannot be subdivided. A project with one very large class still issues a query whose cost grows without bound and eventually exceeds Fuseki's per-request `query-timeout` (20s, `application.conf`), which cancels it mid-stream and fails the whole request. Lowering `ClassChunkSize` cannot reach that case — cost is roughly linear in classes (~120ms fixed + ~100ms per class measured), so smaller chunks only multiply round-trips and total work while leaving the largest single query exactly as slow.

## What changed

The value counts now also chunk by property. `ProjectProperties` discovers the value properties the project's data actually uses — one small anchored query per request, mirroring `ProjectClasses` — and `runCountByGroup` fans out over the cartesian product of class chunks and property chunks.

Binding `?prop` from a `VALUES` table also replaces the `?prop rdfs:subPropertyOf* knora-base:hasValue` traversal that the store would otherwise re-evaluate per candidate row, so each query is cheaper as well as smaller.

`mergeCounts` is unchanged: summing per `(groupId, unit)` already covers the new axis. It stays sound because each counted object falls in exactly one chunk on each axis — a value belongs to one resource, `dedupeRows = true` pins a resource to a single `?resClass` binding, and a value is reached through exactly one property, so property chunks partition the values rather than overlapping them.

Resource-level counts pass an empty `ProjectProperties`, which yields a single chunk: they have no `?prop` to split on. An empty property list keeps the `subPropertyOf*` traversal rather than dropping the constraint — without either, `?prop` would match every predicate on the resource, `rdf:type` and the admin properties included, and inflate every count.

## Measured

Perftest project: 5 classes, 5k resources, 337k triples, on the whole-project value count.

| | |
|---|---|
| one class: traversal → bound properties | 232ms → ~110ms |
| whole-project single query | 654ms |
| 6 chunk queries (3 classes × 4 properties) | slowest **212ms**, sum 946ms |

Peak query time — what the timeout actually budgets — drops 3.1×. Total work rises, as chunking always trades; it is spread over more queries at the existing `MaxConcurrentChunkQueries` cap.

The two-axis fan-out plus sum-merge was verified to reproduce the single query's counts exactly against the perftest data: 10 groups, 6670 values, identical.

## Tests

Seven new tests in a `property chunking` suite mirroring the existing class-chunking one: the partition invariants (no property lost or duplicated, which is what makes summing sound), that an empty list keeps the `subPropertyOf*` fallback, that discovered properties replace the traversal with `VALUES`, that each chunk binds only its own properties, and that both `VALUES` clauses splice into the same `WHERE` block.

#4260 added a `manyClassesTurtle` fixture specifically so that mutating `mergeCounts` from `_ + _` to `math.max` fails the suite — before it, the merge was a no-op in every test. I re-ran that mutation after adding the second axis and it still fails, so the merge coverage survived the change.

`//modules/webapi:test` (1809 tests) and `just check` pass.

## Note

I could not reproduce a timeout locally: the perftest data runs the whole-project query in 0.65s, far inside the budget. This targets the right failure mode and is verified for correctness and relative speedup, but is not verified against a specific failing production request.

`distinctPermissions` is still not chunked at all and is the next place to look if the summary still times out. Separately, the six `(audience, state)` counts re-scan the same rows differing only in the permission `FILTER`; grouping by `?permissions` once and classifying in Scala would cut query volume ~6× and reduce total work rather than redistributing it.

### DIFF
--- a/modules/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsService.scala
+++ b/modules/webapi/src/main/scala/org/knora/webapi/slice/admin/domain/service/ViewRestrictionsService.scala
@@ -183,8 +183,11 @@ final case class ViewRestrictionsService(
     for {
       // Resolved once and reused by every query below: the project's asserted resource classes, which
       // replace two per-row `rdfs:subClassOf` traversals in the queries themselves.
-      classes  <- repo.projectClasses(projectIri)
-      literals <- repo.distinctPermissions(projectIri, itemType, groupBy, classes)
+      classes <- repo.projectClasses(projectIri)
+      // The second axis the grouped value counts are chunked along, so a single large class cannot produce
+      // an unbounded query — see ViewRestrictionsRepo.runCountByGroup.
+      properties <- repo.projectProperties(projectIri)
+      literals   <- repo.distinctPermissions(projectIri, itemType, groupBy, classes)
       // One count per (audience, state). Both are tiny fixed sets — 3 x 2 — and each classification pass
       // runs over the small distinct-literal set, so this stays cheap regardless of project size.
       //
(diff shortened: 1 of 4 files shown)


derschnee68:
@claude review

PR #4268, opened 2026-08-25
feat(grafana): response-duration dashboard — route scoping, Gravsearch trace panel, table fix (DEV-7032)

BalduinLandolt:
Extends the git-synced **DSP-API — Response Duration** dashboard. Linear: [DEV-7032](https://linear.app/dasch/issue/DEV-7032/dsp-api-response-duration-dashboard-gravsearch-slowest-query-trace)

## What changed

**Fix — route-ranking table (panel-6) framing.** The two joined queries rendered as time-series-wide frames (raw label header + `NaN` rows) instead of `method / path / Avg duration / Requests`. Fixed with `format: table` on both queries **and** query `version: v0` (the server silently strips `format` when version is empty). Also guards to routes with >0 requests in range (drops `0/0 = NaN`) and shows **Requests as a count** (`increase`) rather than a per-second rate that rounds to `0.00` for rare routes.

**Scoping / grouping.** Global panels now honour the route/method filters (they previously aggregated over all routes unconditionally). New **Route group** bucket variable and a multi-select **Exclude routes** variable (health/version/export/imports/admin/candelete); both scope every panel. Empty-exclude is safe (PromQL anchors regex matchers).

**Rare-slow-route skew.** Per-route duration graph uses a **log2 y-axis** so a slow-but-rare route (e.g. `/v3/export/resources`) no longer flattens everything else. (An earlier `topk()` was dropped — flickery in a range graph.)

**New panel — Slowest Gravsearch queries (traces).** Tempo (`grafanacloud-dasch-traces`) table of `gravsearch` spans over a duration threshold, sorted slowest-first, scoped by environment/stack. Columns: Duration, Start time, query Shape, schema Predicates, stack, and a **Trace ID** link → trace → `gravsearch` span → Events → `gravsearch.query` (`db.query.text`, the verbatim query).

A dashboard `README.md` and updated panel `description`s document the query-shape rationale (the load-bearing `format`+`version` pairing, the log axis, the exclude anchoring, the gravsearch query-on-event).

## Finding (documented, no code action)

The day/night swing in average duration is a **traffic-mix** effect, not health-check noise: at night ~82% of traffic is fast automated `/v2/resources/*` (~21 ms) and `/v2/node/{listIri}` (~6 ms) reads at steady throughput; daytime layers heavier human-driven search/admin/ontology calls on top. Excluding `/health`+`/version` barely moves it.

## Follow-up (tracked in DEV-7032)

Emit a bounded `gravsearch.project_shortcodes` span attribute (derived from ontology IRIs, a **set** since Gravsearch is cross-project, kept **off** the Alloy spanmetrics dimensions) so panel 7 can show a **Project** column.

## Validation

Built and soak-tested on a live copy dashboard (kept until this merges): https://dasch.grafana.net/d/dsp-api-resp-dur-draft — every panel renders, the table columns/sort are correct, and the trace link opens the trace. Git Sync applies from `main`, so the tracked dashboard updates on merge.

### DIFF
--- a/grafana-dashboards/dsp-api/README.md
+++ b/grafana-dashboards/dsp-api/README.md
@@ -1,0 +1,89 @@
+# DSP-API dashboards
+
+Git-synced Grafana dashboards for DSP-API (folder UID `dsp-api-dashboards`, see `_folder.json`).
+All use the **v2 schema** (`dashboard.grafana.app/v2`): panels under `spec.elements`, positions under
+`spec.layout`, variables under `spec.variables`. `metadata.name` is the dashboard UID and must stay
+stable. Edit the JSON here and push to `main`; the dashboards are read-only in the Grafana UI (see the
+parent [`../README.md`](../README.md) for Git Sync mechanics).
+
+| File | Dashboard | What it answers |
+| --- | --- | --- |
+| `dsp-api-response-duration.json` | DSP-API — Response Duration | Request latency & throughput, global and per route. Detailed below. |
+| `dsp-backend.json` | DSP-API — Backend | JVM / runtime / triplestore-facing backend health. |
+| `dsp-gravsearch.json` | DSP-API — Gravsearch | Gravsearch volume, outcome and duration percentiles by query shape (metrics). |
+| `dsp-route-usage.json` | DSP-API — Route Usage | Which API and dsp-app routes are called, and how often. |
+
+For **how to read a slow Gravsearch trace**, see [`docs/observability/gravsearch-trace-runbook.md`](../../docs/observability/gravsearch-trace-runbook.md)
+and [`traceql-recipes.md`](../../docs/observability/traceql-recipes.md).
+
+## DSP-API — Response Duration
+
+**Purpose:** spot latency regressions and slow endpoints for DSP-API, from the tapir request metrics
+(`tapir_request_*`, `service_name="DSP_svc_api"`). **Audience:** whoever is watching a deploy or
+chasing a slow route.
+
+**Averages, not percentiles.** Every duration figure is `rate(sum)/rate(count)` on `phase="body"`
+(full-request duration). This metric has **no histogram buckets**, so percentiles are not available
+here — use the Gravsearch dashboard / Tempo for tail latency.
+
+**Filters (scope every panel unless noted):**
+
+- `environment`, `stack` — from `label_values` on the metric.
+- `Route group` — a coarse, single-select bucket (`Search`, `Resources`, `Admin`, `Export`, …); the
+  value is a path regex applied as `path=~"${routegroup}"`.
+- `Route (path)`, `Method` — fine multi-select; the per-route panels (5, 6) layer these on top of the
+  route group.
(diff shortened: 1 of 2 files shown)


mpro7 (approved):
Reviewed this while delivering the DEV-6637 Gravsearch baseline ([dasch-specs#181](https://github.com/dasch-swiss/dasch-specs/pull/181)) — no overlap, and three things came back that are useful here.

**1. Your averages agree with the span-metrics pipeline to 0.12%.** Over `2026-08-12` → `2026-08-26`, `tapir_request_duration_seconds` `phase="body"` gives a mean of **0.5864 s** for `/v2/searchextended`; `traces_spanmetrics_latency_*{service="DSP_svc_api"}` gives **0.5871 s** for `POST /v2/searchextended`. Two unrelated instrumentation pipelines, 0.7 ms apart. Worth knowing the tapir choice is sound and that this dashboard won't contradict the baseline of record.

**2. The `Gravsearch duration ≥` picker could be baseline-anchored.** PRD REQ-3.1 asks for a threshold *"expressed relative to the baseline (e.g. > p95 from REQ-5.1) rather than an arbitrary constant"*. Measured over that window, the root `gravsearch` span is **p50 95 ms · p95 1.77 s · p99 4.58 s** — so of `500ms,1s,2s,5s,10s`, **`2s` is the p95-anchored default** and `5s` is roughly the p99 tier. Only a default-value change, and it makes the panel satisfy REQ-3.1 as written.

**3. Your `project_shortcodes` note corrected my PRD amendment — thanks.** I had written that DEV-7031 would make *project* a `topk`-by-shape metric column once the attribute ships. Your README (and DEV-7031 itself) is right that it must be a **set** and stay **off** the Alloy spanmetrics dimension list, so project attribution is permanently per-trace, not a metric dimension. PRD REQ-5.3 now records that it is *met without* a project column rather than waiting on DEV-7031.

**One thing DEV-7031 may be slightly pessimistic about**, if useful for sequencing it: it says the query text lives in an event and so *"can't be `select()`-ed into a column"* — true for a **column**, but the text is **filterable** today. This returns matching spans against prod right now:

```traceql
{span:name="gravsearch" && span.environment = "prod" && event.db.query.text =~ ".*ontology/0806/.*"}
```

So a **project filter** on panel 7 (a `project` dashboard variable interpolated into the regex) is available before DEV-7031 ships — only the rendered *column* needs the span attribute. Two cautions from the same probe: the ontology host differs per environment (`api.dasch.swiss` vs `api.dev.dasch.swiss`), so a shortcode regex wants to be host-agnostic and tested on both; and an unfiltered `event.`-scoped search crosses environments, so keep the `span.environment` predicate in.

For reference, the baseline also lands two structural findings relevant to what this dashboard surfaces: **`has_order_by` appears in 9 of the 10 slowest query shapes** but only 3 of the 8 highest-volume ones, and **`joins:0` in 9 of 10** — slowness is not join-driven, it's ORDER BY (usually plus a FILTER) over few patterns. The worked example is a 3-pattern zero-join date-FILTER query spending 15.3 s (99.68%) in `prequery.execute`, essentially all of it Fuseki.

PR #4269, opened 2026-08-26
feat: emit the target project(s) on the Gravsearch root span (DEV-7031)

mpro7:
[DEV-7031](https://linear.app/dasch/issue/DEV-7031)

## Summary

- The `gravsearch` root span now carries the **target project(s)** as two bounded attributes, so a slow trace says *whose data* it was reading without anyone eyeballing the query text.
- `gravsearch.project_shortcodes` — sorted, de-duplicated shortcodes of the projects the query refers to. `gravsearch.project_restriction` — the `ProjectIri` the request was explicitly scoped to, when it was.
- Panel 7 of **DSP-API — Response Duration** gains a **Project** column reading the first attribute.

## Motivation

Panel 7 could show how slow a Gravsearch was, its shape, and (via the `gravsearch.query` event) the verbatim query — but not which project it targeted. That was recoverable only by reading shortcodes out of ontology IRIs by eye, which a Grafana table cannot select into a column.

## Changes

**Instrumentation** (`SearchResponderV2`)

`setProjectsOnRoot` sits beside `setShapeOnRoot`, called from both string entry points after parse.

Two attributes rather than one plus a flag: "which projects does this query touch" is *inferred* from the query, "which project was this search scoped to" is *stated* by the caller, and a cross-project query has several of the first and none of the second.

Shortcodes come from **all** Knora IRIs, not only ontology ones. dsp-api's internal searches — incoming links, still-image representations, incoming regions — reference built-in ontologies exclusively and interpolate the target resource IRI, so reading ontology IRIs alone would leave every one of their spans blank. Only the shortcode reaches the span, never the IRI behind it.

**Challenges and decisions**

- **Derived via `getProjectShortcode`, not `getProjectCode`.** The latter returns the raw path segment — already upper-cased for an ontology IRI, but the verbatim regex capture for a data IRI — and the project-ID pattern is `\p{XDigit}{4}`, which accepts either case. A query mixing `…/ontology/082a/…` with `http://rdfh.ch/082a/…` would have reported `082A,082a`: one project, two entries. There is a test that fails against the raw code.
- **The restriction is an IRI, not a shortcode.** `limitToProject` arrives from the query string as a `ProjectIri`, and a project IRI created after the shortcode-shaped ones were retired is `http://rdfh.ch/projects/<base64-UUID>`. Resolving it would mean a second abstract member on the `SearchResponderV2` trait plus a repository call on the pre-parse path of every Gravsearch. So the column does not join against `project_shortcodes` for newer projects — a documented limitation, not an oversight.
- **Empty vs. absent is load-bearing.** `project_shortcodes` is always set; an empty value means "no project identifiable" (built-in ontologies only), while an absent attribute means the derivation never ran (interrupted before parse). A test locks both.
- **Panel 7 keeps its query-text filter.** The issue sketched switching the `${project}` variable to the new attribute. A TraceQL attribute predicate does not match a span lacking the attribute, so filtering on it would return nothing until this release is deployed, and would keep hiding every older trace still in retention. The column has no such window. Switching the filter is the remaining follow-up, recorded in the dashboards README.

**Cardinality** — no ops-deploy change needed, and that is the point: the Alloy `otelcol.connector.spanmetrics` dimension list on `ops-deploy` main is exactly `gravsearch.query.shape`, `stack`, `domain`, `environment`. Neither new attribute is on it, so no new Prometheus label. `dsp-gravsearch.json` consumes only `gravsearch_query_shape`, confirming it from the consumer side.

**Docs** — runbook §4 gets both attributes plus why they are two; §7 records that the shape-less-early-interrupt topology lacks the project attribute too. TraceQL recipes §4 gains project filters (regex, not equality — the value is a set). The instrumentation recipe gains a bounded-projection row, since its "never set instance IRIs as attributes" rule would otherwise read as forbidding this change, when what it bans is the IRI itself and not a bounded value derived from one.

## Test Plan

> [!IMPORTANT]
> **No test, compile or format gate was run locally.** This machine has neither `bazel` nor Nix (`just check` exits 127 on a missing binary), Docker is not running, and no Metals/BSP server is live — a deliberate `val x: Int = "string"` probe returned no IDE diagnostics, so an empty diagnostic list proves nothing here. **CI is the first real gate for this PR.**

Compensating hand-verification: every AST case-class arity checked against `SparqlQuery.scala`; `SmartIri.getProjectShortcode`, `Shortcode.value` and `StringValue` visibility confirmed; all five new test queries checked for declared prefixes; scalafmt's 120-column limit and `align.preset = most` arrow columns checked programmatically; dashboard JSON validated with `jq` plus layout-`ElementReference`-to-`elements` and panel-id-uniqueness checks per `grafana-dashboards/CLAUDE.md`.

Tests added — `GravsearchQueryShapeSpec` (pure derivation): single-project; cross-project (several shortcodes, sorted and de-duplicated); built-in-ontology-only (no project); the internal-search shape (built-in ontologies plus an interpolated resource IRI); one project in mixed case reported once; only the shortcode reported, never the instance IRI.

`SearchResponderV2GravsearchSpanE2ESpec` (spans, isolated JVM target): shortcodes on the root span with no restriction when unrestricted; the restriction carrying the project IRI when scoped; no project attribute at all on a parse failure.

## Screenshots

The Project column only populates for traces recorded after this ships, so a screenshot now would show it empty. Worth capturing once the release is deployed.

### DIFF
--- a/grafana-dashboards/dsp-api/dsp-api-response-duration.json
+++ b/grafana-dashboards/dsp-api/dsp-api-response-duration.json
@@ -630,7 +630,7 @@
         "spec": {
           "id": 7,
           "title": "Slowest Gravsearch queries (traces)",
-          "description": "Individual gravsearch executions slower than the Duration ≥ threshold, from Tempo traces (service DSP_svc_api), scoped to the selected environment/stack. Sorted slowest first. To read the actual SPARQL/Gravsearch query: click the Trace ID → in the trace view select the 'gravsearch' span → Events → 'gravsearch.query' → db.query.text (verbatim, untruncated). Tempo returns the most recent matches above the threshold (limit 100), then this table sorts them by duration. Set 'Project shortcode (Gravsearch)' to filter to one project's queries (matches the ontology IRIs in the query text, host-agnostic; empty = all projects). The 'Query (preview)' column shows the start of the query inline (filtering on the event surfaces it); click the Trace ID for the full, untruncated text. A rendered Project column needs the follow-up span attribute (DEV-7032/DEV-7031).",
+          "description": "Individual gravsearch executions slower than the Duration ≥ threshold, from Tempo traces (service DSP_svc_api), scoped to the selected environment/stack. Sorted slowest first. To read the actual SPARQL/Gravsearch query: click the Trace ID → in the trace view select the 'gravsearch' span → Events → 'gravsearch.query' → db.query.text (verbatim, untruncated). Tempo returns the most recent matches above the threshold (limit 100), then this table sorts them by duration. Set 'Project shortcode (Gravsearch)' to filter to one project's queries (matches the ontology IRIs in the query text, host-agnostic; empty = all projects). The 'Query (preview)' column shows the start of the query inline (filtering on the event surfaces it); click the Trace ID for the full, untruncated text. The 'Project' column is the gravsearch.project_shortcodes span attribute (DEV-7031): the projects a query references, read from its ontology IRIs and its resource IRIs alike, so it is populated for dsp-api's internal searches (incoming links, still-image representations, incoming regions) as well. It is blank on traces recorded before that instrumentation was deployed. Note that the 'Project shortcode (Gravsearch)' variable still filters on the query text, so it matches ontology IRIs only; prefer reading the Project column over trusting that filter for internal searches.",
           "links": [],
           "data": {
             "kind": "QueryGroup",
@@ -650,7 +650,7 @@
                       "version": "v0",
                       "spec": {
                         "queryType": "traceql",
-                        "query": "{ span:name = \"gravsearch\" && span.environment =~ \"${environment:regex}\" && span.stack =~ \"${stack:regex}\" && span:duration > ${gsthreshold} && event.db.query.text =~ \".*ontology/${project}.*\" } | select(span:duration, span.gravsearch.query.shape, span.gravsearch.schema_predicates)",
+                        "query": "{ span:name = \"gravsearch\" && span.environment =~ \"${environment:regex}\" && span.stack =~ \"${stack:regex}\" && span:duration > ${gsthreshold} && event.db.query.text =~ \".*ontology/${project}.*\" } | select(span:duration, span.gravsearch.query.shape, span.gravsearch.schema_predicates, span.gravsearch.project_shortcodes)",
                         "limit": 100,
                         "tableType": "spans"
                       }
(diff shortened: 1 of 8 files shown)


grafana-dashboards/dsp-api/dsp-api-response-duration.json
```diff
@@ -650,7 +650,7 @@
                       "version": "v0",
                       "spec": {
                         "queryType": "traceql",
-                        "query": "{ span:name = \"gravsearch\" && span.environment =~ \"${environment:regex}\" && span.stack =~ \"${stack:regex}\" && span:duration > ${gsthreshold} && event.db.query.text =~ \".*ontology/${project}.*\" } | select(span:duration, span.gravsearch.query.shape, span.gravsearch.schema_predicates)",
+                        "query": "{ span:name = \"gravsearch\" && span.environment =~ \"${environment:regex}\" && span.stack =~ \"${stack:regex}\" && span:duration > ${gsthreshold} && event.db.query.text =~ \".*ontology/${project}.*\" } | select(span:duration, span.gravsearch.query.shape, span.gravsearch.schema_predicates, span.gravsearch.project_shortcodes)",
```
BalduinLandolt: since we also emit the limit to project IRI, would it make sense to include that here too?

BalduinLandolt (approved):
Thanks!

grafana-dashboards/dsp-api/dsp-api-response-duration.json
mpro7: Yes — added in acdf67f as a **Scoped to** column next to **Project**.

You also caught a wrong assumption behind leaving it out. I'd reasoned the restriction would be blank on most rows and not worth a column. It won't be: dsp-app's Advanced Search passes `limitToProject` on both the search and the count ([`advanced-search-results.component.ts:150`](https://github.com/dasch-swiss/dsp-app/blob/main/libs/vre/pages/search/advanced-search/src/lib/advanced-search-results.component.ts#L150)), and that's the most Gravsearch-heavy flow in the app — so it's populated on exactly the slow, researcher-composed traces this panel exists to surface.

Showing one of the pair also undercut the reason there are two: **Project** is *inferred* from the query's IRIs (which projects it touches), **Scoped to** is *stated* by the caller. Side by side they answer what neither answers alone — did someone run a project-scoped search, or a query that merely references that project?

Two caveats, now in the panel description and the README:

- Empty **Scoped to** means *unrestricted*, not unknown.
- It holds the `ProjectIri` as received, not a shortcode, so for projects created after shortcode-shaped IRIs were retired it won't visually match **Project**. Resolving it would need a project lookup on the pre-parse path of every Gravsearch (and a second abstract member on the `SearchResponderV2` trait), which didn't seem worth it — happy to revisit if the mismatch is annoying in practice.

PR #4270, opened 2026-08-26
feat!: stepped view-restrictions report, and a property-grouped sibling

derschnee68:
[DEV-6778](https://linear.app/dasch/issue/DEV-6778)

> **BREAKING CHANGE:** `GET /admin/projects/iri/{iri}/view-restrictions/summary` is removed. It shipped in
> v37.6.0 (#4239), so any existing caller now gets a 404. Its replacement is the two stepped endpoints
> below (`/classes` then `/values`); DSP-APP is migrated in the companion PR. Retitled `feat!:` so
> release-please cuts a major rather than a minor.

Two related features. Happy to split if you'd rather review them separately.

## 1. The stepped class report (DEV-6778)

`/summary` returned HTTP 500 on large projects — one request had to finish six `(audience × state)` fan-outs over chunked classes inside a 60s timeout, and one slow class meant an empty screen.

Replaced by two endpoints the frontend drives in sequence:

```
GET .../view-restrictions/classes                              step 1, one query
GET .../view-restrictions/values?resourceClass=&itemType=      step 2, one query per class
```

The enabling change is a query rewrite: grouping by `knora-base:hasPermissions` instead of pre-filtering to a permission subset answers every audience at once, so **12 queries per class collapse to 2** and `distinctPermissions` / `totalResourcesByClass` / all chunking are deleted outright (−1,828 lines).

Verified against the local `anything` project, old vs new on identical data:

| | old `/summary` | new `/classes` |
|---|---|---|
| totalResources | 67 | **67** |
| anon hidden | 9 | **9** |
| anon restrictedView | 5 | **5** |

## 2. Property-grouped report

A property can be used by many classes, so a misconfigured one is scattered across the class report and invisible. Property grouping previously existed as a `groupBy` flag threaded through ~12 signatures; this is separate code with its own routes.

```
GET .../view-restrictions/properties
GET .../view-restrictions/property-values?property=&itemType=
GET .../view-restrictions/property-items?property=&itemType=&page=
```

## Two bugs found and fixed on the way

**Graph scoping silently undercounts.** I initially replaced the `attachedToProject` join with `GRAPH <projectDataGraph>` on `CONVENTIONS.md`'s measured 5.4× (DEV-6827). It's wrong for a project-wide read: a project's resources span one data graph per ontology while `projectDataNamedGraphV2` returns one. On the local project that's 65 resources in `…/anything` and 6 in `…/freetest` — dropped with no error. Reverted. The numbers are now pinned by a two-graph TriG fixture (`ViewRestrictionsServiceSpec`), not only by the no-`GRAPH` substring guard.

**A 27s probe was being paid once per class.** `multiTypedQuery` measured **27.3s on LHTT** returning nothing — its scaladoc claimed `LIMIT 1` bounded it, but there is no hit to stop at when the answer is "no". The stepped design turned a per-report cost into a per-class one. Now gated by a cheap probe (1.4s) and cached per project.

## Measurements behind the property report (local LHTT, 105,983 resources / 859,723 values)

| | |
|---|---|
| Whole-project grouped by property | 24.8s → stepped shape required |
| Property list from data | 14.0s → ontology cache instead (**552ms**) |
| One property, `FILTER` vs **bound** | 3,060ms → **1,053ms** |
| With vs **without** class join | 2,380ms → **1,128ms**, identical counts |

The last one is why the property route touches no `ProjectClasses`, no `VALUES`, no most-specific-class filter — and so is never exposed to the 27s probe.

## Verified on LHTT

- `/properties` → **182 properties in 552ms**, exactly matching SPARQL's `subPropertyOf* knora-base:hasValue`
- `/property-values` for `hasTitle` → `totalValues` 66,484; cross-checked as 66,484 values / 66,484 resources / **1** permission literal

## Test Plan

- `bazel build //modules/webapi:webapi //modules/webapi:test //modules/test-e2e:test` — green
- `ViewRestrictionsByPropertyQuerySpec` — **run**, green. Several assertions are negative, so compiling proves nothing
- `ViewRestrictionsByPropertyServiceSpec` — **run**, green. Pins that classifying a bare permission literal equals classifying a real object across 6 literals × 3 audiences. This is the safety net for deliberately duplicating the permission machinery
- E2E specs written for both reports; not run here (need the stack)

## Known gaps

- The class report lost golden query snapshots when its old queries were deleted; the stepped ones have substring assertions only
- `ViewRestrictionsByPropertyServiceSpec` constructs the service with a `null` repo — every method it exercises is pure, but a stub would be cleaner
- The drill-down is ~6.8s on the worst-case property (my earlier 3.2s estimate omitted the real joins)


derschnee68:
@claude review

derschnee68:
@claude review

derschnee68:
@claude review

domsteinbach (approved):
Could you pls remove the orphaned code? -> a. ViewRestrictionsQuerySpec.scala:23 — golden snapshots removed, fixtures orphaned.

PR #4275, opened 2026-08-27
docs: trim the CLAUDE.md memory files and move the SIPI bump into the docs

mpro7:
## What

Reduces what loads into every Claude Code session in this repo, and removes content that had gone stale.

| Change | Effect |
|---|---|
| Trim derivable content from both `CLAUDE.md` files | Root file est. 4,188 → 3,001 resident tokens (341 → 215 lines); `modules/ingest/CLAUDE.md` est. 890 → 444 (loads only under that module) |
| Move the SIPI/Fuseki bump procedure into the docs | est. 452 resident tokens → a ~15-token pointer |
| Drop the claude-flow leftovers | 36 lines of dead config removed |

Net: **est. ~1,150 fewer tokens of always-loaded context per session.**

## Why

Cut only what a session can reconstruct by reading the repo — module and slice directory listings, the tech-stack list, test-path lists, generic code-style and troubleshooting bullets. Those listings had also rotted, and five claims were verified wrong against the tree:

- the slice list named `lists/` and `shacl/` (both gone) while omitting `api/`, `export/`, `standoff/`
- the module list named 6 of 12 modules
- `modules/webapi/src/test/resources/test.conf` does not exist
- handlers were said to live in `*EndpointsHandler.scala`, of which `webapi` has none — the four that exist are in `modules/ingest`, where that line now lives
- `-old-syntax` appears in no build file; the real toolchain opts are in `bazel/toolchains/BUILD.bazel`

Kept everything the code cannot teach: the Bazel/Nix/RBE build notes, the Metals-MCP bazel-bsp gotcha, the dev-DB workflow, the endpoint-registration recipe, the BagIt RFC pointer, the slice `api`/`domain`/`repo` placement convention, the test-data check, and every doc pointer.

## SIPI bump procedure

Per review feedback (skills are harness and don't belong in the codebase), the procedure lives in the published docs: it extends the existing **Docker Image Versions** section of `docs/05-internals/development/third-party.md`, which already stated that Sipi is digest-pinned in the `oci.pull` blocks and that Fuseki comes from `image_versions.fuseki`. The two new subsections are the how-to for exactly those facts, so no new page and no mkdocs nav change. `CLAUDE.md` keeps a one-line pointer.

Every claim in the new docs re-verified against the tree: `oci_config_label` (`tools/buildinfo/defs.bzl:76`), the two `oci.pull` blocks (`MODULE.bazel:319`/`324`), `FUSEKI_DIST_VERSION` (`MODULE.bazel:297`), and the `## Updating Jena/Fuseki` heading (`modules/fuseki/README.md:30`).

## Notes for review

- Empty `## API Structure` and `## Troubleshooting` headings went with their subsections. The trailing pointer sections (SPARQL, Conventions, Observability, IRI Handling) were already mis-nested as `###` under Troubleshooting and now sit under `## Development Guidelines`.
- `.claude/config.json` held claude-flow's `sparc`/`swarm` blocks. Claude Code never reads that path and nothing in the repo, CI or the justfile references it — #4109 already removed claude-flow from `.claude/settings.json` and missed this file.
- Also fixes four pre-existing MD022/MD032 violations in the ingest file.
- `npx markdownlint-cli -c .markdownlint.yml --disable MD013 MD040` (the CI invocation) passes clean on `CLAUDE.md` and the whole `docs/**/*.md` gate.


CLAUDE.md
```diff
@@ -236,55 +155,10 @@ metals `import-build` tool. Full details in `docs/development/dsp-api-metals-mcp
...
-## Troubleshooting
-
-### Common Issues
-
-- **Docker**: Ensure Docker Desktop is running
-- **Database**: Check Fuseki is accessible at localhost:3030
-- **Tests**: Integration tests require built Docker images
-
-### Debugging
+See the `bump-sipi` skill (`.claude/skills/bump-sipi/SKILL.md`).
```
subotic: Please move what needs to be done for a SIPI update to the docs. Skills should not be used for documentation. If you feel you need a skill, then please go ahead with adding one. Skills are "harness" and don't belong to the codebase.

subotic (changes requested):

CLAUDE.md
mpro7: Ok, moved it into the docs and dropped the skill here [145cad0](https://github.com/dasch-swiss/dsp-api/pull/4275/commits/145cad026100e6fcb2fa5674e3f38b351c82ed5b).

It now extends the existing **Docker Image Versions** section in `docs/05-internals/development/third-party.md`, which already stated that Sipi is digest-pinned in the `oci.pull` blocks and that Fuseki comes from `image_versions.fuseki` — the procedure is just the how-to for those two facts, so it fits there rather than in a new file, and that page is already in the mkdocs nav. `CLAUDE.md` keeps a one-line pointer to it.

One related question while we're here: `.claude/skills/fix-markdownlint/SKILL.md` is also checked in and referenced from `CLAUDE.md`. Should that move out under the same rule? Happy to do it in this PR or leave it for a follow-up.

subotic: Yes, I think so. You can do it as part of this PR or a follow-up. You can decide.

subotic (approved):

CLAUDE.md
mpro7: It is done in follow up hee: https://github.com/dasch-swiss/dsp-api/pull/4283,

PR #4276, opened 2026-08-28
feat(dsp-api): transform rich-text TextValues to standoff in the import pipeline

Nora-Olivia-Ammann:
[DEV-6938](https://linear.app/dasch/issue/DEV-6938)

## Summary

- Transform rich-text `TextValue`s carrying `textValueAsXml` into `knora-base` standoff — reusing `StandoffTagUtilV2.convertXMLtoStandoffTagV2` and the standard mapping exactly as the resource-create path does — reaching xmlupload parity for formatted text (PRD Feature 4; US-6 REQ-6.1/6.2).
- Introduce the injectable `IdSource` service the standoff pass needs for minting (production layer + deterministic test double).
- Set `hasTextValueType` on all text values, and emit system `hasStandoffLinkTo` LinkValues for internal salsah-links with per-target reference counts.

## Changes

- **`IdSource`** — `trait IdSource` + `IdSourceLive` (production) + `IdSourceInMemory` (deterministic test double for exact-isomorphism assertions). Standoff-tag IRIs stay deterministic (`<valueIri>/standoff/<startIndex>`) and are minted directly, not through the service.
- **`OntologyTransformer`** — new effectful `convertRichtextValues` pass: standoff-tag node emission, standard-mapping resolution with custom-mapping and malformed-XML rejection (`ZIO.fail`/`ZIO.attempt` so rejections surface as `TransformerError`, never a defect), plain-text `valueHasString` projection, and dropping `textValueAsXml`/`textValueHasMapping`. A separate `addTextValueType` metadata step. Salsah-link aggregation into system LinkValues with per-target `valueHasRefCount` (within-value `Set` collapse, sorted-by-IRI minting, SystemUser attribution, fixed standoff permissions). The stale `convertLinkValues` "step 5 will merge" docstring is corrected.
- **Layer wiring** — `OntologyTransformer` gains `StandoffMappingService & IdSource`. Because `StandoffMappingService` transitively needs the ontology cache that `OntologyModule` provides, `OntologyTransformer` is lifted out of `OntologyModule` into the `LayersLive` pool to avoid a layer cycle; `toKnoraBase`'s environment is unchanged.

## Test Plan

- Extended `OntologyTransformerSpec` (hermetic unit spec: `TestClock` + `IdSourceInMemory` + exact `Model.isIsomorphicWith`): happy-path standoff, 3-level nested standoff tree with resolved parent IRIs, `hasTextValueType`, salsah-link ref-counts (single, cross-value count, within-value collapse, sorted multi-target minting, user/standoff separation, self-reference), rejection cases (custom mapping, malformed XML, invalid href, >1 incoming edge), and the D4 guard that a minted tag IRI is not classified as a `ValueIri`.
- `nix develop --command bazel test //modules/webapi:test` — full webapi unit suite green; `just check` (scalafmt + license + no-relative-imports) green.
- Multi-perspective review (scala-zio, consistency, simplicity, dune): no Critical findings; actionable items addressed.

## Notes

- Standoff-mapping tests are hermetic (an in-memory `GetMappingResponseV2` stub) — no integration stack / Docker needed.
- All import behaviour stays behind `allow-project-data-import` (off by default); the change is additive — no already-working import regresses.
- Follow-up (advisory): the standoff-link permission string is built identically here and in `CreateResourceV2Handler`; consolidating into a single shared definition would remove the duplication.

## Screenshots

[Attach manually if UI changes]


modules/webapi/src/main/scala/org/knora/webapi/core/LayersLive.scala
```diff
@@ -119,12 +122,16 @@ object LayersLive { self =>
...
+      IdSourceLive.layer,
       InfrastructureModule.layer,
       IriService.layer,
       KnoraResponseRenderer.layer,
       ListsResponder.layer,
       OntologyModule.layer,
       OntologyResponderV2.layer,
+      // Wired here, not inside OntologyModule: OntologyTransformer now needs StandoffMappingService, which
+      // transitively needs the OntologyCache that OntologyModule provides — bundling it there would form a layer cycle.
+      OntologyTransformer.layer,
```
BalduinLandolt: I'm not too familiar with the `OntologyTransformer` layer. But this sounds as if the change makes this layer considerably heavier, with dependencies on the ontology cache. I wonder if it would be possible to extract some things in a separate layer, to keep the dependency graph cleaner. (A dependency circle is typically best broken by pulling out the shared parts.)

modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/IdSource.scala
```diff
@@ -0,0 +1,48 @@
...
+  // ValueIri.makeNew and UUID.randomUUID are impure, so suspend each call in ZIO.succeed rather than computing once:
+  // a hoisted value would reuse the same id on every mint.
+  //
+  // No triplestore uniqueness check (unlike the create path's IriService.makeUnusedIri): import builds a fresh graph
+  // from scratch, so a random 128-bit UUID cannot collide with an existing value IRI. Deliberate, not an oversight.
+  override def freshLinkValueIri(resourceIri: ResourceIri): UIO[ValueIri] =
+    ZIO.succeed(ValueIri.makeNew(resourceIri))
+
+  override def freshStandoffTagUuid: UIO[UUID] =
+    ZIO.succeed(UUID.randomUUID())
```
BalduinLandolt: this breaks with existing naming conventions (see `ValueIri.makeNew`).

Also, if in an effectful context, I think it's best practice to create the UUID using ZIO's random number generator (for testability).

modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/OntologyModule.scala
```diff
@@ -25,19 +25,21 @@ object OntologyModule { self =>
 
   type Dependencies = IriConverter & IriService & StringFormatter & TriplestoreService
 
+  // Note: OntologyTransformer lives in this package but is intentionally NOT part of this module. It depends on
+  // StandoffMappingService, which transitively needs the OntologyCache this module provides, so bundling it here would
+  // form a layer cycle. It is wired directly in core/LayersLive instead.
```
BalduinLandolt: see comment above: this looks a bit suspicious to me

modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/OntologyTransformer.scala
```diff
@@ -340,6 +386,252 @@ final class OntologyTransformer(sf: StringFormatter) { self =>
...
+        )
+      case _ => ZIO.unit
+    }
+  }
+
+  /** Removes the naive-renamed `knora-base#textValueHasMapping` (which has no knora-base equivalent) everywhere. */
+  private def dropTextValueHasMapping(model: Model): Unit = {
+    val textValueHasMapping = model.createProperty(KnoraBase.KnoraBasePrefixExpansion + "textValueHasMapping")
+    val _                   = model.removeAll(null, textValueHasMapping, null)
+  }
```
BalduinLandolt, line 479: I don't suppose it would be possible to not do the rename in this case? Seems redundant

modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/OntologyTransformer.scala
BalduinLandolt, line 1: It's hard for me to review this reasonably by just looking at the code. It looks sane, that's for sure; but the output should be checked for strickt equivalence with regular standoff handling.

modules/webapi/src/test/scala/org/knora/webapi/slice/ontology/IdSourceInMemory.scala
```diff
@@ -0,0 +1,41 @@
...
+ * does not shift the other kind's numbering.
+ */
+final class IdSourceInMemory(linkCounter: Ref[Int], tagCounter: Ref[Int]) extends IdSource {
+
+  override def freshLinkValueIri(resourceIri: ResourceIri): UIO[ValueIri] =
+    linkCounter.getAndUpdate(_ + 1).map(n => ValueIri.unsafeFrom(s"${resourceIri.value}/values/$n"))
+
+  override def freshStandoffTagUuid: UIO[UUID] =
+    tagCounter.getAndUpdate(_ + 1).map(n => new UUID(0L, n.toLong))
+}
```
BalduinLandolt, line 28: with ZIO's rng, this should hopefully be simpler to test

BalduinLandolt (approved):
All in all it looks good, though this one is hard to review with confidence. I added some comments, but nothing critical.

modules/webapi/src/main/scala/org/knora/webapi/core/LayersLive.scala
Nora-Olivia-Ammann: Addressed in be8cec3df. Extracted `StandoffEntityInfoService` — a lean, OntologyCache-only lookup — out of `OntologyResponderV2`. `StandoffMappingService` and `StandoffTagUtilV2` now depend on it instead of the full responder, so the transformer's transitive graph no longer includes `OntologyResponderV2`. The dependency cycle is gone and the graph is leaner, as suggested.

modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/OntologyModule.scala
Nora-Olivia-Ammann: Addressed in be8cec3df. The cycle no longer exists: standoff callers now depend on the extracted `StandoffEntityInfoService` (OntologyCache-only), not `OntologyResponderV2`. The module exposes that service, and this comment is rewritten to describe the new, acyclic arrangement.

modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/IdSource.scala
Nora-Olivia-Ammann: Addressed in be8cec3df. Renamed to `makeLinkValueIri` / `makeStandoffTagUuid` (the effectful makeXxx convention), and `IdSourceLive` now mints via `zio.Random.nextUUID`. I kept the counter-based double — see the reply on IdSourceInMemory.

modules/webapi/src/test/scala/org/knora/webapi/slice/ontology/IdSourceInMemory.scala
Nora-Olivia-Ammann: `IdSourceLive` now uses `zio.Random` (be8cec3df). I kept this deterministic double rather than switching to TestRandom: the transformer's `isIsomorphicWith` fixtures assert readable ids (`/values/1`, `UUID(0,n)`), the repo has no TestRandom precedent (only `Random.setSeed`), and seeding would turn those fixtures into opaque base64. Happy to revisit if you prefer TestRandom.

modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/OntologyTransformer.scala
Nora-Olivia-Ammann: Kept as-is. Stage 1 is a blind schema-wide IRI rename with no per-property carve-out, and `textValueHasMapping` has no knora-base equivalent. Dropping it after the rename is the simplest localized handling; special-casing one property inside the generic stage-1 rename would be more intrusive.

modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/OntologyTransformer.scala
Nora-Olivia-Ammann: Added a cross-emitter equivalence test in be8cec3df: it asserts the transformer's standoff RDF is isomorphic to `InsertValueQueryBuilder.buildStandoffPatterns` (the canonical create-path emission) for the same tags, with `standoffTagHasUUID` projected out.

PR #4277, opened 2026-08-28
fix(export): exclude deleted values and dangling links from the data export (DEV-7008)

derschnee68:
Fixes [DEV-7008](https://linear.app/dasch/issue/DEV-7008). Reported by WordWeb (project 0826) via Nora.

## Problem

The CSV data export could show a link with an **empty label but a populated target IRI**, pointing at a resource that no longer exists. Less visibly, deleted values of *any* type were exported as if they were current.

The issue description carries a correction after review that matters here: the reported rows are **deleted link values**, not live links to a deleted target. Both shapes exist, and they have different causes, so this PR fixes both.

### Cause 1 — deleted values reached the export

The construct query has no value-level `isDeleted false` filter in its non-version branch, so deleted values come back from the triplestore. `/v2/resources` collapses them to `knora-api:DeletedValue` via `markDeletions = true`, but both export read paths left `markDeletions` at its default of `false`:

- `readResourcesSequence` — the CSV export
- `readResourcesSequencePar` — the OAI export

So a deleted value was indistinguishable from a live one downstream. Both now collapse, which covers links *and* every other value type (a deleted text value has no blank-label tell, so it was the silent case). Both methods are export-only in production, so the change is contained.

> Note: the issue names `readResourcesSequencePar` as the CSV path. On current `main` that method serves the OAI export; the CSV export goes through `readResourcesSequence`. Both had the same defect, so both are fixed.

### Cause 2 — a live link to a deleted target left a bare IRI behind

`ExportService.valueColumns` emitted `valueHasString` unconditionally, even when the target resolved to no label — precisely the reported symptom. It now drops the label and the IRI together, so the two columns stay consistent and no dangling reference is exported.

### Supporting change — deleted resources are no longer enumerated for export

`FindResourcesService`'s three queries returned deleted resources. They produced no rows (the read passes `withDeleted = false`), but they still fed the export's cross-batch **link-label map** — so a link pointing at a deleted resource *of the exported class* resolved a label and was exported as if its target were alive, defeating the guard above. Each one also logged a spurious retrieval error. All three queries now filter `isDeleted false`. `FindResourcesService` is consumed only by `ExportService`, so this is contained too.

## Verification

Both shapes are pinned in the self-contained 1612 fixture (no shared dataset touched), and I verified each half of the fix genuinely fails without it:

| Reverted hunk | Observed failure |
|---|---|
| deleted-value collapsing | link cell `Resource2 :: Resource0`, IRI cell carries both IRIs; deleted paragraph text appears; 3 goldens + new test fail |
| `valueColumns` guard | link cell `Resource2 :: ` with `DeletedTargetRes01` in the IRI column — the issue title verbatim |

So the existing golden CSVs double as the regression pin, alongside the explicit `deleted values are excluded from the export` test.

- `bazel test //modules/webapi:test` — full webapi unit suite passes.
- `just check` (CI gate: scalafmt + SPDX headers + no-relative-imports) passes.

## Notes for review

- The deleted-value collapsing changes behaviour *inside* `readResourcesSequence`/`Par` rather than adding a parameter, following the issue's "one place, contained" guidance and keeping `ReadResourcesServiceFake` untouched. Both methods are export-only, but if you'd rather see it explicit at the call site I'm happy to thread a parameter through instead.
- The `FindResourcesService` filter goes slightly beyond the issue's suggested fix. Without it the `valueColumns` guard is toothless for a deleted target that belongs to the exported class, so I judged it necessary to actually deliver the issue's stated expected behaviour ("No dangling reference to a deleted resource"). Easy to split out if you disagree.
- Erased vs soft-deleted targets remain indistinguishable to the export; both are now dropped, which is what the issue asks for.


derschnee68:
@claude review

domsteinbach (approved):

PR #4282, opened 2026-09-01
fix(view-restrictions): stop the class report timing out on large projects

derschnee68:
## Summary

`GET /admin/projects/iri/{iri}/view-restrictions/classes` returned a 500 on a large project. The cause was a Fuseki-cancelled query at the 20s standard tier, reported as `TriplestoreTimeoutException`.

Two changes, one root cause and one guard rail.

## 1. Anchor the class-discovery closure walk in a sub-select

`projectClassesQuery` paired `?resource a ?resClass ; knora-base:attachedToProject <p>` with `?resClass rdfs:subClassOf* knora-base:Resource`. A property path splits the BGP and is never reordered across it, so the store first materialized one row per (resource, asserted type) for the **whole project**, then probed the closure once per row.

That is the exact shape `docs/development/dsp-api-sparql-queries.md` records under "Anchor property paths" — DEV-6803 measured the same `subClassOf* knora-base:Resource` pattern at >60s.

Wrapping the project scan in a sub-select fixes the row count rather than the path: Fuseki evaluates a sub-select bottom-up, so the scan is deduplicated to the classes the project actually uses **before** the closure walk sees them, and the walk runs once per class instead of once per resource-type pair. The result set is identical.

The guard itself stays. `attachedToProject` is also carried by list nodes and by the project's own ontologies, so dropping it would let `knora-base:ListNode` and `owl:Ontology` into the class list and inflate every downstream count. `ViewRestrictionsQuerySpec` pins both the sub-select shape and the retained guard.

## 2. Give the report its own 40s triplestore tier

The report's queries are whole-project scans grouped by permission literal, not the bounded lookups the 20s standard tier is sized for. A tier of its own rather than a longer `query-timeout`, so nothing else silently inherits a doubled budget and the report's durations stay separable.

`view-restrictions-timeout` defaults to 40 seconds, overridable via `KNORA_WEBAPI_TRIPLESTORE_VIEW_RESTRICTIONS_TIMEOUT`. Two validations mirror the search tier's: it must be positive, and it must not exceed `gravsearch-timeout` — an admin-facing read has no business outlasting a Gravsearch query, and the env override makes a bad value injectable at deploy time.

Applied to all nine call sites across both routes: the shared `select` helper and the three class-discovery queries that bypass it in `ViewRestrictionsRepo`, and all four queries in `ViewRestrictionsByPropertyRepo`.

## Testing

`bazel build //modules/webapi:webapi`, `bazel test //modules/webapi:test` (ViewRestrictions filter) and `just check` all pass on this branch.

The fix is verified by query shape, not by a timing measurement against the affected project — the sub-select is what took `projectClassesQuery` off the ledge, but the remaining whole-project scans (`resourceCountsByClass` in particular) have no measured timings on that project. The 40s tier is headroom for those, not a proven bound.

## Notes

Follows #4270, which was squash-merged. Branched fresh off `main` rather than reusing the deleted branch.


derschnee68:
@claude review

domsteinbach (approved):

PR #4286, opened 2026-09-04
perf: pin isResourceInUse so candelete skips project-wide scans (DEV-6885)

BalduinLandolt:
## Summary

- Pin both `isResourceInUse` UNION branches behind GRAPH-scoped subqueries so TDB2 cannot hoist `isDeleted false` into a project-wide scan. Branch 1 (`?s ?p <target>`) cannot be fixed by text reorder — it has one bound term and loses the heuristic.
- Replace the remaining `rdfs:subClassOf* Resource` guard with a cheap drop of `LinkValue` reifications **plus** `FILTER REGEX(STR(?other), ResourceIri.SparqlRegexPattern)`. `NOT EXISTS LinkValue` is **not** equivalent to the Resource path: a live `RegionPreviewValue` sits in branch 1 object position and is not a `LinkValue`, so without the IRI-shape filter `candelete` failed `ResourceIri.from` instead of naming the host.
- Extract the query into `IsResourceInUseQuery` with a golden SPARQL spec. Behavioural ITs for a live incoming link and a live region preview now assert `cannotDoReason` contains the host/source IRI, not a value IRI or a conversion error.
- Record the bound-object wildcard-predicate trap (Fact 1), the connected-`subClassOf*` candidate-count trap, and the `Resource ≢ ¬LinkValue` keep-set (Fact 3) in `docs/development/dsp-api-fuseki-query-execution.md`.

Fixes [DEV-6885](https://linear.app/dasch/issue/DEV-6885/v2resourcescandelete-isresourceinuse-costs-18s-on-prod-sized-projects).

## Stage verification (2026-09-04, `db.stage.dasch.swiss`)

- **Identity, subquery + Resource guard vs original query:** 665/665 targets across 61 projects, 0 diffs (result sizes 0–3840).
- **Identity, original Resource-guard query vs subquery + `NOT EXISTS LinkValue`:** 117/117 targets on `hasLinkTo` hubs (≥200 incoming refs, plus a spread sample), 0 diffs. That sample did not include live region previews.
- **Live region preview (daschland Story, created for this check; will vanish on the next prod mirror):** `NOT EXISTS LinkValue` alone returned the host **and** the live value IRI. With the ResourceIri regex: host only, matching the old Resource guard. A deleted preview on a different Region did not count (`isDeleted true`).
- **Ticket resource** (`0804/ETGHDnh5…`, 0 rows), COUNT wrap, 1 warmup + 5: **1817 ms → 136 ms**.
- Slowest old queries (LIMC hubs): ~9.3–9.5 s → ~135–145 ms.

## Test plan

- [x] `IsResourceInUseQuerySpec` (pinned query text, including the ResourceIri regex)
- [x] `ResourceIriSpec` — SPARQL pattern matches the same strings as `from`
- [x] `ResourcesResponderV2Spec` — 76 tests; both candelete cases assert the reason names the host/source, not `/values/`
- [x] Stage identity + timing + live-preview keep-set as above

### DIFF
--- a/docs/development/dsp-api-fuseki-query-execution.md
+++ b/docs/development/dsp-api-fuseki-query-execution.md
@@ -66,6 +66,13 @@ probes — measured on `/v2/metadata`, where "fixing" the order made it 3× slow
 **disconnection**, not earliness. And the converse of this fact also holds — *removing* a path
 can be far more expensive than keeping it (Fact 11).
 
+**Corollary — a connected `subClassOf*` still scales with candidate count.** After pinning
+the incoming-reference probe (Fact 1 corollary), `/v2/resources/candelete` still spent ~1.5s
+of ~1.6s on a 3840-row dokubib hub inside `?otherClass rdfs:subClassOf* knora-base:Resource`
+over 13k incoming subjects. `FILTER NOT EXISTS { GRAPH <g> { ?other a knora-base:LinkValue } }`
+was byte-identical on that hub and the ticket empty case, ~190ms (DEV-6885). The path was
+doing real work (dropping LinkValues); it was just the expensive way to do it.
+
 ## Fact 4 — `MINUS` evaluates its right side without bindings; `FILTER NOT EXISTS` is per-row
 
 An un-scoped `MINUS { ?x knora-base:isDeleted true }` materializes the deleted-set of the whole
--- a/docs/development/dsp-api-fuseki-query-execution.md
+++ b/docs/development/dsp-api-fuseki-query-execution.md
@@ -32,6 +32,16 @@ Evidence: the tile-permission regression — a shortcode restriction placed afte
 made Fuseki evaluate every left-join for every project first (351ms; 7.9ms with the restriction
 hoisted). See DEV-6796.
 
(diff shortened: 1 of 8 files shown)


siers (approved):
Claude really loves writing this defensive-sounding legalese type text, but I can sorta see what's going on in the code and the numbers of course also speak for themselves

I understand this is essentially a query optimization by simplifying a part of the where clause's search, but reading the md files is not easy for me.

BalduinLandolt:
> Claude really loves writing this defensive-sounding legalese type text, but I can sorta see what's going on in the code and the numbers of course also speak for themselves
> 
> I understand this is essentially a query optimization by simplifying a part of the where clause's search, but reading the md files is not easy for me.

Yea, the markdown is mostly the agent talking to itself. The background for that: The solution proposed in the Linear ticket was subtly wrong; and so the agent now tries to prevent future sessions from repeating the same mistake.

Other than that, I hope the numbers are trustworthy, because I find it hard to reason about the query shape.

siers:
> Other than that, I hope the numbers are trustworthy, because I find it hard to reason about the query shape.

This would have been a great first sentence of the PR description. 😅

PR #4288, opened 2026-09-04
feat!: remove the project active/inactive concept (DEV-7039)

jnussbaum:
## What

Remove the project active/inactive (`status`) concept from dsp-api.

- Drop the `status` field from the project domain model, the admin API DTOs, and the project create/update requests and responses.
- Remove the soft-delete endpoint `DELETE /admin/projects/iri/{iri}`. It only flipped `status` to `false` and deleted nothing. To erase a project, use the separate erase endpoint.
- Narrow the shared `knora-admin:status` property to users and groups. Drop the `owl:cardinality 1` restriction on `knora-admin:knoraProject` and the `status` triples on the built-in projects.
- Bump `knora-base` to v55.

## Migration

A startup upgrade plugin `MigrateRemoveProjectStatus` (registered at version 55) deletes the orphaned `knora-admin:status` triples from existing projects in the admin data graph. The delete matches only `knora-admin:knoraProject`-typed subjects, so user and group status triples survive. The query is idempotent.

## Breaking change

The `status` field no longer appears in project requests or responses, and the soft-delete endpoint is gone. See the migration guide entry `v54 to v55` in `docs/10-migration-guides/index.md`.

## Tests

- `MigrateRemoveProjectStatusSpec` covers the project-only deletion, user/group survival, and idempotency.
- Domain, repo, service, and endpoint specs updated. RDF test fixtures stripped of project `status`.


siers (approved):
LGTM, the migration is simple

PR #4289, opened 2026-09-04
test(import): bulk-import vs single-create parity E2E test (DEV-7149)

Nora-Olivia-Ammann:
[DEV-7149](https://linear.app/dasch/issue/DEV-7149)

## Purpose

Guard against **drift** between the two write paths. The v3 bulk data-import and the v2
single-resource create must produce the same graph for one project. This test imports the same
fixture set through both paths into the same project data graph, then compares the two graphs for
equal triple count, identical resource-IRI set, and RDF isomorphism (normalizing minted
value/standoff IRIs, UUIDs, and creation timestamps; `hasPermissions` is excluded from the compare,
`lastModificationDate` is stripped). If the two paths ever diverge, this test fails.

## What this PR does

Both imports now run and are compared. The bulk import previously failed its own SHACL validation
(`data-shapes.ttl`); this PR fixes those blockers:

- `OntologyTransformer.addResourceMetadata` now strips before setting, so a payload-supplied
  `creationDate` or `hasPermissions` no longer coexists with the synthesized one (data-shapes
  enforces `maxCount 1`). A payload `creationDate` is honored and re-emitted as `xsd:dateTime`,
  matching the create path.
- `data-shapes.ttl` `AttachedToUserNotBuiltInShape` exempts standoff-link LinkValues, which both
  write paths attach to `SystemUser` by design.
- The project fixture registers the copyright holders the file-value fixtures use (`DaSCH`,
  `Wellcome Collection`), which the create path validates against the project allow-list.
- Every file value now carries an explicit `valueHasOrder` in both payloads. The file values
  previously omitted it, so the create path synthesized a positional order and the bulk import did
  not (62 vs 73). Per the design principle that the API synthesizes an order only for values it
  creates itself (e.g. standoff-link LinkValues) and never for values from outside, this is fixed in
  the data, not the transformer.
- The fake `SipiService` mirrors the live one on `numpages = None`. The live service hardcodes
  `numpages = None` (dsp-ingest reports no page count), so neither write path persists
  `knora-base:pageCount`. The fake had returned `Some(3)` for the PDF, which made only the create
  path emit it — a mock artifact, now removed.

## Status — still red (drift guard)

The compare now runs and reports **one residual difference**, and it is a **live-service bug, not a
bulk-import gap** — a follow-up tracked in [DEV-7187](https://linear.app/dasch/issue/DEV-7187):

| Predicate | Compare | Cause |
|---|---|---|
| `hasTextValueType` | A=26 B=25 | The v2 add-value path (`POST /v2/values`) omits it, while the bulk import and the v2 resource-create path both write it. The parity test hits the add-value path via the two-step create of the self-referencing standoff resource, so that text value lacks the triple in the create run. DEV-7187 (probable data clean-up). |

The test stays red on this one difference until DEV-7187 is fixed. The bulk import is correct here.

A fourth item is deferred: full **permission-string parity** — honor the payload `hasPermissions`
string and resolve class/property DOAPs so the strings match, then include `hasPermissions` in the
compare (annotated in `addResourceMetadata`).

On a compare failure the test prints a per-predicate count delta (the canonical line diff
over-reports on a non-isomorphic pair).

## Changes

- Parity spec `BulkImportParityE2ESpec` + shared fixtures + fake `SipiService` + an overridable
  SipiService layer seam on `LayersLive.remainingLayer` / `E2EZSpec`.
- The SHACL-unblock fixes above.

## Test plan

- `bazel test //modules/test-e2e:test --test_filter='.*BulkImportParity.*' --nocache_test_results`
  (stop any local stack first — it holds port 3333).
- Both imports run; the compare fails on the three residual gaps above (per-predicate delta in the
  failure output). The test goes green once the three gaps are closed.

## Screenshots

N/A (no UI changes).


modules/test-e2e/src/test/scala/org/knora/webapi/slice/api/v3/projects/BulkImportParityE2ESpec.scala
```diff
@@ -0,0 +1,510 @@
...
+  override val e2eSpec: Spec[env, Any] = suite("Bulk import vs single-resource create parity")(
+    test("both paths produce isomorphic graphs (permissions excluded)") {
+      for {
+        tester <- ZIO
+                    .serviceWithZIO[UserService](_.findUserByIri(UserIri.unsafeFrom(testerUserIri)))
+                    .someOrFail(new RuntimeException(s"on-behalf-of user $testerUserIri not found"))
+        // Run A: bulk import as root on behalf of the project user, then dump the project data graph.
+        importId <- triggerBulkImport
+        _        <- ZIO.addFinalizer(deleteImportTask(importId).ignore)
+        status   <- pollImportUntilDone(importId).retry(Schedule.spaced(500.millis) && Schedule.recurs(120))
```
BalduinLandolt: the name of this function is misleading, as the retry happens at the call site, not in the function. So it's `pollOnce` not `untilDone`

modules/test-e2e/src/test/scala/org/knora/webapi/slice/api/v3/projects/BulkImportParityE2ESpec.scala
```diff
@@ -0,0 +1,510 @@
...
+  private def pollImportUntilDone(importId: DataTaskId): ZIO[TestApiClient, Throwable, DataTaskStatusResponse] =
+    TestApiClient
+      .getJson[DataTaskStatusResponse](uri"/v3/projects/$projectIri/data-imports/${importId.value}", rootUser)
+      .flatMap(r => ZIO.fromEither(r.body).mapError(new RuntimeException(_)))
+      .flatMap { status =>
+        status.status match {
+          case DataTaskStatus.Completed => ZIO.succeed(status)
+          case DataTaskStatus.Failed    => ZIO.succeed(status)
+          case _                        => ZIO.fail(new RuntimeException("import still in progress"))
+        }
```
BalduinLandolt, line 261: the error handling here does not look right to me. I'm not 100% sure what case will behave how; but if something goes actually wrong, then I think this test will keep retrying rather than failing fast. And modelling "in progress" as a runtime exception feels wrong.

modules/test-e2e/src/test/scala/org/knora/webapi/slice/api/v3/projects/BulkImportParityE2ESpec.scala
```diff
@@ -0,0 +1,510 @@
...
+          if (sentinelPredicates.contains(p)) sentinel
+          else if (st.getObject.isURIResource)
+            blankIds.get(st.getObject.asResource.getURI).fold(st.getObject)(id => out.createResource(id))
+          else st.getObject
+        out.add(subj, st.getPredicate, obj)
+        ()
+      }
+    }
+    out
+  }
```
BalduinLandolt, line 429: I'm not sure how much production code would need to be changed. But if all those random elements were using ZIO's UUID generator, then it would be easy to run the test with a test-RNG, so that the IDs would be fully deterministic.  
If it's a lot of effort, then it's not worth it. But if it's reasonably little effort, then that would be much nicer.

modules/test-e2e/src/test/scala/org/knora/webapi/slice/api/v3/projects/BulkImportParityE2ESpec.scala
```diff
@@ -0,0 +1,510 @@
...
+      .toSet
+
+  private def hasLastModification(model: Model, resourceIri: String): Boolean =
+    model.contains(model.createResource(resourceIri), model.createProperty(lastModProp))
+
+  private def creationDate(model: Model, resourceIri: String): Option[String] =
+    model
+      .listObjectsOfProperty(model.createResource(resourceIri), model.createProperty(kb + "creationDate"))
+      .asScala
+      .toList
```
BalduinLandolt: there's a lot of these conversion operations all over the place. Perhaps a helper function or extension method could streamline this code quite a bit. (But only do it, if it actually becomes nicer.)

modules/test-e2e/src/test/scala/org/knora/webapi/slice/api/v3/projects/BulkImportParityE2ESpec.scala
```diff
@@ -0,0 +1,510 @@
...
+          // has two text values linking one target, so its refcount is 2 on both paths.
+          refCountA.contains(2),
+          refCountB.contains(2),
+        ).label(
+          s"""|counts: A=${normA.size} B=${normB.size}
+              |standoff-link refCount A=$refCountA B=$refCountB
+              |resources only in A: ${(resourcesA -- resourcesB).toList.sorted.mkString(", ")}
+              |resources only in B: ${(resourcesB -- resourcesA).toList.sorted.mkString(", ")}
+              |$diff""".stripMargin,
+        )
```
BalduinLandolt: there's a lot going on in this single assertion. It may be worth looking into multiple and/or composed assertions.

BalduinLandolt (approved):
All in all looks good, I remarked some "stylistics", and maybe the error handling/retry could be improved a bit

Nora-Olivia-Ammann:
## Review comments resolved

Addressed the 5 review threads in `b74ae0763`. All changes are in `BulkImportParityE2ESpec.scala`.

| Line | Comment | Action |
|---|---|---|
| 156 | `pollImportUntilDone` name misleading — retry at call site | Fixed |
| 245 | Error handling models "in progress" as an exception | Fixed |
| 411 | Route minted IDs through ZIO `Random` for a deterministic test-RNG | Declined — see below |
| 427 | Repeated Jena conversion boilerplate | Fixed |
| 214 | Single `assertTrue` does too much | Fixed |

### Poll-loop rewrite (156 + 245)

Both comments share one root cause. The poll now splits into `pollImportOnce` (a single GET) and a `pollImportUntilDone` loop that owns the 500ms delay and the 120-attempt budget. The loop re-polls only on `DataTaskStatus.InProgress`. A transport or deserialization failure from `pollImportOnce` propagates at once, instead of retrying for up to a minute as if the import were still running.

### Dedup helper (427)

Extracted `firstObjectOf(model, subject, predicateUri)`. It removes the repeated `listObjectsOfProperty(...).asScala.toList.headOption` chain.

### Composed assertions (214)

Split the 9-condition `assertTrue(...)` into six named assertions combined with `&&`. Each keeps its original comment. The diagnostic `.label(...)` still applies to the combined result.

### Declined: route minted IDs through ZIO `Random` (411)

`IdSource` already mints standoff UUIDs through ZIO `Random`. But `ResourceIri.makeNew` and `ValueIri.makeNew` are pure `def`s calling `UUID.randomUUID()` directly, from ~90 non-effectful sites (`CreateResourceV2Handler`, `ValuesResponderV2`, and many test specs). Making them effectful cascades a monadic return-type change through all of them. This matches the stated bar: "if it's a lot of effort, then it's not worth it".

### Verification

`bazel build //modules/test-e2e:test` compiles clean. `bazel test //modules/test-e2e:test --test_filter='.*BulkImportParity.*' --nocache_test_results` passed against the local Docker stack.

PR #4294, opened 2026-09-08
fix(import): align bulk-import output with v2-create (parity fixes A–E)

Nora-Olivia-Ammann:
[DEV-7149](https://linear.app/dasch/issue/DEV-7149)

## Summary
- Production fixes that make the v3 bulk data-import write the same graph as the v2 single-resource create. Branched off `main` — PR 1 of a two-PR split; the e2e parity spec follows in PR 2.
- Fixes A and B are extracted byte-identical from the e2e branch (`feature/dev-7149-import-parity-e2e`), so the later `main`→e2e merge is a no-op on them. C, D, E are new.

## Changes
- **Fix A — SHACL shape** (`shacl/data-shapes.ttl`): exempt standoff-link LinkValues from the built-in-SystemUser ban (both write paths attach them to SystemUser by design), via an `sh:or` keyed on `rdf:predicate = knora-base:hasStandoffLinkTo`.
- **Fix B — resource metadata** (`OntologyTransformer.addResourceMetadata`): strip-then-set system metadata so a payload value does not duplicate the synthesized one; honor a payload `creationDate`, re-emitted as `xsd:dateTime`.
- **Fix C — `hasTextValueType` on `POST /v2/values`** (`InsertValueQueryBuilder`, closes DEV-7187): the add-value path now writes the marker the resource-create and bulk paths already write.
- **Fix D — scalar canonicalization** (`OntologyTransformer.canonicalizeScalarLiterals`): re-type `valueHasInteger` to `xsd:integer` and `valueHasTimeStamp` to a UTC `xsd:dateTime`, each with a canonical lexical form. Integer and timestamp only.
- **Fix E — standoff internal references** (`OntologyTransformer.emitStandoff`): resolve an internal reference to the target standoff-tag node IRI, mirroring the create path, instead of writing a relative anchor IRI.

## Test Plan
`bazel test //modules/webapi:test --test_filter='.*(OntologyTransformerSpec|InsertValueQueryBuilderSpec|ProjectMigrationImportServiceSpec|ProjectMigrationImportShaclValidatorSpec).*'` → `OK (211 tests)`. `just check` clean. Fix C regenerates 23 `InsertValueQueryBuilder` golden snapshots (13 FormattedText, 10 UnformattedText).

## Review — deferred / follow-up
- Fix A's positive exemption branch has no unit test here; it is exercised end-to-end by the parity spec in PR 2.
- Two findings are intentionally left in Fix A/B to keep them byte-identical to the e2e branch — tracked in **DEV-7197**: the `sh:or` exemption is broader than the SystemUser/LinkValue case, and an offset-less payload `creationDate` silently falls back to import time.
- Further follow-ups: **DEV-7195** (scalar canonicalization scope — decimal/interval/uri), **DEV-7196** (backfill pre-fix data).


modules/webapi/src/main/resources/shacl/data-shapes.ttl
```diff
@@ -150,12 +150,20 @@ knora-base:AttachedToUserExistsShape
...
 knora-base:AttachedToUserNotBuiltInShape
   a sh:NodeShape ;
   sh:targetSubjectsOf knora-base:attachedToUser ;
-  sh:property [
-    a sh:PropertyShape ;
-    sh:path knora-base:attachedToUser ;
-    sh:not [ sh:in ( knora-admin:SystemUser knora-admin:AnonymousUser ) ] ;
-    sh:message "knora-base:attachedToUser must not reference the built-in SystemUser or AnonymousUser" ;
-  ] .
+  sh:or (
```
Nora-Olivia-Ammann, line 159: This is only a fix so that the test is green, I will take a closer look at the best SHACL solution when I implement that part of the plan.

BalduinLandolt (approved):

PR #4311, opened 2026-09-09
fix(dsp-api): local stack bring-up starts what it names (DEV-7199)

jnussbaum:
Fixes DEV-7199

## Motivation

`just stack-without-app` did not bring up the stack without the app. It depended on
`stack-up`, which runs an unqualified `docker compose up -d`, and stopped `app`
afterwards. All six services share one Compose project, so exclusion depended on the
excluded service starting successfully first. A host path missing from Docker Desktop's
file-sharing allowlist therefore aborted the bring-up of `api`, `sipi` and `ingest` —
the services that were the goal — and the error named only the mount:

```text
Error response from daemon: mounts denied:
The path .../app-config.dev.json is not shared from the host and is not known to Docker.
```

`stack-without-api` and `stack-without-api-and-sipi` shared the design. So did
`stack-start-dev`, which left the `app` container holding port 4200 against a local
Angular dev server.

Separately, recipes needing `bazel` failed with a bare `command not found` outside the
Nix dev shell, and three setup instructions described a stack that no longer exists —
one of them telling the reader to disable Docker Compose V2, which breaks the justfile.

## Summary

- Each subtractive stack recipe now starts exactly the services it names.
- A missing `bazel` now prints the command that activates the Nix dev shell.
- The setup documents describe the stack as it is.

### Stack recipes

`stack-without-app`, `stack-without-api`, `stack-without-api-and-sipi` and
`stack-start-dev` each stop the services they exclude, then pass their own service list
to `docker compose up -d`, then wait only for a service they actually start. None of the
four starts `app`, so none can hit its bind mount. `stack-up` and `stack-up-ci` keep the
full service set; their docstrings gain the `ingest` and `alloy` they always started.

The two `without-api` recipes no longer poll port 3333 for 360 seconds before stopping
the `api` they exclude — they inherited `wait-for-api.sh` from `stack-up`.

### Bazel guard

A private `require-bazel` recipe tests for the executable and aborts with the activation
command. The 20 recipes that invoke `bazel` directly take it as a first dependency, so
`docker-build`, `stack-up`, `test-it` and the rest inherit it. It tests for the
executable, not an environment variable: `flake.nix` sets no marker.

### Documentation

The README's first-run command becomes `just init-db-test` then `just stack-up`, so the
reader runs images built from their own checkout rather than published `:latest` ones.
The Compose V2 instruction is replaced by a pointer to a new Docker Desktop file-sharing
note. `building-and-running.md` lists the six services `docker-compose.yml` defines,
instead of `salsah1`.

### An explicit service list is not enough on its own

**Problem:** Replacing "start everything, then stop one" with an explicit
`docker compose up -d <subset>` passed every cold bring-up test.

**Tried:** The explicit list alone. Measured warm, it fails: `up -d` on a subset is
purely additive and never stops an already-running excluded container. After a prior
`just stack-up`, `just stack-start-dev` still left `app` holding port 4200 — the exact
failure the change set out to remove.

**Solution:** A leading `docker compose stop` for the excluded services. `stop` and `up`
fail differently — only `up` creates a container and resolves its mounts — so the stop
restores guaranteed exclusion without reintroducing the mount failure. It exits 0 when
the container does not exist.

### Guarding on the tool, not the environment

`flake.nix` sets only `PS1`, `SSL_CERT_FILE` and `PATH`, so there is no reliable
dev-shell marker to test. The guard tests `command -v bazel`. CI is unaffected: every CI
job already wraps its bazel recipes in `nix develop --command`.

## Gotchas

- **`just --dry-run` cannot test a shebang recipe's behaviour.** It prints the script
  text, which contains the guard's own error strings — grepping that output proves the
  wiring, not that the guard fires. Run the recipe with and without `bazel` on `PATH`.
- **`docker compose up -d <subset>` never stops anything.** Any recipe that promises a
  service is absent must say so explicitly.
- **A `just` shebang recipe needs a writable `TMPDIR`.** Without one it fails with an IO
  error naming the recipe, which reads like a recipe bug.

## Test Plan

Every recipe was run against a real Docker daemon and its resulting service set
enumerated, both cold and warm (after a full `stack-up`).

- [x] `stack-without-app` → `db sipi ingest api alloy`; port 4200 free; `/health` 200
- [x] `stack-without-api` → `db sipi ingest alloy`; returns without the 360 s API poll
- [x] `stack-without-api-and-sipi` → `db ingest alloy`; port 4200 free
- [x] `stack-start-dev` → `db sipi ingest alloy`; ports 4200 and 3333 free
- [x] All four warm, after a full `stack-up`: each ends with exactly its named set
- [x] Guard fires with exit 1 and the `nix develop` message when `bazel` is absent
- [x] Guard is transparent when `bazel` is present; `just check` runs in the dev shell
- [x] All 20 bazel-calling recipes guarded; `docker-image-tag` correctly not guarded
- [x] `just check` and `just markdownlint` pass

Not verified automatically: the `mounts denied` reproduction needs a Docker Desktop
file-sharing change, per the PRD.

## Follow-up

`stack-without-api` and `stack-without-api-and-sipi` still depend on the full
`docker-build`, which builds the dsp-api image neither recipe starts. Out of scope per
the PRD; it wastes a cached build rather than breaking a bring-up.

### DIFF
--- a/justfile
+++ b/justfile
@@ -170,10 +170,12 @@ stack-start:
     ./modules/webapi/scripts/wait-for-db.sh
     @echo "Stack started"
 
-# Start Stack without API for development
-stack-start-dev: stack-start
-    @echo "Stopping API"
-    docker compose down api
+# Start Stack without API for development: fuseki, sipi, ingest and alloy
+stack-start-dev:
+    @echo "Starting Stack without API"
+    docker compose stop api app
+    docker compose up -d db sipi ingest alloy
+    ./modules/webapi/scripts/wait-for-db.sh
     @echo "Stack started without API"
 
 # Start stack and pull latest images before starting
@@ -268,7 +270,7 @@ structurizer:
 
 ## DSP stack (local dev)
 
-# starts the dsp-stack: fuseki, sipi, api and app
+# starts the dsp-stack: fuseki, sipi, ingest, api, app and alloy
 stack-up: docker-build
     docker compose -f docker-compose.yml up -d db
     ./modules/webapi/scripts/wait-for-db.sh
@@ -279,7 +281,7 @@ stack-up: docker-build
 stack-up-fast: docker-build-dsp-api-image
     docker compose -f docker-compose.yml up -d
 
-# starts the dsp-stack using the 'dsp-repo' repository: fuseki, sipi, api
+# starts the dsp-stack using the 'dsp-repo' repository: fuseki, sipi, ingest, api, app and alloy
 stack-up-ci: docker-build
     docker compose -f docker-compose.yml up -d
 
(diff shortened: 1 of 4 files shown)


justfile
```diff
@@ -10,12 +10,31 @@ scalafmt_targets := "//modules/bagit:bagit //modules/bagit:test //modules/jwt:jw
...
+    echo "This repository builds through Bazel inside the Nix dev shell." >&2
+    echo "" >&2
+    echo "Enter the dev shell, then re-run:" >&2
+    echo "    nix develop" >&2
+    echo "" >&2
+    echo "Or run the recipe through it directly:" >&2
+    echo "    nix develop --command just <recipe>" >&2
+    echo "" >&2
+    echo "With direnv installed, run 'direnv allow' once in the repository root." >&2
+    exit 1
```
BalduinLandolt, line 30: do we really need such a long error message?

BalduinLandolt (approved):
👍  
only one nitpick comment

justfile
jnussbaum: The goal was to get actionable error messages. Under that aspect, the goal is reached, because the caller gets some options how to resolve the issue. 

This could perhaps be shortened, but I don't find it that bad.

PR #4314, opened 2026-09-10
refactor: migrate resource repo query objects to interpolated SPARQL DSL (DEV-7223)

BalduinLandolt:
Part of DEV-7150 (bucket). Resolves DEV-7223. Stacked on #4313.

## What

Migrates the three remaining standalone query objects in `slice/resources/repo` from RDF4J SparqlBuilder to the `sparql"..."` interpolator and lets the callers pass the typed wrappers straight through:

- `IsResourceInUseQuery` now returns `Select`. This is the performance-pinned incoming-reference probe: the two GRAPH-scoped subqueries in a UNION, the `FILTER NOT EXISTS` on `LinkValue`, the resource-IRI regex filter and the outer grouping braces are reproduced exactly (the extra braces matter for join ordering and were kept deliberately). Caller `ResourcesResponderV2`.
- `DeleteValueQuery` now returns `IO[SparqlGenerationException, Update]`. `DELETE { GRAPH <g> … } INSERT { GRAPH <g> … }` with the unscoped WHERE is preserved, as are the indexed `?linkValueN` / `?linkValueUUIDN` variables and the `failIf` precondition checks (now a local helper, since the `QueryBuilderHelper` mixin is gone). Caller `ValuesResponderV2`.
- `GetResourcePropertiesAndValuesQuery` now returns `Construct` from one whole-query template with conditional fragments instead of the previous mix of `s"..."` (prefixes, VALUES) and SparqlBuilder patterns. Every branch keeps its legacy structure, including the `UNION(UNION(values, standoff), links)` nesting when both flags are set. Callers `SearchResponderV2` and `ReadResourcesServiceLive` (timeout choice unchanged via `.copy(timeout = …)`).

No RDF4J import and no `s"..."` SPARQL remain in the three files.

## Verification

- The exact expected strings already embedded in `DeleteValueQuerySpec`, `IsResourceInUseQuerySpec` and `GetResourcePropertiesAndValuesQuerySpec` are kept verbatim as legacy fixtures; the assertions now compare Jena canonical form (`QueryFactory` / `UpdateFactory`, prefix map cleared). No case was removed; the three `failIf` failure cases and the `contains` shape assertions stay.
- The `standoffTagFilter` branch previously had only `contains` checks. Its legacy rendering (both `queryAllNonStandoff` polarities) was captured before editing and added as two new fixtures. Spec count 22 → 24.
- An independent verifier diffed the fixtures against `HEAD`, traced the legacy builder for the two new fixtures, and re-ran everything: `//modules/webapi:test`, `//modules/sparql-builder:test` and `just check` pass.

## Notes for review

- `Fragments.values` throws on an empty resource-IRI list where the legacy code rendered an invalid `VALUES ?resource {  }`. Both callers already guard on non-empty.
- The link-value property is still derived as `toInternalSchema.toIri + "Value"` as before, not switched to `fromLinkPropToLinkValueProp`, to keep this a mechanical port.

## Readability retrofit

Retrofitted after review: `GetResourcePropertiesAndValuesQuery` dropped its two dead parameters (`queryAllNonStandoff` was always true, `maybeValueIri` never set; their three spec cases are removed, all other fixtures untouched) and is now two complete top-level templates (current values / version-date history) with holes only for the dynamic parts, UNION nesting preserved. The search responder got its own `SearchResultResourcesQuery` (two complete inline templates, pinned against the previous renderings in `SearchResultResourcesQuerySpec`).


modules/webapi/src/main/scala/org/knora/webapi/slice/resources/repo/DeleteValueQuery.scala
```diff
@@ -57,153 +54,117 @@ object DeleteValueQuery extends QueryBuilderHelper {
...
+                     |  rdf:subject $resource ;
+                     |  rdf:predicate $linkProperty ;
+                     |  rdf:object $linkTarget ;
+                     |  knora-base:valueHasRefCount ${Literal.int(lu.currentReferenceCount)} ;
+                     |  knora-base:isDeleted false ;
+                     |  knora-base:valueHasUUID ${linkUuidOf(i)} ."""
+            }.joinLines}
+                   |  # Get the resource's last modification date, if it has one, so we can update it
+                   |  OPTIONAL { $resource knora-base:lastModificationDate ?resourceLastModificationDate . }
+                   |}""".render,
```
siers, line 167: is this better?

siers (approved):

modules/webapi/src/main/scala/org/knora/webapi/slice/resources/repo/DeleteValueQuery.scala
BalduinLandolt: there's room for improvement here, but I would say it's still better than the RDF4J thing

siers: there is some room for improvement, but you don't have to edit this, it's fine

PR #4315, opened 2026-09-10
refactor: migrate Lucene search queries and export resource finder to interpolated SPARQL DSL (DEV-7224)

BalduinLandolt:
Part of DEV-7150 (bucket). Resolves DEV-7224. Stacked on #4314.

## What

Migrates the last query sites that mixed `s"..."` SPARQL with RDF4J escaping to the `sparql"..."` interpolator:

- `SearchFulltextQuery`: the fulltext page/count query and the breadth probe are whole-query templates. `build` returns `Select.search(...)` and `buildProbe` returns `Select.searchProbe(...)`, so the timeout tiers now live with the query. `SearchResponderV2` passes the wrappers through and `FulltextBreadthGuard`'s cache key carries the `Select` (a case class, so key equality is unchanged). The performance rationale comments (DEV-6864, DEV-6850, stage measurements) are kept.
- `SearchQueries`: `selectCountByLabel` and `constructSearchByLabel`; the shared project/class restriction is a `Fragment` helper with its `rdfs:subClassOf*` documentation intact.
- `FindResourcesService`: rendering extracted into a package-private `FindResourcesQueries` object taking the project graph and the pre-expanded class IRIs, so it is unit-testable. The subclass-expansion invariant, the deleted-resource exclusion (DEV-7008) and the Gravsearch timeouts are unchanged.

No RDF4J import and no `s"..."` SPARQL remain in these files.

## One behavioural fix

The label queries previously embedded `FusekiLucenceQuery.getQueryString`, which is already SPARQL-escaped, directly between quotes. Wrapping that in a typed literal would have escaped it twice, so the port passes the raw value through `Literal.string` instead. The golden fixtures could not catch this (their term is `Anton*`); a new spec case pins that a term like `O'Brien*` is escaped exactly once. The fulltext query was not affected, since `LuceneQueryString` holds the raw term.

## Verification

- All golden files under `SearchQueriesSpec__*.txt` and `SearchFulltextQuerySpec__*.txt` are byte-unchanged and now serve as legacy fixtures compared through Jena canonical form. No test was removed; the `Knight's` / `say "hello"` escaping assertions and the `failIf` case still pass unchanged, since `Literal.string` escaping is pinned to match `Rdf.literalOf` (`Rdf4jEscapingSpec`).
- `FindResourcesQueriesSpec` embeds the three legacy renderings captured before the port. `ExportServiceSpec` (30 tests, golden-pinned export output produced through these queries) passes with unregenerated goldens.
- An independent verifier re-traced the fixtures and re-ran `//modules/webapi:test`, `//modules/sparql-builder:test` and `just check`.


siers (approved):
Does this introduce more queries than there used to be? It's a bit harder to follow

BalduinLandolt:
> Does this introduce more queries than there used to be? It's a bit harder to follow

should be the same number (I don't see anything new and claude thinks so too).  
There was one PR earlier in the stack., where a query with a lot of conditionals was split up into two queries, but here that's not the case.

siers:
Ok, I went through it again, makes sense.

PR #4318, opened 2026-09-10
refactor: migrate the ValueRepo queries to interpolated SPARQL DSL (DEV-7228)

BalduinLandolt:
Part of DEV-7150 (bucket). Resolves DEV-7228. Stacked on #4317.

## What

Extracts the seven queries embedded in `ValueRepo` into a package-private `ValueQueries` object built from whole-query `sparql"..."` templates, and turns `ValueRepo` into plain orchestration:

- `findById` (CONSTRUCT with the single-hop `rdfs:subClassOf knora-base:Value` guard and three OPTIONALs), `findPreviousValue` (SELECT)
- `eraseValue` now returns the two `WITH <project graph>` DELETEs as a list; `ValueRepo` runs them sequentially so the standoff nodes still go first
- `eraseValueDirectLink`, `updateValuePermissions`
- `reorderValues`, previously an `s"..."` template: the `VALUES (?item ?newOrder)` tuple rows are built as fragments, since the DSL's `Fragments.values` only covers single-variable VALUES

`createValue`, `updateValue` and `checkDuplicateOrder` are untouched (they delegate to `InsertValueQueryBuilder` and the already migrated `CheckDuplicateOrderQuery`). No RDF4J import and no `QueryBuilderHelper` mixin remain in `ValueRepo`.

## Verification

- The eight legacy renderings were captured before editing and are embedded verbatim in the new `ValueQueriesSpec`, which compares Jena canonical form (prefix map cleared) and pins the erase order. An independent verifier re-rendered the legacy builders with RDF4J 5.3.2 and got byte-identical strings.
- The only rendering differences are prefixed names instead of full IRIs, which canonicalisation removes.
- `//modules/webapi:test`, `//modules/sparql-builder:test` and `just check` pass.

## Flagged, not changed

- `findById` matches `rdfs:subClassOf` with a single hop, as the legacy did. A value class two levels below `knora-base:Value` would not match; whether that is intended is a separate question.
- The standoff erase's WHERE requires an incoming triple (`?s ?oo <value>`) to exist, as before.

## Readability retrofit

Retrofitted to the inline-template convention (#4295): prefix vals inlined into every `ValueQueries` template; the two erase statements are literal templates inside the returned list. Rendered SPARQL unchanged.


siers (approved):
👍

PR #4328, opened 2026-09-10
feat: bump Sipi to v9.0.1

subotic:
## What

Bump the pinned `daschswiss/sipi` base image from **v8.0.0 → v9.0.1** (latest on Docker Hub).

- `MODULE.bazel`: both per-arch single-manifest digests (`sipi_base_amd64`, `sipi_base_arm64`) plus the tag and index-digest comment anchors updated to v9.0.1. The version is read from the image's own OCI label, so these digests are the sole source.

## Why the extra changes (v9 breaking change)

Sipi v9 **refuses to start** if `jwt_secret` is empty, shorter than 32 bytes, or equal to the old shipped default `"UP 4888, nice 4-8-4 steam engine"` (now denylisted). Our configs used exactly that literal, which would break the local stack and every container-based test.

This PR moves each runtime site that shared the literal to a distinct ≥32-byte dev secret (`dev-only-insecure-jwt-secret-change-me`) so JWT auth still verifies across dsp-api ↔ dsp-ingest ↔ Sipi:

- 4 Sipi configs under `modules/sipi/config/` + the test-resource copies in `test-it` / `test-e2e`
- `webapi` and `ingest` `application.conf`
- `docker-compose.yml` (`JWT_SECRET`)
- both `DspIngestTestContainer` helpers and `SipiIT`

The pure `JwtCodecSpec` unit test keeps the old literal as an arbitrary codec key — it never talks to a Sipi container.

Production supplies the secret via `SIPI_JWTKEY` (handled in ops-deploy, coordinated with the matching v9 rollout).

## Other v9 changes reviewed, no action needed here

- `--adminuser` / `--adminpasswd` / `SIPI_ADMIN*` removed → not used in dsp-api.
- `token.lua` postMessage route removed → not used.
- `docroot` overlap startup refusal → our `docroot` does not overlap imgroot/scriptdir/tmpdir.
- `admission_mode` now defaults to `advanced` → test uploads are small; can force `SIPI_ADMISSION_MODE=basic` if needed.
- `SIPI_PUBLIC_HOSTS` forwarded-host allowlist → dev reflect-any default is fine; prod handled in ops-deploy.

## Testing

- Rebuilt the derived `knora-sipi` image from the v9 base; its OCI version label reports `v9.0.1`.
- `SipiIT` (12 tests) passes against the v9.0.1 image — container starts (clears the v9 jwt_secret guard) and auth verifies (authorized → Ok, no-permission → Unauthorized, unknown → Not Found).
- Full local suite green: **10/10 Bazel test targets pass** — `webapi`, `ingest`, `bagit`, `jwt`, `shacl-validator`, `sparql-builder` (unit) plus `test-it`, `test-it:test_gravsearch_span`, `test-e2e`, `test-ingest-integration` (container-based). The v9 restricted-image / info.json gating changes did not shift the e2e assertions.

https://claude.ai/code/session_01Fn4MNN6j99c1aNxGKejcFd

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -310,23 +310,23 @@ use_repo(image_versions, "dsp_image_versions")
 
 oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
 
-# Upstream Sipi image (daschswiss/sipi:v8.0.0), pinned per-arch by manifest
+# Upstream Sipi image (daschswiss/sipi:v9.0.1), pinned per-arch by manifest
 # digest. Pulling each platform's single image manifest directly avoids
-# index/variant matching — the arm64 entry of the v8.0.0 index carries a `v8`
+# index/variant matching — the arm64 entry of the v9.0.1 index carries a `v8`
 # variant, which a bare `linux/arm64` platform request does not match. These two
 # digests are the SOLE source of the sipi version: the /version endpoint reads
 # the tag from the image's own `org.opencontainers.image.version` OCI label
 # (//tools/buildinfo:oci_config_label), so there is no duplicated tag string to
 # keep in sync. To bump, see "Bumping the SIPI version" in CLAUDE.md.
-# Index digest: sha256:8cb1098a3fdf2383c5bb3f382f46a1fe58dad4cfed0a56c7aa8c5e06f5ac5e85
+# Index digest: sha256:a115b272f59f93797ff34578a46b6d30af84e4d01c23b1f7ad3861c7670ca601
 oci.pull(
     name = "sipi_base_amd64",
-    digest = "sha256:6fb699b38b21e992719c03d3308e97a1fd4a755758fcb17e966a9c3293539ffd",
+    digest = "sha256:239a0e9ba69b6b1794590fcf2caa712855685f2ad711d8fd3d54cc1c1c04187d",
     image = "index.docker.io/daschswiss/sipi",
 )
 oci.pull(
     name = "sipi_base_arm64",
-    digest = "sha256:069a1955331576d27df0b9efd11711b807228a52449c9532e46a0c9c2df0d5b5",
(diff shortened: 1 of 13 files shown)


BalduinLandolt (approved):

siers (approved):
It's more than a shame that MODULE.bazel doesn't have a tag pinned alongside the hash digest.

subotic:
Closing as superseded by #4329, which merged on 2026-09-16 as `9ddad36b7`.

Everything this PR set out to do is on `main`, and in one case gone further:

| This PR | On `main` now |
|---|---|
| Sipi **v9.0.1** | Sipi **v9.1.0** — newer, and the release that adds the `stream` permission type |
| JWT secret moved off `"UP 4888, nice 4-8-4 steam engine"` | done, with the same replacement value `dev-only-insecure-jwt-secret-change-me` |
| 4 Sipi configs, the `test-it` / `test-e2e` resource copies, `webapi` + `ingest` `application.conf`, `docker-compose.yml`, both `DspIngestTestContainer` helpers, `SipiIT` | all migrated |

Verified rather than assumed: all seven config sites plus `docker-compose.yml` carry the new secret on `main`, and the only surviving `UP 4888` references are `JwtCodecSpec.scala` and stale `target/` build artifacts.

**`JwtCodecSpec` keeping the old literal is deliberate**, and #4329 kept it for the same reason this PR did: it is a jwt-scala interop vector — a pinned token plus the secret that token was signed with — not configuration. It never talks to a Sipi container. Please don't "fix" it.

This PR now reports `DIRTY` precisely because #4329 rewrote the same lines, so resolving the conflicts would leave a no-op diff. Nothing here is unlanded.

The v9 findings in the description that needed no action — `--adminuser` removal, the `token.lua` postMessage route, the `docroot` overlap guard, `admission_mode`, `SIPI_PUBLIC_HOSTS` — still hold and are worth keeping in mind for the ops-deploy side of the rollout.

PR #4329, opened 2026-09-10
feat(admin)!: return an asset access decision instead of a permission code

subotic:
Part of DEV-7244. Supersedes DEV-7155 — this branch's original commit (default a restrict size in `sipi.init.lua`) was dropped, because the model it patched is the one this PR removes.

## Motivation

`GET /admin/files/{shortcode}/{filename}` returned the caller's ACL level as a bare integer, and every consumer invented an access policy from it:

| Caller | Policy it invented | Consequence |
|---|---|---|
| `sipi.init.lua` | code 1 → `allow`, whatever the media type | **`RV` users received the full-resolution unclamped still image** from `/file` |
| `ProjectsEndpointsHandler` | `permissionCode >= 2` → hand over the original | any change to the code's meaning silently widens original-file access |

Three inputs decide what a caller may receive — the ACL level, *which artifact* (original or derivative), and *what kind of media*. The response carried only the first, so each consumer reconstructed the other two, differently, and every defect in this area was one instance of that pattern.

Sipi is the forcing function. Wave-2 refuses a `restrict` decision that restricts nothing, so the ~2,399 `RV` audio, text and document file values measured on stage (2,388 of them audio in project `082A`) would 403 once the wave-2 image rolls. **That is the breakage this PR prevents, not one it introduces:** under the new model those values resolve to `derivative: stream`, so they keep playing, anonymously included, while dsp-ingest still refuses their originals.

## Summary

- `/admin/files/…` now returns an explicit two-channel decision instead of an ACL code: `original` (grant/withhold) and `derivative` (full/clamped/stream/denied). Neither consumer performs permission arithmetic any more.
- An `RV` user can play audio and video and open PDFs, including anonymously, while dsp-ingest still refuses the originals. An `RV` **archive** is now refused on both routes, and an `RV` **still image** is no longer served at full fidelity.
- The project restricted-view setting becomes optional: `DELETE .../RestrictedViewSettings` returns a project to the platform default, `GET` reports `isDefault`, and `pct:100` is rejected and migrated out of the triplestore.

### The endpoint returns a decision
- `MediaKind`, `OriginalAccess`, `DerivativeAccess`, `AssetAccess` under `slice/admin/domain/model/`. `AssetAccess.from` is the single total policy function; the constructor is private, so no caller can assemble an inconsistent pair such as `Grant` with `Denied`.
- The two channels never merge: `original` is read only by dsp-ingest, `derivative` only by Sipi's hook. That decoupling is the whole point — inferring download eligibility from a rendering decision is what let `permissionCode >= 2` widen original access.
- `sipi.init.lua` translates the `derivative` literal onto Sipi's vocabulary and does nothing else: no arithmetic, no media knowledge. Its `permission_code == 1` branch is gone rather than given a better default.
- dsp-ingest gates on `original == "grant"`.

### The restricted view becomes optional
- `KnoraProject.restrictedView` is `Option[RestrictedView]`; the default is applied only at the read boundary.
- Two `DELETE` endpoints (by IRI and shortcode), and all three verbs now return one shape carrying `isDefault`.
- `RestrictedView.Size.from` rejects `pct:100`; `UpgradePluginPR4329` clears both `pct:100` and the `!128,128` that `UpgradePluginPR3112` backfilled, restoring "nothing configured" for 50 of 62 projects. `KnoraBaseVersion` 55 → 56.

### Sipi 9.1.0
- Both per-arch `oci.pull` digests bumped from v8.0.0. This is also the wave-2 rollout — see Gotchas.

### Sipi 9.1.0 refused to start, and the e2e suite could not run at all
**Problem:** After the digest bump the suite reported `Tests run: 0, Failures: 1` — a testcontainer launch failure, not test failures. `knora-sipi` exited 1 with *"refusing startup — jwt_secret is empty, too short (<32 bytes), or the shipped default; set SIPI_JWTKEY"*.
**Tried:** Reading it as a missing-environment problem, i.e. set `SIPI_JWTKEY` in `SipiTestContainer`. That fixes the e2e suite and leaves the local docker-compose stack broken in the same way.
**Solution:** Sipi's `SHIPPED_JWT_SECRET` is `"UP 4888, nice 4-8-4 steam engine"`, and dsp-api used that exact string as its JWT default. Wave-2 turned the shipped default into a fail-closed startup gate. The secret is a **shared trust domain** — dsp-api signs, Sipi and dsp-ingest verify — so it moves in all 14 sites together; a partial move leaves dsp-api signing tokens Sipi rejects, which fails far more confusingly than a clean startup refusal. The new value is Sipi's own test token, `dev-only-insecure-jwt-secret-change-me`, so the two repos now agree. Config is the single mechanism; no `SIPI_JWTKEY` override was added.

### The unmapped-class safety net was dead code
**Problem:** The spec required an RDF file-value class with no `MediaKind` mapping to fail closed *and be counted*. `FileValuePermissionsQuery` constrained `?fileValueClass` to exactly the set `MediaKind` maps, so an unmapped class returned **zero rows** — it 404'd as "file not found" and never reached `failClosed` or the counter.
**Solution:** That same filter was also too loose: it did not stop two *concrete sibling* classes on one node producing two rows, which `getFirstRow` then picked between nondeterministically. One filter, too strict and too loose at once. The filter is removed, the projection is `SELECT DISTINCT` (collapsing the rows the `?objPred` optimizer hint multiplies), and the responder accepts exactly one class — zero-mapping *and* ambiguity both fail closed and increment the counter. Renamed `unmapped_file_value_class` → `unresolved_file_value_class`, since it now counts both. Both paths have fixtures; neither was covered before.

### `AssetAccess.from` could not meet its own acceptance criterion
**Problem:** The spec's own sketch ended its `RestrictedView` arm in a bare catch-all, so a new `MediaKind` would silently inherit `Stream` — while the criterion said adding a kind must fail to compile.
**Solution:** Rewritten with every case spelled out. Verified by adding a ninth kind and watching `-Werror` fail the build.

### A window where the suite was red by design
**Problem:** Between the `sipi.init.lua` rewrite and the digest bump, 13 e2e tests failed. dsp-api fetches assets from `/file` **anonymously**, and the `anything` project's default DOAP grants `RV` to `UnknownUser`, so every anonymous fetch of a non-image asset resolved to `stream` — which Sipi v8.0.0 rejected with a 500. That included the standoff XSLT path (`SipiServiceLive.getTextFileRequest`), an unauthenticated server-to-server fetch nobody had recorded as depending on the old "code 1 means full access" rule.
**Solution:** Diagnosed rather than patched around — confirmed by temporarily returning `allow` instead of `stream` and watching the specs pass, then reverting. The digest bump cleared all 13.

## Gotchas

- **Deploy Sipi and dsp-api together — you already do, structurally.** `knora-sipi` is a thin overlay that copies the Lua scripts onto the upstream Sipi image, and the two `oci.pull` digests are the sole source of the Sipi version, so the hook and the binary ship as one artifact under one dsp-api version. The hook/binary skew that would 500 is therefore a dev and CI hazard only, in the window between rewriting the hook and bumping the digest.
- **This release *is* the wave-2 rollout.** `MODULE.bazel` pinned v8.0.0, which predates wave-2, so wave-2 reaches a deployment only through this bump. Wave-2's stage verification and release note belong to this train, not to a later step.
- **`just check` does not run markdownlint.** Separate recipe, separate CI gate. A green `just check` can still fail CI on a table's column alignment (MD060).
- **`JwtCodecSpec` keeps the old secret on purpose.** It is a jwt-scala interop vector — a pinned token plus the secret that token was signed with — not configuration. Do not "fix" it.
- **Production is unaffected by the JWT change.** Deployments take the secret from the vault (`sipi.prod-config.lua.j2` sets `jwt_secret` from `DSP_JWT_TOKEN`). Only local and test configs moved.
- **`stream` is not a security boundary.** Only still images are transcoded on ingest, so for every other media kind the derivative *is* the original: `original: withhold` + `derivative: stream` withholds no bytes. It closes one route and records an intent. Nothing downstream should be built as if it enforced anything.

## Test Plan

- [x] `just test-e2e` — **OK (945 tests)**, 0 failures
- [x] `just test-unit` — 6/6
- [x] `just test-it` — 2/2, re-run after un-ignoring `streamSuite`; both stream cases now execute and pass against the v9.1.0 image
- [x] `just check` — green
- [x] `just markdownlint` — green
- [x] Unmapped and ambiguous file-value classes each deny, log, and increment the counter (asserted on the counter, not just the failure)
- [x] dsp-ingest's original-download gate refuses every `RV` asset, of every media kind
- [ ] Stage walkthrough after deploy: an `082A` audio file plays, an `ExIK4B…` document opens, dsp-ingest refuses both originals, a `dRHtOC…` still image renders clamped
- [ ] Re-run the stage censuses before deploy to confirm `RV` archives are still zero


BalduinLandolt (commented):
I was under the impression that restricted view but no restricted view settings was our workaround for full resolution but no download button shown in the image viewer.  
Are you sure this is not breaking intended behaviour for some projects?

subotic:
lol, thanks. I will check that and fix it. We should not encode behaviour in such a way.

subotic:
@BalduinLandolt I completely reworked the streaming support. Please take a look.

subotic:
Please also take a look at the upgrade plugin if it looks ok.

jnussbaum (commented):
@subotic There is one problematic implication for 082A sva: The PR description says "Wave-2 refuses a restrict decision that restricts nothing, so the ~2,399 RV audio, text and document file values measured on stage (2,388 of them audio in project 082A) would 403 once the wave-2 image rolls."

However, the current behaviour that non-logged-in users can stream the audios, but not download them, is intended, and should not change.

jnussbaum:
Another small finding: `SipiIT.streamSuite` still carries `@@ TestAspect.ignore`, with a comment saying to drop it in the change that bumps the digests. This PR bumps them. So stream has no direct test here.

BalduinLandolt (approved):
Looks good to me, see @jnussbaum 's comments, though

subotic:
@jnussbaum That behaviour is preserved, and preserving it is the point of the PR. I have clarified the sentence you quoted, which was describing the breakage this change prevents rather than one it introduces.

`AssetAccess.from` decides `RestrictedView` + `Audio` as `derivative: stream`, `original: withhold`. So an anonymous `RV` user on an `082A` audio file still plays it, and dsp-ingest still refuses the original. What would have 403'd is the *old* model under wave-2: it sent `restrict` with no size for non-images, and wave-2 refuses a restriction that restricts nothing.

The `streamSuite` integration tests that pin this are no longer ignored, per your other comment.

subotic:
@jnussbaum Good catch, fixed. The `@@ TestAspect.ignore` and the comment explaining it are gone, folded into the digest-bump commit where they belonged. `streamSuite` now runs against the real v9.1.0 image.

That also gives your other comment its answer in test form: both cases stub `derivative: "stream", original: "withhold"` and assert 200, one on the IIIF route and one on `/file`.

jnussbaum:
@subotic Great, thanks for the 2 clarifications.

PR #4330, opened 2026-09-11
feat: bulk-import permission parity — honor payload hasPermissions and resolve per-entity DOAPs

Nora-Olivia-Ammann:
[DEV-7240](https://linear.app/dasch/issue/DEV-7240)

## Summary

- Align the v3 bulk data-import's permission assignment with the v2 single-resource create path.
- Honor a payload-supplied `hasPermissions` per resource and value; otherwise resolve class-level and property-level DOAPs per entity.
- Flip the bulk-import parity E2E to compare `hasPermissions` — the last deliberate tolerance is gone.

## Changes

- **DoapResolver seam** (`slice/ontology/DoapResolver.scala`): a narrow trait (resolve DOAP + validate a payload string), a live impl backed by `PermissionsResponder`/`PermissionUtilADM`, and an in-memory double, so `OntologyTransformerSpec` stays triplestore-free.
- **Transformer**: `addResourceMetadata`/`addValueMetadata` became `Task[Unit]`; they resolve each distinct DOAP key and each distinct payload literal once (N+1 guard) and, per entity, keep the payload permission or apply the resolved default. `ConversionContext.attachedToUser` widened `UserIri`→`User`.
- **Import service**: dropped the group-level default and the `PermissionsResponder` dependency; removed the now-dead `newDataImportDefaultObjectAccessPermissions`.
- **Parity spec**: removed the `includePermissions` toggle so `hasPermissions` is compared; added `explicit_resource_permission`/`explicit_value_permission` fixtures (payload strings differing starkly from the DOAP default) with a dedicated assertion that the payload wins on both write paths.
- **Docs**: updated `project-data-import.md` to the new per-entity honor-payload-else-DOAP behavior.

## Test Plan

- `bazel test //modules/webapi:test` → `OK (1886 tests)` (includes `OntologyTransformerSpec` with new payload-override cases; no migration-import regression).
- `bazel test //modules/test-e2e:test --test_filter='.*BulkImportParity.*' --nocache_test_results` → PASSED with `hasPermissions` compared, on all fixtures plus the two new ones.
- `just check` (scalafmt + license) green.


modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/DoapResolver.scala
```diff
@@ -0,0 +1,68 @@
...
+ * narrow trait so [[OntologyTransformerSpec]] can substitute an in-memory double and stay triplestore-free. Injecting
+ * the concrete [[PermissionUtilADM]] instead would pull `GroupService` and its triplestore graph into that spec.
+ */
+trait DoapResolver {
+
+  /** The default object-access permission string for a resource of the given class. */
+  def resourceDoap(project: KnoraProject, user: User, resourceClassIri: String): Task[String]
+
+  /** The default object-access permission string for a value of the given property on the given resource class. */
+  def valueDoap(project: KnoraProject, user: User, resourceClassIri: String, propertyIri: String): Task[String]
```
BalduinLandolt, line 29: Ideally, IRIs would not be typed `String`, but it depends a bit on the context and caller, what the better choice would be.

BalduinLandolt (approved):
👍

modules/webapi/src/main/scala/org/knora/webapi/slice/ontology/DoapResolver.scala
Nora-Olivia-Ammann: Kept as `String` here on purpose. `OntologyTransformer` pulls these class and property IRIs out of a Jena `Model` (raw string URIs), and they serve as dedup map keys for the N+1 guard. `DoapResolver` is a narrow test seam over that traversal, not an API boundary, so a typed IRI would be unwrapped again one call later (`PermissionsResponder` takes `SmartIri`).

If we want stronger typing later, the cleaner spot is to convert once in the transformer and map a malformed IRI to a 400.

PR #4331, opened 2026-09-13
docs: add ARCH-MAP.md and CONTEXT.md, adopt ADR-0011 ports contract, move ADRs to docs/adr

subotic:
### Description

Docs and build-tooling change. Three commits, no behaviour change in the API.

**Motivation.** dsp-api had a draft context map and glossary at the root (`CONTEXT-MAP.md`,
`UBIQUITOUS_LANGUAGE.md`) but no machine-checkable component map, and cross-context access was
governed only by prose guardrails. With the modularization plan about to extract the first domain
module, the topology and the access contract need to be written down in the form agents and reviewers
actually consult.

**What changes**

1. `build:` markdownlint now comes from the Nix dev shell (`flake.nix`, pinned by the lock) instead of
   a floating `:latest` Docker image, and the check is folded into `just check`. The `check-formatting`
   CI job already runs inside `nix develop`, so it carries the gate; the Docs Test job no longer lints.
2. `docs:` architecture map and vocabulary.
    - `ARCH-MAP.md` (new): 26 components following current code topology. Every tracked file maps to
      exactly one component (verified with `git ls-files -- ':(glob)...'` per glob). Each entry has
      key entities, a local-context kit (7 files or fewer), symmetric dependency edges, enforcement-tagged
      boundary rules and durable state with its writers. Conventions include the one-way dependency
      arrow, the wiring inventories, and a banned-constructs table. `CLAUDE.md` gets a one-line
      pull-on-demand pointer; the map is never auto-loaded.
    - `CONTEXT.md` (new, replaces `CONTEXT-MAP.md` + `UBIQUITOUS_LANGUAGE.md`): the root file is an
      index (context map, guardrails, shared-kernel terms); each of the eight bounded contexts has its
      own `docs/contexts/<slug>/CONTEXT.md`. They live under `docs/contexts/` only until a context is
      extracted; `MODULARIZATION-PLAN.md` records the move as a Phase 2 work item and completion criterion.
    - ADRs move from `docs/05-internals/design/adr/ADR-NNNN-*.md` to `docs/adr/NNNN-*.md`, get a
      top-level "Decisions (ADR)" nav section, and the missing 0010 is added to the nav.
    - Removed as unused: `.adr-dir` (adr-tools config) and `.kodus-readiness.yml`.
3. `docs:` ADR-0011, cross-context access through consumer-defined ports. Adapts dasch-ops-platform
   ADR-0013 to dsp-api: graph sovereignty, consumer-owned `ports` packages, provider adapters
   `<Port>Live` next to their data, `LayersLive` as pure wiring, identifiers as shared-kernel value
   types (replacing the earlier "small Projects contract"), whole-graph movement as the single platform
   exception. Search is resolved by the same shape: it keeps the Gravsearch language and delivery; SPARQL
   translation and result assembly move behind a `ResourceSearch` port implemented by Resources & Values.
   `CONTEXT.md`, the per-context files, `MODULARIZATION-PLAN.md`, the map's conventions,
   `CONVENTIONS.md`, `REVIEW.md` and the mkdocs nav are aligned.

**Findings surfaced by the map worth a look** (recorded as boundary-rule violations, not fixed here)

- Seven components import `slice.api.*`; api imports resources and export back, so two edges are cycles.
- `webapi-common` depends upward on seven slices (the "false foundation", now measured).
- `webapi-export` links `//modules/ingest:ingest_classes` in-process across the deployable boundary.
- The admin, permissions and project data graphs each have several writers from different contexts.
- A new Bazel module must be hand-registered in three lists or `just check` silently skips it.

**Decisions taken in review conversation**

- Per-context `CONTEXT.md` files under `docs/contexts/` rather than code-adjacent, because Projects and
  Identity & Access still share `slice/admin`.
- dsp-api keeps `<Name>Live` / `<Name>InMemory` naming instead of ADR-0013's `Live<Port>` / `Test<Port>`.

**Breaking change:** none. `just markdownlint` no longer exists as a separate recipe; use `just check`.


subotic:
@BalduinLandolt I allowed myself to add an ADR to dsp-api and to organize things a bit. I also used the dune plugin. Please take a look and see if you agree with what is written, as I had to take some decisions, specifically about search and ports/adapters (you had it a bit differently in the plan).

siers (approved):
I skimmed through a large parts of it. I assume most of this is generated and should be close enough for machine consumption.

docs/architecture/README.md
BalduinLandolt, line 18: still needed? I don't think so

docs/contexts/project-migration/CONTEXT.md
```diff
@@ -0,0 +1,73 @@
...
+_Avoid_: Archival package unless it conforms to the separate Archive contract
+
+**Data Task**:
+A tracked long-running operation with a lifecycle and status.
+_Avoid_: Untracked background job
+
+**Bulk graph movement**:
+An intentional whole-graph transfer using the RDF platform, and the single platform exception to
+graph sovereignty.
+_Avoid_: General permission for cross-context SPARQL
```
BalduinLandolt, line 41: I'm not aware that we have either of thee concepts in the API.  
But what seems to be missing here, is maintenance actions and these pseudo migrations used to re-import the built-in ontologies when they have changed.

BalduinLandolt: (correction: this is probably missing not here, but in the operations file)

docs/contexts/projects/CONTEXT.md
```diff
@@ -0,0 +1,77 @@
...
+- `ProjectIri` is a shared-kernel value type. Project details cross a context boundary only through
+  a port the consuming context declares, and Projects implements those adapters, for example the
+  project snapshot Project Migration needs
+  ([ADR-0011](../../adr/0011-cross-context-access-ports-and-adapters.md)).
+
+## Language
+
+**Project**:
+The ownership and scoping unit for Data Models, Resources, settings, and access administration.
+_Avoid_: Tenant
```
BalduinLandolt, line 31: also, perhaps "dataset"

docs/contexts/resources-values/CONTEXT.md
BalduinLandolt, line 1: maybe the context should be "data", not "resources-values", at least that's our domain term

BalduinLandolt (approved):

BalduinLandolt (approved):

PR #4332, opened 2026-09-14
fix(ingest): write JSON sidecars atomically

siers:
`StorageServiceLive.saveJsonFile` opened the target with `TRUNCATE_EXISTING` and wrote in place. If the process dies mid-write, or a reader picks the file up at the wrong moment, it sees a truncated `.info` sidecar. Noticed while looking at `MigrateSizes`, which rewrites every sidecar in the store and so has a lot of chances to hit it.

Now it writes to a temp file in the target's own directory and renames it over the target with `ATOMIC_MOVE`. Same directory matters — an atomic rename can't cross filesystems. `createTempFileInScoped` cleans up on failure, so a failed write leaves the original sidecar untouched and no stray temp file.

This covers all `saveJsonFile` callers, not just the migration `updateAssetInfoForDerivative` and `backfillOriginalMimeType` commands, so this is now the default mechanism for sidecar file update.

Temp files are named `.<assetid>.info<random>.tmp` and are briefly visible to a concurrent directory scan, but `FileFilters.isInfoFile` matches on the `.info` extension, so scans skip them.

`bazel test //modules/ingest:test` passes


SamuelBoerlin (approved):
Thanks!

BalduinLandolt (approved):

PR #4341, opened 2026-09-16
feat: validate knora-base structure with SHACL data-shapes (DEV-7289)

Nora-Olivia-Ammann:
## What

Add the static SHACL shape file that validates project-migration import data against the `knora-base` structure, plus its test suite. Validation uses fixed `knora-base` shapes only. It does not consider the project ontology.

`data-shapes.ttl` defines:

- `ResourceShape` — required properties on every resource: label, `isDeleted`, `attachedToUser`, `attachedToProject`, `hasPermissions`, `creationDate`, and standoff-link coherence.
- A shape per `knora-base:Value` subclass: Boolean, Color, Date, Decimal, Interval, Int, Link, Time, Uri, Geom, Geoname, List and Text.
- File-value shapes (Archive, Audio, Document, MovingImage, StillImage and its variants, Text) and the `Representation` shapes that link to them.
- `AttachedToUserExistsShape` and `AttachedToUserNotBuiltInShape` — a user reference must resolve to a real, non-built-in user.

## Notable decisions

- **No `sh:closed`.** The validator runs with RDFS inference and a shared `ValueShape`. TopBraid `sh:closed` allows only a shape's own declared paths. So it rejects the shared value properties and the inferred super-properties (`rdf:type`, `knora-base:valueHas`, `knora-base:objectCannotBeMarkedAsDeleted`). The shapes enforce required properties, cardinality, datatypes and node kinds instead.
- Fix two `Representation` shapes that used the wrong file-value class: `DocumentRepresentationShape` used `AudioFileValue`, `TextRepresentationShape` used `MovingImageFileValue`.

## Tests

`ProjectMigrationImportValidatorSpec` covers the resource, value, file-value and representation shapes. Valid data passes. Missing or wrong-typed data fails.

DEV-7289


BalduinLandolt (approved):
Looks reasonable, though I'm no SHACL expert, so I wouldn't notice any issues with these shapes.

One nitpick maybe: The date shapes file carries practically no comments. I don't know how you feel about commenting SHACL shapes; but I feel it would help a not-so-informed reader to find their way around the shapes a bit better.

Nora-Olivia-Ammann:
> Looks reasonable, though I'm no SHACL expert, so I wouldn't notice any issues with these shapes.
> 
> One nitpick maybe: The date shapes file carries practically no comments. I don't know how you feel about commenting SHACL shapes; but I feel it would help a not-so-informed reader to find their way around the shapes a bit better.

Thanks, I took a good look and so did claude, so I am reasonably certain that the shapes are fine, at least they are not erroneous although they do not cover absolutely everything at the moment.

Regarding comments would you want generic one or one on each shape?

BalduinLandolt:
> Regarding comments would you want generic one or one on each shape?

I was more thinking one on each shape, but I don't mind very much _how_ it's done, as long as it helps. And if there are no comments, that's fine by me too.

PR #4342, opened 2026-09-17
feat: record the generated Gravsearch prequery as a gravsearch.prequery trace event (DEV-7302)

BalduinLandolt:
[DEV-7302](https://linear.app/dasch/issue/DEV-7302)

## Summary

The root `gravsearch` span already carries the Gravsearch the client submitted, as a `gravsearch.query`
event. It also carries the **generated SELECT prequery** — the statement actually sent to the
triplestore — as a `gravsearch.prequery` event with `db.query.text`, on both the page path
(`gravsearchV2`) and the count path (`gravsearchCountV2`).

The two together turn a slow trace into something reproducible: the Gravsearch says what was asked
for, the prequery is the text to paste into Fuseki.

It follows the existing `recordQueryOnRoot` mechanism exactly — an **event**, never a span attribute
(the Alloy `otelcol.connector.spanmetrics` dimension list reads span attributes, and generated SPARQL
is unbounded), on the **root span only**, and **never inside a `stageSpan`**, since stage spans are
asserted to carry no events at all (the REQ-1.6 sanitized-error lock).

It is recorded right after `gravsearch.prequery.generate` closes, so it is on the trace even when the
execution that follows is interrupted. It is absent whenever generation never ran (parse failure, early
interrupt) — the same signal as a missing `gravsearch.prequery.generate` span.

## Changes

- `modules/webapi/.../responders/v2/SearchResponderV2.scala` — new `recordPrequeryOnRoot` next to
  `recordQueryOnRoot` (companion helper + trait delegate), called on both paths outside the stage span.
- `modules/test-it/.../SearchResponderV2GravsearchSpanE2ESpec.scala` — asserts the event on the root
  span for the page and count paths, its absence on a parse failure, and extends the stage-span
  no-events lock to cover both payload events and any event name.
- `docs/observability/gravsearch-trace-runbook.md` — § "Read the submitted query" now documents both
  events in one table, plus the TraceQL selector.
- `docs/observability/traceql-recipes.md` — new § 6 "Find the query text" with the event-scope recipes.
- `docs/observability/instrumentation-recipe.md` — § 6 notes that a vertical may capture more than one
  payload event, and the `db.*` semconv caveat is updated now that the generated SPARQL is captured.

## Test Plan

- `bazel build //modules/webapi:webapi //modules/test-it:test` — both compile.
- `just fmt` and `just check` — clean (markdownlint is the CI gate).
- The new assertions live in `SearchResponderV2GravsearchSpanE2ESpec`, a Docker-based integration
  suite that runs in CI's "Build and IT Test" job.

https://claude.ai/code/session_01WwcoHcTT6BWX4hEjAAu7Ay


modules/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -1470,6 +1478,25 @@ object SearchResponderV2 {
...
+   * against Fuseki, the other to attribute the cost back to what was asked for.
+   *
+   * Same three rules as [[recordQueryOnRoot]]: an event rather than a span attribute, because the Alloy
+   * `otelcol.connector.spanmetrics` dimension list reads span attributes and generated SPARQL is
+   * unbounded; on the root span only; and never inside a `stageSpan`, since stage spans are asserted to
+   * carry no events at all (the REQ-1.6 sanitized-error lock).
+   *
+   * Called right after `gravsearch.prequery.generate` closes, so the prequery is on the trace even when
+   * the execution that follows it is interrupted — which is exactly the trace worth reading.
+   */
```
mpro7, line 1477: It looks like the comment rules strengthening doesn't hold.

modules/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
```diff
@@ -114,6 +114,10 @@ trait SearchResponderV2 {
   protected final def recordQueryOnRoot(query: IRI): UIO[Unit] =
     SearchResponderV2.recordQueryOnRoot(tracing, query)
 
+  /** Records the generated SELECT prequery as a `gravsearch.prequery` event on the root span. */
```
mpro7: This comment also translates the method name.

mpro7 (approved):

modules/webapi/src/main/scala/org/knora/webapi/responders/v2/SearchResponderV2.scala
BalduinLandolt: this may be my mistake here, sorry. will fix it as soon as my tokens have recovered.

BalduinLandolt: fixed in https://github.com/dasch-swiss/dsp-api/pull/4342/commits/00eb96b7ac5cc1811f89db786645189c30368fe6

